PR #13, opened 2018-11-06
Add OSX to the list of Travis CI OSes

kpet:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: cpp
 
 os:
   - linux
+  - osx
 
 script:
   - export TOP=$(pwd)


jrprice:
I'll just note that the header directory symlink `ln -s CL OpenCL` *shouldn't* be necessary for macOS with the latest headers, and it might be nice to remove it from the Travis build to help make sure that this is the case. Otherwise this change LGTM.

kpet:
Yup, I should be able to remove this thanks to your change. I'll test without. Thanks for the comment.

kpet:
@jrprice Doesn't work without further changes to the conformance tests. I'll keep the symlink for now.

PR #16, opened 2018-11-07
Add OSX to the list of Travis CI OSes

kpet:


kpet:
@jrprice cl22_trunk seems to fail to build on OSX. I don't have access to a mac and I foresee debugging via Travis CI being quite long. Do you think you could have a look?

kpet:
@jrprice All working with #18. Thanks!

PR #18, opened 2018-11-09
Fix macOS build for 2.2

jrprice:
This should allow #16 to proceed (@kpet).

### DIFF
--- a/test_conformance/clcpp/images/main.cpp
+++ b/test_conformance/clcpp/images/main.cpp
@@ -19,6 +19,10 @@
 #include "test_sample.hpp"
 #include "test_write.hpp"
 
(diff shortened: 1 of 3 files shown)


test_conformance/clcpp/images/main.cpp
```diff
@@ -19,6 +19,10 @@
 #include "test_sample.hpp"
 #include "test_write.hpp"
 
+// FIXME: To use certain functions in test_common/harness/imageHelpers.h
+// (for example, generate_random_image_data()), the tests are required to declare
+// the following variable (hangover from code specific to Apple's implementation):
```
kpet: I'd come to the same conclusion. This is nasty. We should be able to remove this when we integrate these tests into the framework (similar to what ARM's been doing on the 2.0/2.1 branches).

PR #42, opened 2019-01-21
Prevent data races in device_execution/enqueue_flags tests

### LINKED ISSUE #40: cl20 cl21 cl22 device_execution/enqueue_flags data race
At least four of tests' kernels run into data race:
- enqueue_flags_wait_work_group_simple
- enqueue_flags_wait_work_group_event
- enqueue_flags_wait_work_group_local
- enqueue_flags_wait_work_group_event_local

res[gid] is being written by get_num_groups()/get_local_size() threads. And if one of these thread put 2 to the array cell, it still may be overwritten with 1 by another thread. Thus, the test doesn't really test anything.

alexbatashev:
The following kernels ran into data races:
- enqueue_flags_wait_work_group_simple
- enqueue_flags_wait_work_group_event
- enqueue_flags_wait_work_group_local
- enqueue_flags_wait_work_group_event_local
Fix #40

### DIFF
--- a/test_conformance/device_execution/enqueue_flags.cpp
+++ b/test_conformance/device_execution/enqueue_flags.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


alexbatashev:
The problem was with two variables: res and array. Both have been accessed from different threads.
Solutions:
- Every work-group now writes to its own res cell.
- Only one work-group in branch now has access to array cells.
- Local memory fence was changed to global as no operations on local memory had been performed.

kpet:
Can this be rejected now that we have #143 #144 #145 ?

alexbatashev:
Sure

PR #48, opened 2019-02-20
Synchronise with Khronos-private Gitlab branch

kpet:
The maintenance of the conformance tests is moving to Github.

This commit contains all the changes that have been done in
Gitlab since the first public release of the conformance tests.


gwawiork:
Looks good to me 👍

kpet:
Thanks for testing. I'm ready to bet this issue exists on the `cl20_trunk` internal branch as well. Might be worth diff'ing with the Intel branch to extract a patch but that should happen after this PR gets in IMHO.

PR #50, opened 2019-02-20
Synchronise with Khronos-private Gitlab branch

kpet:
The maintenance of the conformance tests is moving to Github.

This commit contains all the changes that have been done in
Gitlab since the first public release of the conformance tests.

Signed-off-by: Kevin Petit kevin.petit@arm.com


readme-spir-v-binaries.txt
```diff
@@ -1,17 +1,14 @@
-To run the 2.2 conformance tests test suite for the C++ features you need
+To run the 2.2 conformance tests test suite for the C++ features you need need
```
pierremoreau: A single “need” should be enough. :slightly_smiling_face:

kpet: Thanks for the review. Pushed a small patch to fix that.

PR #55, opened 2019-03-12
cl12: Compiler - updated extension list for compiler_defines_for_exensions

Jeremy-Kemp:
Updated the list of known extensions to include all extensions referenced in the OpenCL Extension Specification 1.2 - Revision 25 and OpenCL Specification 1.2 - Revision 19 documents.

Also removed extensions from later OpenCL versions.

### DIFF
--- a/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
+++ b/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
@@ -35,6 +35,7 @@ const char *known_extensions[] = {
     "cl_khr_byte_addressable_store",
     "cl_khr_fp16",
     "cl_khr_spir",
+    "cl_khr_depth_images",
 
     //API-only extensions after this point.  If you add above here, modify first_API_extension below.
     "cl_khr_gl_sharing",
@@ -44,15 +45,17 @@ const char *known_extensions[] = {
     "cl_khr_egl_image",
     "cl_khr_egl_event",
     "cl_khr_create_command_queue",
-    "cl_khr_priority_hints",
-    "cl_khr_throttle_hints",
-    "cl_khr_il_program",
-    "cl_khr_mipmap_image",
-    "cl_khr_mipmap_image_writes",
+    "cl_khr_image2d_from_buffer",
+    "cl_khr_dx9_media_sharing",
+    "cl_khr_d3d11_sharing",
+    "cl_khr_gl_depth_images",
+    "cl_khr_initialize_memory",
+    "cl_khr_terminate_context",
+    "cl_khr_gl_msaa_sharing",
 };
 
 size_t num_known_extensions = sizeof(known_extensions)/sizeof(char*);
-size_t first_API_extension = 11;
+size_t first_API_extension = 12;
 
 const char *known_embedded_extensions[] = {
     "cles_khr_int64",
--- a/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
+++ b/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
@@ -45,6 +45,7 @@ const char *known_extensions[] = {
     "cl_khr_egl_image",
     "cl_khr_egl_event",
     "cl_khr_create_command_queue",
+    "cl_khr_il_program",
     "cl_khr_image2d_from_buffer",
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_compiler_defines_for_extensions.cpp
```diff
@@ -44,15 +45,17 @@ const char *known_extensions[] = {
     "cl_khr_egl_image",
     "cl_khr_egl_event",
     "cl_khr_create_command_queue",
-    "cl_khr_priority_hints",
-    "cl_khr_throttle_hints",
-    "cl_khr_il_program",
```
alycm: `cl_khr_il_program` is a valid OpenCL 1.2 extension.

Jeremy-Kemp: Good spot, thanks. I've just commited it back in.

bashbaug:
Especially now, since we have a "unified" extensions spec, shouldn't we include "all" extensions in this list?  A few of them would be a little bit odd to support on an OpenCL 1.x device, but many others would make sense.

In particular, we enable priority hints and throttle hints for some of our OpenCL 1.x devices.

Jeremy-Kemp:
> Especially now, since we have a "unified" extensions spec, shouldn't we include "all" extensions in this list? A few of them would be a little bit odd to support on an OpenCL 1.x device, but many others would make sense.
> 
> In particular, we enable priority hints and throttle hints for some of our OpenCL 1.x devices.

My rationale here was to have the 1.2 conformance branch test for extensions that are defined in the 1.2 OpenCL specs. I haven't yet looked at the unified spec to see what, if any, changes there are regarding when and/or how extensions are supported at particular OpenCL versions. (Is [this](https://github.com/KhronosGroup/OpenCL-Docs) the unified document?)

Further to that, I used the fact that [Clang](https://github.com/llvm-mirror/clang/blob/release_80/include/clang/Basic/OpenCLExtensions.def) defines the minimum supported OpenCL version for each extension (admittedly, Clang has a non-exhaustive list of extensions at the moment, but a good chunk of them are present in the latest release) to inform how to fix this test.

I don't have a strong feeling either way with this one, I'm happy to list every possible khr here in the test, or keep it as a 1:1 mapping to the specific version of OpenCL that the conformance branch is testing against.

Jeremy-Kemp:
>So long as we don't remove the priority hint and throttle hint extensions I'm good.  smiley

In that case, I think we should go the whole hog and have every extension in the test, rather than a scatter-gun approach of adding extensions when required on an ad-hoc basis.

If people agree, I'll commit a change to add every khr from the unified spec.

kpet:
+1 on adding all extensions from the unified spec.

Jeremy-Kemp:
Ok, I've updated the test to include all extensions - thanks for the feedback! :)

Once it's approved, I'll create some more requests for the other branches so that we've got matching behavior.

bashbaug:
I believe that priority hints and throttle hints are API-only extensions, but otherwise LGTM as well.  Thanks!

Jeremy-Kemp:
Apologies for that, I'll move them back to the API-only section of the array. Did not notice in my testing as was using an implementation that does not expose those two extensions.

alycm:
Oh, I built and tested this locally.  It worked for me.

I would merge this, but I have the open question that now that this address the unified extension spec I think that the equivalent change could be made to the cl20_trunk, cl21_trunk and cl22_trunk branches?

kpet:
@Jeremy-Kemp Thanks for the changes. Agree with @alycm, could you prepare PRs for the other branches please?

Jeremy-Kemp:
@kpet what is the standard way for replicating changes onto other branches for CTS? Should I cherry-pick, merge or just create new commits?

kpet:
@Jeremy-Kemp I would start with a cherry-pick and amend where needed.

alycm:
Comment on https://github.com/KhronosGroup/OpenCL-CTS/pull/175#discussion_r275414924 where I realised that `cl_khr_spirv_no_integer_wrap_decoration` is missing from this list.

alycm:
I've retested this, it looks good!  I'll merge this and it's siblings.

PR #68, opened 2019-03-14
cl20: Convert images/clCopyImage to testHarness 

rszymanski:


rszymanski:
That was a wrong change, duplicated the earlier one

Issue #35, opened 2018-11-21
cl12 cl20 cl21 cl22 - User defined size while querying clGetDeviceInfo and CL_DEVICE_EXTENSIONS


gwawiork:
This can generate test errors when extension string is bigger than test expects (then CL_INVALID_VALUE will be returned). 
The following tests will be affected:
- all spir tests 
`char extensions[1024] = {0};
    size_t size;
    // Querying the device for its supported extensions
    cl_int errcode = clGetDeviceInfo(devId,
                                     CL_DEVICE_EXTENSIONS,
                                     sizeof(extensions),
                                     extensions,
                                     &size);`
- computeinfo
For example computeinfo tests defines config_data union where one of the filed is :
`char                            string[1024];`

alycm:
There are PRs open for OpenCL 2.0/2.1/2.2.  Is the patch expected to apply to OpenCL 1.2 as well?

kpet:
All PRs merged, closing.

kpet:
Note: this hasn't been fixed on cl12_trunk.

PR #36, opened 2018-11-30
Fix Issue #35 - ask for capability size. Too small size could cause errors

gwawiork:

PR #95, opened 2019-03-15
cl20: Fix Issue #35 - ask for capability size. Too small size could cause errors

gwawiork:
related with https://github.com/KhronosGroup/OpenCL-CTS/pull/36
Fix Issue #35

### DIFF
--- a/test_conformance/computeinfo/main.c
+++ b/test_conformance/computeinfo/main.c
@@ -21,7 +21,6 @@
 #include "../../test_common/harness/testHarness.h"
 #include "../../test_common/harness/kernelHelpers.h"
 
-
 static int dump_supported_formats;
 
 typedef struct
@@ -378,6 +377,7 @@ int getConfigInfo(cl_device_id device, config_info* info)
             if (err == CL_SUCCESS && config_size_set > 0) {
                 info->config.string = (char*)malloc(config_size_set);
                 err = clGetDeviceInfo(device, info->opcode, config_size_set, info->config.string, &config_size_ret);
+                size_err = config_size_set != config_size_ret;
             }
             break;
         case type_cl_device_svm_capabilities:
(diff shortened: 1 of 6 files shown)


test_conformance/conversions/test_conversions.cpp
```diff
@@ -782,13 +783,20 @@ static int InitCL( void )
...
+    }
+
+    std::vector<char> extensions(set_size);
+    if( (error = clGetDeviceInfo( gDevice, CL_DEVICE_EXTENSIONS, extensions.size(), extensions.data(),  NULL ) ) )
     {
         vlog_error( "FAILURE: unable to get device info for CL_DEVICE_EXTENSIONS!" );
         return -1;
     }
-    else if( strstr( extensions, "cl_khr_fp64" ) )
+    else if( strstr( extensions.data(), "cl_khr_fp64" ) )
```
kpet: `is_extension_available`? I don't think performance is a concern here.

test_conformance/conversions/test_conversions.cpp
```diff
@@ -800,7 +808,7 @@ static int InitCL( void )
     else if( strstr(profile, "EMBEDDED_PROFILE" ) )
     {
         gIsEmbedded = 1;
-        if( !strstr( extensions, "cles_khr_int64" ) )
+        if( !strstr( extensions.data(), "cles_khr_int64" ) )
```
kpet: Same here.

gwawiork:
Ok I will open against master again. 
Regrading cl12_trunk - I see many differences comparing to other branches (e.g conversions tests). Then I will skip this.

kpet:
Thanks. Yes cl12_trunk is a separate problem. We should try to port changes but only when the effort required is reasonable.

PR #96, opened 2019-03-15
cl21: Fix Issue #35 - ask for capability size. Too small size could cause errors

gwawiork:
Fix Issue https://github.com/KhronosGroup/OpenCL-CTS/issues/35


gwawiork:
Force pushed because of rebase needed and solving merge issues.

Issue #98, opened 2019-03-18
test_printf - wrong return value in verifyOutputBuffer


hewj03:
Function verifyOutputBuffer has two 'return false' statements which indicate mismatches. But this function should return non-zero value in case of mismatch.

kpet:
All PRs merged, closing.

PR #99, opened 2019-03-18
cl21: Fix Issue #98 - Fix verifyOutputBuffer return value in printf

hewj03:
This patch solves issue #98 by changing return value from false to 1 in function verifyOutputBuffer to indicate mismatches.


hewj03:
ok

kpet:
Thanks, merging all.

Issue #100, opened 2019-03-18
test_printf - parallel executed tests fail


hewj03:
test_printf use file /tmp/tmpfile or tmpfile to capture OpenCL kernel outputs, which are compared with correct results. If we run multiple printf tests in parallel, they fail because they write to the same file and one test could read another one's outputs.

PR #101, opened 2019-03-18
cl21: Use unique tmp file to allow parallel printf tests

hewj03:
This patch solves #100 by using a unique file for each test, so that parallel printf tests will not fail.

### DIFF
--- a/test_conformance/printf/test_printf.c
+++ b/test_conformance/printf/test_printf.c
@@ -28,6 +28,7 @@
 #define streamDup2(fd1,fd2) dup2(fd1,fd2)
 #endif
 #include <limits.h>
+#include <time.h>
 #include "test_printf.h"
 
 #if defined(_WIN32)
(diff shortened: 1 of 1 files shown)


hewj03:
the patch is rebased

gwawiork:
Have you run this test after rebase (MSVC 2017). I have got exception in function static void releaseOutputStream(int fd):

Unhandled exception at 0x00007FFAB5239E28 (ucrtbase.dll) in test_conformance_printf.exe: An invalid parameter was passed to a function that considers invalid parameters fatal.

gwawiork:
I think the issue is related with patches https://github.com/KhronosGroup/OpenCL-CTS/pull/162 @rszymanski are you able to verify this why on linux there is an exception ?

hewj03:
releaseOutputStream exception is a new issue, see #212

gwawiork:
ok then your patch looks ok. Could you send patches to other branches as well ?

hewj03:
Parallel tests could happen in these cases:
1. run tests on different OpenCL RT
2. one executed in offline mode with spirv usage another in online
3. different people launch same tests which are dispatched to the same machine.

hewj03:
this patch also solves the case when some other process happens to write to /tmp/tmpfile at the same time.

Issue #105, opened 2019-03-20
Compatibility tests - lost fixes for some versions of Visual Studio


gwawiork:
I see some lost fixes in files: 
compatibility\test_common\harness\compat.h
compatibility\test_common\harness\msvc9.c
Lack of this fixes cause erros while compilation on VS2015, VS2017

kpet:
All PRs merged, closing.

PR #106, opened 2019-03-20
cl20: Fix issue 105 - lost fixes for some version of Visual Studio

gwawiork:
Fix for issue https://github.com/KhronosGroup/OpenCL-CTS/issues/105

### DIFF
--- a/test_conformance/compatibility/test_common/harness/msvc9.c
+++ b/test_conformance/compatibility/test_common/harness/msvc9.c
@@ -565,7 +565,7 @@ long int lrintf (float x)
 //                  fenv functions
 //
 ///////////////////////////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


test_conformance/compatibility/test_common/harness/msvc9.c
```diff
@@ -565,7 +565,7 @@ long int lrintf (float x)
 //                  fenv functions
 //
 ///////////////////////////////////////////////////////////////////
-
+#if _MSC_VER < 1900
```
kpet: Should this test for `defined(_MSV_VER)` as well?

kpet:
We're not quite yet at the point where chasing warnings makes sense, merging.

Issue #131, opened 2019-03-21
cl21 cl22 - Improve CRC calculation


gwawiork:
Changes for CRC calculation when offline compiler files (spirv) are used.

kpet:
Both PRs merged, closing.

PR #132, opened 2019-03-21
cl21: Improved CRC calculation

gwawiork:
fix issue https://github.com/KhronosGroup/OpenCL-CTS/issues/131

### DIFF
--- a/test_common/harness/crc32.c
+++ b/test_common/harness/crc32.c
@@ -1,0 +1,104 @@
+/*-
+ *  COPYRIGHT (C) 1986 Gary S. Brown.  You may use this program, or
+ *  code or tables extracted from it, as desired without restriction.
+ *
+ *  First, the polynomial itself and its table of feedback terms.  The
+ *  polynomial is
+ *  X^32+X^26+X^23+X^22+X^16+X^12+X^11+X^10+X^8+X^7+X^5+X^4+X^2+X^1+X^0
+ *
+ *  Note that we take it "backwards" and put the highest-order term in
+ *  the lowest-order bit.  The X^32 term is "implied"; the LSB is the
+ *  X^31 term, etc.  The X^0 term (usually shown as "+1") results in
+ *  the MSB being 1
+ *
+ *  Note that the usual hardware shift register implementation, which
+ *  is what we're using (we're merely optimizing it by doing eight-bit
+ *  chunks at a time) shifts bits into the lowest-order term.  In our
+ *  implementation, that means shifting towards the right.  Why do we
+ *  do it this way?  Because the calculated CRC must be transmitted in
+ *  order from highest-order term to lowest-order term.  UARTs transmit
+ *  characters in order from LSB to MSB.  By storing the CRC this way
+ *  we hand it to the UART in the order low-byte to high-byte; the UART
+ *  sends each low-bit to hight-bit; and the result is transmission bit
+ *  by bit from highest- to lowest-order term without requiring any bit
+ *  shuffling on our part.  Reception works similarly
+ *
+ *  The feedback terms table consists of 256, 32-bit entries.  Notes
+ *
+ *      The table can be generated at runtime if desired; code to do so
+ *      is shown later.  It might not be obvious, but the feedback
+ *      terms simply represent the results of eight shift/xor opera
+ *      tions for all combinations of data and CRC register values
+ *
+ *      The values must be right-shifted by eight bits by the "updcrc
+ *      logic; the shift must be unsigned (bring in zeroes).  On some
+ *      hardware you could probably optimize the shift in assembler by
+ *      using byte-swap instructions
+ *      polynomial $edb88320
+ *
+ *
+ * CRC32 code derived from work by Gary S. Brown.
+ */
+
+#include "crc32.h"
+
+static uint32_t crc32_tab[] = {
+	0x00000000, 0x77073096, 0xee0e612c, 0x990951ba, 0x076dc419, 0x706af48f,
+	0xe963a535, 0x9e6495a3,	0x0edb8832, 0x79dcb8a4, 0xe0d5e91e, 0x97d2d988,
+	0x09b64c2b, 0x7eb17cbd, 0xe7b82d07, 0x90bf1d91, 0x1db71064, 0x6ab020f2,
+	0xf3b97148, 0x84be41de,	0x1adad47d, 0x6ddde4eb, 0xf4d4b551, 0x83d385c7,
+	0x136c9856, 0x646ba8c0, 0xfd62f97a, 0x8a65c9ec,	0x14015c4f, 0x63066cd9,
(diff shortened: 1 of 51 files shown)


kpet:
@StuartDBrady Could you take a look please?

test_common/harness/kernelHelpers.c
```diff
@@ -74,17 +75,10 @@ std::vector<char> get_file_content(const std::string &fileName)
 
 std::string get_kernel_name(const std::string &source)
 {
-    cl_uint crc = 0;
+    cl_uint crc = 0x04c11db7;
```
StuartDBrady: The value of `crc` here of `0x04c11db7` acts as a seed to the CRC32. IMO, it would be better not to have the seed, as the CRC would then be the same as produced by e.g. `jacksum -x -a fcs32`.

To remove the seed, the `crc` argument to `crc32()` should be removed, and the `crc` variable within `crc32` should be initialized with the value of `~0U`, exactly as per the example originally given in `libkern/crc32.c`:

* https://github.com/freebsd/freebsd/blob/master/sys/libkern/crc32.c#L122

I note that `0x04c11db7` is the normal representation of the polynomial for this CRC. However, the polynomial is applied through the use of `crc32_tab`, which has been generated from the polynomial. Altering the initial state of the function does not influence the polynomial that is used, but is instead equivalent to adding a prefix to the data for which the CRC is to be generated, which makes the value of `0x04c11db7` somewhat confusing. Any other non-zero value would suffice equally, although I would recommend, as stated above, not to use a seed value at all.

test_common/harness/kernelHelpers.c
```diff
@@ -74,17 +75,10 @@ std::vector<char> get_file_content(const std::string &fileName)
...
     // Count CRC
-    for (cl_uint i = 0; i < source.size() - source.size() % sizeof(cl_uint); i += sizeof(cl_uint))
-    {
-        cl_uint *ptr = (cl_uint *)&source[i];
-        crc += *ptr;
-    }
-    cl_uint remainder = 0;
-    memcpy(&remainder, &source[0] + source.size() - source.size() % sizeof(cl_uint), source.size() % sizeof(cl_uint));
-    crc += remainder;
+    crc = crc32(crc, &source[0], source.size());
```
StuartDBrady: I would recommend `source.data()` instead of `&source[0]`, for clarity.

test_common/harness/kernelHelpers.c
gwawiork: This is not my code personally, but I guess that this seed was a kind of protection against nulls numbers in input data. Do not you think that having this seed the results of crc calculations will be more 'randomized'?

StuartDBrady: > This is not my code personally, but I guess that this seed was a kind of protection against nulls numbers in input data. Do not you think that having this seed the results of crc calculations will be more 'randomized'?

It is true that the FCS-32 of the empty string is zero. However, I don't really see this as a problem. If anything, if the CTS were ever to create such a file, the hash of "00000000" would be more likely to stand out when looking at log output or a directory listing.

The "seed" is equivalent to prepending the data with a fixed, arbitrary sequence of bytes, and so offers no extra protection of any form. In this case, it happens to be equivalent to prepending the data with a byte sequence of: 0xd5, 0xe3, 0x1c, 0x09.

It seems unlikely that this behaviour was intentional.

gwawiork: I see that this change force us regenerate spirv files we are using now. But if there is no other way to have this PR accepted we will do this :)

test_common/harness/kernelHelpers.c
```diff
@@ -75,10 +75,10 @@ std::vector<char> get_file_content(const std::string &fileName)
 
 std::string get_kernel_name(const std::string &source)
 {
-    cl_uint crc = 0x04c11db7;
+    cl_uint crc = ~0U;
```
StuartDBrady: The code could be simplified slightly by moving this into `crc32`. Otherwise, looks good to me.

test_common/harness/kernelHelpers.c
StuartDBrady: Perhaps you can rename the SPIR-V files instead of regenerating them?

For example:
```
#!/bin/bash

find -name '*.cl' | while read FILE
do
    CHECKSUM=$(jacksum -x -a fcs32 -F '#CHECKSUM' "${FILE}")
    NEW_BASE_NAME=$(sed -r -e 's/^(.*)\.[^.]+\.([^.])+\.cl$/\1.'"${CHECKSUM}"'.\2/' <<< "${FILE}")
    for SUFFIX in bin bc bc32 bc64 spv spv32 spv64 options
    do
        if [[ -e "${FILE%.cl}.${SUFFIX}" ]]
        then
            mv -v "${FILE%.cl}.${SUFFIX}" "${NEW_BASE_NAME}.${SUFFIX}"
        fi
    done
    mv -v "${FILE}" "${NEW_BASE_NAME}.cl"
done
```

gwawiork: Nice :) thank you very much

test_common/harness/kernelHelpers.c
StuartDBrady: Note that in `crc32()`, we do `crc = crc ^ ~0U;`. Propagation of the value of `~0U` here into `crc32()` would give `crc = ~0U ^ ~0U`, i.e. `crc = 0`, which is incorrect, so this line should be removed.

test_common/harness/kernelHelpers.c
```diff
@@ -74,17 +75,10 @@ std::vector<char> get_file_content(const std::string &fileName)
...
     // Count CRC
-    for (cl_uint i = 0; i < source.size() - source.size() % sizeof(cl_uint); i += sizeof(cl_uint))
-    {
-        cl_uint *ptr = (cl_uint *)&source[i];
-        crc += *ptr;
-    }
-    cl_uint remainder = 0;
-    memcpy(&remainder, &source[0] + source.size() - source.size() % sizeof(cl_uint), source.size() % sizeof(cl_uint));
-    crc += remainder;
+    crc = crc32(crc, source.data(), source.size());
```
StuartDBrady: `cl_uint crc = crc32(source.data(), source.size());`

test_common/harness/crc32.c
```diff
@@ -0,0 +1,104 @@
...
+	0xb40bbe37, 0xc30c8ea1, 0x5a05df1b, 0x2d02ef8d
+};
+
+uint32_t
+crc32(uint32_t crc, const void *buf, size_t size)
+{
+	const uint8_t *p;
+
+	p = (const uint8_t*)buf;
+	crc = crc ^ ~0U;
```
StuartDBrady: `uint32_t crc = ~0U;`

test_common/harness/kernelHelpers.c
gwawiork: Fixed

test_common/harness/crc32.c
gwawiork: fixed

PR #133, opened 2019-03-21
cl22: Improved CRC calculation

gwawiork:
fix issue https://github.com/KhronosGroup/OpenCL-CTS/issues/131

### DIFF
--- a/test_common/harness/crc32.c
+++ b/test_common/harness/crc32.c
@@ -1,0 +1,104 @@
+/*-
+ *  COPYRIGHT (C) 1986 Gary S. Brown.  You may use this program, or
+ *  code or tables extracted from it, as desired without restriction.
+ *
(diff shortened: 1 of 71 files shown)


StuartDBrady:
> This crc calculation is executed when offline spirv compilation is done (ASCII kernel source is converted to SPIR-V binary format). Spirv offline compilation is supported since OCL2.1 . Then there is no probably need to add this feature to compatibility tests. It depends what we understand compatibility with previous version or with cl12 version? I am not sure.

The CRC is also needed for "binary" offline compilation, which does not make use of SPIR-V.

The issue here is that if it would appear as though linking of the tests that list kernelHelpers.c but not crc32.c would fail. Can you explain why that is not the case?

In any case, the hash related aspects of this seem fine to me, but I think this may require someone more familiar with the build-related aspects of the CTS to take a look.

Issue #136, opened 2019-03-26
cl20 cl21 cl22: Lost fix - Proposed fix for bug 16239


gwawiork:
Looks like fix for test_half in cl_utils.c proposed by b-sumner was lost

kpet:
All PRs merged, closing.

PR #139, opened 2019-03-26
cl20: Proposed fix for bug 16239

gwawiork:
fix issue: https://github.com/KhronosGroup/OpenCL-CTS/issues/136


kpet:
@b-sumner I think this is blocked on you signing the CLA. Could you have a look please?

Issue #135, opened 2019-03-26
math_brute_force floating-point accuracy error in test_spir 


hewj03:
[Spir test](https://github.com/KhronosGroup/OpenCL-CTS/tree/cl22_trunk/test_conformance/spir) contains some math_brute_force kernels. The test compares SPIR and OpenCL C results bit-by-bit. However, for math_brute_force kernels, specification does not ask for SPIR and OpenCL C path to match bit-to-bit. 
[math_brute_force test](https://github.com/KhronosGroup/OpenCL-CTS/tree/cl22_trunk/test_conformance/math_brute_force) uses ULPs limits to tolerate small mismatches between OpenCL and reference results. Could we apply these ULPs limits to math_brute_force kernels in Spir test as well?

alycm:
Hello, you've opened the MR against 2.1.  This could also be applied to 2.0 and 1.2.   I can test a 1.2 MR if that helps.

alycm:
Thank for the additional branches, I have reviewed https://github.com/KhronosGroup/OpenCL-CTS/pull/152.

hewj03:
Thanks for review. I've submitted a MR against 2.2 as well.

alycm:
All four versions are merged, closing this issue.

PR #140, opened 2019-03-27
cl21: Reuse math_brute_force ulp threshold in spir test

hewj03:
Spir test compares floating-point kernel results bit-by-bit with
correct results. However, for math_brute_force kernels, specification
does not ask for SPIR and OpenCL C path to match bit-to-bit. This patch
reuses ulps threshold from math_brute_force folder for math_brute_force
kernels in spir test. See issue #135

### DIFF
--- a/test_conformance/math_brute_force/FunctionList.c
+++ b/test_conformance/math_brute_force/FunctionList.c
@@ -24,6 +24,36 @@
 
 #define STRINGIFY( _s)                  #_s
 
+// Only use ulps information in spir test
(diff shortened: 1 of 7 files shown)


kpet:
@hewj03 This now needs rebasing. Would you mind?

hewj03:
Fix MSVC 2017 link error and rebased

hewj03:
fixed line ending

gwawiork:
ok looks good

PR #141, opened 2019-03-29
cl12: Bugfix for clGetKernelArgInfo Tests

gmtra:
**Bug:** The expected `CL_KERNEL_ARG_TYPE_QUALIFIER` for some `*const_*_d` tests are incorrectly specified as `CL_KERNEL_ARG_TYPE_NONE`.

**Fix:** Changed the expected qualifier to `CL_KERNEL_ARG_TYPE_CONST` for those cases.

The fix is also applicable for cl20_trunk, cl21_trunk and cl22_trunk.

### DIFF
--- a/test_conformance/api/test_kernel_arg_info.c
+++ b/test_conformance/api/test_kernel_arg_info.c
@@ -1884,16 +1884,16 @@ const char * required_arg_info[][72] = {
     },
     {
         "const_scalar_d",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "char", "constchard",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "uchar", "constuchard",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "uchar", "constunsignedchard",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "short", "constshortd",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "ushort", "constushortd",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "ushort", "constunsignedshortd",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "int", "constintd",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "uint", "constuintd",
-        (const char *)CL_KERNEL_ARG_ADDRESS_PRIVATE, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_NONE), "uint", "constunsignedintd",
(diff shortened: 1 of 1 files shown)


alycm:
I'm not sure that this proposed change is correct, but me saying so relies on examples in the spec rather than clear wording.

To summarise what this proposed change does, it affects the `clGetKernelArgInfo` `CL_KERNEL_ARG_TYPE_QUALIFIER` results like in this example:

| Kernel Argument          | Current CTS                                              | CTS with this change                               |
| ------------------------------|-----------------------------------------------------|-----------------------------------------------------|
|  `char x`                          | `CL_KERNEL_ARG_TYPE_NONE`    | `CL_KERNEL_ARG_TYPE_NONE`   |
|  `const char x`               | `CL_KERNEL_ARG_TYPE_NONE`    | `CL_KERNEL_ARG_TYPE_CONST` |
|  `global const char *x` | `CL_KERNEL_ARG_TYPE_CONST` | `CL_KERNEL_ARG_TYPE_CONST` |

I.e. non-const values are always "none", const pointers are always "const", but this changes const values from "none" to "const".

The existing CTS behaviour actually makes sense to me, because from the perspective of the API (roughly equivalent to the caller) whether or not values are const makes no difference (but of course it does for pointers).

Footnote 18 of the 2.2 spec (or the `CL_KERNEL_ARG_TYPE_QUALIFIER` table entry of the 1.2 spec) says:

> CL_KERNEL_ARG_TYPE_VOLATILE is returned for CL_KERNEL_ARG_TYPE_QUALIFIER if the argument is a pointer and the referenced type is declared with the volatile qualifier. For example, a kernel argument declared as global int volatile *x returns CL_KERNEL_ARG_TYPE_VOLATILE but a kernel argument declared as global int * volatile x does not. Similarly, CL_KERNEL_ARG_TYPE_CONST is returned if the argument is a pointer and the referenced type is declared with the restrict or const qualifier. For example, a kernel argument declared as global int const *x returns CL_KERNEL_ARG_TYPE_CONST but a kernel argument declared as global int * const x does not. CL_KERNEL_ARG_TYPE_RESTRICT will be returned if the pointer type is marked restrict. For example, global int * restrict x returns CL_KERNEL_ARG_TYPE_RESTRICT.

Confusingly footnote 18 is attached to `CL_KERNEL_ARG_TYPE_VOLATILE` but clearly also discusses `CL_KERNEL_ARG_TYPE_CONST`.  I think that this part in particular agrees with my interpretation:

> For example, a kernel argument declared as global int const *x returns CL_KERNEL_ARG_TYPE_CONST but a kernel argument declared as global int * const x does not.

If you agree with my interpretation maybe the correct next step is to try to make the spec clearer?

gmtra:
Thank you for considering my pull request Alastair. 

My reasoning for this change goes as follows:

1. The OpenCL 1.2 specification states on page 169: "If the argument is declared with the constant address space qualifier, the `CL_KERNEL_ARG_TYPE_CONST` type qualifier will be set.". My interpretation of this sentence is that regardless of whether the argument is a const value or pointer to const, if the const qualifier is specified, `CL_KERNEL_ARG_TYPE_CONST` is set. 

2. The footnote on page 169 of the OpenCL 1.2 specification in my opinion deals with the conceptual difference between the use of const * (pointer to constant) and the *const (constant pointer) declarator. Taking the example in the footnote, for `global int const *x` it implies that it is a pointer to a constant and hence must be labeled `CL_KERNEL_ARG_TYPE_CONST` because `x[2] = 1` within the kernel is not legal in this case. For `global int *const x` it implies a constant pointer and must not be labeled `CL_KERNEL_ARG_TYPE_CONST` since `x[2] = 1` is legal in this case. 

3. The footnote does not specify that const values should not be marked `CL_KERNEL_ARG_TYPE_CONST` even though the specification of const makes no difference here. Point 1 seems to cover this case.

alycm:
Hi Guarav, 

Regarding your point 1. I've added a fourth row to my table for the `constant` address space case.  Essentially your quote is talking about pointers to the `constant` address space implicitly having the `const` qualifier, but the tests you are modifying are values in the `private` address space.

| Kernel Argument          | Current CTS                                              | CTS with this change                               |
| ------------------------------|-----------------------------------------------------|-----------------------------------------------------|
|  `char x`                          | `CL_KERNEL_ARG_TYPE_NONE`    | `CL_KERNEL_ARG_TYPE_NONE`   |
|  `const char x`               | `CL_KERNEL_ARG_TYPE_NONE`    | `CL_KERNEL_ARG_TYPE_CONST` |
|  `global const char *x` | `CL_KERNEL_ARG_TYPE_CONST` | `CL_KERNEL_ARG_TYPE_CONST` |
|  `constant char *x`        | `CL_KERNEL_ARG_TYPE_CONST` | `CL_KERNEL_ARG_TYPE_CONST` |

Regarding your point 2: Yes, I agree.  Just to be completely clear, although `global int const *x` is pointing at constant data as you state, it is not pointing at data in the `constant` address space.

Regarding point 3: I don't  agree with your interpretation.   But I am extrapolating from a single example in the footnote where a const value that is a non-const pointer "does not [return CL_KERNEL_ARG_TYPE_CONST]" to mean that all const values that are not a const-pointer (and may not be a pointer at all) should have the same behavior.

Even if I am correct the fact that I am "extrapolating from a single example in the footnote" is a clear sign that the spec could say more about this.

gmtra:
Thanks again for the explanation Alastair. That has cleared things up for me and I agree with you. 

Would be good to point out clearly in the specification that all const values that are not a const-pointer and may not be a pointer at all should have the arg type qualifier CL_KERNEL_ARG_TYPE_NONE. 

I'll close this pull request.

PR #144, opened 2019-04-05
[CL20] Fix data race in Device execution test suite

alexbatashev:
The following kernels ran into data races:
- enqueue_flags_wait_work_group_simple
- enqueue_flags_wait_work_group_event
- enqueue_flags_wait_work_group_local
- enqueue_flags_wait_work_group_event_local
Fix KhronosGroup#40


test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -39,6 +39,8 @@ static const char* enqueue_flags_wait_kernel_simple[] =
     NL, "  size_t lid = get_local_id(0);"
     NL, "  size_t tid = get_global_id(0);"
     NL, ""
+    NL, "  array[index * gs + tid] = array[(index - 1) * gs + tid] + 1;"
```
b-sumner: Why was this moved?  The whole point of this test is to ensure the child waits for the parent.

test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -50,11 +52,9 @@ static const char* enqueue_flags_wait_kernel_simple[] =
     NL, "    }"
     NL, "  }"
     NL, ""
-    NL, "  array[index * gs + tid] = array[(index - 1) * gs + tid] + 1;"
```
b-sumner: This location is correct.

test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -82,7 +82,6 @@ static const char* enqueue_flags_wait_kernel_simple[] =
     NL, "  size_t gs  = get_global_size(0);"
     NL, "  size_t tid  = get_global_id(0);"
     NL, ""
-    NL, "  res[tid] = 0;"
     NL, "  array[tid] = tid;"
```
b-sumner: It would actually be better to move this after the enqueue.

alexbatashev: This will not resolve the data race condition.

test_conformance/device_execution/enqueue_flags.cpp
alexbatashev: done

test_conformance/device_execution/enqueue_flags.cpp
alexbatashev: done

test_conformance/device_execution/enqueue_flags.cpp
b-sumner: Agreed, but it would improve the test.

alexbatashev: Ok, done

test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -185,7 +183,7 @@ static const char* enqueue_flags_wait_kernel_local[] =
     NL, "  size_t tid = get_global_id(0);"
     NL, ""
     NL, "  sub_array[lid] = array[(index - 1) * gs + tid];"
-    NL, "  barrier(CLK_LOCAL_MEM_FENCE);"
+    NL, "  barrier(CLK_GLOBAL_MEM_FENCE);"
```
b-sumner: The local fence is needed here so all threads in the loop below see all the updates

test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -236,7 +234,6 @@ static const char* enqueue_flags_wait_kernel_local[] =
     NL, "  size_t gs  = get_global_size(0);"
     NL, "  size_t tid  = get_global_id(0);"
     NL, ""
-    NL, "  res[tid] = 0;"
     NL, "  array[tid] = tid;"
```
b-sumner: It would improve the test to move this after the enqueue

test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -262,7 +259,7 @@ static const char* enqueue_flags_wait_kernel_event_local[] =
     NL, "  size_t tid = get_global_id(0);"
     NL, ""
     NL, "  sub_array[lid] = array[(index - 1) * gs + tid];"
-    NL, "  barrier(CLK_LOCAL_MEM_FENCE);"
+    NL, "  barrier(CLK_GLOBAL_MEM_FENCE);"
```
b-sumner: The local fence is required here.

test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -499,7 +495,7 @@ static const char* enqueue_flags_wait_work_group_local[] =
     NL, "  size_t gs = get_global_size(0);"
     NL, ""
     NL, "  sub_array[lid] = array[(index - 1) * gs + tid];"
-    NL, "  barrier(CLK_LOCAL_MEM_FENCE);"
+    NL, "  barrier(CLK_GLOBAL_MEM_FENCE);"
```
b-sumner: The local fence is required here.

test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -580,7 +575,7 @@ static const char* enqueue_flags_wait_work_group_event_local[] =
...
```
b-sumner: The local fence is required here

test_conformance/device_execution/enqueue_flags.cpp
alexbatashev: Ah, sorry. It's been a while, I completely forgot there were working barriers

alexbatashev:
@b-sumner do you have any complaints about latest revision?

alexbatashev:
@kpet do these changes make sense to you?

test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -676,7 +672,7 @@ int test_enqueue_flags(cl_device_id device, cl_context context, cl_command_queue
     cl_uint i;
     cl_int err_ret, res = 0;
     clCommandQueueWrapper dev_queue;
-    cl_int kernel_results[MAX_GWS] = { -1 };
+    cl_int kernel_results[MAX_GWS] = { 0 };
```
kpet: Why did you change that line? My understanding of the validation logic below is that both 0 and 1 are expected values in the output. Isn't it better to initialise the results array with a value that can never be valid? I'm afraid this might mask failures.

alexbatashev: The value is being overwritten in kernel anyway. But the way some kernels re-initialize res array causes data races. So, after some discussions here and there I decided to change this value to 0. This should not seriously affect the test. But definitely fixes data races.

kpet: Not sure I'm following. Don't we want to guarantee that the entirety of the array is filled by the kernels? It seems to me that we may be hiding issues in the kernels with this change.

>This should not seriously affect the test. But definitely fixes data races.

I can't reconcile these two statements :).

alexbatashev: The original issue #40 contains detailed explanations with pictures https://github.com/KhronosGroup/OpenCL-CTS/issues/40#issuecomment-456696470 

This is the approach we agreed on with @b-sumner offline. I'll add you to the mail thread.

PR #149, opened 2019-04-10
cl20: Khronos Bug 16236: Support CL_DEPTH images in the image helpers…

kpet:
… when using the border colour


kpet:
Already present in other branches. Merging.

Issue #112, opened 2019-03-20
cl12 cl21 cl22 - Lost fix - Add test for -cl-fp32-correctly-rounded-divide-sqrt to spir subtest.


gwawiork:
Looks like fix in spir tests is missed in opensource and present in internal gitlab

alycm:
Do you happen to know if this fix is also missing from CL 1.2 and 2.0?

gwawiork:
cl20_trunk contains these commits. cl12_trunk not - I send PR to fix this (a little bit different than cl21 and cl22 because math_brute_force.zip was already up to date)

alycm:
Thank for the clarification about 2.0, and for creating the 1.2 PR.

gwawiork:
After review changes needed in cl20 as well.

alycm:
All four versions merged, so closing this issue.

PR #151, opened 2019-04-10
cl12: Add test for -cl-fp32-correctly-rounded-divide-sqrt to spir subtest

gwawiork:
Fix issue #112


test_conformance/spir/run_build_test.cpp
```diff
@@ -366,15 +366,33 @@ bool TestRunner::runBuildTest(cl_device_id device, const char *folder,
...
+    cl_device_fp_config gFloatCapabilities = 0;
+    cl_int err;
+    if ((err = clGetDeviceInfo(device, CL_DEVICE_SINGLE_FP_CONFIG, sizeof(gFloatCapabilities), &gFloatCapabilities, NULL)))
+    {
+      log_info("Unable to get device CL_DEVICE_SINGLE_FP_CONFIG. (%d)\n", err);
+    }
+
+    if (strstr(test_name, "div_cr") || strstr(test_name, "sqrt_cr")) {
+      if ((gFloatCapabilities & CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT) == 0)
+        return true;
```
alycm: This `return true` means that devices that do return `CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT` end up failing conformance_test_spir.   I tried on such a device and got:

```
PASSED 559 of 571 tests.
```

This is because the 12 _cr tests never ran and thus never passed, but they are in the list of tests, so the number of tests that passed in lower than the total number of tests.

The way this is handled for other properties, such as images, is by having a column in this CSV file:
* https://github.com/KhronosGroup/OpenCL-CTS/blob/cl12_trunk/test_conformance/spir/khr.csv

Having an `isFooRequired` function something like this one:
* https://github.com/KhronosGroup/OpenCL-CTS/blob/cl12_trunk/test_conformance/spir/run_services.cpp#L589

And then skipping the test in the TestRunner:
* https://github.com/KhronosGroup/OpenCL-CTS/blob/cl12_trunk/test_conformance/spir/run_build_test.cpp#L326
* https://github.com/KhronosGroup/OpenCL-CTS/blob/cl12_trunk/test_conformance/spir/run_build_test.cpp#L333

My links are for `cl12_trunk`, but I'm pretty sure this code is identical across all branches.  I'll also admit that the current approach seems like overkill, but fixing that seems a much much bigger task than simply adding these tests.

alycm: Actually, making all the above changes would be far more work than the value provided.

If I replace the lines commented on here with this instead:

```c
      if ((gFloatCapabilities & CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT) == 0) {
        (*m_successHandler)(test_name, "");
        std::cout << "Skipped. (Cannot run on device due to missing "
                     "CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT property.)"
                  << std::endl;
        return true;
      } else {
```

then I get

```
PASSED 571 of 571 tests.
```

gwawiork: I see this is general way of handling not supported cases in this test. Then i will fix it in this way.

alycm: Thanks!  I've retested this, it works for me in a configuration where `CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT` is not reported.

alycm:
@gwawiork This just needs CLA signing from your colleagues now.

gwawiork:
@yxsamliu Could you sing CLA agreement? This PR includes commit that you are created some time ago in Khronos internal repository.

PR #152, opened 2019-04-11
cl12: Reuse math_brute_force ulp threshold in spir test

hewj03:
Spir test compares floating-point kernel results bit-by-bit with
correct results. However, for math_brute_force kernels, specification
does not ask for SPIR and OpenCL C path to match bit-to-bit. This patch
reuses ulps threshold from math_brute_force folder for math_brute_force
kernels in spir test. See issue #135

### DIFF
--- a/test_conformance/math_brute_force/FunctionList.c
+++ b/test_conformance/math_brute_force/FunctionList.c
@@ -22,6 +22,34 @@
 
 #define STRINGIFY( _s)                  #_s
 
+// Only use ulps information in spir test
+#ifdef FUNCTION_LIST_ULPS_ONLY
+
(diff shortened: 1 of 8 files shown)


hewj03:
> I could definitely see differences in compile paths cause small but legal differences in results.
We have seen small differences when SPIR and OpenCL C paths have difference choices of whether to vectorize.

hewj03:
Thanks, I've fixed MSVC 2017 link error and rebased to fix conflict

hewj03:
fixed line ending

gwawiork:
ok looks good

alycm:
A few people have now tried this out.  I'm going to accept this and it's siblings.

PR #172, opened 2019-04-12
cl21: Check device address width when deciding which SPIR-V binary to read

aarongreig:
Previously this was determined by checking `sizeof(void *)` but this causes issues when the device address width differs from the host.

### DIFF
--- a/test_conformance/spirv_new/main.cpp
+++ b/test_conformance/spirv_new/main.cpp
@@ -30,7 +30,7 @@ const std::string slash = "/";
 #endif
 
 const std::string spvExt = ".spv";
-const std::string addrWidth = (sizeof(void *) == 4) ? "32" : "64";
+std::string gAddrWidth = "";
 
 std::vector<unsigned char> readBinary(const char *file_name)
 {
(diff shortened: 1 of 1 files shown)


aarongreig:
I did try putting `clGetDeviceInfo()` in `main()` and then storing to the global `addrWidth` string but you need a device ID to call it, which isn't queried until inside `runTestHarnessWithCheck`. Otherwise this was the solution I arrived at that touched the fewest files.

kpet:
Have you tried moving from `runTestHarness` to `runTestHarnessWithCheck` and do it in a  `DeviceCheckFn`? You can plug a custom check function there that is called with the device.

aarongreig:
updated to use a device check

test_conformance/spirv_new/main.cpp
```diff
@@ -30,7 +30,7 @@ const std::string slash = "/";
 #endif
 
 const std::string spvExt = ".spv";
-const std::string addrWidth = (sizeof(void *) == 4) ? "32" : "64";
+std::string addrWidth = "";
```
kpet: Suggest renaming it to `gAddrWidth` to make it clear that it's a global.

kpet:
Thanks for the changes!

PR #177, opened 2019-04-15
cl21: Fix std::vector alignment segfaults on 32-bit

aarongreig:
On 32-bit Release builds assignment to `std::vector<cl_double2>` was
segfaulting due to the underlying storage having incorrect alignment, the
fix is to specialize `std::vector` with a custom allocator
`align_allocator` which uses `align_malloc` and `align_free` for
allocating the underlying storage.

### DIFF
--- a/test_conformance/spirv_new/test_decorate.cpp
+++ b/test_conformance/spirv_new/test_decorate.cpp
@@ -201,8 +201,8 @@ int verify_saturated_results(cl_device_id deviceID,
     size_t in_bytes = sizeof(Ti) * num;
     size_t out_bytes = sizeof(To) * num;
 
-    std::vector<Ti> h_lhs(num);
-    std::vector<Ti> h_rhs(num);
+    std::vector<Ti, align_allocator<Ti>> h_lhs(num);
+    std::vector<Ti, align_allocator<Ti>> h_rhs(num);
 
     To loVal = std::numeric_limits<To>::min();
     To hiVal = std::numeric_limits<To>::max();
@@ -249,7 +249,7 @@ int verify_saturated_results(cl_device_id deviceID,
     err = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, &global, NULL, 0, NULL, NULL);
     SPIRV_CHECK_ERROR(err, "Failed to enqueue cl kernel");
 
-    std::vector<To> h_res(num);
+    std::vector<To, align_allocator<To>> h_res(num);
     err = clEnqueueReadBuffer(queue, res, CL_TRUE, 0, out_bytes, &h_res[0], 0, NULL, NULL);
     SPIRV_CHECK_ERROR(err, "Failed to read to output");
 
@@ -318,7 +318,7 @@ int test_image_decorate(cl_device_id deviceID,
     const int width = 4096;
     const int height = 4096;
 
-    std::vector<cl_uint4> src(width * height);
+    std::vector<cl_uint4, align_allocator<cl_uint4>> src(width * height);
     RandomSeed seed(gRandomSeed);
 
     for (auto &val : src) {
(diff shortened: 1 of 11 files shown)


aarongreig:
> Not sure I'm following. What is requiring the alignment? Why does the issue show up only in 32-bit builds?

It's unclear exactly what's going on since it only happens when building a release configuration and the crash occurs in STL code, but essentially we have found that 32 bit release builds (tested on Ubuntu 16.04) have a tendency to segfault when manipulating `std::vector`s of the cl vector types. For instance while I was investigating this I saw intermittent segfaults in the vector initializer [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/a6fdb7ce6066e7b9f442e61283c06b1ccc16eea0/test_conformance/spirv_new/test_op_undef.cpp#L50) when the test was running for `cl_int3` or `cl_int4`. A quick reproducer is to insert the following lines at the top of `main()` in `spirv_new/main.cpp`:
```
std::vector<cl_int3> a(1024);
std::vector<cl_int4> b(1024);
```
I get an immediate segfault on every run with that change. Applying this patch and instead declaring those 
vectors with an `align_allocator` prevents the segfault and allows tests to run normally. Note that running a 32 bit build on a 64 bit system doesn't reproduce this, I had to use a VM with a 32 bit OS installed.

kpet:
Hmm, sounds like a fun issue. Any sensitivity to the OS / toolchain? Have you tried different versions? I'm worried that we don't really understand what's happening and whether this change is genuinely fixing the issue or whether it just "makes it go away".

aarongreig:
I have reproduced with GCC versions 5.4 and 7.4 and clang versions 3.8 and 6.0 (although clang only segfaults in the op undef test, I wasn't able to coerce it into crashing in main). I was not able to reproduce on Windows with the Microsoft toolchain.

Looking at the assembly where it crashes it seems the cause is the same in binaries from either compiler: a `movaps` instruction attempting to load from an address that isn't aligned to 16, 32 or 64 bytes, which isn't allowed according to the `movaps` spec I found [here](https://www.felixcloutier.com/x86/movaps).
The offending instruction, from a gdb `disassemble` immediately after the segfault happened in a binary generated by clang 3.8:
`=> 0x0808be80 <+112>:   movaps %xmm0,(%esi,%ebx,1)`
and the relevant registers:
```
ebx            0x0      0
esi            0x8331cb8        137567416
```
Note that `esi` doesn't align to 16. I guess it might be a compiler issue if it's generating instructions that are operating on memory that isn't correctly aligned for them? But either way it seems to be an issue in existing versions of popular compilers that can be fixed by ensuring we allocate memory for vector types with appropriate alignment.

kpet:
Thanks for the thorough investigation. I suspected this would be the problem, it typically is on x86 when aligning solves the crash.

The last question to answer from my PoV before we can agree that this change is the way to go is whether we can fix it in the OpenCL headers for all applications instead of working around the issue in conformance only. I'm guessing others must have seen that as well.

aarongreig:
At a very cursory level I did change the declaration of `cl_int4` in `cl_platform.h` to 
` typedef cl_int      __cl_int4       __attribute__((vector_size(16), aligned(16)));`
and this didn't seem to solve the problem, although I haven't had a chance to look into this with much rigor yet. I'll give it a more serious shot and update with my findings.

aarongreig:
adding `aligned(16)` to the header does fix the issue for gcc, but not for the clang versions I mentioned above

kpet:
Ok, I think we probably want to do that then. Have you tested latest clang (8 or trunk) to see if it's worth reporting a bug upstream?

aarongreig:
Tested with clang 8 and it still repros with and without `aligned` in the typedef, I'm putting together a bug report. 
>Ok, I think we probably want to do that then.

I'll open a PR to fix the headers, but I still think it's worth applying this fix here to fix clang builds.

aarongreig:
I think I messed up the testing somewhere, now that I'm trying to write a minimal reproducer for the header PR I can't get `aligned(16)` to fix the issue.

aarongreig:
It seems that the `cl_int4` definition (and the others that make use of `vector_size(16)`) result in types that are over aligned for 32-bit. The following reduced example contains the definition of `cl_int4`, paraphrased from cl_platform.h:
```
#include <type_traits>
#include <cstddef>

typedef int _int4 __attribute__((vector_size(16)));

typedef union{
  int data[4];
  struct {int x, y, z, w;};
  _int4 vec;
} int4;

int main(int, char**) {
static_assert(std::alignment_of<int4>::value <= alignof(max_align_t), "over aligned!");
} 
```
The assert fails. A colleague put me on to the fact that `new` (and by extension a default `std::allocator`) [isn't guaranteed to allocate correctly aligned memory for over aligned types in C++11.](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0035r1.html)

PR #181, opened 2019-04-15
Don't require image support on GPU devices

kpet:
There is no mention of this rule that I could find in the
OpenCL 1.2 specification. cl12_trunk was the only branch
with this behaviour.


kpet:
Thanks for the review.

PR #213, opened 2019-04-22
cl12: Use unique tmp file to allow parallel printf tests

hewj03:
This patch solves #100 by using a unique file for each test, so that parallel printf tests will not fail.


hewj03:
thanks for test on this branch

PR #228, opened 2019-04-25
cl12: Ignore C++11 narrowing errors for now

### LINKED ISSUE #222: cl12: Compiling with Clang-8 requires -Wno-c++11-narrowing
Currently compiling the `cl12_trunk` branch with recent versions of Clang requires the `-Wno-c++11-narrowing` flag, effectively due to the way older C code and C++ code are mixed.   However, the CMake does not set this itself.

Locally when we import the CTS into our project we have something like this in our own CMake:

```
  if(CMAKE_CXX_COMPILER_ID STREQUAL Clang OR
      CMAKE_CXX_COMPILER_ID STREQUAL AppleClang)
    string(APPEND CMAKE_CXX_FLAGS " -Wno-c++11-narrowing")
  endif()
```

pierremoreau:
Fixes #222

### DIFF
--- a/test_conformance/CMakeLists.txt
+++ b/test_conformance/CMakeLists.txt
@@ -28,6 +28,11 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CTS_EXTRA_FLAGS}")
 endif()
 
+# Clang gives C++11 narrowing warnings so suppress these for now
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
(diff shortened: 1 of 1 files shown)


test_conformance/CMakeLists.txt
```diff
@@ -28,6 +28,11 @@ else()
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CTS_EXTRA_FLAGS}")
 endif()
 
+# Clang gives C++11 narrowing warnings so suppress these for now
+if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
```
kpet: Recent versions of GCC also have the issue.

PR #229, opened 2019-04-25
cl12: Do not mistake x86_64 platforms as x86

pierremoreau:
This resulted in “unknown target CPU 'i686'” error messages on my laptop, as `CMAKE_SYSTEM_PROCESSOR` was `x86_64` which matches the `x86` string.

### DIFF
--- a/test_conformance/conversions/CMakeLists.txt
+++ b/test_conformance/conversions/CMakeLists.txt
@@ -29,7 +29,7 @@ set_source_files_properties(
         ../../test_common/harness/errorHelpers.c
         PROPERTIES LANGUAGE CXX)
 
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)")
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "(x86$)|(X86$)")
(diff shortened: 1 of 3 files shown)


alycm:
Surely "we" don't set `CMAKE_CL_64`, but CMake does.  From the docs:

> Set to a true value when using a Microsoft Visual Studio cl compiler that targets a 64-bit architecture.

And these are all non-MSVC paths.

kpet:
Oh, I hadn't realised this was a standard option. Depending on the CMake version, it seems not very clearly defined and recent docs discourage its use. It makes me wonder why we test for non-MSVC as well since recent CMake docs specify that the flag is defined when compiling with MSVC for a 64-bit target. I think I'm fine with the change then but this highlights more build system oddities.

Would welcome another opinion before merging.

pierremoreau:
Thanks @kpet. Is there any public issue/document tracking what’s needed/ongoing regarding merging everything into one branch? I will try to give a hand over the summer.

kpet:
@pierremoreau Thanks for offering :). There is no public document at the moment. I'll try to write something down in the next few days.

pierremoreau:
Thanks for opening #259 and moving cl22_trunk to master!

PR #235, opened 2019-04-26
cl20: CMake improvements

pierremoreau:


kpet:
This doesn't build on macOS. You may want to have a look at the bots.

pierremoreau:
I’m very confused on why it is failing now: this is the faulty code
```
if(CMAKE_COMPILER_IS_GNUCC OR "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
    # -msse -mfpmath=sse to force gcc to use sse for float math,
    # avoiding excess precision problems that cause tests like int2float
    # to falsely fail. -ffloat-store also works, but WG suggested
    # that sse would be better.
    if(CMAKE_ARM_COMPILER OR ANDROID)
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -std=gnu99")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g")
    else(CMAKE_ARM_COMPILER OR ANDROID)
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -std=gnu99 -msse -mfpmath=sse")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -msse -mfpmath=sse")
    endif(CMAKE_ARM_COMPILER OR ANDROID)
else()
    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /D__SSE__")
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /D__SSE__")
endif()
```
which I did not change, and I don’t see how it could work on macOS before.

PR #241, opened 2019-04-29
cl20: Reuse test harness code in kernel_read_write

rszymanski:
Some of the setup functionality is already there in the test harness, so
use that and remove the duplicated code from within the suite.


kpet:
@rszymanski There seems to be a build issue on macOS. Could you take a look please?

rszymanski:
@kpet I know, just haven't had time to fix it.

Issue #250, opened 2019-04-30
test_compiler/execute_after_embedded_header_link fails if directories foo or foo/bar already exists


hewj03:
test_compiler/execute_after_embedded_header_link creates directories foo and foo/bar in /tmp or the current working directory. These directories are deleted later in the test. In some cases, the directories may already exists, e.g. a process other than test_compiler creates them, previous test fails to delete them, or there are parallel test_compiler processes.

foo could be replaced with random string to solve this problem.

PR #251, opened 2019-04-30
cl21: test_compiler: create unique directories for foo/bar

hewj03:
test_compiler/execute_after_included_header_link fails if directories
foo or foo/bar already exists. This happens when another process creates
them, previous test fails to delete them or there are parallel
test_compiler tests. This patch appends a random string to foo. See #250

### DIFF
--- a/test_conformance/compiler/test_compile.c
+++ b/test_conformance/compiler/test_compile.c
@@ -14,12 +14,12 @@
 // limitations under the License.
 //
 #include "testBase.h"
-#if defined(_WIN32)
 #include <time.h>
-#elif  defined(__linux__) || defined(__APPLE__)
+#if  defined(__linux__) || defined(__APPLE__)
 #include <sys/time.h>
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_compile.c
```diff
@@ -2519,6 +2519,33 @@ int test_execute_after_embedded_header_link(cl_device_id deviceID, cl_context co
 #include <direct.h>
 #endif
 
+/*
+ * Create a unique directory fooXXX in the current working directory. XXX is
+ * replaced with a three-letter string that makes the directory unique.
+ */
+static int mkdir_foo(char *foo)
```
rszymanski: Could this function use 'mktemp' to generate unique filename?

hewj03: mktemp has race problem. mkdtemp could be used here, but it is only available in posix compliant systems.

PR #263, opened 2019-05-07
cl21: Fix uninitialized test definition fields

svenvh:
The `selected` field was not initialized, leading to intermittent
"Test X has already been selected" errors when passing a list of tests
on the test_spirv_new command line.

Initialize the selected and result fields with the same values as
ADD_TEST does in test_common/harness/testHarness.h .


svenvh:
Nevermind, this does not apply against this branch.

Issue #259, opened 2019-05-01
Convergence to a single branch


kpet:
The medium/long-term plan for the conformance tests is to converge to a single branch that can be used for all versions of OpenCL. This issue will be updated with the latest plan and progress.

- [x] Replace cl22_trunk with a master branch
- [x] Update Travis config
- [x] Reduce diff between cl21_trunk and master to the minimum
- [x] Make it possible to plug offline vendor compilers for the SPIR-V flow on the master branch
- [x] Introduce a framework change to allow specifying the minimal required version for each test
- [x] Merge cl21_trunk into master
- [x] Reduce diff between cl20_trunk and master to the minimum
- [x] Merge cl20_trunk into master
- [x] Figure out what to do next for cl12_trunk
- [x] Get rid of compatibility suites (#494)
- [x] Change common code on master to support both OpenCL 2.x and 1.x
- [ ] Resolve #517 
- [ ] Mark all tests that require 2.x features
- [x] Merge cl12_trunk into master

pierremoreau:
I started some work on cleaning up the CMake files and allowing the user to select, from CMake, which OpenCL API version to test (as we will want to change the definition of `CL_TARGET_OPENCL_VERSION`, and avoid requiring the offline C++ compiler and libclcxx for versions <= 2.1). I’ll post a MR once I am further along.

kpet:
1. Maybe. I'm hopeful we'll be able to avoid building for a specific version but I'm not 100% sure.
2. Yes, this will need to happen but only right before the merge for each branch so we keep the diff minimal.

kpet:
`cl21_trunk` now merged into `master` \o/.

kpet:
`cl20_trunk` now merged into `master`.

kpet:
Regarding the merge of `cl12_trunk` into `master`, I think the first step is to make it possible to use `master` with a 1.2 implementation. @gwawiork, @alycm is that something you could help with?

alycm:
It's certainly something we'd be interested in having working, I don't have a good handle on how much work it would be though.

To make `master` 1.2 compatible would roughly involve (1) running a 1.2 implementation and seeing what fails, (2) triaging those fails into "tests 2.x feature" and "should not fail", (3) set minimum versions appropriately for the 2.x features, (4) fix the remaining fails.

Step (4) may lead naturally into forward porting any changes that only exist on the `cl12_trunk` branch.

[...]

PR #271, opened 2019-05-13
Add minimal required version functionality

rszymanski:
This adds functionality to define minimal required version through the
ADD_TEST* macros. Tests that don't meet the version requirement will
be skipped.

By default the minimal required version is set to 1.0, subsequent
patches will set the appropriate version for each of the tests.

### DIFF
--- a/test_common/harness/testHarness.c
+++ b/test_common/harness/testHarness.c
@@ -17,6 +17,9 @@
 #include "compat.h"
 #include <stdio.h>
 #include <string.h>
+#include <cassert>
+#include <stdexcept>
+#include <vector>
 #include "threadTesting.h"
 #include "errorHelpers.h"
 #include "kernelHelpers.h"
@@ -703,11 +706,12 @@ test_status callSingleTestFunction( test_definition test, cl_device_id deviceToU
     log_info( "%s...\n", test.name );
     fflush( stdout );
 
-    error = check_opencl_version_with_testname(test.name, deviceToUse);
-    if( error != CL_SUCCESS ) 
-    { 
-        print_missing_feature( error, test.name );
-        return TEST_SKIP; 
+    const Version device_version = get_device_cl_version(deviceToUse);
+    if (test.min_version > device_version)
+    {
+        log_info("%s skipped (requires at least version %s, but the device reports version %s)\n",
+                 test.name, test.min_version.to_string().c_str(), device_version.to_string().c_str());
+        return TEST_SKIP;
     }
 
     error = check_functions_for_offline_compiler(test.name, deviceToUse);
@@ -892,4 +896,30 @@ cl_device_id GetOpposingDevice( cl_device_id device )
 
     // Should never get here
     return NULL;
+}
+
+Version get_device_cl_version(cl_device_id device)
+{
+    size_t str_size;
+    cl_int err = clGetDeviceInfo(device, CL_DEVICE_VERSION, 0, NULL, &str_size);
+    ASSERT_SUCCESS(err, "clGetDeviceInfo");
+
+    std::vector<char> str(str_size);
+    err = clGetDeviceInfo(device, CL_DEVICE_VERSION, str_size, str.data(), NULL);
+    ASSERT_SUCCESS(err, "clGetDeviceInfo");
+
+    if (strstr(str.data(), "OpenCL 1.0") != NULL)
+        return Version(1, 0);
+    else if (strstr(str.data(), "OpenCL 1.1") != NULL)
+        return Version(1, 1);
+    else if (strstr(str.data(), "OpenCL 1.2") != NULL)
+        return Version(1, 2);
+    else if (strstr(str.data(), "OpenCL 2.0") != NULL)
+        return Version(2, 0);
+    else if (strstr(str.data(), "OpenCL 2.1") != NULL)
+        return Version(2, 1);
+    else if (strstr(str.data(), "OpenCL 2.2") != NULL)
+        return Version(2, 2);
+
+    throw std::runtime_error(std::string("Unknown OpenCL version: ") + str.data());
 }
--- a/test_common/harness/testHarness.h
+++ b/test_common/harness/testHarness.h
@@ -18,6 +18,8 @@
 
 #include "threadTesting.h"
 #include "clImageHelper.h"
+#include <string>
+#include <sstream>
 
 #include <string>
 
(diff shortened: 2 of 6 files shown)


AlexeySachkov:
Since we have separate branches for OpenCL 1.2/2.0/2.1 tests, would it be easier just to set appropriate default version on each branch and avoid fine-grained tuning?

test_common/harness/testHarness.c
```diff
@@ -891,3 +900,29 @@ cl_device_id GetOpposingDevice( cl_device_id device )
     // Should never get here
     return NULL;
 }
+
+Version get_device_cl_version(cl_device_id device)
```
gwawiork: There is function for checking CL version. Maybe we should not add this new one:
[get_device_version ](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/kernelHelpers.c#L734)

rszymanski: There's also [this](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/errorHelpers.c#L765), [this](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/api/test_api_min_max.c#L641), and [this](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/computeinfo/main.c#L592). 

I'd prefer to add this simple one, and not over complicate the rest of the patch. Later we can cleanup the duplicates (I'm happy to do that once we merge the branches).

gwawiork: Ok sounds good

rszymanski:
> Since we have separate branches for OpenCL 1.2/2.0/2.1 tests, would it be easier just to set appropriate default version on each branch and avoid fine-grained tuning?

The goal is to have single branch (see #259), and this is one of the steps to achieve that.

test_common/harness/testHarness.c
```diff
@@ -891,3 +900,29 @@ cl_device_id GetOpposingDevice( cl_device_id device )
...
+    else if (strstr(str.data(), "OpenCL 1.2") != NULL)
+        return CL_12;
+    else if (strstr(str.data(), "OpenCL 2.0") != NULL)
+        return CL_20;
+    else if (strstr(str.data(), "OpenCL 2.1") != NULL)
+        return CL_21;
+    else if (strstr(str.data(), "OpenCL 2.2") != NULL)
+        return CL_22;
+
+    throw std::runtime_error(std::string("Unknown OpenCL version: ") + str.data());
```
gwawiork: Is there any reason why errors handling is done with using "throw" and asserts ? I think this is first time we go this way. In other functions we do it in another way. Again, nice to have one consistent procedure.

rszymanski: I know, but the current code is anything but consistent, and I'd rather go "the C++ way" for any new changes. Again, the plan is to cleanup all this mess eventually.

gwawiork: Ok no problem.

AlexeySachkov:
> > Since we have separate branches for OpenCL 1.2/2.0/2.1 tests, would it be easier just to set appropriate default version on each branch and avoid fine-grained tuning?
> 
> The goal is to have single branch (see #259), and this is one of the steps to achieve that.

Sorry, I missed this

test_common/harness/testHarness.c
```diff
@@ -891,3 +900,29 @@ cl_device_id GetOpposingDevice( cl_device_id device )
     // Should never get here
     return NULL;
 }
+
+Version get_device_cl_version(cl_device_id device)
+{
+    size_t str_size;
+    cl_int err = clGetDeviceInfo(device, CL_DEVICE_VERSION, 0, NULL, &str_size);
+    assert(err != CL_SUCCESS && "Could not get CL_DEVICE_VERSION size.");
```
gwawiork: wrong expression, should be err == CL_SUCCESS

test_common/harness/testHarness.c
```diff
@@ -891,3 +900,29 @@ cl_device_id GetOpposingDevice( cl_device_id device )
...
+
+    std::vector<char> str(str_size);
+    err = clGetDeviceInfo(device, CL_DEVICE_VERSION, str_size, str.data(), NULL);
+    assert(err != CL_SUCCESS && "Could not get CL_DEVICE_VERSION string.");
```
gwawiork: wrong expression, should be err == CL_SUCCESS

test_common/harness/testHarness.c
```diff
@@ -655,23 +658,29 @@ int callTestFunctions( test_definition testList[], unsigned char selectedTestLis
...
+                log_info("Skipping %s. Device doesn't support functionality exercised by this test.\n",
+                         testList[i].name);
+                continue;
             }
-            else
+
+            if (testList[i].func == NULL)
             {
                 log_info( "Skipping %s. Test currently not implemented.\n", testList[i].name );
+                continue;
```
gwawiork: Results do not print test result status. It looks strange:
./test_conformance_device_timer CL_DEVICE_TYPE_GPU
Requesting GPU device based on command line for platform index 0 and device index 0
Compute Device Name = Intel(R) Gen9 HD Graphics NEO, Compute Device Vendor = Intel(R) Corporation, Compute Device Version = OpenCL 2.0 NEO , CL C Version = OpenCL C 2.0
sizeof( void*) = 8  (host)
sizeof( void*) = 8  (device)
Skipping timer_resolution_queries. Device doesn't support functionality exercised by this test.
Skipping device_and_host_timers. Device doesn't support functionality exercised by this test.

test_common/harness/testHarness.c
```diff
@@ -655,23 +658,29 @@ int callTestFunctions( test_definition testList[], unsigned char selectedTestLis
...
         {
-            // Skip unimplemented test (can happen when you select all of the tests)
-            if( testList[i].func != NULL )
+            if (testList[i].min_version > device_version)
             {
-                int errors = callSingleTestFunction( testList[i], deviceToUse, forceNoContextCreation,
-                                                     numElementsToUse, queueProps );
-                resultTestList[i] = errors;
-                totalErrors += errors;
+                log_info("Skipping %s. Device doesn't support functionality exercised by this test.\n",
```
kpet: I would suggest making the message more specific. Probably something along the lines of "TEST requires at least version X but the device reported version Y".

test_common/harness/testHarness.c
rszymanski: Yep, silly mistake. This actually shows that we need something better than assert, as it doesn't catch it in release build.

test_common/harness/testHarness.c
rszymanski: Will rework.

test_common/harness/testHarness.c
rszymanski: Need to rework how we deal with versions then.

test_common/harness/testHarness.h
```diff
@@ -18,22 +18,45 @@
...
+#define NOT_IMPLEMENTED_TEST(fn) {NULL, #fn, Version(0, 0)}
 
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
+class Version
+{
+public:
+    Version() : m_major(0), m_minor(0) {}
+    Version(int major, int minor) : m_major(major), m_minor(minor) {}
+    int total() const { return m_major * 10 + m_minor; }
```
kpet: `to_int`? `total` isn't a very self-explanatory name... Suggest aligning on what the headers do as well (OpenCL 1.2 is 120).

rszymanski: Agreed, `to_int` sounds much better.

PR #273, opened 2019-05-16
Use test_status for test results

rszymanski:
This moves from numeric values to enum values to keep result state of
the test, so we can easily save this information in JSON results.

### DIFF
--- a/test_common/harness/testHarness.c
+++ b/test_common/harness/testHarness.c
@@ -526,7 +526,7 @@ static int find_matching_tests( test_definition testList[], unsigned char select
 }
 
 static int saveResultsToJson( const char *fileName, const char *suiteName, test_definition testList[],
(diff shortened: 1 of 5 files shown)


test_common/harness/testHarness.c
```diff
@@ -717,29 +704,41 @@ int callSingleTestFunction( test_definition test, cl_device_id deviceToUse, int
     fflush( stdout );
 
     error = check_opencl_version_with_testname(test.name, deviceToUse);
-    test_missing_feature(error, test.name);
-
-    error = check_functions_for_offline_compiler(test.name, deviceToUse);
-    test_missing_support_offline_cmpiler(error, test.name);
```
kpet: Are we no longer performing that check?

rszymanski: Missed it when cherry-picking from 2.0 branch, now restored.

PR #282, opened 2019-05-24
Fix test_host_queue_order not to rely on bit accurate float opsHost order

nikhiljnv:
Currently, test_host_queue_order relies on bit accurate float ops.
It expects expects that sqrt((float)i * i) == i.
This is not always true due to floating point precision limitations.
The test does not really need the sqrt operation and it can be removed
instead of trying to correct the floating point check.

Replace problematic float operation with constant value of 1.


nikhiljnv:
Closing this pull request in favor of correct pull request #295

Issue #284, opened 2019-05-24
CTS building - changes dedicated for MSVC2017 and Intel Compiler v.19.0


gwawiork:
Some patches needed to compile CTS using MSVC2017 and Intel Compiler version 19.0 (available in Intel System Studio 2019):

- cmake changes - add compilation flags for long double compilation
- changes in compat.h - defines already defined in Intel Compiler
- fix assert in test_SVM - strange complicated code for checking OCL version. It has been simplified
- bruteforce tests - fix for Intel compiler - signbitl is available in Intel Compiler.
Expect 4 pull requests for each branch.

gwawiork:
I have set of additional commits dedicated for  cl12_trunk branch to solve compilation errors for ICC compiler and errors while comparing double precision tests in math_brute_force.

- some of changes are similar to changes introduced for branches cl20+
- I will propose to upgrade reference_math.c file to be more compatible with cl20+ branches
- I noticed many compilation errors because not adding ../../test_common/harness/mt19937.c in function  set_source_files_properties
- I noticed compilation error in test_geometrics usage MAKE_HEX_FLOAT instead of MAKE_HEX_DOUBLE like we have on branches cl20+

I hope these changes allow to compile tests using MSVC, double precision tests will not fail.

PR #288, opened 2019-05-24
Intel compiler bruteforce -  fix signbitl function

gwawiork:
Fix Issue #284 . When Intel compiler is used then embeded signbitl will be used


gwawiork:
I will check this

gwawiork:
It does not work for me. The test signbit fp64 is failing when std:signbit is used ( I have MSVC 2017)

Issue #299, opened 2019-05-28
Test SVM gap - test clEnqueueSVMMemcpy in case of host memory allocation


gwawiork:
The test test_enqueue_api checks if API function clEnqueueSVMMemcpy works when both (source and destination) memory pointers are SVM allocations. We should check also case when one of the allocation (source or destination) is a host memory. The discussion about this topic inside the Khronos group was [here](https://www.khronos.org/members/login/bugzilla/show_bug.cgi?id=14598)
My proposal it to extend the existing test case equeue_api to test host memory as a source or destination memory.

kpet:
Fixed on all relevant branches. Closing.

PR #300, opened 2019-05-28
test clEnqueueSVMMemcpy in case of host memory allocation

gwawiork:
Fix issue #299

### DIFF
--- a/test_conformance/SVM/test_enqueue_api.cpp
+++ b/test_conformance/SVM/test_enqueue_api.cpp
@@ -118,25 +118,22 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
     svm
   };
 
-  typedef struct _TestType {
+  struct TestType {
     allocationTypes srcAlloc;
     allocationTypes dstAlloc;
-    _TestType(allocationTypes type1, allocationTypes type2): srcAlloc(type1), dstAlloc(type2){}
-  } TestType;
+    TestType(allocationTypes type1, allocationTypes type2): srcAlloc(type1), dstAlloc(type2){}
+  };
 
- typedef std::vector<TestType> TestTypes;
+ std::vector<TestType> testTypes;
 
- TestTypes testTypes;
  testTypes.push_back(TestType(host, svm));
  testTypes.push_back(TestType(svm, host));
  testTypes.push_back(TestType(svm, svm));
 
- size_t array_size_2 = sizeof(typeSizes) / sizeof(typeSizes[0]);
-
   for (const auto test_case : testTypes)
   {
     log_info("clEnqueueSVMMemcpy case: src_alloc = %s, dst_alloc = %s\n", test_case.srcAlloc == svm ? "svm" : "host", test_case.dstAlloc == svm ? "svm" : "host");
-    for (size_t i = 0; i < sizeof(typeSizes) / sizeof(typeSizes[0]); ++i)
+    for (size_t i = 0; i < ARRAY_SIZE(typeSizes); ++i)
     {
       //generate initial data
       std::vector<cl_uchar> fillData0(typeSizes[i]), fillData1(typeSizes[i]);
@@ -167,14 +164,12 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
       cl_uchar * dst_ptr;
       if (test_case.srcAlloc == host) {
         src_ptr = srcHostData.data();
-      }
-      if (test_case.srcAlloc == svm) {
+      } else if (test_case.srcAlloc == svm) {
         src_ptr = srcBuffer;
       }
       if (test_case.dstAlloc == host) {
         dst_ptr = dstHostData.data();
-      }
-      if (test_case.dstAlloc == svm) {
+      } else if (test_case.dstAlloc == svm) {
         dst_ptr = dstBuffer;
       }
       clEventWrapper eventMemcpy;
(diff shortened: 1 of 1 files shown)


gwawiork:
@kpet  Please review new changes. The time execution is really short and this change has no impact

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -113,115 +113,166 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
-    cl_uchar *dstBuffer = (cl_uchar *)clSVMAlloc(context, CL_MEM_READ_WRITE, elementNum * typeSizes[i], 0);
+  typedef struct _TestType {
+    allocationTypes srcAlloc;
+    allocationTypes dstAlloc;
+    _TestType(allocationTypes type1, allocationTypes type2): srcAlloc(type1), dstAlloc(type2){}
+  } TestType;
 
-    clEventWrapper userEvent = clCreateUserEvent(context, &error);
-    test_error(error, "clCreateUserEvent failed");
+ typedef std::vector<TestType> TestTypes;
```
kpet: I think that type isn't needed.

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -113,115 +113,166 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
-    generate_data(fillData0, typeSizes[i], seed);
-    generate_data(fillData2, typeSizes[i], seed);
+  enum allocationTypes {
+    host,
+    svm
+  };
 
-    cl_uchar *srcBuffer = (cl_uchar *)clSVMAlloc(context, CL_MEM_READ_WRITE, elementNum * typeSizes[i], 0);
-    cl_uchar *dstBuffer = (cl_uchar *)clSVMAlloc(context, CL_MEM_READ_WRITE, elementNum * typeSizes[i], 0);
+  typedef struct _TestType {
```
kpet: This is C++ code. `struct TestType` is enough.

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -113,115 +113,166 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
-    test_error(error, "clEnqueueSVMMemFill failed");
+ TestTypes testTypes;
+ testTypes.push_back(TestType(host, svm));
+ testTypes.push_back(TestType(svm, host));
+ testTypes.push_back(TestType(svm, svm));
 
-    clEventWrapper eventMemcpy;
-    error = clEnqueueSVMMemcpy(queue, CL_FALSE, dstBuffer, srcBuffer, elementNum * typeSizes[i], 1, &eventMemFill, &eventMemcpy);
-    test_error(error, "clEnqueueSVMMemcpy failed");
+ size_t array_size_2 = sizeof(typeSizes) / sizeof(typeSizes[0]);
```
kpet: This doesn't seem to be used.

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -113,115 +113,166 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
-    test_error(error, "clCreateUserEvent failed");
+ typedef std::vector<TestType> TestTypes;
 
-    clEventWrapper eventMemFill;
-    error = clEnqueueSVMMemFill(queue, srcBuffer, &fillData0[0], typeSizes[i], elementNum * typeSizes[i], 1, &userEvent, &eventMemFill);
-    test_error(error, "clEnqueueSVMMemFill failed");
+ TestTypes testTypes;
+ testTypes.push_back(TestType(host, svm));
+ testTypes.push_back(TestType(svm, host));
+ testTypes.push_back(TestType(svm, svm));
```
kpet: What about host/host? We might as well throw it in and have exhaustive coverage.

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -113,115 +113,166 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
-    error = clEnqueueSVMMemcpy(queue, CL_FALSE, dstBuffer, srcBuffer, elementNum * typeSizes[i], 1, &eventMemFill, &eventMemcpy);
-    test_error(error, "clEnqueueSVMMemcpy failed");
+ size_t array_size_2 = sizeof(typeSizes) / sizeof(typeSizes[0]);
 
-    error = clSetUserEventStatus(userEvent, CL_COMPLETE);
-    test_error(error, "clSetUserEventStatus failed");
+  for (const auto test_case : testTypes)
+  {
+    log_info("clEnqueueSVMMemcpy case: src_alloc = %s, dst_alloc = %s\n", test_case.srcAlloc == svm ? "svm" : "host", test_case.dstAlloc == svm ? "svm" : "host");
+    for (size_t i = 0; i < sizeof(typeSizes) / sizeof(typeSizes[0]); ++i)
```
kpet: My initial comment still applies. Please use the `ARRAY_SIZE` macro.

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -113,115 +113,166 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
+
+      error = clSetUserEventStatus(userEvent, CL_COMPLETE);
+      test_error(error, "clSetUserEventStatus failed");
+
+      cl_uchar * src_ptr;
+      cl_uchar * dst_ptr;
+      if (test_case.srcAlloc == host) {
+        src_ptr = srcHostData.data();
+      }
+      if (test_case.srcAlloc == svm) {
```
kpet: `else if`?

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -112,52 +112,74 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
-      if (test_case == bothSvmAlloc) {
-        error = clEnqueueSVMMemcpy(queue, CL_FALSE, dstBuffer, srcBuffer, elementNum * typeSizes[i], 1, &eventMemFill, &eventMemcpy);
-        test_error(error, "clEnqueueSVMMemcpy failed - source memory: SVM allocation, destination memory: SVM allocation");
+      if (test_case.dstAlloc == host) {
+        dst_ptr = dstHostData.data();
       }
-
-      error = clSetUserEventStatus(userEvent, CL_COMPLETE);
-      test_error(error, "clSetUserEventStatus failed");
+      if (test_case.dstAlloc == svm) {
```
kpet: Same here

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -222,44 +224,52 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
-        return -1;
+        return TEST_FAIL;
+      }
+
+      error = clGetEventInfo(eventMemFillList[1], CL_EVENT_COMMAND_TYPE, sizeof(cl_command_type), &commandType, NULL);
+      test_error(error, "clGetEventInfo failed");
+      if (commandType != CL_COMMAND_SVM_MEMFILL)
+      {
+          log_error("Invalid command type returned for clEnqueueSVMMemFill\n");
+          return TEST_FAIL;
```
kpet: Please use a loop instead of copy/pasting the block of code.

test_conformance/SVM/test_enqueue_api.cpp
gwawiork: I discussed this with @bashbaug . We think this case is not very useful and probably spec clarification is needed to turn off such case.

kpet: Hmm, I'd argue that it is clearly allowed by the spec and has been since 2.0. I don't think we can change this now and though I agree it doesn't seem very useful I think we need to cover it here.

PR #311, opened 2019-06-10
cl20: Intel compiler bruteforce -  fix signbitl function

gwawiork:
Fix issue #284


kpet:
Only Appveyor failed. Since it shouldn't be triggering anyway, I'll merge this.

Issue #312, opened 2019-06-12
copy_image and fill_image tests use incorrect stride for 1D arrays


robquill:
The copy_image (for OpenCL 1.2) and fill_image (for 1.2 and 2.x) conformance tests incorrectly use the row pitch instead of the slice pitch when mapping the image.

kpet:
Fixed on all branches, closing.

PR #313, opened 2019-06-12
CL20: Use slice pitch instead of row pitch for 1D arrays.

robquill:
The copy_image (for OpenCL 1.2) and fill_image (for 1.2 and 2.x) conformance tests incorrectly use the row pitch instead of the slice pitch when mapping the image.

See #312

### DIFF
--- a/test_conformance/images/clFillImage/test_fill_generic.cpp
+++ b/test_conformance/images/clFillImage/test_fill_generic.cpp
@@ -194,12 +194,26 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
     }
     else {
         // Else copy one scan line at a time.
+        size_t dstPitch2D = 0;
+        switch (imageInfo->type)
+        {
+            case CL_MEM_OBJECT_IMAGE3D:
+            case CL_MEM_OBJECT_IMAGE2D_ARRAY:
+            case CL_MEM_OBJECT_IMAGE2D:
+                dstPitch2D = mappedRow;
+                break;
+            case CL_MEM_OBJECT_IMAGE1D_ARRAY:
+            case CL_MEM_OBJECT_IMAGE1D:
+                dstPitch2D = mappedSlice;
+                break;
+        }
+
         for ( size_t z = 0; z < depth; z++ )
         {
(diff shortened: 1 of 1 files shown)


test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -194,12 +194,26 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
+        size_t dstPitch2D = 0;
+        switch (imageInfo->type)
+        {
+            case CL_MEM_OBJECT_IMAGE3D:
+            case CL_MEM_OBJECT_IMAGE2D_ARRAY:
+            case CL_MEM_OBJECT_IMAGE2D:
+                dstPitch2D = mappedRow;
+                break;
+            case CL_MEM_OBJECT_IMAGE1D_ARRAY:
+            case CL_MEM_OBJECT_IMAGE1D:
```
jlewis-austin: Why add IMAGE1D here? It just distracts from the real fix.

robquill: dstPitch2D is initialised to 0 so need to set it to something valid for all cases in the switch.

There are other ways we could fix this bug, e.g. by swapping height and depth at line 148:

        case CL_MEM_OBJECT_IMAGE1D_ARRAY:
            height = imageInfo->arraySize;
            depth = 1;

(This was my original fix when I was fixing this for 1.2 but the fix proposed in the pull request matches the equivalent fix that is already present in clCopyImage/test_copy_generic.cpp)

jlewis-austin: OK, sounds reasonable.

kpet: Agree that this isn't right, that it should work and that reducing the diff with similar code is probably more helpful given the situation we're in.

kpet:
@robquill Could you prepare a PR for the master branch please?

robquill:
> @robquill Could you prepare a PR for the master branch please?

Yep, see pull request #331

kpet:
Thanks!

PR #331, opened 2019-06-18
master: Use slice pitch instead of row pitch for 1D arrays.

robquill:
The copy_image (for OpenCL 1.2) and fill_image (for 1.2 and 2.x) conformance tests incorrectly use the row pitch instead of the slice pitch when mapping the image.


fengxuegt:
hi all, I want to know the reason why we use slice pitch instead of row pitch for 1Darray? Can you explain it a little? Thanks a lot.

Issue #332, opened 2019-06-19
GL sharing tests -  compilation on Windows fixes


gwawiork:
I noticed issues with compilation on Windows - cl12_trunk branch. 
Binary names are not compatible with others - branches cl20+
Some clarification how to compile on Windows

kpet:
All PRs merged. Closing.

PR #336, opened 2019-06-19
GL sharing: Test name adjustment and Windows compilation clarification.

gwawiork:
Fix issue #332


test_conformance/gl/CMakeLists.txt
```diff
@@ -1,10 +1,21 @@
...
-list(APPEND CLConform_LIBRARIES  glut32 opengl32 glu32 )
+# Windows MSVC/Intel compiler hits: 
+# - add x64/x86 binaries with glew and glut/freeglut libraries (.lib) to the directory lib/x64 or lib/x86. If names are different then modify list CLConform_LIBRARIES accordingly
+# - GL headers files should be added to project source directory and GL subdirectory
+# - Add .dll files: glut glew to the test binary location.
+    if(CMAKE_CL_64)
+        link_directories("../../lib/x64")
+    else(CMAKE_CL_64)
+        link_directories("../../lib/x86")
+    endif(CMAKE_CL_64)
```
kpet: Hmm, I'm wondering whether that shouldn't be a CMake macro passed when configuring the build.

gwawiork: You mean CMAKE_CL_64 ? I did not add this variable and this is added when cmake project generation by adding CMAKE_CL_64=ON

kpet: No, I was thinking of the hardcoded relative paths. It would be cleaner to use a CMake variable instead. `CLConform_GL_LIBRARIES_DIR`?

test_conformance/gl/CMakeLists.txt
```diff
@@ -37,42 +45,29 @@ set (GL_SOURCES
...
     set_source_files_properties(
-        ${GL_SOURCES}
+        ${${MODULE_NAME}_SOURCES}
         COMPILE_FLAGS -DGLEW_STATIC)
     include_directories("./GLEW/")
 endif(MINGW)
 
-if (MSVC)
-    if(CMAKE_CL_64)
-    list(APPEND CLConform_LIBRARIES glew64)
```
alycm: I know nothing about GLEW really, but both this and the cl12 version of the PR remove all reference of `glew64` , leaving only `glew32`.   Is that intentional?   I.e. will it still work as long as `CLConform_GL_LIBRARIES_DIR` is set appropriately?

gwawiork: Here we have some names confusion.
These libraries are  downloaded from the Internet http://glew.sourceforge.net/ .  Check for example glew-2.1.0 . It includes for x64 and x86 in both directories glew32 files. For me it works.
We have just remeber that for windows libraries for x64 or x86 bitness should have names with 32 numbers.

gwawiork:
Quick instruction how to build GL tests on Windows:
- download from the Internet : glew-2.1.0-win32.zip freeglut-MSVC-3.0.0-2.mp.zip
- extract include\GL header files to OpenCL-CTS\test_conformance\gl\GL\
- extract library files (glew32 and freeglut) to OpenCL-CTS\lib\x64\
- generate solution adding params: -DGL_IS_SUPPORTED=ON -DCLConform_GL_LIBRARIES_DIR=lib\x64
- run compilation
- finally copy dlls (freeglut.dll glew32.dll) from zip packages to binary localization

Ubuntu:
- freeglut and glew packages installation needed
- generate solution adding params: -DGL_IS_SUPPORTED=ON
- no header/library files copy needed
- run compilation

PR #349, opened 2019-06-24
Fix enqueue_flags test to use correct barrier type.

nikhiljnv:
Currently, enqueue_flags test uses CLK_LOCAL_MEM_FENCE.
Use CLK_GLOBAL_MEM_FENCE instead as all threads across work-groups
need to wait here.


raunraun:
LGTM

Issue #325, opened 2019-06-17
Ambiguous function calls in c11_atomics


kpet:
From @StuartDBrady:

The CTS generates code to call atomic_fetch_add() with the type
atomic_ulong * for the first parameter, and uint for the second
parameter.
    
On targets where atomic_ptrdiff_t and atomic_uintptr_t are equivalent
to long and ulong respectively, such calls are ambiguous: the type
of the second parameter may be long or ulong, and either of these is
sufficient to represent the full range of uint.

kpet:
Fixed on all relevant branches, closing.

PR #354, opened 2019-07-04
cl21: fix ambiguous call atomic_fetch_add in c11_atomics

hewj03:
Cast operand type to AddSubOperandType. See #325


hewj03:
Only cl21 and master branches are affected. PR to master: https://github.com/KhronosGroup/OpenCL-CTS/pull/361
These two PRs will make file test_atomics.cpp identical as in cl20 trunk.

kpet:
Thanks!

Issue #355, opened 2019-07-04
cl21: spirv path broken in math_brute_force tests


gwawiork:
I see that after moving tests to harness we lost ability run this test using spirv path

gwawiork:
Looks like fixed

PR #357, opened 2019-07-04
Fix spirv path in bruteforce tests

gwawiork:
Fix issue #355

### DIFF
--- a/test_conformance/math_brute_force/main.c
+++ b/test_conformance/math_brute_force/main.c
@@ -1621,18 +1621,11 @@ int InitILogbConstants( void )
     "   out[1] = FP_ILOGBNAN;\n"
     "}\n";
 
-    cl_program query = clCreateProgramWithSource(gContext, 1, &kernel, NULL, &error);
-    if( NULL == query || error)
+    cl_program query;
+    error = create_single_kernel_helper(gContext, &query, NULL, 1, &kernel, NULL);
+    if (NULL == query || error)
     {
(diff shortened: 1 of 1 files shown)


test_conformance/math_brute_force/main.c
```diff
@@ -1843,26 +1810,13 @@ int MakeKernels( const char **c, cl_uint count, const char *name, cl_uint kernel
       strcat(options, " -cl-fast-relaxed-math");
     }
 
-    *p = clCreateProgramWithSource( gContext, count, c, NULL, &error );
-    if( NULL == *p || error )
+    error = create_single_kernel_helper(gContext, p, NULL, count, c, NULL, options);
+    if ( error != CL_SUCCESS )
     {
         vlog_error( "\t\tFAILED -- Failed to create program. (%d)\n", error );
         return -1;
```
kpet: Should this be changed to `return error` as well?

gwawiork:
In my opinion not, offline spirv compilation is available since CL2.1

kpet:
Sure, but there is nothing SPIR-V specific that I could spot in the patch. If the patch applies I think we should port it, if anything to reduce the number of places where the code sequence for building a program exists.

gwawiork:
Ok I will prepare those fixes.
Additionally I've noticed more places in main.c where clCreateProgramWithSource has to be exchanged with create_single_kernel_helper. Fix in progress.

gwawiork:
All fixes have been pushed out

kpet:
Thanks! Merging all PRs.

Issue #373, opened 2019-07-12
cl21 test_spirv_new - failures after recent changes in offline compilations


gwawiork:
It looks that after recent changes in offline compilation some of these tests are failing. Some of them uses both types of compilations in the same time (online and offline). I will propose changes to allow user run online compilation though global variable is set to offline compilation.

gwawiork:
no tests on master branch

PR #375, opened 2019-07-12
cl21 Fix test spirv_new - allow user run online and offline compilation in one application run.

gwawiork:
Fix issue #373

### DIFF
--- a/test_common/harness/kernelHelpers.c
+++ b/test_common/harness/kernelHelpers.c
@@ -486,6 +486,17 @@ int create_single_kernel_helper_create_program(cl_context context,
                                                       kernelProgram, buildOptions, gCompilationMode);
 }
 
+int create_single_kernel_helper_create_program_force_cmpl(cl_context context,
+    cl_program *outProgram,
+    unsigned int numKernelLines,
+    const char **kernelProgram,
+    const char *buildOptions,
+    CompilationMode compilationMode)
+{
+    return create_single_kernel_helper_create_program(context, outProgram, numKernelLines,
+        kernelProgram, buildOptions, compilationMode);
+}
+
 int create_single_kernel_helper_with_build_options(cl_context context,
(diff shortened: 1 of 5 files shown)


gwawiork:
This is a specific test suit because it contains offline files already prepared. In some cases it needs run at the same time online + offline compilation.
I've rethinked once more time my changes and your comment I discovered that this fix is not needed :). The source of my changes was issue that -ILPath option stop working. However I see that instead of this I can give option --compilation-cache-path and test works as previously. I thought that this new option work only when --compilation-mode is given (there is a comment in help). 
 

> For offline compilation (binary and spir-v modes) only:

        --compilation-cache-mode <cache-mode>  Specify a compilation caching mode:
                                 compile-if-absent  Read from cache if already populated, or
                                                    else perform offline compilation (default)
                                 force-read      Force reading from the cache
                                 overwrite       Disable reading from the cache
        --compilation-cache-path <path>   Path for offline compiler output and CL source

> 

Finally I see this works for me fine:
./test_spirv_new --compilation-cache-path spirv_bin CL_DEVICE_TYPE_GPU
If you have no objection in such running this case I will close the issue and corresponding MRs

kpet:
>This is a specific test suit because it contains offline files already prepared. In some cases it needs run at the same time online + offline compilation.

Yes, I'd somehow forgotten that the SPIR-V files accessed by `get_program_with_il` are generated from the assembly and not from the C code which, as you say, doesn't exist for all tests. The test using `gCompilationCachePath` to access the SPIR-V files was enough to fool me during a quick scan.

>If you have no objection in such running this case I will close the issue and corresponding MRs

Re-using the compilation cache option is a dirty hack that is only going to confuse people. `spirv_new` should have its own private option to specify the location of the SPIR-V files.

gwawiork:
I will close this PR and  new will be open with new proposal

Issue #381, opened 2019-07-22
cl21: test_svm fails if 64-bit atomics extensions are supported


hewj03:
Two errors:
1. kernel build error
`1:3:2: error: use of type 'atomic_uintptr_t' (aka '_Atomic(unsigned long)') requires cl_khr_int64_base_atomics extension to be enabled
 atomic_uintptr_t pNext;
 ^
1:3:2: error: use of type 'atomic_uintptr_t' (aka '_Atomic(unsigned long)') requires cl_khr_int64_extended_atomics extension to be enabled
Compilation failed`

2. run segfault
`0x00007FF7F1234EC2 (0x0000006EA0F7E9A8 0x0000006EA0F7EA10 0x0000006EA0F7E9E8 0x0000006E00000001), launch_kernels_and_verify() + 0x4C2 bytes(s), C:\Users\wenjuhe\cnf\github-OpenCL-CTS\OpenCL-CTS-21\test_conformance\SVM\test_fine_grain_memory_consistency.cpp, line 106 + 0xA byte(s)`

kpet:
All PRs merged, closing.

PR #382, opened 2019-07-22
cl21: Fix SVM if 64-bit atomic extensions are supported

hewj03:
Need to enable 64-bit atomic extensions in kernel source.
Use InterlockedCompareExchange64 for 64-bit type. See issue #381

This patch modifies test_fine_grain_memory_consistency.cpp to be the same as in cl20_branch.


hewj03:
PR to master: https://github.com/KhronosGroup/OpenCL-CTS/pull/396
Also need to fix common.h in cl20 branch: https://github.com/KhronosGroup/OpenCL-CTS/pull/397

kpet:
Thanks!

Issue #387, opened 2019-07-23
Media sharing test - add to build for Windows 


gwawiork:
The current test version do not compile. The following changes are required:

- adjust cases to ADD_TEST marco
- modify cmakes files
- enable build only for Windows
- In case extension cl_khr_dx9_media_sharing not supported skip the test

gwawiork:
Merged Issue can be closed

PR #389, opened 2019-07-23
Fix building media sharing tests

gwawiork:
Fix issue #387

### DIFF
--- a/test_extensions/CMakeLists.txt
+++ b/test_extensions/CMakeLists.txt
@@ -1,4 +1,1 @@
-# Remember curren source directory (`test_conformance').
-set( CLConf_Install_Base_Dir "${CMAKE_CURRENT_SOURCE_DIR}" )
-
 add_subdirectory( media_sharing )
--- a/test_extensions/CMakeLists.txt
+++ b/test_extensions/CMakeLists.txt
@@ -1,0 +1,4 @@
+# Remember curren source directory (`test_conformance').
+set( CLConf_Install_Base_Dir "${CMAKE_CURRENT_SOURCE_DIR}" )
+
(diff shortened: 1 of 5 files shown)


test_extensions/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+# Remember curren source directory (`test_conformance').
+set( CLConf_Install_Base_Dir "${CMAKE_CURRENT_SOURCE_DIR}" )
```
kpet: Where is this used?

test_extensions/media_sharing/procs.h
```diff
@@ -0,0 +1,31 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+
+#ifndef __PROCS_H__
```
kpet: `__PROCS_H__` is a reserved identifier. See https://en.cppreference.com/w/c/language/identifier. Please use `#pragma once` or `MEDIA_SHARING_PROCS_H` for example.

test_extensions/CMakeLists.txt
gwawiork: I've removed this

test_extensions/media_sharing/procs.h
gwawiork: fixed

kpet: You missed the point :). The problem was the double underscore followed by a capital letter...

Issue #388, opened 2019-07-23
Compilation errors in D3D cases


gwawiork:
There is wrong path to kernelHelpers in harness.h file

kpet:
All PRs merged, closing.

PR #392, opened 2019-07-23
Fix compilation in D3D cases

gwawiork:
Fix issue #388


gwawiork:
Ok. I see that in CmakeLists.txt we have this path already included :

`include_directories (${CLConform_SOURCE_DIR}/test_common/harness`

then enough is to add:

`#include "testHarness.h"`

PR #398, opened 2019-07-29
cl12: Add aarch64 to system processor if statement in CMakeLists

aarongreig:
Aarch64 devices also need `-fsigned-char` but previously would not have hit this branch of the if statement.

### DIFF
--- a/test_conformance/CMakeLists.txt
+++ b/test_conformance/CMakeLists.txt
@@ -22,7 +22,7 @@ else()
     # set it, so setting it here once means we have consistent compiliation
     # flags across the entire conformance_test_* executables.
     set(CTS_EXTRA_FLAGS "${CTS_EXTRA_FLAGS} -mfpmath=sse -msse2")
-  elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(ARM)|(ARM64)")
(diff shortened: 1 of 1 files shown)


test_conformance/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ else()
     # set it, so setting it here once means we have consistent compiliation
     # flags across the entire conformance_test_* executables.
     set(CTS_EXTRA_FLAGS "${CTS_EXTRA_FLAGS} -mfpmath=sse -msse2")
-  elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(ARM)|(ARM64)")
+  elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "(ARM)|(ARM64)|(aarch64)")
```
kpet: I'm guessing that `ARM64` is redundant then, isn't it?

alycm: A quick code search on GitHub find ~10k using aarch64, and ~1k using ARM64.   But both do seem to be used.

kpet: Thanks. No harm in keeping both anyway.

Issue #327, opened 2019-06-17
cl21 select doesn't work in offline mode


kpet:
`parseCustomParam` isn't called.

PR #401, opened 2019-07-30
Generate kernel sources only mode

boazo:
This introduces a new generate kernel sources mode. In this mode the conformance dumps the .cl files into a folder but does not attempt to run the full conformance. The change is comprised from the introduction of the new flag and forcing the individual conformance tests to continue running all tests even if compilation fails. A few suites still need to be modified to support this change.

### DIFF
--- a/test_common/harness/kernelHelpers.c
+++ b/test_common/harness/kernelHelpers.c
@@ -86,7 +86,7 @@ static std::string get_kernel_content(unsigned int numKernelLines, const char *c
     return kernel;
 }
 
-std::string get_kernel_name(const std::string &source)
+std::string get_kernel_name(const std::string &source, const std::string &options)
 {
     // Count CRC
     cl_uint crc = crc32(source.data(), source.size());
@@ -172,6 +172,10 @@ std::string get_kernel_name(const std::string &source)
         oss << kernelsList;
     }
     oss << std::hex << std::setfill('0') << std::setw(8) << crc;
+    if(!options.empty()) {
+        crc = crc32(options.data(), options.size());
+        oss << '.' << std::hex << std::setfill('0') << std::setw(8) << crc;; 
+    }
     return oss.str();
 }
 
@@ -391,6 +395,12 @@ static int get_offline_compiler_output(std::ifstream &ifs,
             return -1;
         }
 
+        if(gExecutionMode == kGenerate) {
+            // in this mode we don't emit the compiler
+            // we fake a compilation error instead
+            return -1;
+        }
+
         // write source to input file
         ofs.write(kernel.c_str(), kernel.size());
         ofs.close();
@@ -423,13 +433,14 @@ static int create_single_kernel_helper_create_program_offline(cl_context context
 {
     int error;
     std::string kernel = get_kernel_content(numKernelLines, kernelProgram);
-    std::string kernelName = get_kernel_name(kernel);
 
     // set build options
     std::string bOptions;
     bOptions += buildOptions ? std::string(buildOptions) : "";
 
+    std::string kernelName = get_kernel_name(kernel, bOptions);
     kernelName = add_build_options(kernelName, buildOptions);
+
 
     std::ifstream ifs;
     error = get_offline_compiler_output(ifs, context, kernel, openclCXX, compilationMode, bOptions, kernelName);
@@ -608,6 +619,13 @@ int create_single_kernel_helper(cl_context context,
             return error;
         }
     }
+
+    if(gExecutionMode == kGenerate)
+    {
+        log_info("INFO: KERNEL %s WAS GENERATED\n", kernelName);
+        return -1;
+    }
+
     // Remove offline-compiler-only build options
     std::string newBuildOptions;
     if (buildOptions != NULL)
--- a/test_common/harness/parseParameters.cpp
+++ b/test_common/harness/parseParameters.cpp
@@ -30,6 +30,7 @@ using namespace std;
 CompilationMode      gCompilationMode = kOnline;
 CompilationCacheMode gCompilationCacheMode = kCacheModeCompileIfAbsent;
 std::string          gCompilationCachePath = ".";
+ExecutionMode        gExecutionMode = kExecute;
 
 void helpInfo ()
 {
@@ -47,6 +48,8 @@ void helpInfo ()
              "                                 force-read      Force reading from the cache\n"
              "                                 overwrite       Disable reading from the cache\n"
              "        --compilation-cache-path <path>   Path for offline compiler output and CL source\n"
+             "        --generate-sources-only Use this flag to dump the .cl and build options files used by the test.\n"
+             "                                This feature is still under development (Beta)"
              "\n");
 }
 
@@ -158,6 +161,12 @@ int parseCustomParam (int argc, const char *argv[], const char *ignore)
                 return -1;
             }
         }
+        else if (!strcmp(argv[i], "--generate-sources-only"))
+        {
+            delArg++;
+            log_info("Generating sources only\n");
+            gExecutionMode = kGenerate;
+        }
 
         //cleaning parameters from argv tab
         for (int j = i; j < argc - delArg; j++)
@@ -172,6 +181,11 @@ int parseCustomParam (int argc, const char *argv[], const char *ignore)
         return -1;
     }
 
+    if (gExecutionMode == kGenerate && (gCompilationCacheMode != kCacheModeOverwrite || gCompilationMode == kOnline))
+    {
+        log_error("Compilation cache mode can only be overwrite when in generate sources only mode.\n");
+        return -1;
+    }
     return argc;
 }
 
--- a/test_common/harness/parseParameters.h
+++ b/test_common/harness/parseParameters.h
@@ -19,6 +19,11 @@
 #include "compat.h"
 #include <string>
 
+enum ExecutionMode
+{
+    kExecute = 0,
+    kGenerate ,
+};
 enum CompilationMode
 {
     kOnline = 0,
@@ -36,6 +41,7 @@ enum CompilationCacheMode
 extern CompilationMode gCompilationMode;
 extern CompilationCacheMode gCompilationCacheMode;
 extern std::string gCompilationCachePath;
+extern ExecutionMode gExecutionMode;
 
 extern int parseCustomParam (int argc, const char *argv[], const char *ignore = 0 );
 
--- a/test_conformance/buffers/test_buffer_fill.c
+++ b/test_conformance/buffers/test_buffer_fill.c
@@ -779,8 +779,6 @@ int test_buffer_fill_struct( cl_device_id deviceID, cl_context context, cl_comma
                 buffers[0] = clCreateBuffer(context, flag_set[src_flag_id],  ptrSize * num_elements, NULL, &err);
             if ( err ){
                 print_error(err, " clCreateBuffer failed\n" );
-                clReleaseEvent( event[0] );
-                clReleaseEvent( event[1] );
                 free( (void *)pattern );
                 align_free( (void *)inptr );
                 align_free( (void *)hostptr );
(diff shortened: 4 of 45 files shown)


test_common/harness/parseParameters.cpp
```diff
@@ -172,6 +181,11 @@ int parseCustomParam (int argc, const char *argv[], const char *ignore)
         return -1;
     }
 
+    if (gExecutionMode == kGenerate && (gCompilationCacheMode != kCacheModeOverwrite || gCompilationMode == kOnline))
```
gwawiork: Why in online mode we cannot generate .cl ?

test_conformance/buffers/test_buffer_fill.c
```diff
@@ -779,8 +779,6 @@ int test_buffer_fill_struct( cl_device_id deviceID, cl_context context, cl_comma
                 buffers[0] = clCreateBuffer(context, flag_set[src_flag_id],  ptrSize * num_elements, NULL, &err);
             if ( err ){
                 print_error(err, " clCreateBuffer failed\n" );
-                clReleaseEvent( event[0] );
```
gwawiork: What is the purpose of removing these clReleaseEvent functions in this file ?

test_conformance/geometrics/test_geometrics.cpp
```diff
@@ -265,10 +268,11 @@ int test_geom_cross(cl_device_id deviceID, cl_context context, cl_command_queue
 
     if(!is_extension_available(deviceID, "cl_khr_fp64")) {
         log_info("Extension cl_khr_fp64 not supported; skipping double tests.\n");
-        return 0;
+        return error;
```
gwawiork: you should use enum test_status . In this case TEST_SKIP ?

test_conformance/geometrics/test_geometrics.cpp
```diff
@@ -502,17 +506,16 @@ int test_geom_dot(cl_device_id deviceID, cl_context context, cl_command_queue qu
...
     }
 
-    if (retVal)
-        return retVal;
 
     if(!is_extension_available(deviceID, "cl_khr_fp64"))
     {
         log_info("Extension cl_khr_fp64 not supported; skipping double tests.\n");
-        return 0;
+        return retVal;
```
gwawiork: you should use enum test_status . In this case TEST_SKIP ?

test_conformance/integer_ops/test_abs.c
```diff
@@ -324,6 +326,10 @@ int test_integer_abs(cl_device_id device, cl_context context, cl_command_queue q
         log_info("Failed on %d types\n", fail_count);
         return -1;
     }
+    if(compile_count) {
+        log_info("Failed to compile %d kernels\n", compile_count);
+        return -1;
```
gwawiork: We should start using  enum test_status. See testHarness.h

test_common/harness/parseParameters.h
```diff
@@ -19,6 +19,11 @@
 #include "compat.h"
 #include <string>
 
+enum ExecutionMode
+{
+    kExecute = 0,
+    kGenerate ,
+};
```
kpet: Couldn't this be another compilation mode instead? `kGenerateSource`?

test_conformance/buffers/test_buffer_fill.c
kpet: +1, this looks odd.

test_common/harness/kernelHelpers.c
```diff
@@ -391,6 +395,12 @@ static int get_offline_compiler_output(std::ifstream &ifs,
             return -1;
         }
 
+        if(gExecutionMode == kGenerate) {
+            // in this mode we don't emit the compiler
+            // we fake a compilation error instead
+            return -1;
```
kpet: If we're going to keep that approach, I think it's worth introducing a symbolic name for this error code.

test_conformance/commonfns/test_binary_fn.c
```diff
@@ -141,7 +141,7 @@ int test_binary_fn( cl_device_id device, cl_context context, cl_command_queue qu
         }
         const char *ptr = programSrc;
         err = create_single_kernel_helper( context, &program[ i ], &kernel[ i ], 1, &ptr, "test_fn" );
-        test_error( err, "Unable to create kernel" );
+        test_failure_warning(CL_SUCCESS, err, "Unable to create kernel" );
```
kpet: Using `test_failure_warning` is confusing. Maybe introducing a new error checking macro to deal with this is justified. What you want isn't so much warnings but deferred errors.

test_conformance/commonfns/test_degrees.c
```diff
@@ -79,6 +79,19 @@ const char *degrees3_kernel_code =
...
+    int err;
+    err = test_degrees_float(device, context, queue, n_elems);
+    if( ! is_extension_available( device, "cl_khr_fp64" ) )
+    {
+        log_info( "Skipping double -- cl_khr_fp64 is not supported by this device.\n" );
+    } else {
+        err |= test_degrees_double(device, context, queue, n_elems);
+    }
+    return err;
+}
```
kpet: Nice! This should be a separate patch.

test_conformance/geometrics/test_geometrics.cpp
```diff
@@ -265,10 +268,11 @@ int test_geom_cross(cl_device_id deviceID, cl_context context, cl_command_queue
 
     if(!is_extension_available(deviceID, "cl_khr_fp64")) {
         log_info("Extension cl_khr_fp64 not supported; skipping double tests.\n");
-        return 0;
+        return error;
     } else {
         log_info("Testing doubles...\n");
-        return test_geom_cross_double( deviceID,  context,  queue,  num_elements, seed);
+        error |= test_geom_cross_double( deviceID,  context,  queue,  num_elements, seed);
+        return error;
```
kpet: Have you tried to refactor this in a similar way to what you did for `test_degrees`?

test_conformance/geometrics/test_geometrics.cpp
```diff
@@ -502,17 +506,16 @@ int test_geom_dot(cl_device_id deviceID, cl_context context, cl_command_queue qu
...
     if(!is_extension_available(deviceID, "cl_khr_fp64"))
     {
         log_info("Extension cl_khr_fp64 not supported; skipping double tests.\n");
-        return 0;
+        return retVal;
     }
 
     log_info("Testing doubles...\n");
-    return test_geom_dot_double( deviceID,  context,  queue,  num_elements, seed);
+    retVal |= test_geom_dot_double( deviceID,  context,  queue,  num_elements, seed);
```
kpet: Same question here.

test_conformance/integer_ops/test_abs.c
```diff
@@ -324,6 +326,10 @@ int test_integer_abs(cl_device_id device, cl_context context, cl_command_queue q
         log_info("Failed on %d types\n", fail_count);
         return -1;
     }
+    if(compile_count) {
+        log_info("Failed to compile %d kernels\n", compile_count);
```
kpet: This will print a confusing message when running the tests in source generation mode.

test_conformance/integer_ops/test_abs.c
```diff
@@ -177,7 +177,7 @@ static const char * dest_stores[] = {
 int test_integer_abs(cl_device_id device, cl_context context, cl_command_queue queue, int n_elems)
 {
     cl_int *input_ptr, *output_ptr, *p;
-    int err;
+    int err, compile_count = 0;
```
kpet: s/compile_count/compilation_errrors/

test_conformance/integer_ops/test_int_basic_ops.c
```diff
@@ -319,7 +319,11 @@ cl_int perThreadDataInit(perThreadData * pThis, ExplicitType type,
                                           &(pThis->m_program[ i ]),
                                           &(pThis->m_kernel[ i ]), 1,
                                           &ptr, "test" );
-        test_error( err, "Unable to create test kernel" );
+        if(err) {
```
kpet: It seems you've had to make the same change multiple times. Could the tests be refactored to avoid this?

test_conformance/select/test_select.c
```diff
@@ -582,6 +587,11 @@ int main(int argc, char* argv[])
         log_error( "Failed to allocate memory for argList array.\n" );
         return 1;
     }
+    argc = parseCustomParam(argc, argv);
```
kpet: This should be a separate change (see https://github.com/KhronosGroup/OpenCL-CTS/issues/327). There is a patch on the gitlab `cl21_Arm` branch that we haven't upstream yet IIRC.

test_common/harness/parseParameters.cpp
boazo: We can add this to online compilation  mode as well if this is useful.
I'll add it to the to-do list

test_conformance/buffers/test_buffer_fill.c
boazo: The negative flow of this and a few other tests is faulty (it has been copy pasted). The resources being freed are actually resources that never got allocated at the point of this check. Since the generate sources only mode returns an error (the compilation doesn't succeed), I actually trigger this negative flow and get a segmentation fault when running the conformance in this use case. Please have a look at the entire code base and let me know if I missed something.

test_conformance/geometrics/test_geometrics.cpp
boazo: are you referring to the log_info macro?
I don't think this relates to my change, does it?

test_conformance/geometrics/test_geometrics.cpp
boazo: same as above reply

test_conformance/integer_ops/test_abs.c
boazo: I will try using it in places I modify.

test_common/harness/parseParameters.h
boazo: I am okay with adding this as another compilation mode - will follow up on this

test_common/harness/kernelHelpers.c
boazo: sure, could you point me to where the right place would be to add this symbolic name?

test_conformance/commonfns/test_binary_fn.c
boazo: sure, and agreed

test_conformance/integer_ops/test_abs.c
boazo: ok, so this relates to adding a dedicated macro symbolic name as you suggested above, right?

test_conformance/integer_ops/test_abs.c
boazo: not sure I understood this comment

test_conformance/integer_ops/test_int_basic_ops.c
boazo: need to think about this more carefully

test_conformance/select/test_select.c
boazo: sure

boazo:
Maybe, now that you have seen all of the changes as one big chunk of code. It would make sense to upload patches in the following manner:
1. First patch: introduce the changes in the framework. Most of the test suites would work out of the box so there will already be testable cases to demonstrate this capability
2. Sequence of patches: One patch per test suites that need to be modified. This will allow a focused discussion. 

What do you think?

test_conformance/geometrics/test_geometrics.cpp
boazo: will have a look at it

gwawiork:
I am ok with such splitting

test_conformance/geometrics/test_geometrics.cpp
gwawiork: I mean enum from testHarness.h -> typedef enum test_status . In case extension is not supported we TEST_SKIP error code should be returned

test_conformance/buffers/test_buffer_fill.c
gwawiork: Ok I see the issue and agree with the your proposal. However this should be reported as separate Issue and fix.

test_common/harness/parseParameters.cpp
gwawiork: I think this will help quick debugging Thanks

zakaria6868:
source-only  generation flow looks good , better to upload each conformance executable changes in a separate patch

PR #421, opened 2019-08-05
Move printing sub-test information into test harness

radszy:
This removes all the duplicated code from each test, and moves it to
test harness so that we have single place where this information is
printed.


kpet:
Raised https://github.com/KhronosGroup/OpenCL-CTS/issues/427 so we review the list of required image formats.

PR #429, opened 2019-08-07
Fix parsing of offline compilation options in test_select

StuartDBrady:
The local parsing of arguments in test_select.c throws away any unknown arguments.  Pass these to parseCustomParam() before the local argument parsing, so that the offline compilation options work as expected.

### DIFF
--- a/test_conformance/select/test_select.c
+++ b/test_conformance/select/test_select.c
@@ -58,6 +58,8 @@ static int doTest(cl_command_queue queue, cl_context context,
 
 static void printUsage( void );
 
+static void TestFinishAtExit(void);
+
 //-----------------------------------------
 // Definitions and initializations
 //-----------------------------------------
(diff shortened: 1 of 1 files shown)


test_conformance/select/test_select.c
```diff
@@ -573,8 +575,18 @@ test_definition test_list[] = {
...
 
-int main(int argc, char* argv[])
+int main(int argc, const char* argv[])
 {
+    test_start();
+
+    argc = parseCustomParam(argc, argv);
+    if (argc == -1)
+    {
+        test_finish();
```
kpet: `test_finish` doesn't seem to do much these days but it's all over the place so we might as well keep it here now that you've done the work.

test_conformance/select/test_select.c
```diff
@@ -573,8 +575,18 @@ test_definition test_list[] = {
...
+        return -1;
```
kpet: `EXIT_FAILURE`

PR #430, opened 2019-08-07
Build the harness as a static library

### LINKED ISSUE #193: Build the test harness and other common code as a static library
...and link it into individual binaries. This would shorten the build time, remove one source of mistakes and avoid having to change all CMake files for all tests when adding to the common code (e.g. #133).

kpet:
This reduces by a factor of roughly two the number of files to build.

Fixes #193.


kpet:
The PR succeeded on Travis but the result isn't showing up here for some reason, merging.

PR #434, opened 2019-08-12
Add support for arm printf extension.

Leporacanthicus:
The cl_arm_printf extension uses a call-back to tell the user-code
that some printf output has appeared from the execution of the CL
kernel.

The extsiting Printf tests assume output to stdout.

This change adds code to detect the cl_arm_printf extension, and an
implementation of the callback code to write to stdout.

Also fix a couple of formatting errors and don't "poll" when waiting
for events.

Signed-off-by: Mats Petersson <mats.petersson@arm.com>

### DIFF
--- a/test_conformance/printf/test_printf.c
+++ b/test_conformance/printf/test_printf.c
@@ -43,6 +43,8 @@
 #include "harness/mt19937.h"
 #include "harness/parseParameters.h"
 
+#include <CL/cl_ext.h>
+
 typedef  unsigned int uint32_t;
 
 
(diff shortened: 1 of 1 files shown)


test_conformance/printf/test_printf.c
```diff
@@ -1076,7 +1076,21 @@ test_status InitCL( cl_device_id device )
 
     gFd = acquireOutputStream();
 
-    gContext = clCreateContext(NULL, 1, &device, notify_callback, NULL, NULL);
+    cl_context_properties printf_properties[] =
```
kpet: This could have been called `arm_printf_properties`.

Issue #440, opened 2019-08-21
SVM - svm_fine_grain_memory_consistency fail in case one of SVM devices do not support 64bit atomics 


gwawiork:
The test svm_fine_grain_memory_consistency  takes all visible devices into account while running. In situation when one device on the platform supports extensions and second not it returns failures. This is incorrect situation, all devices should be verified if required extensions are supported.
Solution:
Move checking extension exists for device to function create_cl_objects which discovers capable_devices for the test.

kpet:
PR merged, closing.

PR #441, opened 2019-08-21
Fix verification if atomics 64bit extensions supported by device

gwawiork:
Fix issue #440


gwawiork:
My opinion (I looked through the code) this is the first time we need to check both extensions in the one time

Issue #249, opened 2019-04-30
cl12 cl20 cl21 cl22 bruteforce macro_unary float tests can't fail


jlewis-austin:
(migrated from gitlab)
The TestFloat() function in macro_unary.c has an exit block that overwrites any incoming errors. Failing test cases will be logged but the overall test reports PASSED if the cleanup code completes without error. Test cases affected: isfinite, isinf, isnan, isnormal, signbit, not

kpet:
Fixed on all branches, closing.

PR #444, opened 2019-08-22
Fix for Khronos https://github.com/KhronosGroup/OpenCL-CTS/issues/249

lpavank:
Fix for Khronos https://github.com/KhronosGroup/OpenCL-CTS/issues/249 where incoming errors are overwritten by exit block

Co-contributor jlewis-austin

### DIFF
--- a/test_conformance/math_brute_force/macro_unary.c
+++ b/test_conformance/math_brute_force/macro_unary.c
@@ -396,6 +396,7 @@ static cl_int TestFloat( cl_uint job_id, cl_uint thread_id, void *data )
     int     ftz = job->ftz;
     cl_uint j, k;
     cl_int error = CL_SUCCESS;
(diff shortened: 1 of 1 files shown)


jlewis-austin:
Right, these all need attention.  We just wanted to explicitly address the false positive issue first by bringing this one into line with similar files. Will do a cl12_trunk PR.

test_conformance/math_brute_force/macro_unary.c
```diff
@@ -564,6 +564,7 @@ static cl_int TestFloat( cl_uint job_id, cl_uint thread_id, void *data )
     }
 
 exit:
+    cl_int ret = error;
```
jlewis-austin: This caused a build error in cl12_trunk. I'll push a fix to declare it at the top instead.

Issue #452, opened 2019-09-11
image_streams write tests don't allow flushing half float denormals


jlewis-austin:
Related to #28; this issue covers CL_HALF_FLOAT code paths

jlewis-austin:
Per discussion with kpet, this fix is not trivially portable to cl12_trunk, so will leave this open until 1.2 is validated in master branch.

PR #453, opened 2019-09-12
Allow CL_HALF_FLOAT denorm flushing for write tests (#452)

jlewis-austin:
* On mismatch, add relaxation when denormal half result is expected
* Refactor to use common validation function
* Clean up some diagnostics

### DIFF
--- a/test_common/harness/conversions.h
+++ b/test_common/harness/conversions.h
@@ -117,7 +117,7 @@ static inline int IsDoubleSubnormal( double x )
 #endif
 }
 
-static inline int IsHalfSubnormal( uint16_t x )
+static inline int IsHalfSubnormal( cl_half x )
 { 
     return ( ( x & 0x7fffU ) - 1U ) < 0x03ffU; 
 }
--- a/test_common/harness/conversions.h
+++ b/test_common/harness/conversions.h
@@ -117,6 +117,11 @@ static inline int IsDoubleSubnormal( double x )
 #endif
 }
 
+static inline int IsHalfSubnormal( uint16_t x )
+{ 
+    return ( ( x & 0x7fffU ) - 1U ) < 0x03ffU; 
+}
+
 #if defined(__cplusplus)
 }
 #endif
--- a/test_common/harness/imageHelpers.h
+++ b/test_common/harness/imageHelpers.h
@@ -640,7 +640,14 @@ protected:
 
 extern int  DetectFloatToHalfRoundingMode( cl_command_queue );  // Returns CL_SUCCESS on success
 
-int inline is_half_nan( cl_ushort half ){ return (half & 0x7fff) > 0x7c00; }
+// sign bit: don't care, exponent: maximum value, significand: non-zero
+static int inline is_half_nan( cl_ushort half ){ return ( half & 0x7fff ) > 0x7c00; }
+
+// sign bit: don't care, exponent: zero, significand: non-zero
+static int inline is_half_denorm( cl_ushort half ){ return IsHalfSubnormal( half ); }
+
+// sign bit: don't care, exponent: zero, significand: zero
+static int inline is_half_zero( cl_ushort half ){ return ( half & 0x7fff ) == 0; }
 
 cl_ushort convert_float_to_half( cl_float f );
 cl_float  convert_half_to_float( cl_ushort h );
--- a/test_conformance/images/kernel_read_write/test_iterations.cpp
+++ b/test_conformance/images/kernel_read_write/test_iterations.cpp
@@ -1171,6 +1171,25 @@ bool validate_float_write_results( float *expected, float *actual, image_descrip
     return pass;
 }
 
+bool validate_half_write_results( cl_ushort *expected, cl_ushort *actual, image_descriptor *imageInfo )
+{
+    bool pass = true;
+    // Compare half floats
+    if (memcmp(expected, actual, 2 * get_format_channel_count(imageInfo->format)) != 0) {
+
+        // 8.3.2 Fix up cases where we have NaNs or generated half denormals
+        for ( size_t j = 0; j < get_format_channel_count( imageInfo->format ); j++ ) {
+            if ( is_half_nan( expected[j] ) && is_half_nan( actual[j] ) )
+                continue;
+            if ( is_half_denorm( expected[j] ) && is_half_zero( actual[j] ) )
+                continue;
+            pass = false;
+            break;
+        }
+    }
+    return pass;
+}
+
 int test_read_image_2D( cl_context context, cl_command_queue queue, cl_kernel kernel,
                         image_descriptor *imageInfo, image_sampler_data *imageSampler,
                        bool useFloatCoords, ExplicitType outputType, MTdata d )
--- a/test_conformance/images/kernel_read_write/test_iterations.cpp
+++ b/test_conformance/images/kernel_read_write/test_iterations.cpp
@@ -1155,7 +1155,7 @@ bool validate_float_write_results( float *expected, float *actual, image_descrip
 {
     bool pass = true;
     // Compare floats
-    if( memcmp( expected, actual, 4 * get_format_channel_count( imageInfo->format ) ) != 0 )
+    if( memcmp( expected, actual, sizeof( cl_float ) * get_format_channel_count( imageInfo->format ) ) != 0 )
     {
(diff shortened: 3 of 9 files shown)


Jeremy-Kemp:
Could you resolve the conflicts please? I presume its because of your other PR #456 given that they're very similar changes.

test_conformance/images/kernel_read_write/test_write_image.cpp
```diff
@@ -448,38 +449,22 @@ int test_write_image( cl_device_id device, cl_context context, cl_command_queue
...
-                                log_error( "    Actual:   0x%04x 0x%04x 0x%04x 0x%04x\n", a[0], a[1], a[2], a[3] );
-                                if( inputType == kFloat )
-                                {
-                                    float *p = (float *)(char *)imagePtr;
-                                    log_error( "    Source: %a %a %a %a\n", p[ 0 ], p[ 1 ], p[ 2 ], p[ 3 ] );
-                                    log_error( "          : %12.24f %12.24f %12.24f %12.24f\n", p[ 0 ], p[ 1 ], p[ 2 ], p[ 3 ] );
-                                }
-                                if( ( --numTries ) == 0 )
-                                    return 1;
+                                float *p = (float *)(char *)imagePtr;
```
EwanC: seems like the `(char *)` is unecessary, could just be  `float *p = (float *)imagePtr`?

test_conformance/images/kernel_read_write/test_write_1D_array.cpp
```diff
@@ -401,38 +403,22 @@ int test_write_image_1D_array( cl_device_id device, cl_context context, cl_comma
...
-                                log_error( "    Actual:   0x%04x 0x%04x 0x%04x 0x%04x\n", a[0], a[1], a[2], a[3] );
-                                if( inputType == kFloat )
-                                {
-                                    float *p = (float *)(char *)imagePtr;
-                                    log_error( "    Source: %a %a %a %a\n", p[ 0 ], p[ 1 ], p[ 2 ], p[ 3 ] );
-                                    log_error( "          : %12.24f %12.24f %12.24f %12.24f\n", p[ 0 ], p[ 1 ], p[ 2 ], p[ 3 ] );
-                                }
-                                if( ( --numTries ) == 0 )
-                                    return 1;
+                                float *p = (float *)(char *)imagePtr;
```
EwanC: `float *p = (float *)imagePtr`?

jlewis-austin: That's existing code and is used all over the place, but I can go ahead and fix these instances. I think it probably came from matching the BufferOwningPtr<char> casts, where the double casting is necessary.

jlewis-austin:
> Could you resolve the conflicts please? I presume its because of your other PR #456 given that they're very similar changes.

Will do. I wasn't sure which set would go in first so left them unresolved.

test_conformance/images/kernel_read_write/test_write_1D_array.cpp
EwanC: aah, just stood out in the isolated diff. Thanks 👍

test_common/harness/imageHelpers.h
```diff
@@ -631,7 +631,14 @@ class CoordWalker
...
 
-int inline is_half_nan( cl_ushort half ){ return (half & 0x7fff) > 0x7c00; }
+// sign bit: don't care, exponent: maximum value, significand: non-zero
+static int inline is_half_nan( cl_ushort half ){ return ( half & 0x7fff ) > 0x7c00; }
+
+// sign bit: don't care, exponent: zero, significand: non-zero
+static int inline is_half_denorm( cl_ushort half ){ return IsHalfSubnormal( half ); }
+
+// sign bit: don't care, exponent: zero, significand: zero
+static int inline is_half_zero( cl_ushort half ){ return ( half & 0x7fff ) == 0; }
```
kpet: Please use `cl_half` instead of `cl_short`. These functions are also good candidates to move to the headers.

test_conformance/images/kernel_read_write/test_iterations.cpp
```diff
@@ -1171,6 +1171,25 @@ bool validate_float_write_results( float *expected, float *actual, image_descrip
     return pass;
 }
 
+bool validate_half_write_results( cl_ushort *expected, cl_ushort *actual, image_descriptor *imageInfo )
+{
+    bool pass = true;
+    // Compare half floats
+    if (memcmp(expected, actual, 2 * get_format_channel_count(imageInfo->format)) != 0) {
```
kpet: s/2/sizeof(cl_half) :).

test_conformance/images/kernel_read_write/test_iterations.cpp
```diff
@@ -1171,6 +1171,25 @@ bool validate_float_write_results( float *expected, float *actual, image_descrip
     return pass;
 }
 
+bool validate_half_write_results( cl_ushort *expected, cl_ushort *actual, image_descriptor *imageInfo )
```
kpet: `cl_half`?

test_common/harness/conversions.h
```diff
@@ -117,6 +117,11 @@ static inline int IsDoubleSubnormal( double x )
 #endif
 }
 
+static inline int IsHalfSubnormal( uint16_t x )
+{ 
+    return ( ( x & 0x7fffU ) - 1U ) < 0x03ffU;
```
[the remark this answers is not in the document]
jlewis-austin: It's actually depending on the subtract rolling over to handle the case of 0 not being a subnormal. That's pretty obscure I know, but these conversions are duplicates of those in half/cl_utils.h so I left the  algorithm alone. I definitely agree that the headers would be a better location for these--it seems like any algorithm tweaks would fit better with that change, though?

test_conformance/images/kernel_read_write/test_iterations.cpp
jlewis-austin: This also applies to validate_float_write_results. Should I fix that one as well?

test_common/harness/conversions.h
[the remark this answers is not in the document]
kpet: Oh yes, I read that a bit too quickly. I think we should definitely do the following:
- Move the code from `half/cl_utils.h` to the common code (`test_common/harness/conversions.h` maybe?)
- Add a comment explaning the -1 trick :)

test_conformance/images/kernel_read_write/test_iterations.cpp
kpet: If you don't mind. It's a small-ish thing but could prevent some head-scratching.

test_common/harness/conversions.h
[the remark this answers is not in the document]
kpet: We have several copies of the double and float variants of these functions as well. I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/511 so we don't lose track of the fact that this needs fixing as well.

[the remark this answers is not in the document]
jlewis-austin: Right, I should have created that myself. I was trying to isolate the half/float bug fixes from the refactoring so I extended the existing double/float duplication pattern to half, with the idea that the double/float/half utility consolidation should all be done together in a separate PR. Would it be possible to get this functionality fix in as-is and extend #511 to cover the half issues as well?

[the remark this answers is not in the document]
kpet: Works for me.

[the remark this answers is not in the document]
jlewis-austin: OK, just to confirm, I'll push an update to use sizeof() and document the trick for this instance of IsHalfSubnormal. I'll update #511 to cover the remaining comments.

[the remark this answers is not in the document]
kpet: Fine by me, thanks!

kpet:
@jlewis-austin Where are we with this one? Do you need anything from others to progress?

jlewis-austin:
No, I just got pulled off on other work. Will try to get to it next week.

kpet:
Cool, thanks! (I was also checking that I wasn't the blocker :)).

test_common/harness/imageHelpers.h
jlewis-austin: Added to #511

jlewis-austin:
@kpet @EwanC sorry for the delay. All comments should be addressed or covered under other issues.

PR #461, opened 2019-09-27
Don't use the hadd built-in function in the basic:barrier test

kpet:
It is not strictly required and tested by the integer_ops suite.

This change enables passing the barrier test with clvk+clspv ahead
of adding support for hadd in clspv.

### DIFF
--- a/test_conformance/basic/test_barrier.c
+++ b/test_conformance/basic/test_barrier.c
@@ -40,7 +40,7 @@ const char *barrier_kernel_code =
 "        tmp_sum[tid] += a[i];\n"
 "     \n"
 "     // updated to work for any workgroup size \n"
-"    for (i=hadd(lsize,1); lsize>1; i = hadd(i,1))\n"
+"    for (i=(lsize+1)/2; lsize>1; i = (i+1)/2)\n"
 "    {\n"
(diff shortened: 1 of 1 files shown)


b-sumner:
Do we really need to check workarounds into the trunk?  Why can't this stay in a branch?

kpet:
The reason I'm proposing it for inclusion instead of keeping it on my stack of workarounds is that it gets us closer to unit-type testing. The built-in function under test is `barrier` here.

b-sumner:
Seems like converting existing conformance tests to be more unit-like is an independent project.  Are we meaning to start that here?  The conformance tests are already pretty simplistic and their occasional use of features not strictly being tested has been valuable.

kpet:
>Seems like converting existing conformance tests to be more unit-like is an independent project.

Yes and no. I don't see any of us doing this just for the sake of it. I think it's more likely to happen gradually as part of work towards other goals.

>Are we meaning to start that here?

Why not? :)

>The conformance tests are already pretty simplistic and their occasional use of features not strictly being tested has been valuable.

Agree, I've found that to be valuable as well in some cases. We do need to be careful with this but as long as we check that there is adequate coverage otherwise (which I believe is the case here), I think we shouldn't be afraid.

Issue #462, opened 2019-10-01
test_svm svm_enqueue_api - possible crash


gwawiork:
Looks like C++ pitfall. Need to exchange
`auto check_event : eventMemFillList`
to 
`auto &check_event : eventMemFillList`

kpet:
PR merged, closing.

PR #463, opened 2019-10-01
Fix Issue 462 - test_svm crash

gwawiork:
Fix Issue #462


kpet:
@gwawiork One small comment though: for the issue to be automatically closed when the PR is merged you need to use `Fixes #462`. `Fix Issue #462` doesn't work. See https://help.github.com/en/articles/closing-issues-using-keywords for a detailed guide.

PR #467, opened 2019-10-02
Fix inconsistent dll linking warnings for VS 2015/17 (#466)

### LINKED ISSUE #466: Fix windows "inconsistent dll linkage" build warnings
With VS2015 and above, building the CTS produces several MB worth of warnings pertaining to duplicate definitions, e.g.
`7>C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\ucrt\corecrt_math.h(576): note: see previous definition of 'rintf'
7>C:\wsl\ocl_cts\OpenCL-CTS\test_common\harness/compat.h(158): warning C4273: 'rintl': inconsistent dll linkage`

Affected functions should be removed from compat.h and msvc9.c for newer MSVC versions.

jlewis-austin:
-- #ifdef out duplicate definitions
-- remove make_nan() dead code

Fixes #466

### DIFF
--- a/test_common/harness/compat.h
+++ b/test_common/harness/compat.h
@@ -151,83 +151,83 @@ typedef long long           int64_t;
         #define     isinf( _x)      ((_x) == INFINITY || (_x) == -INFINITY)
     #endif
         
-#if ! defined( __INTEL_COMPILER )
+    #if _MSC_VER < 1900 && ! defined( __INTEL_COMPILER )
 
-double rint( double x);
-float  rintf( float x);
-long double rintl( long double x);
+        double rint( double x);
+        float  rintf( float x);
+        long double rintl( long double x);
 
-float cbrtf( float );
-double cbrt( double );
+        float cbrtf( float );
+        double cbrt( double );
 
-int    ilogb( double x);
-int    ilogbf (float x);
-int    ilogbl(long double x);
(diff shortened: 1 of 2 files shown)


jlewis-austin:
BTW, I noticed that test_conversions and test_geometrics still compile with debug flags. Some of the implementations removed here (e.g. rintf) use algorithms that look like they could be brittle to optimizations. I'm wondering if those flags can be removed for the newer compilers at least (or better yet fix the affected algorithms)? I don't know how to verify that on all possible compilers though...

gwawiork:
Could you give me one day more for deep verification please :) ? In general looks ok

kpet:
@gwawiork Thanks!

@jlewis-austin I'm tempted to suggest we remove the debug flags and start from there. They shouldn't be hardcoded, especially if we can't tie their presence to a specific compiler/issue.

jlewis-austin:
> I'm tempted to suggest we remove the debug flags and start from there. They shouldn't be hardcoded, especially if we can't tie their presence to a specific compiler/issue.

@kpet I agree; will open a new issue for that.

gwawiork:
verification passed. can be merged

kpet:
@jlewis-austin @gwawiork  Thanks! Merging.

PR #473, opened 2019-11-10
Remove dead code from test_userevents

kpet:
This code probably predates the move to the test framework and is
no longer useful.


kpet:
This is a trivial removal of simple commented out code, merging.

PR #474, opened 2019-11-10
Use clCreateCommandQueue when possible (in a few tests)

kpet:
In all these cases, the new entrypoint is not necessary. These
changes enable the tests to work or are a necessary step to get
the tests to work on an OpenCL 1.2 implementation.

While this may not be the final approach we want to solve this
specific compatibility issue, it also has the nice property of
reducing the diff with cl12_trunk until we merge.


kpet:
@aarongreig  Thanks for the review. Merging.

PR #475, opened 2019-11-10
Mark SVM and pipes tests as requiring OpenCL 2.0

kpet:


kpet:
Thanks for the review!

PR #481, opened 2019-11-10
Fix delete calls

kpet:
Both allocations are made with new[].


kpet:
@alycm  Thanks for the reviews!

PR #483, opened 2019-11-10
select: don't forget to count tests when failing to create programs

kpet:
A test writer really shouldn't have to think about this...


kpet:
Thanks for the review, merging.

PR #485, opened 2019-11-11
A few error handling and flag test fixes for mem_host_flags

kpet:
- Add missing error checks
- Add missing return
- Fix flag check


kpet:
Thanks for the review, merging.

PR #487, opened 2019-11-14
Fixes #486 - add build option for devices OCL2.X

### LINKED ISSUE #486: test_compiler - compile kernel as opencl C 2.0 code for devices 2.0 and higher
I noticed issues when test compiler runs case compiler_defines_for_extensions on device OCL2.1 . No build option "-cl-std=CL2.0" cause that kernel is compiled as OpenclC 1.2 This should be fixed

gwawiork:
Fixes #486


kpet:
Thanks!

PR #490, opened 2019-11-18
harness: select queue creation function based on OpenCL version

kpet:
Use clCreateCommandQueue before OpenCL 2.0.

This enables most of the binaries to run to completion on a
1.2 implementation.


kpet:
Thanks for the review!

Issue #494, opened 2019-11-19
Get rid of compatibility test suites


kpet:
As discussed within Khronos, these suites are mostly duplicates of the api and basic suites. They can be removed for the most part. The valuable tests (testing deprecated APIs) should be moved to the main api and basic suites.

PR #495, opened 2019-11-19
Remove exact duplicates from the compatibility suites

kpet:
Contributes to #494.


Jeremy-Kemp:
Are we restricting this to be _exact_ only? If not, tests like api/test_binary are very similar as well and could be removed.

kpet:
No. I just started with exact duplicates because they're the easiest to spot. I'll be going from most obvious to least obvious and, yes, I agree that many other tests should be removed.

test_conformance/compatibility/test_conformance/basic/CMakeLists.txt
```diff
@@ -39,16 +37,11 @@ set(${MODULE_NAME}_SOURCES
...
     test_async_copy.cpp
     test_sizeof.c
-    test_vector_creation.cpp
     test_vec_type_hint.c
-    test_numeric_constants.cpp
     test_constant_source.cpp
     test_bufferreadwriterect.c
     test_async_strided_copy.cpp
-    test_preprocessors.cpp
     test_kernel_memory_alignment.cpp
```
Jeremy-Kemp: This one only has white space diffs. We could remove that too?

kpet: I will remove all of these in the next change. I'm trying to slice this into small-ish, (kind of) reviewable chunks.

kpet:
Thanks for the review!

PR #500, opened 2019-11-22
Fixes issue #499 - Test basic from master branch on OCL 1.2 device

### LINKED ISSUE #499: Test basic from master branch on OCL 1.2 device
Work related with Issue #259
Need to set ADD_TEST_VERSION for:
    ADD_TEST_VERSION( wg_barrier, Version(2, 0) ),
    ADD_TEST_VERSION( progvar_prog_scope_misc, Version(2, 0) ),
    ADD_TEST_VERSION( progvar_prog_scope_uninit, Version(2, 0) ),
    ADD_TEST_VERSION( progvar_prog_scope_init, Version(2, 0) ),
    ADD_TEST_VERSION( progvar_func_scope, Version(2, 0) ),
    ADD_TEST_VERSION( global_linear_id, Version(2, 0)  ),
    ADD_TEST_VERSION( local_linear_id , Version(2, 0) ),
    ADD_TEST_VERSION( enqueued_local_size, Version(2, 0) ),
    ADD_TEST_VERSION( simple_read_image_pitch, Version(2, 0) ),
    ADD_TEST_VERSION( simple_write_image_pitch, Version(2, 0) ),
    ADD_TEST_VERSION( get_linear_ids, Version(2, 0) ),
    ADD_TEST_VERSION( rw_image_access_qualifier, Version(2, 0) ),

Need to fix test_basic sizeof - do not use "-cl-std=CL2.0" build option

gwawiork:
Fixes #499

### DIFF
--- a/test_conformance/basic/main.c
+++ b/test_conformance/basic/main.c
@@ -148,8 +148,8 @@ test_definition test_list[] = {
     ADD_TEST_VERSION( local_linear_id, Version(2, 0) ),
     ADD_TEST_VERSION( enqueued_local_size, Version(2, 0) ),
 
-    ADD_TEST_VERSION( simple_read_image_pitch, Version(2, 0) ),
-    ADD_TEST_VERSION( simple_write_image_pitch, Version(2, 0) ),
+    ADD_TEST( simple_read_image_pitch ),
+    ADD_TEST( simple_write_image_pitch ),
 
 #if defined( __APPLE__ )
     ADD_TEST( queue_priority ),
--- a/test_conformance/basic/main.c
+++ b/test_conformance/basic/main.c
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -63,7 +63,7 @@ test_definition test_list[] = {
     ADD_TEST( mri_multiple ),
     ADD_TEST( image_r8 ),
     ADD_TEST( barrier ),
-    ADD_TEST( wg_barrier ),
+    ADD_TEST_VERSION( wg_barrier, Version(2, 0) ),
     ADD_TEST( int2float ),
     ADD_TEST( float2int ),
     ADD_TEST( imagereadwrite ),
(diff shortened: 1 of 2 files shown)


test_conformance/basic/test_sizeof.c
```diff
@@ -49,8 +49,15 @@ cl_int get_type_size( cl_context context, cl_command_queue queue, const char *ty
...
+    size_t major = 0;
+    size_t minor = 0;
+    bool deviceLt20 = false;
+    int error = get_device_version(device, &major, &minor);
+    if (major < 2) {
+        deviceLt20 = true;
+    }
 
-    cl_int err = create_single_kernel_helper_with_build_options(context, &p, &k, 4, sizeof_kernel_code, "test_sizeof", "-cl-std=CL2.0");
+    cl_int err = create_single_kernel_helper_with_build_options(context, &p, &k, 4, sizeof_kernel_code, "test_sizeof", deviceLt20 ? "" : "-cl-std=CL2.0");
```
kpet: As far as I can tell with a quick read of the code, the only thing that requires CL2 in this test is the `void*` pointer case. The compatibility suite has got `global void*` (which was needed for one implementation). I'm fine with what you've done but we'll need to make the creation of the table of pointer types conditional to the version when folding the basic compatibility suite into the main basic suite.

test_conformance/basic/main.c
```diff
@@ -136,27 +136,27 @@ test_definition test_list[] = {
...
-    ADD_TEST( local_linear_id ),
-    ADD_TEST( enqueued_local_size ),
+    ADD_TEST_VERSION( global_linear_id, Version(2, 0) ),
+    ADD_TEST_VERSION( local_linear_id, Version(2, 0) ),
+    ADD_TEST_VERSION( enqueued_local_size, Version(2, 0) ),
 
-    ADD_TEST( simple_read_image_pitch ),
-    ADD_TEST( simple_write_image_pitch ),
+    ADD_TEST_VERSION( simple_read_image_pitch, Version(2, 0) ),
+    ADD_TEST_VERSION( simple_write_image_pitch, Version(2, 0) ),
```
kpet: Why do the two `{read,write}_image_pitch` require CL2? I couldn't spot anything obvious skimming the code.

gwawiork: My assumption was that this test was not present in cl12_trunk test suite. But you are right there is no reason why it should not be enabled. Moreover we have this test on internal branch enabled for CL1.2.

PR #503, opened 2019-11-25
Fixes issue #502 - Bruteforce and Integer ops tests from master branch on OCL 1.2 device

### LINKED ISSUE #502: Bruteforce and Integer ops tests from master branch on OCL 1.2 device
Work related with Issue #259
Need to disable build option "-cl-std=CL2.0"

gwawiork:
Fixes #502

### DIFF
--- a/test_conformance/integer_ops/test_integers.cpp
+++ b/test_conformance/integer_ops/test_integers.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


test_conformance/math_brute_force/main.c
```diff
@@ -1745,9 +1755,18 @@ int MakeKernels( const char **c, cl_uint count, const char *name, cl_uint kernel
...
+    size_t minor = 0;
+    error = get_device_version(gDevice, &major, &minor);
+    if (error != CL_SUCCESS)
+    {
+        vlog_error("\t\tFAILED -- Failed to get devcie version. (%d)\n", error);
+        return error;
+    }
+    if (major >= 2) {
+        strcpy(options, "-cl-std=CL2.0");
+    }
```
kpet: A couple of comments:
- This duplicates even more code. We should implement `MakeKernels` using `MakeKernel` before making further changes.
- I couldn't spot any new built-in function added on master compared to cl12_trunk. Do we even need to pass `-cl-std=CL2.0` at all?

test_conformance/integer_ops/test_integers.cpp
```diff
@@ -85,10 +85,23 @@ int test_single_param_integer_kernel(cl_command_queue queue, cl_context context,
...
+    error = get_device_version(deviceID, &major, &minor);
+    if (error != CL_SUCCESS)
+    {
+        vlog_error("\t\tFAILED -- Failed to get devcie version. (%d)\n", error);
+        return error;
+    }
+
+    if (major < 2) {
+        deviceLt20 = true;
+    }
```
kpet: This approach won't work. The `integer_ctz` test (the only one to require 2.x AFAICT) will run and its kernel fail to build. The only change needed I think is to mark this test as requiring 2.x.

Also, for future reference, please use `get_device_cl_version` instead of `get_device_version` as this removes the need to handle errors.

kpet:
@gwawiork Do you intend to split/update that PR?

gwawiork:
Yes I do :)

kpet:
Ok, thanks :).

test_conformance/math_brute_force/main.c
gwawiork: I assume that each test which is build on device 2.X should add to this flag to the build options? Am I wrong ?

kpet: Well, it's only really required for programs that use 2.x features.

gwawiork:
I will open new PRs splitted with changes suggested in review.

Issue #508, opened 2019-12-01
Object queries tests should be exhaustive where possible


kpet:
Some tests for object queries are currently only testing one value which means that an implementation that returns a fixed value would pass the test. In many of these cases the tests could be exhaustive and go over all combinations without any noticeable impact on run time. Examples include:
- [ ] get_sampler_info (and compatibility version)
- [ ] get_command_queue_info (and compatibility version)

We should probably also do the following as part of this work:
- [ ] Turn `TEST_COMMAND_QUEUE_PARAM` macros into functions and/or templates
- [ ] Reassess whether querying/comparing `CL_DEVICE_VENDOR_ID` to compare devices is a good thing to do. What is wrong with just comparing the `cl_device_id`?
- [ ] Merge `test_queries.cpp` and  `test_queries_compatibility.cpp`.
- [ ] Find a problem to the core vs. extension properties discussed in https://github.com/KhronosGroup/OpenCL-CTS/pull/666

PR #509, opened 2019-12-01
Move sampler and command queue compatibility query tests to the main api suite

kpet:
...and mark tests that exercise new-style entrypoints as requiring 2.0

These tests give us (or rather will give us once made exhaustive, see
#508) coverage for both new-style and old-style creation entrypoints,
making the entrypoint used in other tests irrelevant as long as the
old-style entrypoints are required to be implemented.

Also get rid of duplicate compatibility tests that were in the same file.

Contributes to #494.


kpet:
This is just shuffling code around as agreed on the WG calls, merging to get forward progress.

PR #510, opened 2019-12-01
Use old-style sampler creation in basic suite and remove duplicate compatibility tests

kpet:
As agreed in the WG calls, the query tests are sufficient to cover
both old-style and new-style sampler and command queue creation
entrypoints.

Use old-style entrypoints in all tests that don't require 2.x
features to get compatibility with 1.x implementations.
This makes it possible to remove duplicate compatibility tests.

Contributes to #494.


kpet:
The changes are trivial, merging this to get forward progress.

PR #521, opened 2019-12-17
Move cl_khr_create_command_queue test to the main api suite

kpet:
In theory, it could exist on a 2.x implementation.


gwawiork:
Looks good, I think this test is available for 1.2 too

kpet:
Thanks for the review! Yes no version restriction on this one.

Issue #532, opened 2019-12-23
New extended subgroups - test development


gwawiork:
This issue is to cover test development of the new extensions:

- [x] cl_khr_subgroup_extended_types
- [ ] cl_khr_subgroup_non_uniform_vote
- [ ] cl_khr_subgroup_ballot
- [ ] cl_khr_subgroup_non_uniform_arithmetic
- [ ] cl_khr_subgroup_shuffle
- [ ] cl_khr_subgroup_shuffle_relative
- [ ] cl_khr_subgroups_clustered_reductions

rjodinchr:
Fixed: Dedicated test implementations for all 7 subgroup extensions are fully implemented and registered in `test_conformance/subgroups/main.cpp`.

PR #533, opened 2019-12-23
Extended subgroups - Add support for extension cl_khr_subgroups_extended_types

gwawiork:
Contributes to #532 
Initial version. Add support for extension cl_khr_subgroups_extended_types

### DIFF
--- a/test_conformance/subgroups/subhelpers.h
+++ b/test_conformance/subgroups/subhelpers.h
@@ -158,6 +158,9 @@ run_kernel(cl_context context, cl_command_queue queue, cl_kernel kernel, size_t 
     error = clEnqueueWriteBuffer(queue, in, CL_FALSE, 0, isize, idata, 0, NULL, NULL);
     test_error(error, "clEnqueueWriteBuffer failed");
 
+    error = clEnqueueWriteBuffer(queue, xy, CL_FALSE, 0, msize, mdata, 0, NULL, NULL);
+    test_error(error, "clEnqueueWriteBuffer failed");
+
     error = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, &global, &local, 0, NULL, NULL);
     test_error(error, "clEnqueueNDRangeKernel failed");
 
@@ -189,7 +192,7 @@ struct test {
         clProgramWrapper program;
         clKernelWrapper kernel;
         cl_platform_id platform;
-        cl_int sgmap[2*GSIZE];
+        cl_int sgmap[4*GSIZE];
         Ty mapin[LSIZE];
         Ty mapout[LSIZE];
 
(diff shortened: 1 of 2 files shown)


test_conformance/subgroups/subhelpers.h
```diff
@@ -22,60 +22,737 @@
...
+    struct cl_half3 {
+        ::cl_half3 data;
+    };
+    struct cl_half4 {
+        ::cl_half4 data;
+    };
+    struct cl_half8 {
+        ::cl_half8 data;
+    };
+    struct cl_half16 {
```
kpet: Why do we need these special types?

test_conformance/subgroups/test_workgroup.cpp
```diff
@@ -40,7 +40,30 @@ static const char * bcast_source =
...
+//"printf(\"gid = %d, sub group local id = %d, sub group id = %d, x form in = %d, new_set = %d, out[gid] = %d , x = %d\\n\",gid,xy[gid].x, xy[gid].y, x, xy[gid].z, out[gid], x);"
+"}\n";
+static const char * bcast_first_source =
+"__kernel void test_bcast_first(const __global Type *in, __global int4 *xy, __global Type *out)\n"
+"{\n"
+"    int gid = get_global_id(0);\n"
+"    XY(xy,gid);\n"
+"    Type x = in[gid];\n"
+"    out[gid] = sub_group_broadcast_first(x);\n"
+//"printf(\"gid = %d, sub group local id = %d, sub group id = %d, x form in = %d, new_set = %d, out[gid] = %d , x = %d\\n\",gid,xy[gid].x, xy[gid].y, x, xy[gid].z, out[gid], x);"
```
kpet: These changes shouldn't be in this PR.

test_conformance/subgroups/subhelpers.h
gwawiork: This is because compilation errors. When one type is used to define another one.
For example the vector cl_uint3 is defined as follows (in cl_platform.h):
typedef  cl_uint4  cl_uint3;
or cl_half is defined as:
typedef unsigned __int16        cl_half;

Using next this types in below templates cause error : type is already defined.

gwawiork:
> Thanks for pushing this! My initial reaction is that it's a bit on the big side for a single PR. A few comments:
> 
> * I can't tell the purpose of the first commit. This will at least need a commit message.
> * The second commit contains changes that don't relate to cl_khr_subgroups_extended_types. These should be removed.
> * There are a number of trivial (whitespace, float->cl_float, etc) changes. Maybe we could land these  first to reduce the amount of noise? Although there aren't that many of them...
> * I'm concerned with the maintainability of all these templates which seem to duplicate a lot of code. Have you considered refactoring the existing test to use data tables for all the types? This should make extending the list of types pretty simple.


* This is needed in further implementation - need to switch to vector of 4 elements. Then I have ability to send information to work item kernel e.g which broadcast id should be. Previously solution when generated data and broadcast id is the same value was problematic.
* ok I will prepare once again but not sure which you mean are not related with cl_khr_subgroups_extended_types 
* ok I will try to separate them
* regarding refactoring. I was thinking about this but I was lost in this Kingdom of Templates :) Because there was a lot of work with this group of subgroup extensions  I decided first implement with adjustment to existing code. My recommendation is to leave it as is and create new task for refactoring. I think code duplication is hard to reduce in this case.

PR #535, opened 2020-01-06
Fix max limit query for test_3D_2D_array.

nikhiljnv:
test_3D_2D_array uses max limits of 3D images
for both 3D images and 2D_array.
This leads to testing of sizes larger than
2D array limits.
Use correct CL_MEM_OBJECT_IMAGE_TYPE to avoid
test-cases with invalid sizes.


nikhiljnv:
> 
> 
> LGTM, thanks! `get_max_sizes` is a horrible function :(. Is this fix needed on `cl12_trunk` as well?

Thanks for the quick reply. This PR is for cl12_trunk. This will be needed in master as well. Will create a separate PR for the same.

kpet:
Yes, that's what I meant. Thanks!

PR #536, opened 2020-01-07
Fix max limit query for test_3D_2D_array.

nikhiljnv:
test_3D_2D_array uses max limits of 3D images
for both 3D images and 2D_array.
This leads to testing of sizes larger than
2D array limits.
Use correct CL_MEM_OBJECT_IMAGE_TYPE to avoid
test-cases with invalid sizes.


test_conformance/device_execution/enqueue_flags.cpp
```diff
@@ -54,7 +54,7 @@ static const char* enqueue_flags_wait_kernel_simple[] =
     NL, ""
     NL, "  if((index + 1) == BITS_DEPTH)"
     NL, "  {"
-    NL, "    barrier(CLK_LOCAL_MEM_FENCE);"
+    NL, "    barrier(CLK_GLOBAL_MEM_FENCE);"
```
kpet: Hmm, I'm guessing you didn't intend these changes to be part of the PR, did you?

nikhiljnv: Looks like there was some diff between Khronos master and my fork which led to these unnecessary changes. I have merged Khronos:master to my local fork.
Let me know if anything else is messed up.

PR #538, opened 2020-01-08
Build CTS on windows or linux - Python script proposal

### LINKED ISSUE #537: Build CTS on windows or linux - Python script proposal
I am not sure if this is proper place to have such build script. The script is especially dedicated for windows but can build binaries on Linux as well.

gwawiork:
Fixes #537

### DIFF
--- a/build_win_lnx.py
+++ b/build_win_lnx.py
@@ -1,0 +1,176 @@
+import argparse
+import os
+import shutil
+import subprocess
+from sys import platform
+
+'''
+    Assuming the script is placed inside the cloned OpenCL-CTS directory. 
+    Script is able to build whole OpenCL-CTS suite on Windows (MSVC 2019, MSVC 2017, icc 2019) 
+    or Linux operating system (gcc)
+    Requirements:
+        python 3.X, cmake, git, MSVC 2017/2019 (Windows), GCC (Linux)
+    Examples:
+        #Build whole suite using default settings
+        build_win_lnx.py
+        #Build whole suite using MSVC 2019 on Windows, binaries in bitness x86
+        build_win_lnx.py --compiler msvc --bitness x86 --msvc_version 2019
+        #As above but: use bitness x64 and do not clone repositories
+        build_win_lnx.py --compiler msvc --bitness x64 --msvc_version 2019 --skip_clone
+        #As above but: build only test_half. Do not build ICD
+        build_win_lnx.py --compiler msvc --bitness x64 --msvc_version 2019 --skip_clone –build_target test_half --skip_icd
+'''
+
+
+root_dir = os.getcwd()
+
+
+def run():
+    args = process_command_line()
+    compiler = args.compiler
+    msvc_version = args.msvc_version
+    bitness = args.bitness
+    skip_clone = args.skip_clone
+    skip_icd = args.skip_icd
+    skip_cts = args.skip_cts
+    build_target = args.build_target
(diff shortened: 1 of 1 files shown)


build_win_lnx.py
```diff
@@ -0,0 +1,176 @@
+import argparse
```
EwanC: `#!/usr/bin/python3`  ?

build_win_lnx.py
```diff
@@ -0,0 +1,176 @@
...
+def clean_path(path):
+    if os.path.exists(path):
+        print("Cleaning path: ", path)
+        shutil.rmtree(path)
+    else:
+        print("The path: ", path, " does not exist")
+
+
+def process_command_line():
+    parser = argparse.ArgumentParser()
```
EwanC: a `description` argument here could he helpful

build_win_lnx.py
```diff
@@ -0,0 +1,176 @@
...
+    parser.add_argument('--msvc_version', default="2017", choices=['2017', '2019'],
+                        help='choose msvc version')
+    parser.add_argument('--compiler', default="msvc", choices=['msvc', 'icc'],
+                        help='choose compiler')
+    parser.add_argument('--bitness', default="x64", choices=['x64', 'x86'],
+                        help='choose bitness')
+    parser.add_argument('--skip_clone', action='store_true',  help='skip cloning repositories')
+    parser.add_argument('--skip_icd', action='store_true',  help='skip building ICD project')
+    parser.add_argument('--skip_cts', action='store_true', help='skip building CTS project')
+    parser.add_argument('--build_target', default="ALL_BUILD", help='target to build (only Windows)')
```
EwanC: if this option only works on Windows can this `parser.add_argument` be wrapped in a `platform == 'win32':` so it users don't accidentally expect behaviour from it? Same for the `--msvc_version` and `--compiler` options.

build_win_lnx.py
```diff
@@ -0,0 +1,176 @@
...
+        elif bitness == "x86" and msvc_version == "2019":
+            icd_cmd = icd_solution_cmd_win_x86_msvc_2019
+            if compiler == "msvc":
+                cts_cmd = cts_solution_cmd_win_x86_msvc_2019
+            elif compiler == "icc":
+                cts_cmd = cts_solution_cmd_win_x86_icc_2019
+
+    elif platform[:5] == "linux":
+        print("Linux operating system")
+        compiler = "gcc"
```
EwanC: Is this always true, or could the compiler be `clang` if an environment var like `CC/CXX` is set?

gwawiork: I've added only support for gcc. Feel free to extend support for any compiler you want

EwanC: Only supporting gcc is fine, was mostly concernced about the script printing `Compiler: gcc` when another compiler was actually being used by cmake. But now that message is guarded to windows only on line 117, so avoiding the potential for misleading output.

lakshmih:
We would like to review this and will post comments shortly.

lakshmih:
We would like to stick to cmake-based builds rather than a build script. We recommend moving any necessary logic into the CMake build and documentation to the readme (refer https://github.com/KhronosGroup/OpenCL-CTS/pull/1021)

nikhiljnv:
Any reason we need separate scripts and can not reuse or build on top of https://github.com/KhronosGroup/OpenCL-CTS/blob/master/presubmit.sh ?

gwawiork:
There are already scripts in repo

PR #544, opened 2020-01-10
Fixes #498 - Extend progvar_prog_scope_uninit checks

### LINKED ISSUE #498: Test program scope variable with default initialiser
This issue relates to KhronosGroup/OpenCL-Docs#163, which proposes to clarify what is the default value of program-scope/static variables in the global address space when no explicit initialiser is given.

The CTS should most likely be updated to ensure implementation zero-initialise those variables.

mantognini:
This adds checks to progvar_prog_scope_uninit which ensure that
program-scope variables are default-initialized to the zero-value for
their type.

### DIFF
--- a/test_conformance/basic/test_progvar.cpp
+++ b/test_conformance/basic/test_progvar.cpp
@@ -1,6 +1,6 @@
 //
-// Copyright (c) 2017 The Khronos Group Inc.
-// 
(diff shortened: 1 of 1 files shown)


kpet:
At least three implementers have run the test and are happy with it. Since this aligns with standard C behaviour, I'm merging this to get forward progress.

PR #552, opened 2020-01-15
Fixed build errors for aarch64 platforms

bhargavdas:
While building for aarch64 platforms the test suites math_brute_force,
contractions and conversions fails as it cannot find the expansion of
"SYS__sysctl" macro.

So ignoring the block containing "SYS__sysctl" macro if the platform is
aarch64.

Signed-off-by: bhargavdas <bhargav_das@mentor.com>

### DIFF
--- a/test_conformance/CMakeLists.txt
+++ b/test_conformance/CMakeLists.txt
@@ -94,3 +94,13 @@ add_subdirectory(vec_step)
 
 # Add any extension folders
 add_subdirectory(spir)
+
+# Copy all csv files
+FILE(GLOB CSV_FILES
+ 	 "${CMAKE_CURRENT_SOURCE_DIR}/*.csv"
+)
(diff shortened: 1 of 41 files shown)


bhargavdas:
> Thanks for the patch! I'm not sure why you don't have the definition on AArch64. What toolchain are you using?
> Two comments:
> 
> * We have three copies of that code. Would you mind moving the code to a file in `test_common/harness`?
> * The code should use `uname` instead of `sysctl`. This would be simpler and more portable.

* I'm using `arm-poky-linux-gnueabi-gcc` toolchain which is available for aarch64.
* So just for some more clarification, should I create a new file inside `test_common/harness` or is there any file available in which I can append the code?

kpet:
> should I create a new file inside test_common/harness or is there any file available in which I can append the code?

I'd move the code to `test_common/harness/deviceInfo.cpp` or `test_common/harness/testHarness.c`. A lot of that code needs reshuffling anyway so it doesn't matter too much. The important bit is really to make the code common.

PR #558, opened 2020-01-16
Remove headers tests from the conformance tests

kpet:
These tests have been moved to the headers repository. None of these
tests exercise the OpenCL implementation. They are just testing the
headers and host toolchain.

Whether they are required to build and run as part of the conformance
process will be discussed under
https://github.com/KhronosGroup/OpenCL-Headers/issues/58.


kpet:
Merging, as agreed during the F2F meeting.

Issue #517, opened 2019-12-12
image_streams tests don't handle CL_DEVICE_TYPE_DEFAULT correctly


jlewis-austin:
The tests have codepaths specific to GPU devices, but can fail if the CL_DEVICE_TYPE_GPU isn't explicitly defined via the command line or environment variable. This works properly in cl12_trunk via a clGetDeviceInfo call to get the actual device type, but the call is missing in the master branch.

kpet:
Thanks for reporting this. I've added it to the checklist for the merge of cl12_trunk into master so we don't lose track of this.

kpet:
Even the code on cl12_trunk isn't fully correct: implementations that return GPU|DEFAULT as the device type for example aren't handle correctly.

kpet:
Fixed by https://github.com/KhronosGroup/OpenCL-CTS/pull/611 (and previous changes). Remaining issues around comparing the device type for equality covered by https://github.com/KhronosGroup/OpenCL-CTS/issues/574.

PR #560, opened 2020-01-17
checkDeviceTypeOverride will now detect a cl device if not given one

jiabaxie:
Previously multiple test_image_stream.exe tests were failing because the test would not use the right device type if not given one. Now, when the test uses checkDeviceTypeOverride to check for device types, it now detects the device type in addition to previously checking for environmental variables. 

All image tests passed when ran.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -27,6 +27,8 @@ RoundingMode gFloatToHalfRoundingMode = kDefaultRoundingMode;
 static cl_ushort float2half_rte( float f );
 static cl_ushort float2half_rtz( float f );
 
+cl_device_type gDeviceType = CL_DEVICE_TYPE_DEFAULT;
+
 double
 sRGBmap(float fc)
 {
--- a/test_conformance/api/main.c
+++ b/test_conformance/api/main.c
@@ -28,7 +28,6 @@
 // FIXME: To use certain functions in harness/imageHelpers.h
 // (for example, generate_random_image_data()), the tests are required to declare
 // the following variables (<rdar://problem/11111245>):
-cl_device_type gDeviceType = CL_DEVICE_TYPE_DEFAULT;
 bool gTestRounding = false;
 
 test_definition test_list[] = {
--- a/test_conformance/basic/main.c
+++ b/test_conformance/basic/main.c
@@ -28,7 +28,6 @@
 // FIXME: To use certain functions in harness/imageHelpers.h
 // (for example, generate_random_image_data()), the tests are required to declare
 // the following variables (<rdar://problem/11111245>):
-cl_device_type gDeviceType = CL_DEVICE_TYPE_DEFAULT;
 bool gTestRounding = false;
 
 test_definition test_list[] = {
--- a/test_conformance/clcpp/images/main.cpp
+++ b/test_conformance/clcpp/images/main.cpp
@@ -22,7 +22,6 @@
 // FIXME: To use certain functions in test_common/harness/imageHelpers.h
 // (for example, generate_random_image_data()), the tests are required to declare
 // the following variable (hangover from code specific to Apple's implementation):
-cl_device_type gDeviceType = CL_DEVICE_TYPE_DEFAULT;
 
 int main(int argc, const char *argv[])
 {
--- a/test_conformance/images/clCopyImage/main.cpp
+++ b/test_conformance/images/clCopyImage/main.cpp
@@ -36,7 +36,6 @@ bool gTestMipmaps;
 int gTypesToTest;
 cl_channel_type gChannelTypeToUse = (cl_channel_type)-1;
 cl_channel_order gChannelOrderToUse = (cl_channel_order)-1;
-cl_device_type    gDeviceType = CL_DEVICE_TYPE_DEFAULT;
 
 extern int test_image_set( cl_device_id device, cl_context context, cl_command_queue queue, MethodsToTest testMethod );
 
--- a/test_conformance/images/clFillImage/main.cpp
+++ b/test_conformance/images/clFillImage/main.cpp
@@ -34,7 +34,6 @@ bool gEnablePitch;
 int  gTypesToTest;
 cl_channel_type  gChannelTypeToUse = (cl_channel_type)-1;
 cl_channel_order gChannelOrderToUse = (cl_channel_order)-1;
-cl_device_type   gDeviceType = CL_DEVICE_TYPE_DEFAULT;
 
 extern int test_image_set( cl_device_id device, cl_context context, cl_command_queue queue, MethodsToTest testMethod );
 static void printUsage( const char *execName );
--- a/test_conformance/images/clGetInfo/main.cpp
+++ b/test_conformance/images/clGetInfo/main.cpp
@@ -31,7 +31,6 @@ bool gTestMaxImages;
 bool gTestRounding;
 int  gTypesToTest;
 cl_channel_type gChannelTypeToUse = (cl_channel_type)-1;
-cl_device_type    gDeviceType = CL_DEVICE_TYPE_DEFAULT;
 
 extern int test_image_set( cl_device_id device, cl_context context, cl_mem_object_type image_type );
 static void printUsage( const char *execName );
--- a/test_conformance/images/clReadWriteImage/main.cpp
+++ b/test_conformance/images/clReadWriteImage/main.cpp
@@ -34,7 +34,6 @@ bool gTestMipmaps;
 int  gTypesToTest;
 cl_channel_type gChannelTypeToUse = (cl_channel_type)-1;
 bool            gEnablePitch = false;
-cl_device_type    gDeviceType = CL_DEVICE_TYPE_DEFAULT;
 
 #define MAX_ALLOWED_STD_DEVIATION_IN_MB        8.0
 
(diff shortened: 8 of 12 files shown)


Jeremy-Kemp:
The change looks to get whatever the default device is, however checkDeviceTypeOverride is passed CL_DEVICE_TYPE_DEFAULT by all tests that call it.

Are you setting CL_DEVICE_TYPE in your environment anywhere that could cause the problem you're seeing?

TheFlyingSushi:
Hi,
No I have not set a CL_DEVICE_TYPE in my environment anywhere. 
The image tests have different validation logic depending on the device type, and CL_DEVICE_TYPE_DEFAULT is too generic for some of the subtests. One of the failing subtests easily reproducible by running "test_image_streams.exe read int CL_R CL_SIGNED_INT8 CL_FILTER_NEAREST CL_ADDRESS_REPEAT NORMALIZED 1D".

jiabaxie:
Hi,
No I have not set a CL_DEVICE_TYPE in my environment anywhere.
The image tests have different validation logic depending on the device type, and CL_DEVICE_TYPE_DEFAULT is too generic for some of the subtests. One of the failing subtests easily reproducible by running "test_image_streams.exe read int CL_R CL_SIGNED_INT8 CL_FILTER_NEAREST CL_ADDRESS_REPEAT NORMALIZED 1D".

jlewis-austin:
It looks like this should address https://github.com/KhronosGroup/OpenCL-CTS/issues/517

jiabaxie:
Hi, is there any updates on this review? Or is there any problems with this commit? Thanks

test_common/harness/testHarness.c
```diff
@@ -797,8 +797,43 @@ void checkDeviceTypeOverride( cl_device_type *inOutType )
             *inOutType = CL_DEVICE_TYPE_ACCELERATOR;
         else if( strcmp( force_cpu, "CL_DEVICE_TYPE_DEFAULT" ) == 0 )
             *inOutType = CL_DEVICE_TYPE_DEFAULT;
-    }
-
+    } else {
+        //Detect the device type
+		cl_platform_id  platform;
+		int error = clGetPlatformIDs(1, &platform, NULL);
```
bashbaug: I think this should check for `CL_PLATFORM_INDEX` vs. unconditionally choosing the first platform?

jiabaxie:
Hi, I'm not too clear on the changes you've requested.
- Calling checkDeviceTypeOverride in test harness doesn't set gDeviceType, which is the variable used by test_read_1D.cpp. It'd be difficult to set gDeviceType from testharness.c
- What specifically should i do to have the framework decide the functionality of command line?
- How do i get the device type from the queue passed by the test framework? Or if i get the test to provide the device type doesn't the test need to enumerate platforms and devices again?

kpet:
So, looking at this a bit more, things are even worse than I thought. `gDeviceType` has no bearing on what device type gets selected. It's a side mechanism for tests to know the device type. The value set there has to match the type of the actual device that was selected elsewhere for the tests to be correct. There's a lot of code to remove and change to fix all of that cleanly but only a small number of them are required to fix your issue (one step per PR to keep things reviewable). Here's what I think is a plausible plan:
1. Define and declare `gDeviceType` once in the common code. I would move this to `test_common/harness/imageHelpers.{cpp,h}`.
1. Get rid of `checkDeviceTypeOverride` (possibly moving some of the logging) and set `gDeviceType` in `runTestHarnessWithCheck` from `device_type`. At this point, your issue should be fixed.
1. Remove all code setting `gDeviceType` in tests, possibly still accepting the command-line parameters for a transition period.
1. Remove all command line parameters used to declared a device type.
1. Have tests get the device type from the queue provided by the framework or the device(s) they selected themselves.
1. Modify the image helpers that use `gDeviceType` to take a device or device type as an argument.
1. Remove the last remnants of `gDeviceType`.

jiabaxie:
Hi,
I've moved gDeviceType to imageHelper.cpp/.h and it seems to be ok, but I have a bit of an issue trying to set gDeviceType in checkDeviceTypeOverride, in testharness.c., as testharness.c does not have an include for imagehelper.h, and including that brought some unresolved external symbols. Should I try resolve these symbols, or should something else be done?

kpet:
Hi,

>I've moved gDeviceType to imageHelper.cpp/.h and it seems to be ok

Great! I suggest you create a PR already for that change alone. This is unambiguously good and we can get this out of the way.

>but I have a bit of an issue trying to set gDeviceType in checkDeviceTypeOverride, in testharness.c.

Hmm, why did you keep `checkDeviceTypeOverride`?

>including that brought some unresolved external symbols. and other conflicts. Should I try resolve these symbols and conflicts, or should something else be done?

It's hard for me to help without seeing the exact issues/conflicts. Could you comment with some specifics?

jiabaxie:
Hi,
Sorry for the clutter above. I misclicked a few things there.
A new PR was created for the first step here: https://github.com/KhronosGroup/OpenCL-CTS/pull/575.
When I mentioned checkDeviceTypeOverride, I mean runTestHarnesswithCheck. I mixed those two up by mistake. I've talked with someone about the issue I was having, and I think I have a solution for step 2. I will try to finish that, and report if I have any other issues.

kpet:
No worries!

jiabaxie:
Hello,
I have an update on step 2. Currently I have what it seems to an issue between header files. In order to access the variable gDeviceType from imageHelper.cpp/.h from testharness.c, I have written an include statement. There were several attempts to then assign gDeviceType, but they have all errored.

The first attempt was to define gDeviceType as a global in testHarness.c and then assign it in runTestHarnessWithCheck, but the build errored with gDeviceType already defined.

The second attempt was to not define gDeviceType as a global, and just straight up assign it in the function. However, the build errored with unresolved external symbol "bool gTestRounding", which is declared as an extern in imageHelper.cpp.

I have also tried other ways to access gDeviceType, such as declaring an extern cl_device_type gDevicetype in testharness.cpp or .h, but anytime it seems try to access the _correct_ gDeviceType, comes with a conflict of gTestRounding.

How should I proceed with this conflict?

kpet:
Hi,

I'm not sure I understand the problem. What's wrong with including `imageHelpers.h` in `testHarness.c`?

jiabaxie:
Hi, 
  
I'm not exactly sure what exactly are you not sure about. When I try to access gDeviceType from imageHelpers.h, when building, it errors with a unresolved external symbol "bool gTestRounding". It is a strange situation as gTestRounding was not touched at all. 

In addition to the #include, Here's the change I've done to assign gDevicetype, just for testing purposes:
gdevicetype
8-d449c7530f81.png)

jiabaxie:
Ok, I've pinpointed the issue.

gTestRounding is declared as an extern used in a function in imageHelper.cpp, set somewhere else. Whatever includes imageHelpers.h, has access to that function (being generate_random_image_data), therefore needs gTestRounding defined and set. 
When I try to include imageHelpers.h in testharness.c, not all that use testharness.c (usually using the function runTestHarnesswithCheck) has a gTestRounding set, and consequently erroring due to gTestRounding not set for generate_random_image_data.

Setting gTestrounding in testharness.c does not work, as it yields an error saying the variable "gTestRounding" already declared. The brute force way is to set gTestRounding in every file that has an instance where testharness.c is used but has not yet set gTestRounding. This solves the issue, but produces redundant code. 

Is there a better way you think could resolve this issue better?

kpet:
Have you tried to move `gTestRounding` to `imageHelper.{cpp,h}`? Looking around, it seems that this should work.

jiabaxie:
I'm looking into that, but I ran into more errors as I solve gTestRounding:
harness.lib(imageHelpers.obj) : error LNK2005: "int __cdecl DetectFloatToHalfRoundingMode(struct _cl_command_queue *)" (?DetectFloatToHalfRoundingMode@@YAHPEAU_cl_command_queue@@@Z) already defined in utils.obj [C:\Users\jiabaxie\Deskto
p\ConformTest\OpenCL-CTS\build\test_extensions\media_sharing\test_media_surface_sharing.vcxproj]

There's a couple of these errors with the same message, namingly:
int __cdecl DetectFloatToHalfRoundingMode(struct _cl_command_queue *)
unsigned short __cdecl convert_float_to_half(float)
float __cdecl convert_half_to_float(unsigned short)

I'm not too sure how to resolve these.

kpet:
Hmm, we have two copies of that function (https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_extensions/media_sharing/utils.cpp#L1588 and https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/imageHelpers.cpp#L2929). They've diverged but looking at the diff it seems the copy in the media sharing tests just wasn't updated as helpers were introduced in the common code. Two options:
1. Fully solve the issue now: get rid of the copy in the media sharing tests and use the one from the image helpers there. This probably should be a separate pre-requisite change.
2. Hack your way around it and defer the full resolution: rename the function in the media sharing tests and create an issue to address this.

kpet:
@jiabaxie Are you still working on this issue?

jiabaxie:
Hi kpet,

Yes I am. I recently had to switch locations and I got hung up on some setup. My plan is to attempt to solve the issue and if too much trouble comes across I will create an issue to address it instead.

Thanks

kpet:
Great, thanks!

jiabaxie:
Ok, I took a look at these functions, and they are identical except DetectFloatToHalfRoundingMode, which has a small amount of difference. Simply removing these three functions from utils.cpp seems ok. No errors from building occurred. Do you know a reliable way of testing that the tests using utils.cpp functions correctly?
After moving gTestRounding to imageHelpers.cpp/.h everything seemed ok. I've set gDeviceType in testharness.c. I will make a new PR for this shortly
Thanks

PR #571, opened 2020-01-22
test_svm - enqueue_api - lack of synchronization points

### LINKED ISSUE #570: test_svm - enqueue_api - lack of synchronization points
There is a need to add additional clEnqueueSVMMap synchronization point to be sure that data is visible on device and host in case of coarse-grained SVM

gwawiork:
Fixes #570

### DIFF
--- a/test_conformance/SVM/test_enqueue_api.cpp
+++ b/test_conformance/SVM/test_enqueue_api.cpp
@@ -198,10 +198,10 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
         }
       }
       clEventWrapper eventUnmap[2];
-      error = clEnqueueSVMUnmap(queue, srcBuffer, 0, 0, &eventUnmap[0]);
+      error = clEnqueueSVMUnmap(queue, srcBuffer, 0, nullptr, &eventUnmap[0]);
       test_error(error, "clEnqueueSVMUnmap srcBuffer failed");
 
-      error = clEnqueueSVMUnmap(queue, dstBuffer, 0, 0, &eventUnmap[1]);
+      error = clEnqueueSVMUnmap(queue, dstBuffer, 0, nullptr, &eventUnmap[1]);
       test_error(error, "clEnqueueSVMUnmap dstBuffer failed");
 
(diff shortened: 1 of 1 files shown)


Jeremy-Kemp:
Is it enough to guarantee that we're dealing with coarse grained buffers due to the call to create_cl_objects?

If so, LGTM.

gwawiork:
If other types of SVM would be supported nothing wrong. The test (the same specification)  requires at least coarse grain to be supported by device.

Jeremy-Kemp:
I was just thinking out loud that we'd have overkill synchronisation if the test were suddenly to test fine grained buffers, but it's not an issue really!

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -177,11 +177,16 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
...
       test_error(error, "clEnqueueSVMMemcpy failed");
 
-      clEventWrapper eventMap;
-      error = clEnqueueSVMMap(queue, CL_FALSE, CL_MAP_READ | CL_MAP_WRITE, dstBuffer, elementNum * typeSizes[i], 1, &eventMemcpy, &eventMap);
-      test_error(error, "clEnqueueSVMMap failed");
+      //coarse grain only supported. Synchronization required using map
+      clEventWrapper eventMap[2];
 
-      error = clWaitForEvents(1, &eventMap);
+      error = clEnqueueSVMMap(queue, CL_FALSE, CL_MAP_READ | CL_MAP_WRITE, srcBuffer, elementNum * typeSizes[i], 1, &eventMemcpy, &eventMap[0]);
```
kpet: `CL_MAP_WRITE` is not necessary though it's harmless. You could also introduce a variable for `elementNum * typeSizes[i]`, it's used in quite a few places.

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -192,10 +197,12 @@ int test_svm_enqueue_api(cl_device_id deviceID, cl_context c, cl_command_queue q
             return TEST_FAIL;
         }
       }
+      clEventWrapper eventUnmap[2];
+      error = clEnqueueSVMUnmap(queue, srcBuffer, 0, 0, &eventUnmap[0]);
```
kpet: Please use `nullptr` instead of `0`.

PR #572, opened 2020-01-29
Remove last compatibility tests

### LINKED ISSUE #494: Get rid of compatibility test suites

kpet:
They were a duplicate of the non-compatibility version.

The non-compatibility version assumed that BGRA formats aren't
required for implementations that support the embedded profile,
while the compatibility version didn't.

The unified specification doesn't currently document any format
requirements for the embedded profile but the OpenCL 1.2 specification
did and no BGRA format was part of the list. The consensus from the
Working Group is that BGRA formats are not a requirement of the embedded
profile and the specification will be changed to reflect this
(see https://github.com/KhronosGroup/OpenCL-Docs/issues/201).

Closes #494.


kpet:
Thanks for the review!

PR #575, opened 2020-02-04
Moved all instances of gDeviceType to imageHelper.cpp

jiabaxie:
This is the first step in trying to refactor gDeviceType, from comments in this Closed PR: https://github.com/KhronosGroup/OpenCL-CTS/pull/560

gDeviceType is now moved to imageHelper.cpp, and only in imageHelper.cpp/.h are they declared and defined.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -27,6 +27,8 @@ RoundingMode gFloatToHalfRoundingMode = kDefaultRoundingMode;
 static cl_ushort float2half_rte( float f );
 static cl_ushort float2half_rtz( float f );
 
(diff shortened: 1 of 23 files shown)


jiabaxie:
Hi kpet,
I'm not sure what you mean. I did a mass search on the entire list of CTS files and the only location I found extern cl_device_type gDeviceType was in imageHelpers.h. Perhaps there's a few I missed? Since I can't seem to find any. This also bring me to the next question: what do you mean by add one to imageHelpers.h? There is already extern cl_device_type gDeviceType in there.
Thanks.

kpet:
>but my second question remains: there's already a extern cl_device_type gDeviceType in imageHelpers.h, what did you mean by add one?

Hadn't spotted that one. No need to add it again if it's already there.

PR #576, opened 2020-02-07
Computeinfo: optional images capabilities

darmroz:
* according to the specification part of
device capabilities depend on CL_DEVICE_IMAGE_SUPPORT
* behaviour regarding lack of images support is not
defined
* this change skips checking images related capabilities
 when they are not supported

### DIFF
--- a/test_conformance/computeinfo/main.c
+++ b/test_conformance/computeinfo/main.c
@@ -109,15 +109,17 @@ typedef struct
     const char*           opcode_name;
     int                   config_type;
     config_data           config;
+    bool                  images_support_required;
 } config_info;
 
-#define CONFIG_INFO( major, minor, opcode, type ) { { major, minor }, opcode, #opcode, type_ ## type, { 0 } }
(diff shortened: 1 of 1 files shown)


alycm:
Intuitively I would expect these queries to return `0` in the case that `CL_DEVICE_IMAGE_SUPPORT` is `CL_FALSE`.

If being pedantic I'm not sure that needs to be specified, if images are supported there is a mandated minimum, if images are not supported you return whatever number is accurate (very likely to be `0`).

But that would still allow you to query the property and not get an error code, you just can't pass/fail the test based on what the value returned is.

kpet:
Agree with @alycm. My reading of the spec is that these queries are always supported but the spec only constrains the value when `CL_DEVICE_IMAGE_SUPPORT` returns `CL_TRUE`.

Should we make this explicit, require that implementations return `0` when `CL_DEVICE_IMAGE_SUPPORT` returns `CL_FALSE` and test for this?

darmroz:
Thanks, I think specification could be more detailed about that but I see your point and agree.

PR #578, opened 2020-02-08
Pipes optional

boazo:


test_conformance/pipes/main.c
```diff
@@ -15,72 +15,94 @@
...
+  cl_uint max_packet_size;
+  error = clGetDeviceInfo(device, CL_DEVICE_PIPE_MAX_PACKET_SIZE,
+                          sizeof(max_packet_size), &max_packet_size, NULL);
+  if (error != CL_SUCCESS) {
+    print_error(error, "Unable to get pipe max packet size");
+    return TEST_FAIL;
+  }
+
+  if (max_packet_size == 0) {
+    return TEST_SKIP;
```
kpet: Hmm, just spotted another problematic case: a reported size of zero should fail the test on 2.0-2.2 implementations, not skip it. I suggest adding ` && (version > Version(2,2))` to the condition.

PR #581, opened 2020-02-08
Subgroups optional

boazo:


test_conformance/subgroups/main.cpp
```diff
@@ -41,6 +41,20 @@ static test_status checkSubGroupsExtension(cl_device_id device)
         return TEST_SKIP;
     }
 
+    cl_uint max_sub_groups;
+    int error;
+
+    error = clGetDeviceInfo(device, CL_DEVICE_MAX_NUM_SUB_GROUPS,
```
kpet: This query doesn't exist prior to 2.1. The whole block needs to be run only for >2.2 and moved lower down.

kpet:
This needs to be rebased.

PR #597, opened 2020-02-15
set gDeviceType in testharness.c

jiabaxie:
There's a couple of things I did to achieve the main goal of setting gDeviceType in testharness.c
1. I included imagehelpers.h to testharness.c in order to access gDeviceType
2. I moved gTestRounding to imageHelpers.cpp/.h as other files that uses testharness had a duplicate definition of gTestRounding. I also removed all mentions of extern gTestRounding as that was not necessary.
3. I deleted the three functions because they were an outdated version of a duplicate in imagehelpers.cpp/.h:
int __cdecl DetectFloatToHalfRoundingMode(struct _cl_command_queue *)
unsigned short __cdecl convert_float_to_half(float)
float __cdecl convert_half_to_float(unsigned short)

This PR was made in a series of PRs for managing gDeviceType. Original thread: https://github.com/KhronosGroup/OpenCL-CTS/pull/560#issuecomment-585289636


test_common/harness/testHarness.c
```diff
@@ -328,6 +320,24 @@ int runTestHarnessWithCheck( int argc, const char *argv[], int testNum, test_def
     }
 
     device = devices[choosen_device_index];
+	err = clGetDeviceInfo(device, CL_DEVICE_TYPE, sizeof(device_type), &device_type, NULL);
```
kpet: You don't need to query the type of the device since the device was selected above from the type.

test_common/harness/testHarness.c
```diff
@@ -328,6 +320,24 @@ int runTestHarnessWithCheck( int argc, const char *argv[], int testNum, test_def
...
+	switch (device_type)
+	{
+	case CL_DEVICE_TYPE_GPU:            log_info("Requesting GPU device "); break;
+	case CL_DEVICE_TYPE_CPU:            log_info("Requesting CPU device "); break;
+	case CL_DEVICE_TYPE_ACCELERATOR:    log_info("Requesting Accelerator device "); break;
+	case CL_DEVICE_TYPE_DEFAULT:        log_info("Requesting Default device "); break;
+	default:                            log_error("Requesting unknown device "); return EXIT_FAILURE;
+	}
+	log_info(based_on_env_var ? "based on environment variable " : "based on command line ");
+	log_info("for platform index %d and device index %d\n", choosen_platform_index, choosen_device_index);
```
kpet: I think you didn't need to move that block of code.

test_common/harness/testHarness.c
jiabaxie: The selection of the device type above this query was done by getting the environmental variables and looking at command line arguments, but in the case that none are set, I think it needs to be queried so that it doesn't simply run with CL_DEVICE_TYPE_DEFAULT if it is not specified.

jiabaxie: However, the query might need to happen before testharness checks for command line arguments and env vars

kpet: Yes, there are indeed cases where the query is required but even with the query the code can't work in the general case as most tests are testing for equality between `gDeviceType` and a specific device type (like CPU or GPU) but the `CL_DEVICE_TYPE` query could return a combination of these (CPU | GPU | DEFAULT is apparently legal for example) so ultimately we will need to change all uses to check a single bit and fail or request command line info in non-sensical cases. I'll raise a spec bug to clarify whether that was intended.

jiabaxie: Let me know what specific action I should take to address that, since I think there should be something done about that

kpet: Yes, it will take a while to fix all of that. @JohnKesapidesARM will add the query in #611.

PR #529, opened 2019-12-20
Support for fp16 in bruteforce tests

gwawiork:
Fixes #528

PR #600, opened 2020-02-19
Permit half overflow within allowable ULP

EwanC:
Modify the algorithm for calculating half precision ULP error so
that it duplicates the behaviour of the [single precision ULP algorithm](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/errorHelpers.c#L407-L428),
in regards to allowing overflow within the defined ULP error.

In the case where the test value is infinity, but the reference is
finite, pretend the test value is 63336.0 and calculate the ULP error
against that.

Encountered this while testing half precision `hypot()` in [PR !529](https://github.com/KhronosGroup/OpenCL-CTS/pull/529),
for inputs `hypot(-48864.0, 43648.0)` which has reference
`65519.755799`. With RTE rounding this only just rounds to `65504` as half,
and returning INF is currently infinite ULP error. Using the leniency
introduced by this change however the error is `~0.5` within the `2` ULP
bounds defined by the spec.

### DIFF
--- a/test_common/harness/errorHelpers.cpp
+++ b/test_common/harness/errorHelpers.cpp
@@ -307,8 +307,6 @@ const char *GetQueuePropertyName(cl_command_queue_properties property)
 #define scalbnl(_a, _i )    ldexpl( _a, _i )
 #endif
 
-static float Ulp_Error_Half_Float( float test, double reference );
-
 // taken from math tests
 #define HALF_MIN_EXP    -13
 #define HALF_MANT_DIG    11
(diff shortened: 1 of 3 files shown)


jlewis-austin:
It looks like it's at least as valid as the existing float hack. Probably should pair this with a spec clarification because the current wording doesn't seem to allow for these cases.

EwanC:
Thanks for giving this a look @jlewis-austin, attempted some matching spec language here https://github.com/KhronosGroup/OpenCL-Docs/pull/231

jlewis-austin:
I guess this doesn't need to wait on the spec change since it's just making the behavior consistent with floats. LGTM

bashbaug (approved):
LGTM - the logic added here for fp16 matches the existing logic for fp32.

Issue #595, opened 2020-02-13
De-duplicate and improve PrintArch functions


kpet:
See https://github.com/KhronosGroup/OpenCL-CTS/pull/552

kpet:
Proposed PR against cl12_trunk: https://github.com/KhronosGroup/OpenCL-CTS/pull/596

bhargavdas:
New PR proposed for master #601

kpet:
Solved by #601

PR #601, opened 2020-02-20
Improved PrintArch functions (#595)

bhargavdas:
Moved all duplicate 'PrintArch' functions to testHarness.c
Replaced sysctl system call used in 'PrintArch' function with uname
system call

Signed-off-by: bhargavdas <bhargav_das@mentor.com>

### DIFF
--- a/test_common/harness/testHarness.cpp
+++ b/test_common/harness/testHarness.cpp
@@ -28,6 +28,8 @@
 #include "typeWrappers.h"
 #include "imageHelpers.h"
 #include "parseParameters.h"
+#include <sys/utsname.h>
+#include <sys/sysctl.h>
 
 #if !defined(_WIN32)
 #include <unistd.h>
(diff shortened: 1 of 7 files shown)


test_common/harness/testHarness.c
```diff
@@ -945,3 +946,53 @@ Version get_device_cl_version(cl_device_id device)
...
+
+    int type = 0;
+    size_t typeSize = sizeof( type );
+    sysctlbyname( "hw.cputype", &type, &typeSize, NULL, 0 );
+    vlog( "cpu type:\t%d\n", type );
+    typeSize = sizeof( type );
+    sysctlbyname( "hw.cpusubtype", &type, &typeSize, NULL, 0 );
+    vlog( "cpu subtype:\t%d\n", type );
+
+#elif defined( __linux__ ) && !defined(__aarch64__)
```
kpet: Can you remove the second condition?

bhargavdas:
> Many thanks for porting the patch quickly. It seems the change doesn't build on macOS (see https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/sysctlbyname.3.html). I think you're probably missing some headers (see https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/sysctlbyname.3.html)

Yes I have also observed that. Looks like 'sys/sysctl.h' header is required.

test_common/harness/testHarness.c
bhargavdas: > Can you remove the second condition?

Ok.

kpet: This comment still needs resolving.

kpet:
There's one comment left to address and the change needs to be rebased but, apart from that, it's good to go.

test_common/harness/testHarness.c
```diff
@@ -945,3 +947,53 @@ Version get_device_cl_version(cl_device_id device)
...
+    vlog( "ARCH:\tppc64\n" );
+#elif defined( __PPC__ )   
+    vlog( "ARCH:\tppc\n" );
+#elif defined( __i386__ )   
+    vlog( "ARCH:\ti386\n" );
+#elif defined( __x86_64__ )   
+    vlog( "ARCH:\tx86_64\n" );
+#elif defined( __arm__ )   
+    vlog( "ARCH:\tarm\n" );
+// Add 64 bit support
```
Jeremy-Kemp: Tiny nit-pick, but this comment could be removed.

test_common/harness/testHarness.c
bhargavdas: > This comment still needs resolving.

Done

test_common/harness/testHarness.c
bhargavdas: > Tiny nit-pick, but this comment could be removed.

Done

kpet:
The change is currently conflicting though, you still need to rebase it.

bhargavthriler:
> The change is currently conflicting though, you still need to rebase it.



> The change is currently conflicting though, you still need to rebase it.

Ok, I'll rebase as soon as possible

bhargavdas:
> > The change is currently conflicting though, you still need to rebase it.
> 
> Ok, I'll rebase as soon as possible

Done

PR #602, opened 2020-02-20
Get rid of checkDeviceTypeOverride

kpet:
This is now done in the test harness.


kpet:
Thanks for the review, done.

Issue #470, opened 2019-11-05
Test matrix


kpet:
We don't currently have an agreed set of supported test configurations for the conformance tests. From previous discussions, it seems the following parameters and values are of interest:

Supported build environments:
- Linux
- macOS
- Windows
- Android

Supported targets:
- x86 (32/64-bit)
- Arm (32/64-bit)

Supported toolchains:
- GCC
- Clang
- MSVC
- MinGW
- ICC ( Intel® C++ Compiler ) - https://software.intel.com/en-us/c-compilers

Anything else? Any thoughts?

gwawiork:
Hi 
Regrading toolchains/compiler we could add ICC ( Intel® C++ Compiler ) - https://software.intel.com/en-us/c-compilers

EwanC:
Could clarify the `Arm (32/64-bit)`  point as v7 & v8, rather than aarch32 & aarch64

Jeremy-Kemp:
Is there any interest in Android being a supported environment? We do have the Android build script as part of CTS.

kpet:
Yes, I'll include Android.

PR #603, opened 2020-02-20
Fix warnings reported by -Wconversion-null

kpet:
Fix error reporting in clFillImage as a result.


kpet:
Thanks for the review! FWIW, this was enabled by default in my environment. I'd rather not add too many compiler-specific flags before we have https://github.com/KhronosGroup/OpenCL-CTS/issues/470 in place.

PR #611, opened 2020-02-21
Global variable gDeviceType correctly set

JohnKesapidesARM:
Add function to set gDeviceType  on kernel_read_write tests(image_streams)
as the device type affects the way validation is performed in the test.


JohnKesapidesARM:
Reworking the patch atm..

PR #612, opened 2020-02-21
Failure in copy_images max_images

JohnKesapidesARM:
An existing workaround on the max_image size calulcation that disallows
the width of an image to be less than 16 can stress the calculcation of
the remainder dimension to be less than 1.0 in size. In three dimentional
objects (3d,2d array) where one dimention is set to max and the other is
set to 16 there might not be enough space left for the 3rd one.
This workaround clamps the third dimension to a minimum of 1.0

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -15,6 +15,7 @@
 //
 #include "imageHelpers.h"
 #include <limits.h>
+#include <assert.h>
 #if defined( __APPLE__ )
 #include <sys/mman.h>
 #endif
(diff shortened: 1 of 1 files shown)


JohnKesapidesARM:
I'd be happy to add that. My only concern is that if any the other two dimensions M and x0 are 0 then that would probably be a problem with the CL Implementation rather than the test-code. Would an assert for the other two dimensions work better?

Jeremy-Kemp:
Yes, you're right. It looks like both M and x0 are the result of various calls to `clGetDeviceInfo`.

Given that there isn't a way to error from `get_max_sizes`, an assert would be a good compromise.

kpet:
We could also throw an exception in the error cases.

JohnKesapidesARM:
@kpet I opted for an assertion for now, because looking for assertions in the tests, there are some classes defined in the spirv tests. We should probably first move those under test_common first, so that can be used widely in the tests.

test_common/harness/imageHelpers.cpp
```diff
@@ -15,6 +15,7 @@
 //
 #include "imageHelpers.h"
 #include <limits.h>
+#include <assert.h>
```
kpet: Should use `<cassert>` but not a big deal.

PR #614, opened 2020-02-21
enable Appveyor Windows builds

### LINKED ISSUE #11: Add support for Windows continuous integration using Appveyor

bashbaug:
Enables AppVeyor Windows builds, using similar build steps as for Travis Linux and OSX builds.  This will help to prevent, or at least identify, future Windows build breaks.

Example build output:

* https://ci.appveyor.com/project/bashbaug/opencl-cts

I've enabled builds for Visual Studio 2017, for 32-bit and 64-bit builds, for Release builds only, for now at least.  It's reasonably straightforward to add support for other builds if we so desire.

Fixes #11 (I think).

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -3,10 +3,6 @@ os:
 
 shallow_clone: true
 
-branches:
-  except:
-    - gh-pages
-
 platform:
   - Win32
   - x64
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,51 @@
+os:
+  - Visual Studio 2017
+
+# do not build on tags
+#skip_tags: true
+
+# clone only the top level commit
+shallow_clone: true
+
+# branches to build: all but pages
+branches:
+  except:
+    - gh-pages
+
+platform:
+  - Win32
+  - x64
+
+configuration:
+  #- Debug
+  - Release
+
+before_build:
+  # Setup environment:
+  - ps: $env:TOP = $env:APPVEYOR_BUILD_FOLDER
+  - ps: $env:TOP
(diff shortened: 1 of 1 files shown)


.appveyor.yml
```diff
@@ -0,0 +1,51 @@
...
+# do not build on tags
+#skip_tags: true
+
+# clone only the top level commit
+shallow_clone: true
+
+# branches to build: all but pages
+branches:
+  except:
+    - gh-pages
```
kpet: copy/paste?

.appveyor.yml
```diff
@@ -0,0 +1,51 @@
+os:
+  - Visual Studio 2017
+
+# do not build on tags
+#skip_tags: true
```
kpet: remove?

.appveyor.yml
bashbaug: I don't see a copy-paste error here, but I did remove the comment since it wasn't adding much.  Let me know if there's still a bug somewhere I'm just not seeing.

.appveyor.yml
bashbaug: Yes - fixed.

bashbaug:
I've rebased these changes and it looks like builds are passing again.

.appveyor.yml
kpet: We don't have a `gh-pages` branch and I don't think we want any branch exclusions at the moment. I would remove the whole `branches` tree.

bashbaug: Sure, no problem, I've removed this whole section.  If we ever do add github pages for the CTS we should add it as a branch exclusion, though.

PR #615, opened 2020-02-23
Make it possible to run kernel_read_write on a 1.x implementation

kpet:


gwawiork:
LGTM. I noticed similar issue [here ](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/gl/test_images_write_common.cpp#L372)

kpet:
Thanks for the review! Yes, I haven't touch interop suites at all. There is work needed there.

PR #622, opened 2020-02-24
CMake Build: Tidy up when -msse2 is passed to gcc

Jeremy-Kemp:
Conversions and brute force only check when compiling for Android. This also needs to occur for ARM Linux builds to ensure that msse2 is not passed to the compiler.


Jeremy-Kemp:
My intent was to get the build working without changing the existing behaviour as much as possible.

I can remove `CMAKE_ARM_COMPILER` entirely and add `-msse2` globally for x86/amd64 architectures (based on `CMAKE_SYSTEM_PROCESSOR`) if that'd be more desirable.

Do you know why the current guard for msse2 is just `ANDROID`? Wondering if that can also be removed so we rely entirely on x86/amd64.

kpet:
Understood. I don't believe the current code is the result of careful conscious decisions :). That's why I'd rather start from something clean and generic and deal with potential problems as they arise. Worst case, we can always revert...

Jeremy-Kemp:
I've made the requested changes and given them a brief test with bruteforce and basic. Haven't seen anything untoward.

Issue #623, opened 2020-02-24
BGRA Images are Unconditionally Tested


Jeremy-Kemp:
readimage3d in basic is unconditionally testing BGRA images. This format is not a requirement for Embedded Profile implementations, so this should be relaxed.

Jeremy-Kemp:
Fixed in #624. Closing.

Issue #427, opened 2019-08-05
The tables of required image formats is incomplete


kpet:
See `check_minimum_supported`.

Jeremy-Kemp:
#629 tackles read _or_ write. Read _and_ write still needs working on.

bashbaug:
I am closing this issue since the remaining task (checking for read-and-write images) is tracked in #639.

PR #624, opened 2020-02-24
Conditionally test BGRA in Basic readimage3d (#623)

Jeremy-Kemp:
This change adds checks to see if testing against an embedded implementation and if so, queries whether BGRA is supported or not.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -3811,6 +3811,26 @@ void build_required_image_formats(cl_mem_flags flags,
 	}
 }
 
+bool is_image_format_required(cl_image_format format,
+                              cl_mem_flags flags,
+                              cl_mem_object_type image_type,
+                              cl_device_id device)
+{
+	std::vector<cl_image_format> formatsToSupport;
+	build_required_image_formats(flags, image_type, device, formatsToSupport);
+
+	for (auto &formatItr: formatsToSupport)
+	{
+		if (formatItr.image_channel_order == format.image_channel_order &&
+		    formatItr.image_channel_data_type == format.image_channel_data_type)
+		{
+			return true;
+		}
+	}
+
+	return false;
+}
+
 bool check_minimum_supported(cl_image_format *formatList,
                              unsigned int numFormats,
                              cl_mem_flags flags,
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -3664,15 +3664,14 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
     return false;
 }
 
-bool check_minimum_supported(cl_image_format *formatList,
-                             unsigned int numFormats,
-                             cl_mem_flags flags,
-                             cl_mem_object_type image_type,
-                             cl_device_id device)
+void build_required_image_formats(cl_mem_flags flags,
+                                  cl_mem_object_type image_type,
+                                  cl_device_id device,
+                                  std::vector<cl_image_format>& formatsToSupport)
 {
-	bool passed = true;
-	std::vector<cl_image_format> formatsToSupport;
 	Version version = get_device_cl_version(device);
+
+	formatsToSupport.clear();
 
 	// Required embedded formats.
 	static std::vector<cl_image_format> embeddedProfReadOrWriteFormats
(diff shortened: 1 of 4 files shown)


Jeremy-Kemp:
How about creating a function like `isImageFormatRequired`. This would return a bool whether the requested format is required or not based on the profile of the device. It could be placed in `image_helpers.cpp`. 

I could put that in a separate PR and then update _this_ PR with the new function.

kpet:
Yes, that sounds good. A few things worth keeping in mind:
- The new function shouldn't duplicate what `check_minimum_supported` does so we don't have the list of required formats in multiple places.
- The list of formats that `check_minimum_supported` has is way outdated. Please create another issue if you feel it's too much of a stretch to tackle this as part of this PR.

kpet:
Linked related https://github.com/KhronosGroup/OpenCL-CTS/issues/427 issue.

Jeremy-Kemp:
Didn't mean to do that to the pull. I can close and re-create the pull or just revert the commit. Any preferences?

test_common/harness/imageHelpers.cpp
```diff
@@ -3810,18 +3809,50 @@ bool check_minimum_supported(cl_image_format *formatList,
...
-    }
+bool is_image_format_required(cl_image_format format,
+                              cl_mem_flags flags,
+                              cl_mem_object_type image_type,
+                              cl_device_id device)
+{
+	std::vector<cl_image_format> formatsToSupport;
+	build_required_image_formats(flags, image_type, device, formatsToSupport);
+
+	for (auto &formatItr: formatsToSupport)
```
kpet: You could use `std::find` though it's not that much of an improvement.

test_common/harness/imageHelpers.cpp
```diff
@@ -3810,18 +3809,50 @@ bool check_minimum_supported(cl_image_format *formatList,
...
+		    formatItr.image_channel_data_type == format.image_channel_data_type)
+		{
+			return true;
+		}
+	}
+
+	return false;
+}
+
+bool check_minimum_supported(cl_image_format *formatList,
```
kpet: This probably should be moved to `test_conformance/images/clGetInfo/test_loops.cpp` now.

test_conformance/basic/test_readimage3d.cpp
```diff
@@ -113,120 +111,114 @@ prepare_reference(unsigned char * input_ptr, int w, int h, int d)
...
-    free(output_ptr);
-  free(ref_ptr[0]);
-  free(ref_ptr[1]);
-
-    return err;
+	cl_mem streams[2];
+	cl_program program;
+	cl_kernel kernel;
+	cl_sampler sampler;
+	static cl_image_format img_format[2] =
```
kpet: You could drop the 2.

test_conformance/basic/test_readimage3d.cpp
```diff
@@ -113,120 +111,114 @@ prepare_reference(unsigned char * input_ptr, int w, int h, int d)
...
+	static const char* kernelName[] =
+	{
+		"test_bgra8888",
+		"test_rgba8888",
+	};
+	static const char* kernelSourceString[] =
+	{
+		bgra8888_kernel_code,
+		rgba8888_kernel_code,
+	};
```
kpet: Maybe worth defining  a structure for image format/kernel name/source. Or maybe use `std::tuple`. Not sure that would make the code that much better.

test_conformance/basic/test_readimage3d.cpp
```diff
@@ -113,120 +111,114 @@ prepare_reference(unsigned char * input_ptr, int w, int h, int d)
...
+	int img_height = 64;
+	int img_depth = 64;
+	int i, err;
+	size_t origin[3] = {0, 0, 0};
+	size_t region[3] = {img_width, img_height, img_depth};
+	size_t length = img_width * img_height * img_depth * 4 * sizeof(float);
+
+	PASSIVE_REQUIRE_3D_IMAGE_SUPPORT( device )
+
+	for (uint32_t i = 0; i < 2; i++)
```
kpet: The 2 should probably be replaced with an `ARRAY_SIZE`. This could become a range for if you introduced a structure.

test_conformance/basic/test_readimage3d.cpp
```diff
@@ -113,120 +111,114 @@ prepare_reference(unsigned char * input_ptr, int w, int h, int d)
...
+	{
+		if (is_image_format_required(img_format[i], CL_MEM_READ_ONLY, CL_MEM_OBJECT_IMAGE3D, device))
```
kpet: You could reverse the logic and use `continue` to reduce the indentation of the entire block below.

test_common/harness/imageHelpers.cpp
Jeremy-Kemp: I originally had that, but didn't want to also overload `==` for `cl_image_format`.

test_conformance/basic/test_readimage3d.cpp
Jeremy-Kemp: Good spot, I missed that.

test_common/harness/imageHelpers.cpp
kpet: I wasn't expecting this to be required. If it is, it's probably not worth it then.

test_common/harness/imageHelpers.cpp
Jeremy-Kemp: As part of this PR? I'm trying to just fix BGRA here.

kpet: Fair enough. It's become possible as a result of your change but it's not strictly required. That being said, it's not even 2 minutes of copy/paste :). Your call.

test_conformance/basic/test_readimage3d.cpp
Jeremy-Kemp: I've made a struct with a slightly dodgy name but am not sure of anything more suitable.

test_conformance/basic/test_readimage3d.cpp
```diff
@@ -110,123 +108,123 @@ prepare_reference(unsigned char * input_ptr, int w, int h, int d)
     return ptr;
 }
 
+struct testFormat
```
kpet: You could have scoped the declaration to the function. Doesn't matter too much.

test_conformance/basic/test_readimage3d.cpp
kpet: Works for me.

test_conformance/basic/test_readimage3d.cpp
```diff
@@ -110,123 +108,123 @@ prepare_reference(unsigned char * input_ptr, int w, int h, int d)
...
+					break;
+				}
+				default:
+				{
+					err = CL_OUT_OF_RESOURCES;
+					test_error(err, "Unhandled case");
+					break;
+				}
+			}
+		}
```
kpet: It's a shame to have a big switch just to print a pass message with the format name. We probably have a format to string function somewhere.

PR #629, opened 2020-02-26
Update table of required image formats (#427)

Jeremy-Kemp:
This commit updates the table of required image formats. The table is built depending on the profile of the device, the requested image type, and the avaiablitiy of relevent extensions.


Jeremy-Kemp:
This is a pre-requisite change for #624

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,45 +3671,201 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+		{ CL_RGBA, CL_FLOAT },
+	};
+
+	/*
+		Required full profile formats.
+		This array does not contain any full profile
+		formats that have restrictions on when they
+		are required.
+	*/
+	cl_image_format fullProfReadOrWriteFormats[] =
```
kpet: This list is not valid for an OpenCL 1.2 implementation (see https://github.com/KhronosGroup/OpenCL-Docs/issues/229).

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,45 +3671,201 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
-        { CL_BGRA, CL_UNORM_INT8} };
+bool check_minimum_supported(cl_image_format *formatList,
+                             unsigned int numFormats,
+                             cl_mem_flags flags,
+                             cl_mem_object_type image_type,
+                             cl_device_id device)
+{
+	bool passed = true;
+	bool bDepth = false;
+	bool bsRGBA = false;
```
kpet: These could maybe have a more explicit name. I'm thinking something like `depth_required` (or better).

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,45 +3671,201 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+			uiTotalFormatSize += sizeof(fullProfReadOrWriteDepthFormats);
+			bDepth = true;
+		}
+
+		// sRGB writes not required for 1DImage Buffers
+		if (image_type != CL_MEM_OBJECT_IMAGE1D_BUFFER)
+		{
+			if (flags == CL_MEM_WRITE_ONLY)
+			{
+				// sRGB writes are only required with the cl_khr_srgb_image_writes extension
```
kpet: I think that's incorrect. Supporting the extension doesn't mean that the formats become required.

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,45 +3671,201 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+	}
+
+	cl_image_format *readFormatsToSupport = (cl_image_format*)malloc(uiTotalFormatSize);
+	cl_image_format *writeFormatsToSupport = (cl_image_format*)malloc(uiTotalFormatSize);
+
+	if (readFormatsToSupport == NULL || writeFormatsToSupport == NULL)
+	{
+		passed = false;
+		goto exit;
+	}
```
kpet: I would suggest to use `std::vector` to simplify that code.

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,45 +3671,201 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+        testCount = uiTotalFormatSize / sizeof( readFormatsToSupport[ 0 ] );
     }
     else
     {
         formatsToTest = writeFormatsToSupport;
-        testCount = sizeof( writeFormatsToSupport ) / sizeof( writeFormatsToSupport[ 0 ] );
+        testCount = uiTotalFormatSize / sizeof( writeFormatsToSupport[ 0 ] );
     }
 
     for( unsigned int i = 0; i < testCount; i++ )
```
kpet: The message below needs to be updated with the actual version supported by the device.

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,45 +3671,201 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+	}
+	else
+	{
+		size_t uiReadOffset = 0;
+		size_t uiWriteOffset = 0;
+
+		memcpy(readFormatsToSupport + uiReadOffset,
+		       fullProfReadOrWriteFormats,
+		       sizeof(fullProfReadOrWriteFormats));
+		uiReadOffset += sizeof(fullProfReadOrWriteFormats) / sizeof(fullProfReadOrWriteFormats[0]);
```
kpet: You should use `ARRAY_SIZE`.

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,45 +3671,201 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+                             cl_mem_flags flags,
+                             cl_mem_object_type image_type,
+                             cl_device_id device)
+{
+	bool passed = true;
+	bool bDepth = false;
+	bool bsRGBA = false;
+
+	// Required embedded formats.
+	cl_image_format embeddedProfReadOrWriteFormats[] =
```
kpet: All these tables should be `static`.

test_common/harness/imageHelpers.cpp
Jeremy-Kemp: Are the new names OK?

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,56 +3674,173 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+		{ CL_sRGBA, CL_UNORM_INT8 },
+	};
+
+	static std::vector<cl_image_format> fullProfReadOrWriteDepthFormats
+	{
+		{ CL_DEPTH, CL_UNORM_INT16 },
+		{ CL_DEPTH, CL_FLOAT },
+	};
+
+	static std::vector<cl_image_format> fullProfSRGBFormats
```
Jeremy-Kemp: This needs to be update based on the outcome of [KhronosGroup/OpenCL-Docs/issues/229](https://github.com/KhronosGroup/OpenCL-Docs/issues/229).

Jeremy-Kemp: I've made sRGB match the layout of sRGBA for now.

test_common/harness/imageHelpers.cpp
Jeremy-Kemp: Agree, I mis-read the spec. Updated to only require `CL_MEM_READ_ONLY`.

test_common/harness/imageHelpers.cpp
Jeremy-Kemp: Should be correct now based on the version of the device.

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,56 +3677,181 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+		(specifically 2.x).
+		There are cases whereby the format isn't required.
+	*/
+	static std::vector<cl_image_format> fullProf2XSRGBFormats
+	{
+		{CL_sRGB, CL_UNORM_INT8},
+	};
+
+	// Full profile
+	if (!gIsEmbedded)
```
kpet: You could reverse the condition for clarity.

test_common/harness/imageHelpers.cpp
```diff
@@ -3671,56 +3677,181 @@ bool find_format( cl_image_format *formatList, unsigned int numFormats, cl_image
...
+	// Embedded profile
+	else
+	{
+		copy(embeddedProfReadOrWriteFormats.begin(),
+		     embeddedProfReadOrWriteFormats.end(),
+		     back_inserter(readFormatsToSupport));
+
+		copy(embeddedProfReadOrWriteFormats.begin(),
+		     embeddedProfReadOrWriteFormats.end(),
+		     back_inserter(writeFormatsToSupport));
```
kpet: You could de-duplicate the copy code and assign a pointer depending on the profile.

Jeremy-Kemp: I've taken this further. Given that we're currently only looking at read _or_ write, and that the read or write formats are nearly identical, we can just add everything and special case what we need.

PR #631, opened 2020-02-27
Remove duplicate definition of align_{malloc,free}

### LINKED ISSUE #326: Duplicate definition for align_{malloc,free}
math_brute_force redefines the functions present in the kernel helpers which it needs to include as well.

kpet:
Also use it instead of duplicating the code.

Fixes #326

### DIFF
--- a/test_common/harness/alloc.h
+++ b/test_common/harness/alloc.h
@@ -1,0 +1,68 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


jrprice:
Maybe opening a can of worms here, but can we consider moving to C++17, where we could just use `aligned_alloc` everywhere? IMO it's pretty widely supported now. That said, for Android it'd bump the minimum version to P (which I don't mind, but others might?).

I don't object to this change going in though; we could make this change later if we did decide to adopt C++17.

Jeremy-Kemp:
Could we hold off on C++17? We're still supporting some platforms with old GCC which might make this a little tricky.

Will need to investigate further to confirm.

kpet:
C++17 wouldn't be possible for us either at the moment but we'd support a move to C++14. I'll get this in.

PR #604, opened 2020-02-20
Rename test .c sources to .cpp where necessary

jrprice:
Remove hacks to force language from CMake files.

Closes KhronosGroup/OpenCL-CTS#25

Issue #25, opened 2018-11-09
Many files have got a .c extension but are built as C++


kpet:
This triggers warnings with some compilers, requires CMake trickery and is bad practise. We ought to rename all the files to their rightful extension.

PR #637, opened 2020-02-28
Fplib build fix

Jeremy-Kemp:
Related to #25 and #604.

The change re-names fplib.c to fplib.cpp and also changes the conditions under which it is built (to match when it is invoked by conversions).


Jeremy-Kemp:
Yes I think having some sort of arch variable that we define in the top level `CMakeLists.txt` will make any dependant build decisions a lot easier to maintain.

I'm looking forward to `travis-arm`, it'll be a great step forward for CTS :).

kpet:
I've got the Arm and AArch64 CI ready (and failing because of this issue) now. I'll rebase the change and create a PR once we get a fix for this issue in.

kpet:
@Jeremy-Kemp Do you want to add the arch variable as part of this PR or get this in as is and do it afterwards?

Jeremy-Kemp:
> @Jeremy-Kemp Do you want to add the arch variable as part of this PR or get this in as is and do it afterwards?

I can do this now. I'll cobble something together and update this review.

CMakeLists.txt
```diff
@@ -121,14 +121,26 @@ endif (GL_IS_SUPPORTED AND CLConform_GL_LIBRARIES_DIR)
 include(CheckFunctionExists)
 include(CheckIncludeFiles)
 
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm.*|ARM.*)" OR CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64.*|AARCH64.*|arm64.*|ARM64.*)")
+    set(CLConform_TARGET_ARCH ARM)
```
kpet: I suggest we use two different values for `ARM` and `ARM64`.

test_conformance/conversions/CMakeLists.txt
```diff
@@ -13,9 +13,9 @@ set (${MODULE_NAME}_SOURCES
       Sleep.cpp test_conversions.cpp basic_test_conversions.cpp
 )
 
-if(ANDROID)
-    list(APPEND CONVERSIONS_SOURCES fplib.c)
-endif(ANDROID)
+if(CMAKE_COMPILER_IS_GNUCC AND (${CLConform_TARGET_ARCH} EQUAL "ARM" OR ${CLConform_TARGET_ARCH} EQUAL "ARM64"))
```
kpet: This file is actually not required for ARM64.

PR #638, opened 2020-03-02
Add Travis CI jobs for Arm and AArch64

kpet:
- Use fixed Linaro toolchains
- Move travis script to a shell file


travis.sh
```diff
@@ -0,0 +1,68 @@
...
+TOOLCHAIN_FILE=${TOP}/toolchain.cmake
+touch ${TOOLCHAIN_FILE}
+
+# Prepare toolchain if needed
+if [[ ${JOB_ARCHITECTURE} != "" ]]; then
+    TOOLCHAIN_URL_VAR=TOOLCHAIN_URL_${JOB_ARCHITECTURE}
+    TOOLCHAIN_URL=${!TOOLCHAIN_URL_VAR}
+    wget ${TOOLCHAIN_URL}
+    TOOLCHAIN_ARCHIVE=${TOOLCHAIN_URL##*/}
+    tar xvf ${TOOLCHAIN_ARCHIVE}
```
Jeremy-Kemp: Could we drop the v? Its quite a spammer.

Jeremy-Kemp:
Seeing the gcc 7.5 toolchain got me thinking about how to catch new language features being added that might not work on older gcc's.

I'll create an issue where we can start thinking about what minimum versions of gcc we want to support.

jrprice:
Is there a reason these jobs are being done via cross-compilation rather than native `arm64` jobs (which Travis CI apparently supports)?

travis.sh
kpet: Yeah, it runs quickly enough. I'll remove it.

kpet:
We could do native builds for `arm64` but it seems Travis doesn't support `arm` so we'd still need cross-compilation for that.

Jeremy-Kemp:
We should probably just stick with the script for both `arm` and `arm64` then to avoid introducing additional complexity to the build.

kpet:
Thanks for the review, I'll give @jrprice a chance to comment before merging this.

Issue #491, opened 2019-11-18
Fix bruteforce's fast math mode domains


lakshmih:
Cast array to floats before setting NANs to prevent large nonsensical
numbers outside of the valid sin, cos, sincos, and reciprocal domain's
from being tested

PR #641, opened 2020-03-02
[QC] cl20: bruteforce: Fix domains used in fast math mode.

lakshmih:
Cast input array to floats before setting NANs which are also floats to
prevent large nonsensical numbers outside of the valid domains of
several math functions from being tested.

Khronos Bug: https://github.com/KhronosGroup/OpenCL-CTS/issues/491
Test Suite Affected: bruteforce Subtests: sin, cos, sincos, reciprocal

### DIFF
--- a/test_conformance/math_brute_force/unary.cpp
+++ b/test_conformance/math_brute_force/unary.cpp
@@ -491,13 +491,13 @@ static cl_int TestFloat( cl_uint job_id, cl_uint thread_id, void *data )
         if ( strcmp(fname,"sin")==0 || strcmp(fname,"cos")==0 )  //the domain of the function is [-pi,pi]
         {
           if( fabs(p_j) > M_PI )
-            p[j] = NAN;
+            ((float*)p)[j] = NAN;
         }
 
         if ( strcmp( fname, "reciprocal" ) == 0 )
         {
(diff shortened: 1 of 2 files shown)


lakshmih:
@bashbaug @b-sumner  Can you review/merge this or add any reviewers I need to have on the request?

bashbaug:
> Can you review/merge this or add any reviewers I need to have on the request?

This change looks reasonable to me.  I do wonder though if we should choose a different value than `NAN` in the out-of-range + fast relaxed math cases, since `-cl-fast-relaxed-math` implies `-cl-finite-math-only` therefore implementations do not necessarily need to handle `NAN` inputs.

Also tagging @kpet - is there a usual reviewer for math / ULP related changes?

kpet:
I've had a first quick look. I _think_ the approach the existing code is following is not the right one. I haven't tried to find the code checking the results but changing inputs that are out of the domain for which the acceptable error is defined to a value that is also out of the domain shouldn't allow the verification code to assume that all outputs have to be within the acceptable error threshold so my guess is that the verification code is wrong as well or has an ugly special case. Furthermore, as @bashbaug  pointed out `-cl-fast-relaxed-math` implies `-cl-finite-math-only` so using `NAN` isn't guaranteed to work. We should understand what is done there before making a decision.

lakshmih:
Proposing a different fix.

PR #649, opened 2020-03-05
check offline compiler for clcpp tests

ijkmn:
disable clcpp tests when khronos offline compiler is not available


kpet:
I don't think we want to do that at the moment. These tests can't be run without an offline compiler but they can be built and I think we should keep building them always so they don't rot as we change common code. Any reason you want to make that change?

ijkmn:
when i build cts without offline compiler, met build error. but now i can't reproduce after revert changes, and build successfully.
thank you.

PR #650, opened 2020-03-05
update glew to 2.1.0

ijkmn:
update glew to 2.1.0 for missing some api and macro
fix build and link error


Jeremy-Kemp:
Perhaps we should be getting these headers from external repo, just like we do with the CL headers (aka passing `CL_INCLUDE_DIR` to `CMake`) - rather than having to maintain them directly inside CTS.

gwawiork:
I've tested this change in my Windows GL build env. Build passed, execution passed then LGTM

kpet:
Agree with @Jeremy-Kemp, we should probably track these as an external dependency. @ijkmn I think  this can go in if people that support GLES are fine with the change but could you create an issue to cover managing these files as an external dependency?

PR #651, opened 2020-03-05
fix gles build error

ijkmn:
1. gles 3.x also define GL_ES_VERSION_2_0
2. ICD don't export clGetGLContextInfoKHR, should get extension function address
3. remove ' include gl2extQCOM.h '


kpet:
I'm guessing this should now be rebased on top of the GLEW update. I'm also wondering whether we want to split this into multiple PRs.

PR #662, opened 2020-03-06
Build: Ensure compilers required directory structure is present.

Jeremy-Kemp:
The compiler test requires that directories named "includeTestDirectory" and "secondIncludeTestDirectory" are present with the binary.


kpet:
Thanks for the issue, merging.

PR #665, opened 2020-03-09
Add a test that clSetKernelArg does not retain the arg

alan-baker:
* Add `retain_mem_object_set_kernel_arg` to test that calling
clSetKernelArg with a cl_mem object does not retain that mem object

This issue came up from https://github.com/kpet/clvk/issues/157. @kpet requested I add a conformance test to cover it. Tested on a desktop implementation and clvk.

### DIFF
--- a/test_conformance/api/test_retain.cpp
+++ b/test_conformance/api/test_retain.cpp
@@ -232,3 +232,44 @@ int test_retain_mem_object_multiple(cl_device_id deviceID, cl_context context, c
     return 0;
 }
 
+int test_retain_mem_object_set_kernel_arg(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
+{
+    int err;
+    cl_mem buffer = nullptr;
+    cl_program program;
+    cl_kernel kernel;
+    static volatile uint32_t sValue;
+    sValue = 0;
+    auto callback = []( cl_mem, void * ) {
+      ++sValue;
+    };
+    const char *testProgram[] = { "__kernel void sample_test(__global int *data){}" };
+
+    buffer = clCreateBuffer( context, CL_MEM_READ_ONLY, 32, NULL, &err );
+    test_error( err, "Unable to create buffer to test with" );
+
+    err = clSetMemObjectDestructorCallback( buffer, callback, nullptr );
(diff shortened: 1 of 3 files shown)


jlewis-austin:
Maybe out of the scope of this PR, but it seems the CTS should also be testing that the mem object _is_ held by the implementation while executing, e.g.:
```
buf = clCreateBuffer
clSetKernelArg(..., buf)
clEnqueueNDRangeKernel
clReleaseMemObject(buf)
...
clFlush
```
I'm not aware of any tests that do this, but also don't really know how to search for it.

alan-baker:
@jlewis-austin, the test you describe seems to be covered by `release_during_execute` in test_retain_program.cpp.

jlewis-austin:
@alan-baker yes indeed, thanks!

test_conformance/api/test_retain.cpp
```diff
@@ -232,3 +232,44 @@ int test_retain_mem_object_multiple(cl_device_id deviceID, cl_context context, c
     return 0;
 }
 
+int test_retain_mem_object_set_kernel_arg(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
+{
+    int err;
+    cl_mem buffer = nullptr;
```
kpet: You don't need to initialise `buffer`.

test_conformance/api/test_retain.cpp
```diff
@@ -232,3 +232,44 @@ int test_retain_mem_object_multiple(cl_device_id deviceID, cl_context context, c
...
+    buffer = clCreateBuffer( context, CL_MEM_READ_ONLY, 32, NULL, &err );
+    test_error( err, "Unable to create buffer to test with" );
+
+    err = clSetMemObjectDestructorCallback( buffer, callback, nullptr );
+    test_error( err, "Unable to set destructor callback" );
+
+    err = create_single_kernel_helper( context, &program, nullptr, 1, testProgram, nullptr );
+    test_error( err, "Unable to build sample program" );
+
+    kernel = clCreateKernel( program, "sample_test", &err );
```
kpet: You could have got `create_single_kernel_helper` to create the kernel for you.

Issue #642, opened 2020-03-02
Ignore vendor extensions when querying host command queue properties


lakshmih:
The presence of the flags related to vendor extensions on host queue properties is causing the api and api_compatibility tests to fail.  Ignore vendor extensions when querying host command queue properties.

lakshmih:
@bashbaug @kpet  requesting review for this.

kpet:
@lakshmih I did comment on your PR :).

lakshmih:
This same fix needs to be made in the compatibility api test as well

alycm:
I believe this was addressed by merging: https://github.com/KhronosGroup/OpenCL-CTS/pull/957

Closing.

PR #666, opened 2020-03-09
cl20: api:  Allow vendor flags

lakshmih:
Mask off vendor extensions when querying host command queue properties.

Khronos Bug: https://github.com/KhronosGroup/OpenCL-CTS/issues/642
Test Suite Affected: api

### DIFF
--- a/test_conformance/api/test_queries.cpp
+++ b/test_conformance/api/test_queries.cpp
@@ -237,6 +237,9 @@ int test_get_command_queue_info(cl_device_id deviceID, cl_context context, cl_co
     clGetDeviceInfo(deviceID, CL_DEVICE_QUEUE_ON_HOST_PROPERTIES, sizeof(device_props), &device_props, NULL);
     log_info("CL_DEVICE_QUEUE_ON_HOST_PROPERTIES is %d\n", (int)device_props);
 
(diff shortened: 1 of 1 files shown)


lakshmih:
@bashbaug

lakshmih:
We have a queue property (in one of our extensions) that is incompatible with some of the standard OpenCL properties. 
This gets returned in the query but the spec doesn't require that you should be able to create a command queue with "all" of the properties returned. 
We think the conformance test should only test creation of the queue with properties that are part of the core spec.

kpet:
I had a suspicion that was the case. I'm a bit worried about how maintainable the current patch is going to be (will we think about updating the mask as we add new properties?). A couple of ideas/proposals:
- If the extension that adds the property in question is public we could add a test for the extension string and mask out the problematic property.
- Maybe we could introduce a mask in the headers to cover all core properties. This would assume that we won't add incompatible core properties in the future though.

Any opinions?

lakshmih:
The extension isn't public (as you likely guessed as well :)) and even if it were I'd like to keep extension-properties separate from core ones.
We are okay to add a mask but I couldnt find a common enough header in which to add it. It should be available to api and computeinfo at least, and potentially to other tests as well. In the absence of such a common header, would you consider accepting the fix as is?

PR #667, opened 2020-03-10
Fix test skip handling while test initialization - device_timer

### LINKED ISSUE #664: No test status when TEST_SKIP in case of InitCL
Hi 
I see truncated log from test binary in case of cases for OpenCL 2.X and executed on platform OpenCL 1.X
example:
`test_device_timer.exe
Compute Device Name = Intel(R) UHD Graphics 605, Compute Device Vendor = Intel(R) Corporation, Compute Device Version = OpenCL 1.2 NEO , CL C Version = OpenCL C 1.2
sizeof( void*) = 8  (host)
sizeof( void*) = 4  (device)
`
This situation is related with PR #628 

We should print something on output in situation of `return TEST_SKIP;` 
However I am wondering how and what . The problem is that we do not reach calling function https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/testHarness.cpp#L471
where print_results function is called. 
Do you have any suggestion what should be printed in case of 
this skip:
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/device_timer/main.cpp#L41

and this:
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/device_timer/main.cpp#L62

gwawiork:
This is part of the fix #664

### DIFF
--- a/test_common/harness/testHarness.cpp
+++ b/test_common/harness/testHarness.cpp
@@ -67,6 +67,21 @@ int runTestHarness( int argc, const char *argv[], int testNum, test_definition t
                           ( imageSupportRequired ) ? verifyImageSupport : NULL );
 }
 
+int skip_init_info(int count) {
+    log_info("Test skipped while initialization\n");
+    log_info("SKIPPED %d of %d tests.\n", count, count);
+    return EXIT_SUCCESS;
+}
+
+int fail_init_info(int count) {
+    log_info("Test failed while initialization\n");
+    log_info("FAILED %d of %d tests.\n", count, count);
+    return EXIT_FAILURE;
+}
+void version_expected_info(const char * test_name, const char * expected_version, const char * device_version) {
+    log_info("%s skipped (requires at least version %s, but the device reports version %s)\n",
+        test_name, expected_version, device_version);
+}
 int runTestHarnessWithCheck( int argc, const char *argv[], int testNum, test_definition testList[],
                              int forceNoContextCreation, cl_command_queue_properties queueProps,
                              DeviceCheckFn deviceCheckFn )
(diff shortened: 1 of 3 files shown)


test_common/harness/testHarness.cpp
```diff
@@ -67,6 +67,21 @@ int runTestHarness( int argc, const char *argv[], int testNum, test_definition t
                           ( imageSupportRequired ) ? verifyImageSupport : NULL );
 }
 
+int skip_init_info() {
+    log_info("Test skipped while initialization\n");
+    log_info("PASSED %s.\n", "test");
```
kpet: As per previous comments, I think my preference would be to say `SKIPPED` here with the total number of tests. We should have a clear distinction between passing and skipping tests.

Also, I don't understand why `test` is printed via a format string.

gwawiork: regarding `test` - its a mistake.  Are we interested in printing full list of tests skipped ?

gwawiork: if no list of tests then:

Test skipped (requires at least version 2.1, but the device reports version 1.2)
Test skipped while initialization
SKIPPED 2 of 2 tests.

kpet: I wouldn't try to print the list. This really belongs in the JSON results. The sample output looks good to me.

gwawiork: updated

gwawiork:
Thanks. Then please merge. It allows me go with changes in other tests.

kpet:
I was just giving others a chance to comment. Merging to enable forward progress.

PR #672, opened 2020-03-11
Update printf to query the correct device version

Jeremy-Kemp:
Printf was requiring only a 1.2 implementation. It should really be >= 1.2.


Jeremy-Kemp:
> The old code was not incorrect (have a look at `check_opencl_version`) 

The crux was really that `check_opencl_version` doesn't support `2.2`, but that doesn't really matter anyway as we're dropping it :)

Thanks for reviewing!

kpet:
Yes, true, not much of a practical problem though :).

Issue #675, opened 2020-03-13
Printf test is not checking scientific notation correctly


Jeremy-Kemp:
The C spec says the following:

> The E conversion specifier produces a number with E instead of e introducing the exponent. The exponent always contains **at least two digits**, and only as many more digits as necessary to represent the exponent.

Conformance has a few cases where we test scientific notation, but testing against the wrong value.

For example,
```C
    printf("%+#21.15E\n", 789456123.0);
```
expects a result of `+7.894561230000000E+8`. However, the result should be `+7.894561230000000E+08`.

jlewis-austin:
Just to clarify, from the code it looks like the issue is not that it doesn't _allow_ the correct 2-digit answer, but that it doesn't _require_ it, is that correct? I'm looking at the snippet that skips leading zeroes before comparing to the reference.

Jeremy-Kemp:
Your comment has alerted me to the fact that this issue is one of my own creation that I spotted once I began fixing #673 .

The existing printf behaviour is to have reference results with all leading zeros stripped from the exponent. The test then also strips leading zeros from the actual result read back from the device.

With my changes, the reference result now contains leading zeros, but the test was still stripping the leading zeros from the device. It was at this point I noticed the "bug" and raised this issue.

Now I actually think that stripping leading zeros from the reference and the device result is desirable behaviour, as there is no comment on how many leading zeros there may be.

I'll update #680 to strip leading zeros from the reference and the actual result, and then close this issue.

Jeremy-Kemp:
Closing as per previous comment.

Issue #673, opened 2020-03-11
Printf floating point tests do not consider device rounding mode


Jeremy-Kemp:
The reference results for the `printf` test are hard-coded. 

For implementations not implementing `CL_FP_ROUND_TO_NEAREST`, floating point tests can appear to fail.

For example, the first float test `printf("%f",10.3456)` expects a result of `10.345600`.  https://github.com/KhronosGroup/OpenCL-CTS/blob/8a00f4ecede75f15a822cfdd0321ad69e3acea8f/test_conformance/printf/util_printf.cpp#L238

On an implementation with `CL_FP_ROUND_TO_ZERO` the correct result is actually `10.345599`.

Jeremy-Kemp:
@kpet can we start utilising Labels so that we can filter out serious issues/PR's from the not-so-serious? 

My first though to solve this is by using `std::stringstream` to generate the reference result (with the correct rounding mode set that matches the device). 

I'd appreciate some input about how we want to solve this.

Below snippet for clarity:

```c
#include <cfenv>
#include <string>  
#include <iostream> 
#include <sstream>   
#include <iomanip>

int main()
{
	// Save the current rounding mode
	const int old_round = std::fegetround();

	// Match this to whatever CL_DEVICE_SINGLE_FP_CONFIG, CL_DEVICE_SINGLE_FP_CONFIG or CL_DEVICE_HALF_FP_CONFIG is
	std::fesetround(FE_TOWARDZERO);

	// Build the reference result
	std::stringstream test1;
	test1 << std::fixed << std::setprecision(6) << 10.3456 << std::endl;

	// Restore rounding mode
	std::fesetround(old_round);

	// Save the reference result
	std::string reference = test1.str();

	// Print the reference result
	// Will print 10.345599 in this example
	std::cout <<  reference;

	return 0;
}
```

kpet:
@Jeremy-Kemp Yes, I'm all for using the labels. I've created one for cleanups. I'll start labeling some issues. This one is a bug.

On this particular issue, what you're proposing seems reasonable at first glance. Tagging @jlewis-austin and @EwanC who've worked on several FP-related issues.

jlewis-austin:
It seems reasonable, but with a few strings to pull wrt what the specs say on the subject. One concern is that some platforms could ignore the currently set rounding mode in order to provide exact round-trip behavior for printf--we might consider accepting the correctly-rounded answer in any case. 
[...]

PR #680, opened 2020-03-17
Printf fixes for reference results #673 #675

Jeremy-Kemp:
The PR aims to fix issues identified in #673 and #675.

The reference results had two issues:
- They did not take into account the rounding mode of the device.
- Scientific notation results did not have trailing zero's, meaning that the exponent could be a single digit, despite the requirement being at least two digits.

This PR introduces runtime generated reference results for types with numerical representations (float, int, hex, etc) where a direct mapping to standard C99 printf is possible to execute on the CPU (as opposed to tests such as address space, vectors etc which cannot be directly mapped to C99 printf).

### DIFF
--- a/test_conformance/printf/util_printf.cpp
+++ b/test_conformance/printf/util_printf.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -14,9 +14,17 @@
 // limitations under the License.
 //
 #include "harness/compat.h"
+#include "harness/rounding_mode.h"
 
 #include "test_printf.h"
+#include <assert.h>
 
+// Helpers for generating runtime reference results
+static void intRefBuilder(printDataGenParameters&, char*, const size_t);
+static void floatRefBuilder(printDataGenParameters&, char* rResult, const size_t);
+static void octalRefBuilder(printDataGenParameters&, char*, const size_t);
+static void unsignedRefBuilder(printDataGenParameters&, char*, const size_t);
+static void hexRefBuilder(printDataGenParameters&, char*, const size_t);
 
 //==================================
 
@@ -30,7 +38,7 @@
 
 //------------------------------------------------------
 
-struct printDataGenParameters printIntGenParameters[] = {
+std::vector<printDataGenParameters> printIntGenParameters = {
 
     //(Minimum)Five-wide,default(right)-justified
 
@@ -70,37 +78,6 @@ struct printDataGenParameters printIntGenParameters[] = {
 
 };
 
-//------------------------------------------------
-
-// Lookup table - [string]int-correct buffer     |
-
-//------------------------------------------------
-
-std::vector<const char*> correctBufferInt = {
-    "   10",
-
-    "10   ",
-
-    "00010",
-
-    "  +10",
-
-    "+10  ",
-
-    "00100",
-
-    " 00100",
-
-    "100   ",
-
-    " 00100"
-
-};
-
-
-
-
-
 //-----------------------------------------------
 
 //test case for int                             |
@@ -109,13 +86,15 @@ std::vector<const char*> correctBufferInt = {
 
 testCase testCaseInt = {
 
-    sizeof(correctBufferInt)/sizeof(char*),
-
     TYPE_INT,
 
     correctBufferInt,
 
-    printIntGenParameters
+    printIntGenParameters,
+
+    intRefBuilder,
+
+    kint
 
 };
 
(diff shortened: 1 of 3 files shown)


jlewis-austin:
LGTM; see https://github.com/KhronosGroup/OpenCL-CTS/issues/673#issuecomment-600712622 regarding potential issue with host vs device rounding modes.

test_conformance/printf/util_printf.cpp
```diff
@@ -859,16 +786,130 @@ size_t verifyOutputBuffer(char *analysisBuffer,testCase* pTestCase,size_t testId
...
+        log_error("clGetDeviceInfo failed");
+        return;
+    }
+
+    // Map device rounding to CTS rounding type
+    if (fpConfig & CL_FP_ROUND_TO_NEAREST)
+    {
+        deviceRound = kRoundToNearestEven;
+    }
+    else if (fpConfig & CL_FP_ROUND_TO_ZERO)
```
kpet: What if an implementation reports several supported rounding modes? This isn't forbidden by the spec. There doesn't seem to be a way for the implementation to report which is the default though.

test_conformance/printf/util_printf.cpp
```diff
@@ -859,16 +786,130 @@ size_t verifyOutputBuffer(char *analysisBuffer,testCase* pTestCase,size_t testId
...
+        deviceRound = kRoundTowardZero;
+    }
+    else if (fpConfig & CL_FP_ROUND_TO_INF)
+    {
+        // Can be either round up or round down
+        deviceRound = kRoundUp;
+    }
+    else
+    {
+        assert(true && "Unreachable");
```
kpet: I'm guessing you meant `assert(false && "Unreachable")`

Jeremy-Kemp: Good spot, yeah this should be false. I'll update it.

test_conformance/printf/util_printf.cpp
Jeremy-Kemp: Yes this is a tricky one. Without a way of knowing which mode will actually be used by the implementation, I'm not sure we can solve this at the moment (aside from attempting to detect the mode in-kernel like bruteforce does).

Jeremy-Kemp:
> Any specific reason you use stream capture and `printf` rather than something like `snprintf` to generate the reference strings?

I was re-using the infrastructure that was already present in the test for capturing the stream (its how the current result is read back from the implementation. As usual, I was trying to have the least amount of impact on the rest of the test by trying to fit with what was already present.

It was very tempting to re-write the entire test, but I've not got enough time to do so at the moment.

Jeremy-Kemp:
Does anyone have any comments/objections to this? Its blocking our switch over from the internal repo so am keen to work through it to get something approved.

test_conformance/printf/util_printf.cpp
kpet: There is no way to solve this properly without changing the spec I think. This approach works because it covers the most common case (an embedded profile that won't support RTE but will support RTZ). It's essentially the same logic as `get_default_rounding_mode` and I think we probably ought to be using that to limit the number of places where we hard-code that assumption.

I suggest you create a Gitlab issue to add a query to report the default rounding mode or at least discuss this. We should definitely make all of this a lot clearer. What does it mean to support multiple rounding modes for example if there can be only one default? Can the default rouding mode change in some circumstances, etc? :).

kpet:
>I was re-using the infrastructure that was already present in the test for capturing the stream (its how the current result is read back from the implementation).

I think there is no reason to tie ourselves to all that complexity for reference result generation. I also suspect that using `snprintf` for result generation would reduce the size of the diff substantially.

As you say, it is tempting to make more sweeping changes to that test in general :). Related to this, I've been working on a printf extension to tidy up some of this. I can say more about this by email if you're intested.

Jeremy-Kemp:
> I think there is no reason to tie ourselves to all that complexity for reference result generation.

In that case I'll tidy it as it'll be cleaner.

> Related to this, I've been working on a printf extension to tidy up some of this. I can say more about this by email if you're intested.

Sounds good, yes please!

Jeremy-Kemp:
I've switched to snprintf which means that all the stream capturing functions have moved back to their original location (and reduced the diff in doing so).

Jeremy-Kemp:
Is there anything else that needs addressing here?

kpet:
I hadn't noticed your latest commit. I think this probably just needs a last review then.

test_conformance/printf/util_printf.cpp
```diff
@@ -14,13 +14,18 @@
 // limitations under the License.
 //
 #include "harness/compat.h"
+#include "harness/rounding_mode.h"
+#include "harness/kernelHelpers.h"
 
 #include "test_printf.h"
+#include <assert.h>
```
kpet, line 21: Could use `<cassert>`

Jeremy-Kemp:
Thanks very much @kpet @jlewis-austin

PR #681, opened 2020-03-17
Add a minimum required version to the pragma_unroll compiler test.

aarongreig:
This test is for a 2.0 feature so it should have a 2.0 version requirement.


gwawiork:
Strange. I've have already pushed such change 
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/main.cpp#L90

Is your local branch up to date ? Maybe rebase needed.

aarongreig:
ah you ninja'd me, my bad for not pulling

PR #682, opened 2020-03-17
Fix test_subgroups - test as core feature.

### LINKED ISSUE #46: cl21 - The subgroups test is not aligned with OpenCL 2.1 spec
Internal MR (102) should be moved to opensource to fix subgroups testing to be aligned with specification OpenCL2.1
IFP - independent forward progress

- current version looks like only for platforms supporting OpenCL2.0
- condition for test execution - if implementation shows cl_khr_subgroups extension then use OpenCL 2.0 subgroups in device is OpenCL 2.X and higher then core subgroups should be used
- in OpenCL2.0 version we require extension cl_khr_subgrups and IFP is expected to be supported so test IFP should be executed
- in core subgroups we should not use function/enums with KHR suffixes (proper only for OpenCL2.0).
- in core subgroups version no extension, because this is a core feature and IFP is not obligatory . Only if capability CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS is set to 1 then start IFP testing

gwawiork:
Fixes #46

### DIFF
--- a/test_conformance/subgroups/main.cpp
+++ b/test_conformance/subgroups/main.cpp
@@ -74,7 +74,7 @@ static test_status InitCL(cl_device_id device)
     {
         return ret;
     }
-    if (version > Version(2, 2))
+    if (version >= Version(3, 0))
     {
         cl_uint max_sub_groups;
         int error;
--- a/test_conformance/subgroups/main.cpp
+++ b/test_conformance/subgroups/main.cpp
@@ -30,50 +30,17 @@ test_definition test_list[] = {
     ADD_TEST_VERSION(work_group_functions_ext, Version(2, 0)),
     ADD_TEST_VERSION(work_group_functions_core, Version(2, 1)),
     ADD_TEST_VERSION(barrier_functions_ext, Version(2, 0)),
-    ADD_TEST_VERSION(barrier_functions_core, Version(2, 1))
+    ADD_TEST_VERSION(barrier_functions_core, Version(2, 1)),
+    ADD_TEST_VERSION(ifp_ext, Version(2, 0)),
+    ADD_TEST_VERSION(ifp_core, Version(2, 1))
 };
 
 const int test_num = ARRAY_SIZE(test_list);
-bool gUseCoreSubgroups = false;
-bool gTestIFP = true;
-
-static test_status checkIFPSupport(cl_device_id device)
-{
-    cl_uint ifp_supported;
-    cl_uint error;
-    error = clGetDeviceInfo(device,
-                            CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS,
-                            sizeof(ifp_supported), &ifp_supported, NULL);
-    if (error != CL_SUCCESS)
-    {
-        print_error(
-            error,
-            "Unable to get CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS "
-            "capability");
-        return TEST_FAIL;
-    }
-    // skip testing ifp
-    if (ifp_supported != 1)
-    {
-        log_info("INDEPENDENT FORWARD PROGRESS not supported...\n");
-        gTestIFP = false;
-    }
-    else
-    {
-        log_info("INDEPENDENT FORWARD PROGRESS supported...\n");
-    }
-    return TEST_PASS;
-}
 
 static test_status InitCL(cl_device_id device)
 {
     auto version = get_device_cl_version(device);
     test_status ret = TEST_PASS;
-    ret = checkIFPSupport(device);
-    if (ret != TEST_PASS)
-    {
-        return ret;
-    }
     if (version >= Version(3, 0))
     {
         cl_uint max_sub_groups;
--- a/test_conformance/subgroups/main.cpp
+++ b/test_conformance/subgroups/main.cpp
@@ -30,6 +30,8 @@ test_definition test_list[] = {
 };
 
 const int test_num = ARRAY_SIZE(test_list);
+bool use_core_subgroups = false;
+bool test_ifp = true;
 
 static test_status checkSubGroupsExtension(cl_device_id device)
 {
@@ -56,8 +58,22 @@ static test_status checkSubGroupsExtension(cl_device_id device)
 
     if ((version > expected_min_version) && !hasExtension)
     {
-        log_error("'cl_khr_subgroups' is a required extension, failing.\n");
-        return TEST_FAIL;
+        log_info("'cl_khr_subgroups' not found. Using OpenCL 2.1 core subgroups.\n");
+        use_core_subgroups = true;
+        cl_uint ifp_supported;
+        cl_uint error;
+        error = clGetDeviceInfo(device, CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS,
+            sizeof(ifp_supported), &ifp_supported, NULL);
+        if (error != CL_SUCCESS) {
+            print_error(error, "Unable to get CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS capability");
+            return TEST_FAIL;
+        }
+        //skip testing ifp
+        if (ifp_supported != 1) {
+            log_info("INDEPENDENT FORWARD PROGRESS not supported...\n");
+            test_ifp = false;
+        }
+        return TEST_PASS;
     }
 
     return TEST_PASS;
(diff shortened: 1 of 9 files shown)


test_conformance/subgroups/main.cpp
```diff
@@ -30,6 +30,8 @@ test_definition test_list[] = {
 };
 
 const int test_num = ARRAY_SIZE( test_list );
+bool use_core_subgroups = false;
+bool test_ifp = true;
```
kpet: We probably want to follow the `gVariable` naming convention for globals.

gwawiork:
> Shouldn't we test both the core and extension entrypoints when both are supported?

Maybe additional test should be added in test_api for this purpose with minimal version checking OpenCL 2.1 ?

kpet:
I'm not sure I understand how duplicating code into the api suite addresses my comment :).

If these subgroup tests run quickly enough, why wouldn't we modify them to run twice using the core feature _and_ the extension?

My concern is that your patch removes all coverage for the extension on 2.1+ implementations.

gwawiork:
If extension is reported on 2.1+ ten test will go the path as today works (KHR in api calls and IFP obligatory).

kpet:
Yes, I meant the opposite: this doesn't add coverage for the core feature when the extension is present.

gwawiork:
@bashbaug How do you think. Does the test should cover situation when extension is present and device is 2.1+ as Kevin writes (verify old and new API calls)? For me presenting extension means use previous API. And this is written in specification https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Ext.html#cl_khr_subgroups

No extension and device 2.1+ use new API

gwawiork:
I've committed change that removes PRAGMA definitions in case of core subgroups

bashbaug:
> @bashbaug How do you think. Does the test should cover situation when extension is present and device is 2.1+ as Kevin writes (verify old and new API calls)? For me presenting extension means use previous API.

I think @kpet is right that the extension codepath and the core subgroup codepaths could be slightly different, at least through the frontend compiler.  So, if these tests don't take too long to execute, it seems reasonable to test both codepaths if the implementation advertises support for both the subgroup extension and core subgroups.

If the tests take too long to execute we still may want to consider something like a subgroup "wimpy mode" to at least get some coverage going through both paths.  I just ran the subgroup test on my laptop though and it ran in a couple of minutes, so hopefully this isn't necessary.

gwawiork:
I've added changes to test both path - core and extensions.

gwawiork:
Summing up latest changes.
- split tests cases to test core and extension code path
  - core path
    - do not use KHR functions
    - do not use pragma
    - IFP optionally
  - extension path
    - use KHR functions
    - use pragma
    - IFP obligatory

test_conformance/subgroups/test_workgroup.cpp
```diff
@@ -1038,7 +1037,42 @@ int test_work_group_functions(cl_device_id device, cl_context context,
...
         device, context, queue, num_elements, "test_scexmin", scexmin_source);
 
-    error |= test<cl_int, IFP, G, L>::run(device, context, queue, num_elements,
-                                          "test_ifp", ifp_source, NUM_LOC + 1);
+
+    if (gTestIFP)
+    {
+        error |=
+            test<cl_int, IFP, G, L>::run(device, context, queue, num_elements,
+                                         "test_ifp", ifp_source, NUM_LOC + 1);
```
kpet, line 1045: I think this should be a separate test.

test_conformance/subgroups/subhelpers.h
```diff
@@ -352,33 +354,53 @@ struct test
         if (local > LSIZE) local = LSIZE;
 
         // Get the sub group info
-        clGetKernelSubGroupInfoKHR_fn clGetKernelSubGroupInfoKHR_ptr;
-        clGetKernelSubGroupInfoKHR_ptr = (clGetKernelSubGroupInfoKHR_fn)
-            clGetExtensionFunctionAddressForPlatform(
-                platform, "clGetKernelSubGroupInfoKHR");
-        if (clGetKernelSubGroupInfoKHR_ptr == NULL)
+        if (gUseCoreSubgroups)
```
kpet, line 357: Maybe the two paths could be reconciled by passing a structure with function pointers instead of duplicating the code. It's not too bad but it will make maintenance slightly harder. Please verify that the core and KHR enumerants are the same if you want to attempt this though.

You could also have a single utility functions to fill that structure.

gwawiork:
Recent changes include:
* do not use global variables
* make IFP as separate tests
* use stringstream data type
* use special class to load function pointer for specific API

test_conformance/subgroups/test_barrier.cpp
```diff
@@ -139,21 +139,44 @@ template <int Which> struct BAR
...
+int test_barrier_functions_ext(cl_device_id device, cl_context context,
+                               cl_command_queue queue, int num_elements)
+{
+    bool hasExtension = is_extension_available(device, "cl_khr_subgroups");
+
+    if (!hasExtension)
+    {
+        log_info(
+            "Device does not support 'cl_khr_subgroups'. Skipping the test.\n");
+        return TEST_SKIP;
```
kpet, line 178: Should be TEST_SKIPPED_ITSELF. This will fail the test.

test_conformance/subgroups/test_queries.cpp
```diff
@@ -176,3 +183,24 @@ int test_sub_group_info(cl_device_id device, cl_context context,
...
+int test_sub_group_info_ext(cl_device_id device, cl_context context,
+                            cl_command_queue queue, int num_elements)
+{
+    bool hasExtension = is_extension_available(device, "cl_khr_subgroups");
+
+    if (!hasExtension)
+    {
+        log_info(
+            "Device does not support 'cl_khr_subgroups'. Skipping the test.\n");
+        return TEST_SKIP;
```
kpet, line 202: Same here.

test_conformance/subgroups/test_workgroup.cpp
```diff
@@ -884,161 +634,242 @@ int test_work_group_functions(cl_device_id device, cl_context context,
...
+int test_work_group_functions_ext(cl_device_id device, cl_context context,
+                                  cl_command_queue queue, int num_elements)
+{
+    bool hasExtension = is_extension_available(device, "cl_khr_subgroups");
+
+    if (!hasExtension)
+    {
+        log_info(
+            "Device does not support 'cl_khr_subgroups'. Skipping the test.\n");
+        return TEST_SKIP;
```
kpet, line 871: Same here.

test_conformance/subgroups/test_workitem.cpp
```diff
@@ -301,3 +307,25 @@ int test_work_item_functions(cl_device_id device, cl_context context,
...
+int test_work_item_functions_ext(cl_device_id device, cl_context context,
+                                 cl_command_queue queue, int num_elements)
+{
+    bool hasExtension = is_extension_available(device, "cl_khr_subgroups");
+
+    if (!hasExtension)
+    {
+        log_info(
+            "Device does not support 'cl_khr_subgroups'. Skipping the test.\n");
+        return TEST_SKIP;
```
kpet, line 326: And here.

gwawiork:
> Thanks for the changes. The names you picked for the core/ext function/enum unification have really made this a lot less readable but I suggest we don't address this here. I think we're both tired of iterating on this PR :).
> 
> Just needs fixing the use of `TEST_SKIP`. This will fail the test instead of skipping it. See https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/testHarness.cpp#L729

Hi 
I am ready to fix if you propose better names. At least test cases names. Regarding  _ext _core postfix -  maybe I should be compatible with previous version and old name should be core, and new test cases should contain "_ext" in name. Additionally I could change names from work_group_functions to subgroup_functions. What do you think ?

kpet:
These are fine IMHO. I was really thinking of `f_ptr` and `subgroupsApiSet.get_enum_max_size()`.The core and KHR enums have the same value so there's no need for such a long-winded abstraction. We could just use the core enums (potentially with a static assert to guarantee that they have the same value as the KHR enums but even that is maybe superfluous).

`f_ptr` could contain the name of the function to make things more readable.

gwawiork:
Yes I see this they are the same. Two doubts here:
- what if somebody would compile tests against OpenCL 2.0  where only one set of enums are defined
- then it looks only IFP test make sens prepare in doubled version. Others could be single , testing core API ?

kpet:
>what if somebody would compile tests against OpenCL 2.0 where only one set of enums are defined

We've agreed in the working group that having the CTS depend on the latest headers was fine.

>then it looks only IFP test make sens prepare in doubled version. Others could be single , testing core API ?

Well, we still want to test both sets of API calls. I agree that the difference of coverage is small compared to having to duplicate everything but I think we don't really have a choice. TBH, these tests are quick enough to run that I'm not overly worried.

gwawiork:
Hope this time will be better :)

PR #687, opened 2020-03-18
Fix memory leak in api - zero_sized_enqueue

radszy:
The test enqueues kernel and requests event associated with its
execution, but then never releases the event.


gwawiork:
LGTM

Issue #684, opened 2020-03-17
Reduce the run time of tests


kpet:
Some tests take a long time to run, making them harder to use for regression testing. We should look for opportunities to reduce the execution time while retaining enough coverage.

Jeremy-Kemp:
Thread dimensions uses a kernel (`clear_memory_kernel`) to memset an enormous buffer before running each "real" kernel. Changing this to a `clEnqueueWriteBuffer` would improve runtimes considerably.

(I have this already as a patch and will be more than happy to provide).

kpet:
That sounds great! One could probably use `clEnqueueFillBuffer` as well. Maybe worth creating a separate issue for this one.

PR #694, opened 2020-03-20
Thread Dimensions: Improve execution time.

### LINKED ISSUE #692: Reduce thread dimensions execution time
From #684 

> Thread dimensions uses a kernel (`clear_memory_kernel`) to memset an enormous buffer before running each "real" kernel. Changing this to a `clEnqueueWriteBuffer` would improve runtimes considerably.

> (I have this already as a patch and will be more than happy to provide).

My patch is actually to use `clEnqueueFillBuffer` as @kpet mentioned on the original issue. I'll post a PR.

Jeremy-Kemp:
Replaced a kernel used for memseting a large buffer with clEnqueueFillBuffer.

Fixes #692. Addresses #684.

### DIFF
--- a/test_conformance/thread_dimensions/test_thread_dimensions.cpp
+++ b/test_conformance/thread_dimensions/test_thread_dimensions.cpp
@@ -139,12 +139,6 @@ static const char *thread_dimension_kernel_code_atomic_long =
 "        if (error)\n"
 "                atom_or(&dst[t_address-start_address], error);\n"
 "\n"
-"}\n"
-"\n"
-"__kernel void clear_memory(__global uint *dst)\n\n"
-"{\n"
-"            dst[get_global_id(0)] = 0;\n"
-"\n"
 "}\n";
 
 static const char *thread_dimension_kernel_code_not_atomic_long =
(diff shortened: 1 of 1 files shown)


Jeremy-Kemp:
Without wanting to give specific figures, the speedup is significant. Especially so in environments where device code is slow as you mentioned (we've been running this patch for quite some time to reduce the cost of running the test).

I've not run this on other implementations, but I'd always thought that unified memory implementations would see a greater benefit than those with discrete memory (which perhaps may explain why your testing on Nvidia doesn't seem to produce much benefit).

gwawiork:
I will check the improvement on my side.

gwawiork:
Tested on Intel Coffee Lake integrated gfx , Windows 10:
Before fix (2 runs)
PASSED 12 of 12 tests.
command took 0:6:48.35 (408.35s total)
PASSED 12 of 12 tests.
command took 0:7:19.11 (439.11s total)

With fix (2 runs)
PASSED 12 of 12 tests.
command took 0:7:14.99 (434.99s total)
PASSED 12 of 12 tests.
command took 0:7:24.33 (444.33s total)

I do not see improvement.

kpet:
@gwawiork Thanks! No improvement but no obvious regression either. I'm merging this.

PR #699, opened 2020-03-21
[QC] cl20: bruteforce: Fix domains used in fast math mode. 

lakshmih:
Clamp input arrays to valid domains of several math functions being
tested.

Khronos Bug: https://github.com/KhronosGroup/OpenCL-CTS/issues/491
Test Suite Affected: bruteforce Subtests: sin, cos, sincos, reciprocal

### DIFF
--- a/test_conformance/math_brute_force/unary.cpp
+++ b/test_conformance/math_brute_force/unary.cpp
@@ -506,14 +506,18 @@ static cl_int TestFloat( cl_uint job_id, cl_uint thread_id, void *data )
         float p_j = *(float *) &p[j];
         if ( strcmp(fname,"sin")==0 || strcmp(fname,"cos")==0 )  //the domain of the function is [-pi,pi]
         {
-          if( fabs(p_j) > M_PI )
-            p[j] = NAN;
+            if (fabs(p_j) > M_PI) ((float *)p)[j] = NAN;
         }
 
         if ( strcmp( fname, "reciprocal" ) == 0 )
         {
-          if( fabs(p_j) > 0x7E800000 ) //the domain of the function is [2^-126,2^126]
-            p[j] = NAN;
+            const float l_limit = HEX_FLT(+, 1, 0, -, 126);
+            const float u_limit = HEX_FLT(+, 1, 0, +, 126);
+
+            if (fabs(p_j) < l_limit
+                || fabs(p_j)
+                    > u_limit) // the domain of the function is [2^-126,2^126]
+                ((float *)p)[j] = NAN;
         }
       }
     }
--- a/test_conformance/math_brute_force/unary_two_results.cpp
+++ b/test_conformance/math_brute_force/unary_two_results.cpp
@@ -239,8 +239,7 @@ int TestFunc_Float2_Float(const Func *f, MTdata d, bool relaxedMode)
             if (relaxedMode && strcmp(f->name, "sincos") == 0)
             {
               float pj = *(float *)&p[j];
-              if(fabs(pj) > M_PI)
-                p[j] = NAN;
+              if (fabs(pj) > M_PI) ((float *)p)[j] = NAN;
             }
           }
         }
@@ -252,8 +251,7 @@ int TestFunc_Float2_Float(const Func *f, MTdata d, bool relaxedMode)
             if (relaxedMode && strcmp(f->name, "sincos") == 0)
             {
               float pj = *(float *)&p[j];
-              if(fabs(pj) > M_PI)
-                p[j] = NAN;
+              if (fabs(pj) > M_PI) ((float *)p)[j] = NAN;
             }
           }
         }


lakshmih:
@kpet @bashbaug  please review

lakshmih:
We were able to build and test this for Android and for VS2015. I'm not sure what platform the integration build failed on.
@bashbaug @kpet can you please review and help merge?

kpet:
Linking the previous PR (https://github.com/KhronosGroup/OpenCL-CTS/pull/641) so we don't lose track of what was already said. Please don't re-create new PRs, it makes it hard to follow the dicussion.

test_conformance/math_brute_force/unary.cpp
```diff
@@ -491,13 +491,13 @@ static cl_int TestFloat( cl_uint job_id, cl_uint thread_id, void *data )
...
           if( fabs(p_j) > M_PI )
-            p[j] = NAN;
+            ((float*)p)[j] = M_PI;
         }
 
         if ( strcmp( fname, "reciprocal" ) == 0 )
         {
           if( fabs(p_j) > 0x7E800000 ) //the domain of the function is [2^-126,2^126]
-            p[j] = NAN;
+            ((float*)p)[j] = 0x7E800000;
```
kpet: Presumably this results in an integer to floating-point conversion which is probably not what you want.

lakshmih: Thanks, I'll check the integer to floating point comparison. 

As to the verification approach being proposed, if the generated input was outside the domain of acceptable inputs, we are setting it to one that is within (in this case the last in the range of acceptable inputs). This should void the need for any special case handling in the verification code as well.  Does that address the concerns you had on the previous pull request?

lakshmih: Addressed the comment about the cast

lakshmih:
@kpet @bashbaug Could you please take another look? I believe the patch now addresses your questions while still fixing an existing bug.

jlewis-austin:
My only concern is if we are losing coverage for NaN inputs when not in fast math mode (which seems likely since there is no special-case treatment of fast math mode here).
Is the purpose of this change to pass conformance, or simply to allow the tests to work with fast math mode on?

jlewis-austin:
There is a lengthy special-case in the verification code (https://github.com/KhronosGroup/OpenCL-CTS/blob/efc9f38fdf263f9855f35196f1836c9bd8ce6630/test_conformance/math_brute_force/unary.cpp#L605) that handles gTestFastRelaxed, including filtering out NaN and Inf input results. So as currently written, it appears that leaving the invalid inputs as NaN would be correct and avoid changing inputs of non-relaxed tests.

lakshmih:
@jlewis-austin  I agree that we should leave invalid inputs as NAN. However, the code as currently written has a bug where we assign NAN (a float value) to an int array, resulting in an undesired implicit conversion. This also means the code that filters out NAN inputs in the code you referenced doesn't recognize such inputs as NAN. 

The original pull request fixes this so we are assigning float to float. Please see below: 

https://github.com/KhronosGroup/OpenCL-CTS/pull/641/commits/17b110e803d6a32ba9d5eb70cacd3c4a5be58dd3

jlewis-austin:
> 
> The original pull request fixes this so we are assigning float to float. Please see below:
> 
> [17b110e](https://github.com/KhronosGroup/OpenCL-CTS/commit/17b110e803d6a32ba9d5eb70cacd3c4a5be58dd3)

@lakshmih yes I understand that. I was attempting to help you answer the question raised in that review. @kpet mentioned that the verification code "is wrong as well or has an ugly special case". My comment is that it does indeed have an ugly special case that appears to handle NaN inputs correctly. So, aside from potential refactoring, the original fix looks correct.

test_conformance/math_brute_force/unary.cpp
```diff
@@ -506,14 +506,15 @@ static cl_int TestFloat( cl_uint job_id, cl_uint thread_id, void *data )
...
+            if (fabs(p_j) > M_PI) ((float *)p)[j] = NAN;
         }
 
         if ( strcmp( fname, "reciprocal" ) == 0 )
         {
-          if( fabs(p_j) > 0x7E800000 ) //the domain of the function is [2^-126,2^126]
-            p[j] = NAN;
+            cl_uint limit = 0x7E800000;
+            if (fabs(p_j) > *((float *)&limit)) // the domain of the function is
+                                                // [2^-126,2^126]
```
jlewis-austin, line 516: We're missing equivalent code to exclude small numbers from the domain

lakshmih:
> Looks like a good fix. Did the formatting changes come from clang-format?

yes

test_conformance/math_brute_force/unary.cpp
lakshmih: Fixed.

lakshmih:
@jlewis-austin  Can you review and help merge? Thanks.

jlewis-austin:
LGTM, but you'll have to get @kpet approval. It's showing as blocked by an unresolved comment he made on an earlier iteration.

kpet (dismissed):
Approving. While this is definitely "hiding" deeper issues with this code, this PR does fix an immediate problem.

lakshmih:
Thanks @kpet   @jlewis-austin 
Can you merge as well, or should I? Not sure what the protocol is. 
I'm not sure I agree with the check-format changes but let me know if you want me to make them.

kpet:
@lakshmih The format issues need to be fixed. The whole point of automated formatting is to avoid these discussions :). Agree the line breaks aren't the best, you can try to move the comment to improve this.

lakshmih:
@kpet Can you merge if everything looks good? Thanks!

kpet (approved):

kpet:
All looks good, thanks! Merging.

PR #702, opened 2020-03-23
Add minimum version for relaxed precision bruteforce tests.

aarongreig:
OpenCL 1.2 does not specify any ULP requirements when fast relaxed math
is enabled.

### DIFF
--- a/test_conformance/math_brute_force/main.cpp
+++ b/test_conformance/math_brute_force/main.cpp
@@ -682,6 +682,10 @@ int test_divide( cl_device_id deviceID, cl_context context, cl_command_queue que
 {
     return doTest( "divide" );
 }
+int test_divide_no_relaxed( cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements )
+{
+    return doTest( "divide_no_relaxed" );
+}
 int test_divide_cr( cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements )
 {
     return doTest( "divide_cr" );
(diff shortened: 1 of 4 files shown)


test_conformance/math_brute_force/main.cpp
```diff
@@ -774,7 +774,7 @@ test_definition test_list[] = {
     ADD_TEST( sinh ),
     ADD_TEST( sinpi ),
     ADD_TEST( sqrt ),
-    ADD_TEST( sqrt_cr ),
+    ADD_TEST_VERSION( sqrt_cr, Version(2, 0) ),
```
alycm: It seems like this will exclude the `sqrt_cr` test completely?   But the ULP should still be tested for the non-relaxed case.

Sorry, I've not looked at how these kind of cases are generally handled.

aarongreig: from what I can tell the `_cr` variants are the relaxed math tests, the regular `sqrt` test above remains unchanged

alycm: I believe the `_cr` is for "correctly rounded": https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT

aarongreig:
This is a suggested solution to this problem, I'm aware that it isn't ideal and I welcome feedback from those better versed in this corner of the codebase

kpet:
I believe this has been superseded by https://github.com/KhronosGroup/OpenCL-CTS/pull/819. @FranklandJack Could you confirm?

FranklandJack:
> I believe this has been superseded by #819. @FranklandJack Could you confirm?

Yep, I can confirm that is the case. @aarongreig are you happy for this to be closed?

aarongreig:
yep, closing

Issue #468, opened 2019-10-03
Remove debug flags from conversions and geometrics release tests


jlewis-austin:
These test sets override any existing CMAKE_C_FLAGS_RELEASE flags and replace with flags disabling optimizations. 
The comment in conversions CMakeLists.txt is "# VS2005 AMD64 WAR (basic_test_conversions.c causes internal compiler error)" but it overrides the flags for all platforms. It builds fine on VS2015 and up at least.
For geometrics, the comment is "# VS2005 optimization WAR for geom_length" and flags are overridden for MSVC only.
Per @kpet comment, proposal is to remove all flag overrides and go from there. If any fixes are required, they should be done generically in source if possible or targeted to specific compilers/versions and documented as such.

kpet:
Fixed by #704.

PR #704, opened 2020-03-24
Preserve existing release flags in MSVC

EwanC:
Without this the `/MD` MSVC Release build flag gets lost, resulting in the below error when using a Ninja CMake generator with VS2019 compiler. 

```
harness.lib(testHarness.cpp.obj) : error LNK2038: mismatch detected for 'RuntimeLibrary': value 'MD_DynamicRelease' doesn't match value 'MT_StaticRelease' in test_conversions.cpp.obj
```

Before replacement in this example `CMAKE_C{XX}_FLAGS_RELEASE` is `/MD /O2 /Ob2 /DNDEBUG`, after replacement `/MD /Od /Ob0 /DNDEBUG`


kpet:
It's seems we'd already been round that loop: https://github.com/KhronosGroup/OpenCL-CTS/issues/468 :).

EwanC:
Well spotted finding that issue :slightly_smiling_face:   I'm just waiting for the conversions tests to finish on my local machine before pushing the proposed change removing these flags. Unfortunately on non-wimpy mode this takes a while to run.

PR #706, opened 2020-03-25
Skip, don't fail, image tests on no image support

### LINKED ISSUE #693: image tests fail when not supported by device
When testing using the `opencl_conformance_tests_12_quick.csv` the following test executables fail when the OpenCL device under test does not supported images even though image support is an optional feature:

* `test_cl_copy_images`
* `test_cl_fill_images`
* `test_cl_get_info`
* `test_cl_read_write_images`
* `test_kernel_image_methods`
* `test_image_streams`
* `test_samplerless_reads`

This is a behaviour change from `cl12_trunk` where the tests are reported as being skipped if images are not supported.

kbenzie:
Fixes #693 by returning `TEST_SKIP` instead of `TEST_FAIL` from the
`verifyImageSupport()` kernel helper function when
`checkForImageSupport()` reports the device does not support images.

### DIFF
--- a/test_common/harness/kernelHelpers.cpp
+++ b/test_common/harness/kernelHelpers.cpp
@@ -1367,12 +1367,17 @@ size_t get_pixel_bytes( const cl_image_format *fmt )
 
 test_status verifyImageSupport( cl_device_id device )
 {
-    if( checkForImageSupport( device ) )
+    int result = checkForImageSupport( device );
+    if( result == 0 )
     {
-        log_error( "ERROR: Device does not supported images as required by this test!\n" );
-        return TEST_FAIL;
+        return TEST_PASS;
     }
-    return TEST_PASS;
+    if( result == CL_IMAGE_FORMAT_NOT_SUPPORTED )
+    {
+        log_error( "SKIPPED: Device does not supported images as required by this test!\n" );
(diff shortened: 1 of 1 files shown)


test_common/harness/kernelHelpers.cpp
```diff
@@ -1369,8 +1369,8 @@ test_status verifyImageSupport( cl_device_id device )
 {
     if( checkForImageSupport( device ) )
     {
-        log_error( "ERROR: Device does not supported images as required by this test!\n" );
-        return TEST_FAIL;
+        log_error( "SKIPPED: Device does not supported images as required by this test!\n" );
+        return TEST_SKIP;
```
kpet: This doesn't quite work. With the code as is you'd end up skipping the test if the query in `checkForImageSupport` wasn't successful. I suggest testing for `CL_IMAGE_FORMAT_NOT_SUPPORTED` explicitly:
- 0 => `TEST_PASS`
- `CL_IMAGE_FORMAT_NOT_SUPPORTED` => `TEST_SKIP`
- else => `TEST_FAIL`

kbenzie: Okay, I'll make these changes.

PR #708, opened 2020-03-25
Generic Address Space: Skip test on unsupported devices.

Jeremy-Kemp:
Where a device does not have to support this test, skip it.


Jeremy-Kemp:
Sure. Do you want it as part of the "normal" build (aka add to [travis.sh](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/travis.sh)), or as additional jobs (aka in [.travis.yml](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/.travis.yml))

kpet:
I think we should add that as part of the standard build. I would expect the flag to mostly cover additional code.

PR #712, opened 2020-03-25
Generic Address Space: Skip tests utlilising SVM on devices that do n…

Jeremy-Kemp:
…ot support SVM.

Where a device does not support SVM, do not run the generic address space test(s) that rely on SVM.

### DIFF
--- a/test_common/harness/threadTesting.h
+++ b/test_common/harness/threadTesting.h
@@ -23,6 +23,7 @@
 #endif
 
 #define TEST_NOT_IMPLEMENTED        -99
+#define TEST_SKIPPED_ITSELF         -100
 
 typedef int (*basefn)(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements);
 extern int test_threaded_function( basefn fnToTest, cl_device_id device, cl_context context, cl_command_queue queue, int numElements );
--- a/test_conformance/generic_address_space/advanced_tests.cpp
+++ b/test_conformance/generic_address_space/advanced_tests.cpp
@@ -947,59 +947,8 @@ int test_generic_advanced_casting(cl_device_id deviceID, cl_context context, cl_
     return test.Execute(deviceID, context, queue, num_elements);
 }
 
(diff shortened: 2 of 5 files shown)


test_conformance/generic_address_space/advanced_tests.cpp
```diff
@@ -1003,10 +1003,15 @@ int test_generic_ptr_to_host_mem(cl_device_id deviceID, cl_context context, cl_c
...
     /* Test SVM capabilities and select matching tests */
     cl_device_svm_capabilities caps;
+    auto version = get_device_cl_version(deviceID);
+    auto expected_min_version = Version(2, 0);
 
     cl_int error = clGetDeviceInfo(deviceID, CL_DEVICE_SVM_CAPABILITIES, sizeof(caps), &caps, NULL);
     test_error(error, "clGetDeviceInfo(CL_DEVICE_SVM_CAPABILITIES) failed");
 
+    if ((version < expected_min_version) || (version > Version(2,2) && caps == 0))
+        return result;
```
kpet: The problem with this approach is that the test will be reported as passing instead of skipped. We should add support for tests skipping themselves until we have a better way of managing features (I've now got a prototype for this). I suggest we just add a special case to https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/testHarness.cpp#L748, similar to `TEST_NOT_IMPLEMENTED`. `TEST_SKIPPED_ITSELF`?

Jeremy-Kemp: Sure, this will be a useful stepping stone to the more generic solution. I'll put something together today.

test_common/harness/threadTesting.h
```diff
@@ -23,6 +23,7 @@
 #endif
 
 #define TEST_NOT_IMPLEMENTED        -99
+#define TEST_SKIPPED_ITSELF         -100
```
kpet: These really shouldn't be defined here :). Let's ignore this for now.

PR #719, opened 2020-03-31
New spirv_new test case OpSpecConstant 

### LINKED ISSUE #718: Test specialization constants  in spirv_new CTS test suite
I will add coverage for testing specialization constants in spirv_new test suite for devices > 2.1. Test coverage:
data types: 
- char, uchar, short, ushort, int, uint, long, ulong, float, half, double - OpSpecConstant
- boolean - OpSpecConstantTrue, OpSpecConstantFalse

gwawiork:
OpSpecConstant OpSpecConstantTrue OpSpecConstantfalse
Fixes #718


test_conformance/spirv_new/main.cpp
```diff
@@ -81,7 +81,12 @@ void spirvTestsRegistry::addTestClass(baseTestClass *test, const char *testName)
     test_definition testDef;
     testDef.func = test->getFunction();
     testDef.name = testName;
-    testDef.min_version = Version(2, 1);
+    if (strstr(testName, "op_spec_constant_")) {
```
kpet, line 84: This is a nasty hack :). We will need to change the test registration logic to be able to pass an optional min version. I'd like to reuse some of that registration logic a bit more widely to reduce the amount of boilerplate so I'd like to keep it reasonably clean. I suggest doing this as a separate pre-requisite patch.

gwawiork: Ok thanks. Yes it is nasty but works :D.

kpet:
This needs rebasing / merging master into to fix the CI.

test_conformance/spirv_new/spirv_asm/op_spec_constant_uint_simple.spvasm32
```diff
@@ -0,0 +1,74 @@
...
+               OpName %arrayidx1 "arrayidx1"
+               OpDecorate %37 Alignment 4
+         %37 = OpDecorationGroup
+               OpDecorate %__spirv_BuiltInGlobalInvocationId BuiltIn GlobalInvocationId
+               OpDecorate %__spirv_BuiltInGlobalInvocationId Constant
+               OpDecorate %__spirv_BuiltInGlobalInvocationId LinkageAttributes "__spirv_BuiltInGlobalInvocationId" Import
+               OpGroupDecorate %37 %numbers_buffer_addr %spec_constant_value_addr %id
+       %uint = OpTypeInt 32 0
+      %uchar = OpTypeInt 8 0
+     %uint_0 = OpConstant %uint 0
```
kpet, line 34: I'm guessing you intended this to be a spec constant. It would be good to understand why the test passed (it's not really fit for purpose if it lets that through).

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -0,0 +1,182 @@
...
+                references[i], device_results[i]);
+            return TEST_FAIL;
+        }
+    }
+    return TEST_PASS;
+}
+
+
+#define TEST_SPEC_CONSTANT(NAME, type, value, spec_constant_value)             \
+    TEST_SPIRV_FUNC_VERSION(op_spec_constant_##NAME##_simple, Version(2, 1))   \
```
kpet, line 131: Why pass the version since you're always passing the default value? Why was the change to the test registration code needed?

test_conformance/spirv_new/spirv_asm/op_spec_constant_uint_simple.spvasm32
gwawiork: Sorry this is my mistake. The test was failing. I fixed 64 bit version but forgot about 32. Thanks for catching this.

test_conformance/spirv_new/test_op_spec_constant.cpp
gwawiork: This is my typo. This should be Version(2,2). In comparison to other test cases which requires Version(2,1). This is why I need modify registration code (this solves your comment https://github.com/KhronosGroup/OpenCL-CTS/pull/719#discussion_r402922370)

gwawiork:
> I've had more of a look at the tests. I'm concerned that these tests will be hard to maintain due to the sheer volume of duplication across the SPIR-V assembly files. This is not something new but we should start thinking of a solution to this (most of these are just variations where only one type changes). I'm not saying we absolutely need to solve this for this change but we should think of how we could mitigate that, for example:
> 
> * Could we shorten the code?
> * Do we really need all the `OpName`?

Thank you for review of this code and spending your time. Sure something here could be improved. However from my point of view this is additional work and should be done as a additional issue. Regarding OpName I think I can remove them.

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -0,0 +1,182 @@
...
+}
+
+TEST_SPIRV_FUNC(op_spec_constant_false_simple)
+{
+    // 1-st ndrange use default spec const (false) value = value - 1 (first
+    // check verifies that values are different) 2-nd ndrange sets spec const
+    // true so value = value + 1
+    cl_uchar value = (cl_uchar)8;
+    std::vector<cl_uchar> inputs(1024, value);
+    std::vector<cl_uchar> references(1024, value);
```
kpet, line 178: Not saying this needs to change but: why is this a vector? This will run 1024 times the same test. I'm not sure there's a lot of value in that. We could have much simpler SPIR-V if we changed these tests to operate on a single value.

gwawiork: From validation experience more is better then less :). If you insist on change I could prepare.

gwawiork:
> You could have kept _some_ of the OpNames for readability, I was asking a question, not requesting that you remove all of them really :). Sorry if that wasn't clear.

If you do not mind I would leave this as is now :/

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -0,0 +1,182 @@
...
+                      i, references[i], device_results[i]);
+            return TEST_FAIL;
+        }
+    }
+
+    err = clSetProgramSpecializationConstant(prog, 101, sizeof(T),
+                                             &spec_constant_value);
+    SPIRV_CHECK_ERROR(err, "Failed to run clSetProgramSpecializationConstant");
+
+    err = clBuildProgram(prog, 1, &deviceID, NULL, NULL, NULL);
```
StuartDBrady, line 95: This can fail with CL_INVALID_OPERATION, because there are already kernel objects attached to the program. Please use a new program, _without_ a call to `clBuildProgram()` between the call to `clCreateProgramWithIL()` and the call to `clSetProgramSpecializationConstant()`.

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -0,0 +1,182 @@
...
+    SPIRV_CHECK_ERROR(err, "Failed to create numbers_buffer");
+
+    clMemWrapper spec_buffer =
+        clCreateBuffer(context, CL_MEM_USE_HOST_PTR,
+                       sizeof(spec_constant_value), &spec_constant_value, &err);
+    SPIRV_CHECK_ERROR(err, "Failed to create spec_buffer");
+
+    err = clSetKernelArg(kernel, 0, sizeof(clMemWrapper), &numbers_buffer);
+    SPIRV_CHECK_ERROR(err, "Failed to set kernel argument inputs_buffer");
+    err = clSetKernelArg(kernel, 1, sizeof(clMemWrapper), &spec_buffer);
```
StuartDBrady, line 66: This `spec_buffer` argument does not seem to be used for anything, in the kernels that I have looked at. Could you explain its purpose, please, or remove it if it has none?

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -0,0 +1,182 @@
...
+        std::vector<type> references(1024,                                     \
+                                     (type)value + (type)spec_constant_value); \
+        return test_spec_constant(deviceID, context, queue,                    \
+                                  "op_spec_constant_" #NAME "_simple", inputs, \
+                                  references, (type)spec_constant_value);      \
+    }
+
+// type name, type, value init, spec constant value
+TEST_SPEC_CONSTANT(int, cl_int, 20, -11)
+TEST_SPEC_CONSTANT(uint, cl_uint, 25, 43)
```
StuartDBrady, line 143: It is of no real value to have separate signed and unsigned versions of the tests for each size of integer type, because signedness is not visible within SPIR-V, nor is it visible to the `clSetProgramSpecializationConstant()` API function. Please could you remove _either_ the signed tests _or_ the unsigned tests?

test_conformance/spirv_new/spirv_asm/op_spec_constant_char_simple.spvasm32
```diff
@@ -0,0 +1,69 @@
...
+         %25 = OpLoad %_ptr_CrossWorkgroup_uchar %numbers_buffer_addr Aligned 4
+         %26 = OpLoad %uint %id Aligned 4
+   %arrayidx = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_uchar %25 %26
+         %28 = OpLoad %uchar %arrayidx Aligned 1
+       %conv = OpSConvert %uint %28
+         %30 = OpLoad %_ptr_CrossWorkgroup_uchar %spec_constant_value_addr Aligned 4
+         %31 = OpLoad %uchar %30 Aligned 1
+      %conv1 = OpSConvert %uint %31
+        %add = OpIAdd %uint %conv %conv1
+      %conv2 = OpUConvert %uchar %add
```
StuartDBrady, line 61: This could be simplified by removing the `OpSConvert` and `OpUConvert` instructions.

test_conformance/spirv_new/spirv_asm/op_spec_constant_short_simple.spvasm32
```diff
@@ -0,0 +1,71 @@
...
+         %26 = OpLoad %_ptr_CrossWorkgroup_ushort %numbers_buffer_addr Aligned 4
+         %27 = OpLoad %uint %id Aligned 4
+   %arrayidx = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_ushort %26 %27
+         %29 = OpLoad %ushort %arrayidx Aligned 2
+       %conv = OpSConvert %uint %29
+         %31 = OpLoad %_ptr_CrossWorkgroup_ushort %spec_constant_value_addr Aligned 4
+         %32 = OpLoad %ushort %31 Aligned 2
+      %conv1 = OpSConvert %uint %32
+        %add = OpIAdd %uint %conv %conv1
+      %conv2 = OpUConvert %ushort %add
```
StuartDBrady, line 63: This could be simplified by removing the OpSConvert and OpUConvert instructions.

test_conformance/spirv_new/spirv_asm/op_spec_constant_short_simple.spvasm64
```diff
@@ -0,0 +1,73 @@
...
+         %26 = OpLoad %_ptr_CrossWorkgroup_ushort %numbers_buffer_addr Aligned 8
+         %27 = OpLoad %ulong %id Aligned 8
+   %arrayidx = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_ushort %26 %27
+         %29 = OpLoad %ushort %arrayidx Aligned 2
+       %conv = OpSConvert %uint %29
+         %32 = OpLoad %_ptr_CrossWorkgroup_ushort %spec_constant_value_addr Aligned 8
+         %33 = OpLoad %ushort %32 Aligned 2
+      %conv1 = OpSConvert %uint %33
+        %add = OpIAdd %uint %conv %conv1
+      %conv2 = OpUConvert %ushort %add
```
StuartDBrady, line 65: This could be simplified by removing the OpSConvert and OpUConvert instructions.

test_conformance/spirv_new/spirv_asm/op_spec_constant_uchar_simple.spvasm64
```diff
@@ -0,0 +1,71 @@
...
+         %25 = OpLoad %_ptr_CrossWorkgroup_uchar %numbers_buffer_addr Aligned 8
+         %26 = OpLoad %ulong %id Aligned 8
+   %arrayidx = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_uchar %25 %26
+         %28 = OpLoad %uchar %arrayidx Aligned 1
+       %conv = OpUConvert %uint %28
+         %31 = OpLoad %_ptr_CrossWorkgroup_uchar %spec_constant_value_addr Aligned 8
+         %32 = OpLoad %uchar %31 Aligned 1
+      %conv1 = OpUConvert %uint %32
+        %add = OpIAdd %uint %conv %conv1
+      %conv2 = OpUConvert %uchar %add
```
StuartDBrady, line 63: This could be simplified by removing the OpUConvert instructions.

test_conformance/spirv_new/spirv_asm/op_spec_constant_uchar_simple.spvasm32
```diff
@@ -0,0 +1,69 @@
...
+         %25 = OpLoad %_ptr_CrossWorkgroup_uchar %numbers_buffer_addr Aligned 4
+         %26 = OpLoad %uint %id Aligned 4
+   %arrayidx = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_uchar %25 %26
+         %28 = OpLoad %uchar %arrayidx Aligned 1
+       %conv = OpUConvert %uint %28
+         %30 = OpLoad %_ptr_CrossWorkgroup_uchar %spec_constant_value_addr Aligned 4
+         %31 = OpLoad %uchar %30 Aligned 1
+      %conv1 = OpUConvert %uint %31
+        %add = OpIAdd %uint %conv %conv1
+      %conv2 = OpUConvert %uchar %add
```
StuartDBrady, line 61: This could be simplified by removing the OpUConvert instructions.

test_conformance/spirv_new/spirv_asm/op_spec_constant_ushort_simple.spvasm32
```diff
@@ -0,0 +1,71 @@
...
+         %26 = OpLoad %_ptr_CrossWorkgroup_ushort %numbers_buffer_addr Aligned 4
+         %27 = OpLoad %uint %id Aligned 4
+   %arrayidx = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_ushort %26 %27
+         %29 = OpLoad %ushort %arrayidx Aligned 2
+       %conv = OpUConvert %uint %29
+         %31 = OpLoad %_ptr_CrossWorkgroup_ushort %spec_constant_value_addr Aligned 4
+         %32 = OpLoad %ushort %31 Aligned 2
+      %conv1 = OpUConvert %uint %32
+        %add = OpIAdd %uint %conv %conv1
+      %conv2 = OpUConvert %ushort %add
```
StuartDBrady, line 63: This could be simplified by removing the OpUConvert instructions.

test_conformance/spirv_new/spirv_asm/op_spec_constant_ushort_simple.spvasm64
```diff
@@ -0,0 +1,73 @@
...
+         %26 = OpLoad %_ptr_CrossWorkgroup_ushort %numbers_buffer_addr Aligned 8
+         %27 = OpLoad %ulong %id Aligned 8
+   %arrayidx = OpInBoundsPtrAccessChain %_ptr_CrossWorkgroup_ushort %26 %27
+         %29 = OpLoad %ushort %arrayidx Aligned 2
+       %conv = OpUConvert %uint %29
+         %32 = OpLoad %_ptr_CrossWorkgroup_ushort %spec_constant_value_addr Aligned 8
+         %33 = OpLoad %ushort %32 Aligned 2
+      %conv1 = OpUConvert %uint %33
+        %add = OpIAdd %uint %conv %conv1
+      %conv2 = OpUConvert %ushort %add
```
StuartDBrady, line 65: This could be simplified by removing the OpUConvert instructions.

test_conformance/spirv_new/test_op_spec_constant.cpp
gwawiork: Could you clarify? My understanding is that changing value by clSetProgramSpecializationConstant requires program rebuilding using clBuildProgram. You can read about this [here](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_setting_spir_v_specialization_constants). Noticed that clSetProgramSpecializationConstant is run second time.

gwawiork: In my opinion there is no need to create program object once again. However I create once again kernel object and clSetKernelArg calls are executed

StuartDBrady: > Could you clarify? My understanding is that changing value by clSetProgramSpecializationConstant requires program rebuilding using clBuildProgram. You can read about this [here](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_setting_spir_v_specialization_constants). Noticed that clSetProgramSpecializationConstant is run second time.

`clSetProgramSpecializationConstant()` must be called _before_ `clBuildProgram()`. `clBuildProgram()` is called by `get_program_with_il()`.

> In my opinion there is no need to create program object once again.

The [specification states](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clBuildProgram) that `clBuildProgram()` returns:
> * `CL_​INVALID_​OPERATION` if there are kernel objects attached to _program_.

The test must not call `clBuildProgram()` for a program after calling `clCreateKernel()` for that same program.

> However I create once again kernel object and clSetKernelArg calls are executed

It is not clear to me what you are saying, here. Are you simply describing what the code in the PR currently does? If so, I've read the code and understand what is written: there's no real need to turn it into English, for me, at least.

gwawiork: Looks that second usage the objects I've made is forbidden.  Then I will create new objects. Thanks

test_conformance/spirv_new/test_op_spec_constant.cpp
gwawiork: Good catch. It looks really not needed.

gwawiork:
@StuartDBrady I've fixed recent review remarks. The changes in summary:
* no redundant OpUConvert OpSConvert
* no redundant OpName
* avoid using program that has already kernel program attached
* remove not used buffer
* Simplified test case - instead of buffers use scalar variable

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -1,5 +1,5 @@
 /******************************************************************
-Copyright (c) 2016 The Khronos Group Inc. All Rights Reserved.
+Copyright (c) 2020 The Khronos Group Inc. All Rights Reserved.
```
StuartDBrady, line 2: Should we only be _adding_ new years? @kpet?

test_conformance/spirv_new/test_op_spec_constant.cpp
StuartDBrady: Note: "_each size_ of integer type". Could you please do this for the other SPIR-V Integer types, i.e. 8-bit (uchar), 16-bit (ushort) and 32-bit (ulong), too? Thanks.

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -0,0 +1,162 @@
...
+
+template <typename T>
+int run_case(cl_device_id deviceID, cl_context context, cl_command_queue queue,
+             const char *name, T init_buffer, T spec_constant_value,
+             T final_value, bool use_spec_constant,
+             bool (*notEqual)(const T &, const T &) = isNotEqual<T>)
+{
+    clProgramWrapper prog;
+    cl_int err = CL_SUCCESS;
+    err = get_program_with_il(prog, deviceID, context, name);
```
StuartDBrady, line 28: As I previously explained, this does not work, as `get_program_with_ir()` calls `clBuildProgram()`.

test_conformance/spirv_new/spirv_asm/op_spec_constant_uint_simple.spvasm32
StuartDBrady: @kpet Wondering what your views on this are now. It's a lot better than it was, although there's still some room for improvement. Would it be okay to address this separately, once we've removing the binaries, do you think?

kpet: @StuartDBrady Assuming you intended to reply to that comment https://github.com/KhronosGroup/OpenCL-CTS/pull/719#pullrequestreview-412082959 and not this thread, I think we should indeed address that after removing the binaries. Getting forward progress on this PR is definitely more important IMHO.

StuartDBrady:
Looks okay to me, assuming @kpet is happy with the resolution of the changes he requested.

StuartDBrady:
@gwawiork, this needs a rebase before it can be merged.

gwawiork:
@StuartDBrady. Thanks a lot for review. After rebase I had to add additional fixing commit. Additionally I've added setting minimal version type as 2.2 for bool type tests

kpet:
This now needs rebasing again.

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+#include "testBase.h"
+#include "types.hpp"
+
+
+template <typename T>
+int run_case(cl_device_id deviceID, cl_context context, cl_command_queue queue,
+             const char *name, T init_buffer, T spec_constant_value,
+             T final_value, bool use_spec_constant,
+             bool (*notEqual)(const T &, const T &) = isNotEqual<T>)
```
bashbaug, line 24: Minor: notEqual doesn't seem to be used?

test_conformance/spirv_new/test_op_spec_constant.cpp
```diff
@@ -0,0 +1,159 @@
...
+    SPIRV_CHECK_ERROR(err, "Failed to create output_buffer");
+
+    err = clSetKernelArg(kernel, 0, sizeof(clMemWrapper), &output_buffer);
+    SPIRV_CHECK_ERROR(err, "Failed to set kernel argument output_buffer");
+
+    size_t work_size = 1;
+    err = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, &work_size, NULL, 0,
+                                 NULL, NULL);
+    SPIRV_CHECK_ERROR(err, "Failed to enqueue kernel");
+    clFinish(queue);
```
bashbaug, line 57: Minor: consider checking the return value from `clFinish`.  Or, could it be omitted entirely?  The subsequent call to `clEnqueueReadBuffer` is blocking.

kpet (approved):
Agree with Ben here. This PR could use some more improvements but it's better to merge it as is to get forward progress and address the remaining comments later.

kpet:
Created #911 to cover the remaining comments, merging.

Issue #625, opened 2020-02-24
Inconsistent code formatting


Jeremy-Kemp:
The tests have an inconsistent formatting style which could be improved.

The repo already has a formatting script [clean_tests.py](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/clean_tests.py). Perhaps this could be run as a pre-commit hook?

Any thoughts on this?

kpet:
I would get rid of that file and use `clang-format` but I think we should defer this until people have upstreamed their local changes (I think @gwawiork has quite a few for example) to reduce rebasing pain.

jrprice:
I second the use of `clang-format`. @kpet Are you proposing that we do a full format of the whole code base? The alternative would be to just apply the format to new changes to avoid the rebase pain and destruction of history (this is the LLVM approach).

Jeremy-Kemp:
Agree that this should be left until things quiet down a bit to make everyone's lives easier.

clang-format would be great, was worried about opening that can of worms but it might be for the best in the long run.

kpet:
@jrprice I was thinking of a full reformat since we don't have that much history in Github but what you're proposing sounds like it could be a good compromise and allow us to get started sooner.

>was worried about opening that can of worms 

I would hope that people see the benefit of having _a_ consistent style, even if braces aren't where they like to see them :).

jrprice:
As discussed in #724, we'd like to do a full reformat of the codebase, but given the potential disruption to those with large changes that have not yet been upstreamed we want to get consensus from active contributors before we do so.

Below is a list of member companies who have contributed to CTS (scraped from commit logs, let me know if I've missed anyone). Please check your box if/when you are happy for us to apply a full reformat to the entire codebase (if you don't have edit permissions here just let me know and I'll do it for you).

- [ ] AMD
- [ ] Arm
- [ ] Codeplay
- [x] Google
- [ ] Imagination
- [ ] Intel
- [ ] NVIDIA
- [ ] Qualcomm
- [ ] Samsung

If you are not a member company but you feel you would be negatively impacted by a full reformat please let us know.

gwawiork:
@jrprice I do not have edit permission. You can check Intel checkbox

Jeremy-Kemp:
@jrprice could you check the Imagination box for me please?

jlewis-austin:
[...]

PR #724, opened 2020-04-03
Add clang-format config and check-format.sh script

jrprice:
Travis CI is now configured to run `check-format.sh` and the job will fail if any format violations are detected. This will only check the format of new/modified code; existing code that violates the format
will not cause a CI failure.

As for the config: I generated this using [whatstyle](https://github.com/mikr/whatstyle), so in theory it's the style that most closely matches the existing codebase (though given that we are currently inconsistent across the codebase I'm not sure what this actually means). I don't have any strong opinions about the style itself, so if anybody wants to change anything about it please feel free to make suggestions and we can argue about them here :-)

We started discussing this in #625. This PR does not reformat the full codebase, it just ensures that future contributions adhere to a consistent code style. I've separately applied the format to the whole codebase [here](https://github.com/jrprice/OpenCL-CTS/commit/53166187cce2ef88f102cb9b71d23cfd40b530e7), so you can see how this style will affect your favorite source files. If people _would_ like to do a full reformat of the codebase I'm fine with that too, it just incurs a bit of rebase pain for the PRs that are currently open.


Jeremy-Kemp:
This is great, thanks for adding it.

I'll have a look at `.clang-format` but likewise don't really mind too much about the format. It'll just be refreshing to have a consistent style :).

I'd be in favour of having it applied to the entire code-base. Perhaps we can get this in first and see if there is any strong reaction (positive or negative) to the style. That'll let us amend anything before applying it to the entire codebase which we could do once any outstanding (active) PR's are resolved.

.travis.yml
```diff
@@ -10,6 +10,8 @@ jobs:
     env: JOB_ARCHITECTURE=arm
   - os: linux
     env: JOB_ARCHITECTURE=aarch64
+  - os: linux
```
Jeremy-Kemp, line 13: Could we make this the first job so that we'll fail earlier?

.clang-format
```diff
@@ -0,0 +1,124 @@
...
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+StatementMacros:
+  - Q_UNUSED
+  - QT_REQUIRE_VERSION
+TabWidth:        8
```
Jeremy-Kemp, line 122: Small preference for 4.

.clang-format
```diff
@@ -0,0 +1,124 @@
...
+UseTab:          Never
```
Jeremy-Kemp, line 123: Small preference for indentation only.

.travis.yml
jrprice: Done. I don't think Travis CI has the concept of job dependencies though, so I think it'll still run all of the jobs and will probably not send the failure email until they've all finished.

.clang-format
```diff
@@ -0,0 +1,124 @@
...
+BreakBeforeBinaryOperators: NonAssignment
+BreakBeforeBraces: Custom
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     0
```
kpet, line 51: Maybe we should have a limit here. 80? 100?

.clang-format
```diff
@@ -0,0 +1,124 @@
...
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+StatementMacros:
+  - Q_UNUSED
+  - QT_REQUIRE_VERSION
+TabWidth:        4
+UseTab:          ForIndentation
```
kpet, line 123: I'd rather we go back to `Never` personally. Without tabs, the code always looks right, regardless of editor configuration.

jrprice: I also had a slight preference for `Never` but gave Jeremy's preference higher weight. It's two against one now though. Sorry Jeremy!

jrprice:
Updated the config again, hopefully we are mostly converged now. For reference I've also updated the branch with the [format applied to the whole codebase](https://github.com/jrprice/OpenCL-CTS/commit/1b77f921abd440916305936b339e51c9e993557c). The column limit change has made it a little more disruptive.

> Ultimately, I think we should reformat the entire code base but I'd rather we hold off for now. There are a few large PRs in flight and several vendors still have rather large changes to upstream (at least arm and Intel AFAIK).

How are we going to decide when it's OK to do this? There will probably always be multiple large changes in flight or pending upstreaming, so at some point we'll have to bite the bullet.

Jeremy-Kemp:
I suggest making a GitLab issue so that we can discuss when to apply this to the entire codebase.

kpet:
I agree with @Jeremy-Kemp that an issue is a good way of tracking this. I see no reason not to have it on Github though. I suggest a list of checkboxes, one per company that usually contributes at least (not ruling out non-members showing up with a request to defer the formatting though it seems unlikely). We can highlight this in the next couple of working group calls so everyone is aware.

jrprice:
I've added the suggested checkbox list to the [existing issue](https://github.com/KhronosGroup/OpenCL-CTS/issues/625#issuecomment-614072214).

Are we happy for this change to go in now? This will help advertise to existing contributors that we are now enforcing a format (because everyone with open PRs will probably have to reformat...).

kpet:
Thanks! You beat me to it :). Merging this.

PR #730, opened 2020-04-11
Make it possible for enqueue_map tests to catch more USE_HOST_PTR issues

### LINKED ISSUE #700: basic:enqueue_map_{buffer,image} can't catch missing host_ptr copy-on-map bugs
These tests update the memory passed as host_ptr instead of computing a reference result in a separate allocation which means that an implementation that uses copies to manage the host_ptr and never updates as part of map/unmap commands the memory pointed to by host_ptr can pass the test when it shouldn't.

kpet:
Don't update the memory passed as host_ptr with the correct result
and compute the reference result using a separate allocation so that
implementations that use copies to manage the host_ptr and never update
as part of map/unmap commands the memory pointed to by host_ptr fail
the test.

Rename initialData to hostPtrData for clarity.

Fixes #700

### DIFF
--- a/test_conformance/basic/test_enqueue_map.cpp
+++ b/test_conformance/basic/test_enqueue_map.cpp
@@ -46,7 +46,8 @@ int test_enqueue_map_buffer(cl_device_id deviceID, cl_context context, cl_comman
     int error;
     const size_t bufferSize = 256*256;
     MTdataHolder d{gRandomSeed};
-    BufferOwningPtr<cl_char> initialData{malloc(bufferSize)};
+    BufferOwningPtr<cl_char> hostPtrData{malloc(bufferSize)};
+    BufferOwningPtr<cl_char> referenceData{malloc(bufferSize)};
     BufferOwningPtr<cl_char> finalData{malloc(bufferSize)};
 
     for (int src_flag_id=0; src_flag_id < ARRAY_SIZE(flag_set); src_flag_id++)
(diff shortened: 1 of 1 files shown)


test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -112,9 +114,9 @@ int test_enqueue_map_buffer(cl_device_id deviceID, cl_context context, cl_comman
 
         for( size_t q = 0; q < bufferSize; q++ )
         {
-            if( initialData[ q ] != finalData[ q ] )
+            if( referenceData[ q ] != finalData[ q ] )
             {
-                log_error( "ERROR: Sample %d did not validate! Got %d, expected %d\n", (int)q, (int)finalData[ q ], (int)initialData[ q ] );
+                log_error( "ERROR: Sample %d did not validate! Got %d, expected %d\n", (int)q, (int)finalData[ q ], (int)hostPtrData[ q ] );
```
Jeremy-Kemp, line 119: Nit-pick, can we print from `referenceData` instead of `hostPtrData`? I think it reads a little better based on the previous condition using `referenceData` and `finalData` (also the memory is not the same).

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -211,10 +215,10 @@ int test_enqueue_map_image(cl_device_id deviceID, cl_context context, cl_command
 
     for( size_t q = 0; q < imageSize * imageSize * 4; q++ )
     {
-      if( initialData[ q ] != finalData[ q ] )
+      if( referenceData[ q ] != finalData[ q ] )
       {
         log_error( "ERROR: Sample %d (coord %d,%d) did not validate! Got %d, expected %d\n", (int)q, (int)( ( q / 4 ) % imageSize ), (int)( ( q / 4 ) / imageSize ),
-                                    (int)finalData[ q ], (int)initialData[ q ] );
+                                    (int)finalData[ q ], (int)hostPtrData[ q ] );
```
Jeremy-Kemp, line 221: Nit-pick, can we print from `referenceData` instead of `hostPtrData`? I think it reads a little better based on the previous condition using `referenceData` and `finalData` (also the memory is not the same).

test_conformance/basic/test_enqueue_map.cpp
kpet: Nice catch, will do!

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -197,9 +202,11 @@ int test_enqueue_map_image(cl_device_id deviceID, cl_context context, cl_command
           mappedPtr[ ( y * rowPitch/sizeof(cl_uint) ) + x ] = value;
 
           // Also update the initial data array
-          value = referenceData[ ( ( offset[ 1 ] + y ) * imageSize + offset[ 0 ] ) * 4 + x ];
+          value =
```
jlewis-austin, line 205: I definitely don't want to start formatting wars, but this section gets harder to read with the new format. I'm not sure if it's the splitting at the =, the max line length, or the combination of the two.

Jeremy-Kemp: @jlewis-austin could you put any formatting discussion points in #625 ? That way we can look at them all in one place.

kpet: Sorry I missed this before merging. I agree the formatting doesn't look great there. Indenting all but the first line of the expressions would make for more readable code. I'm sure there's an option for this.

jlewis-austin: @Jeremy-Kemp done

kpet: This is probably fixed by #739

PR #734, opened 2020-04-15
Remove check_opencl_version function

### LINKED ISSUE #527: Replace all uses of check_opencl_version with get_device_cl_version
We don't need two functions and `check_opencl_version` isn't used much.

ellnor01:
Removing all references to check_opencl_version as similar
get_device_cl_version() can be used instead.

Fixes #527


kpet:
@ellnor01 This needs rebasing.

PR #736, opened 2020-04-15
Use clang-format-9 for Travis CI formatting checks

jrprice:
The default version on Xenial is old, and it doesn't understand the non-boolean version of `AllowShortIfStatementsOnASingleLine`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,12 @@ os:
 jobs:
   include:
   - os: linux
+    dist: bionic
     env: JOB_CHECK_FORMAT=1
+    addons:
+        apt:
+          packages:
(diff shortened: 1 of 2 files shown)


kpet:
Would moving to bionic fix it? Would removing the option be a good resolution? (FWIW, using external LLVM packages was a bit flaky in clvk's CI.)

jrprice:
> Would moving to bionic fix it?

The default clang-format seems to be 6.0 on Bionic, but 9.0 appears to be available from the core repositories so at least we wouldn't need to use the LLVM apt repository.

> Would removing the option be a good resolution?

That's an option too, if we don't care about this option.

Issue #723, opened 2020-04-02
Mismatched new[]/delete and new/free


jrprice:
The `test_loops.cpp` files in the image tests perform mismatched allocation/deallocation in their helper routines.

In clCopyImage/clFillImage/kernel_image_methods, we [allocate with `new[]`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/images/clCopyImage/test_loops.cpp#L199) and then [deallocate with `delete`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/images/clCopyImage/test_loops.cpp#L255) (instead of `delete[]`).
clReadWriteImage used to do this but was fixed in #481.

In clGetInfo/kernel_read_write, we [allocate with `new []` and then convert to a `BufferOwningPtr`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/images/clGetInfo/test_loops.cpp#L184-L185), which [deallocates with `free`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/typeWrappers.h#L290).

I suspect `BufferOwningPtr` is used to avoid leaking memory if the function returns early for any reason. Using `std::vector` for these allocations would be a straightforward way of avoids leaks and fixing both of the above mismatches.

I also note that the helper routines in these files are all fairly similar (although not identical), and can probably be refactored into a single common header (though I haven't examined the differences closely yet).

jrprice:
PR #732 by @ellnor01  broke most of CTS for us because of this same issue. A `malloc` was switched to a `new[]`, the result is captured by a `BufferOwningPtr`, which means it is later deallocated with `free`.

This issue stems from the fact that we are in a strange limbo between C and C++. Can we commit to moving towards C++ only at this point, starting by removing all `malloc`/`free` usages?

kpet:
Yes, moving towards C++ sounds good.

PR #737, opened 2020-04-15
Use std::string for get_device_*_string helpers

jrprice:
Removes need to use BufferOwningPtr to avoid memory leaks.

Contributes to #723.

### DIFF
--- a/test_common/harness/deviceInfo.cpp
+++ b/test_common/harness/deviceInfo.cpp
@@ -13,13 +13,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+
 #include <stdexcept>
+#include <vector>
+
 #include "deviceInfo.h"
 #include "errorHelpers.h"
 #include "typeWrappers.h"
 
-/* Helper to allocate and return a buffer containing device information for the specified device info parameter. */
-static void *alloc_and_get_device_info( cl_device_id device, cl_device_info param_name, const char *param_description )
+/* Helper to return a string containing device information for the specified
+ * device info parameter. */
+static std::string get_device_info(cl_device_id device,
+                                   cl_device_info param_name)
 {
     size_t size = 0;
     int err;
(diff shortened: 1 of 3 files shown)


test_common/harness/deviceInfo.cpp
```diff
@@ -13,13 +13,18 @@
...
 #include "deviceInfo.h"
 #include "errorHelpers.h"
 #include "typeWrappers.h"
 
-/* Helper to allocate and return a buffer containing device information for the specified device info parameter. */
-static void *alloc_and_get_device_info( cl_device_id device, cl_device_info param_name, const char *param_description )
+/* Helper to return a string containing device information for the specified
+ * device info parameter. */
+static std::string get_device_info(cl_device_id device,
+                                   cl_device_info param_name)
```
kpet, line 27: This signature only really works when the returned value is a string. This works for all current uses so I'm inclined to leave it as is.

test_common/harness/deviceInfo.cpp
```diff
@@ -29,44 +34,40 @@ static void *alloc_and_get_device_info( cl_device_id device, cl_device_info para
         throw std::runtime_error("clGetDeviceInfo failed\n");
     }
 
-    if (0 == size)
-        return NULL;
+    if (0 == size) return "";
```
kpet, line 37: We should probably throw an exception here instead of testing against an empty string in all users.

test_common/harness/deviceInfo.cpp
```diff
@@ -29,44 +34,40 @@ static void *alloc_and_get_device_info( cl_device_id device, cl_device_info para
...
 /* Determines if an extension is supported by a device. */
 int is_extension_available(cl_device_id device, const char *extensionName)
 {
-    char *extString = alloc_and_get_device_extensions_string(device);
-
-    BufferOwningPtr<char> extStringBuf(extString);
-
-    return strstr(extString, extensionName) != NULL;
+    std::string extString = get_device_extensions_string(device);
+    return extString.find(extensionName) != std::string::npos;
```
kpet, line 54: Technically this is wrong (the function would report true for "cl_khr_prefix" on an implementation that supports "cl_khr_prefix_and_more"). We should really split the string on spaces and match against individual "words". Since this replicates the existing behaviour, I'm fine with doing this as a separate PR if you prefer.

test_common/harness/deviceInfo.cpp
jrprice: Renamed to `get_device_info_string` to make this more obvious.

test_common/harness/deviceInfo.cpp
jrprice: This occurred to me too but I couldn't think of any examples where this would be the case. Still, makes sense to be robust, so I've done this in this PR.

test_common/harness/deviceInfo.cpp
jrprice: I'm not sure about this. While this might be fine for all current use cases (every path to this function will always result in a non-empty string on a correct OpenCL implementation), this might not always be true. For example, `get_device_il_version_string` could be used to detect the presence of an optional IL (i.e. empty string would be a valid, non-fatal result), or it might be valid for some future implementation to not support any extensions (thus validly returning an empty string). Also this helper function could be used for any of the other device queries, some of which also could produce empty strings.

kpet: An empty string will still have the `NUL` character and be of size 1 ;).

jrprice: Good point 😳 

Done, PTAL.

PR #738, opened 2020-04-16
Remove duplicate format functions

radszy:
This moves the filter_formats and get_format_list functions to a common
file. These functions were roughly the same, with an optional filtering
in some tests for testing mipmaps.

### DIFF
--- a/test_conformance/images/common.cpp
+++ b/test_conformance/images/common.cpp
@@ -1,0 +1,117 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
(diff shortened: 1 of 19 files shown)


test_conformance/images/common.cpp
```diff
@@ -0,0 +1,102 @@
+#include "common.h"
```
kpet, line 1: Please add a license header.

test_conformance/images/common.h
```diff
@@ -0,0 +1,19 @@
+#ifndef IMAGES_COMMON_H
```
kpet, line 1: Please add a license header.

PR #740, opened 2020-04-17
Fix buffer overflow in compiler suite

jrprice:
See the usage of this array [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/e72cff3e7d4af0127c69db7b105bf200faac37cb/test_conformance/compiler/test_build_helpers.cpp#L157).

### DIFF
--- a/test_conformance/compiler/test_build_helpers.cpp
+++ b/test_conformance/compiler/test_build_helpers.cpp
@@ -26,8 +26,6 @@ const char *sample_kernel_code_single_line[] = {
 "\n"
 "}\n" };
 
-size_t sample_single_line_lengths[1];
-
 const char *sample_kernel_code_multi_line[] = {
 "__kernel void sample_test(__global float *src, __global int *dst)",
 "{",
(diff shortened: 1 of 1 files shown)


jrprice:
> We should change the [usage](https://github.com/KhronosGroup/OpenCL-CTS/blob/e72cff3e7d4af0127c69db7b105bf200faac37cb/test_conformance/compiler/test_build_helpers.cpp#L157) to use `ARRAY_SIZE` as well.

Good point. I've changed the way the line lengths arrays are declared/initialized to use `ARRAY_SiZE` for all the functions in this file, and also removed the magic 'number of lines' value passed to `clCreateProgramWithSource`. PTAL.

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -240,18 +246,22 @@ int test_load_discreet_length_source(cl_device_id deviceID, cl_context context,
...
+    size_t line_lengths[num_lines];
+    for (i = 0; i < num_lines; i++)
     {
-        sample_bad_multi_line_lengths[ i ] = strlen( sample_kernel_code_bad_multi_line[ i ] );
+        line_lengths[i] = strlen(sample_kernel_code_bad_multi_line[i]);
     }
 
     /* Now force the length of the third line to skip the actual error */
-    sample_bad_multi_line_lengths[2] -= strlen("thisisanerror");
+    line_lengths[2] -= strlen("thisisanerror");
```
Jeremy-Kemp, line 259: We could statically assert that `line_lengths` is at least 3 elements.

jrprice: Done.

PR #742, opened 2020-04-20
is_extension_available - extension not found in case of space in string pattern

### LINKED ISSUE #741: is_extension_available - white characters
After recent changes checking extension does not work in this [case](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/d3d11/harness.cpp#L77). 
Open question: Should I remove spaces in the string or add method which will remove any white space automatically inside of is_extension_available ?

gwawiork:
Fixes #741


gwawiork:
Ok will push only one commit fixing the issue.

jrprice:
LGTM - needs formatting fix though.

gwawiork:
Thanks. Format fixed

kpet:
All the Travis jobs passed but Travis never reported back to Github. Merging.

PR #744, opened 2020-04-22
Turn on d3d binaries - Windows

gwawiork:
Changes in appveyor configuration file to be able build d3d10 and d3d11 tests.


gwawiork:
I am confused with this change because these tests uses [ARCH ](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/d3d11/CMakeLists.txt#L5)variable. Now I am setting for both platforms -DARCH=x86_64. However for _win32 platform there is a need to set i686  Do you know how to pass in appveyor.yml different variable for different platform? Maybe it can stay as is now? Maybe these settings in CMakeLists.txt are obsoleted and can be removed?

gwawiork:
Finally I've found solution how to set these variables

PR #746, opened 2020-04-24
Fix initialization of input data for api test

jrprice:
Caught by MSAN.


jrprice:
> If you're feeling like making more changes to this test, it would be nice to get rid of that hard-coded `10`

Sure, I'll do that as a follow-up patch for the whole file.

PR #752, opened 2020-04-28
Accept OpenCL 3.0 in version parsing code and use where appropriate

### LINKED ISSUE #751: Add support for parsing OpenCL 3.0 to Version and replace uses of 2.2 experimental with 3.0
There are a number of conditions comparing the reported version to 2.2 instead of 3.0.

kpet:
There were a number of tests against 2.2 that are clearer against 3.0.

Fixes #751

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ else(CMAKE_BUILD_TYPE STREQUAL "release")
     set (BUILD_FLAVOR "debug")
 endif(CMAKE_BUILD_TYPE STREQUAL "release")
 
-add_definitions(-DCL_TARGET_OPENCL_VERSION=220)
+add_definitions(-DCL_TARGET_OPENCL_VERSION=300)
 add_definitions(-DCL_USE_DEPRECATED_OPENCL_2_1_APIS=1)
 add_definitions(-DCL_USE_DEPRECATED_OPENCL_2_0_APIS=1)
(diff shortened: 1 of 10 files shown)


gwawiork:
LGTM, only format error need to be fix.
I think we should run format conversion for all files to avoid such errors

jrprice:
Now that the headers have been updated to move the 3.0 symbols from `CL_EXPERIMENTAL` to `CL_VERSION_3_0`, the code guarded by these dynamic checks won't compile under Travis CI (hence why other PRs are failing CI). We'd either need to change the `#ifdef CL_EXPERIMENTAL` guards to `#ifdef CL_VERSION_3_0`, or just make 3.0 the default and remove the `#ifdef CL_EXPERIMENTAL` guards altogether.

kpet:
I think my vote goes to removing the guards. Should I do that under this PR and widen the scope to "basic changes to support CL3"?

jrprice:
> I think my vote goes to removing the guards. Should I do that under this PR and widen the scope to "basic changes to support CL3"?

Yes, sounds good to me, thanks!

jrprice:
The `CMakeLists.txt` file will also need to be updated to enable 3.0 in the headers.

kpet:
Yup just spotted that I forgot to commit this.

kpet:
Merging this to fix CI.

PR #754, opened 2020-04-28
Replace magic constants in api/kernels tests

jrprice:
Also remove one unused test function.

### DIFF
--- a/test_conformance/api/test_kernels.cpp
+++ b/test_conformance/api/test_kernels.cpp
@@ -194,10 +194,12 @@ int test_execute_kernel_local_sizes(cl_device_id deviceID, cl_context context, c
     clKernelWrapper kernel;
     clMemWrapper            streams[2];
     size_t    threads[1], localThreads[1];
-    cl_float inputData[100];
-    cl_int outputData[100];
     RandomSeed seed( gRandomSeed );
     int i;
+
+    num_elements = 100;
+    std::vector<cl_float> inputData(num_elements);
+    std::vector<cl_int> outputData(num_elements);
 
     /* Create a kernel to test with */
     if( create_single_kernel_helper( context, &program, &kernel, 1, sample_single_test_kernel, "sample_test" ) != 0 )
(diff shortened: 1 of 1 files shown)


test_conformance/api/test_kernels.cpp
```diff
@@ -370,87 +388,6 @@ int test_set_kernel_arg_by_index(cl_device_id deviceID, cl_context context, cl_c
     return 0;
 }
 
-int test_set_kernel_arg_struct(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
```
gwawiork, line 373: We do not need this test ?

jrprice: This test was passing a structure containing `cl_mem` objects as a kernel argument, which is not valid. I assume there is a (valid) test that covers this in the SVM suite. The test was not being used, presumably because it is not valid and would not pass on many implementations.

kpet: This is worth double-checking (the SVM coverage). Also, should we be removing the program source as well? I'm tempted to suggest this should have been a separate PR.

test_conformance/api/test_kernels.cpp
```diff
@@ -194,30 +194,34 @@ int test_execute_kernel_local_sizes(cl_device_id deviceID, cl_context context, c
...
+
     /* Create a kernel to test with */
     if( create_single_kernel_helper( context, &program, &kernel, 1, sample_single_test_kernel, "sample_test" ) != 0 )
     {
         return -1;
     }
 
     /* Create some I/O streams */
-    streams[0] = clCreateBuffer(context, (cl_mem_flags)(CL_MEM_READ_WRITE),  sizeof(cl_float) * 100, NULL, &error);
+    streams[0] = clCreateBuffer(context, (cl_mem_flags)(CL_MEM_READ_WRITE),
```
kpet, line 211: We should probably remove all these `cl_mem_flags` casts at some point. Probably best done as a global sed job in a dedicated PR.

test_conformance/api/test_kernels.cpp
jrprice: [`test_set_kernel_exec_info_svm_ptrs.cpp`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/SVM/test_set_kernel_exec_info_svm_ptrs.cpp) does this the right way.

I can submit the removal of this test (and the source - good catch!) as a separate PR. Just need to fix CI for OpenCL 3.0 first ;-)

test_conformance/api/test_kernels.cpp
jrprice: Opened https://github.com/KhronosGroup/OpenCL-CTS/issues/759 for this.

jrprice:
I've merged in master to removes the aforementioned test and its source. PTAL.

PR #757, opened 2020-04-30
Refactor basic/intmath tests into a single file

jrprice:
These six tests were duplicating almost all of their code in separate files. This brings them into a single file with templated routines to perform the tests.

This change also makes use of more C++ features and utilities from the harness library to make the code cleaner. Hopefully these changes are not too controversial (the code still isn't perfect, but it's a lot better IMO).

I switched the data types to unsigned integers, which addresses a signed integer overflow issue.

This also fixes the verification in the vector variants, which were previously not verifying the full dataset.


jrprice:
> * Where was the undefined behaviour happening?

The input arrays were generated with random integers in the range [0,2^32). These arrays were then added or multiplied together, which sometimes generated signed integer overflow. This was caught by `-fsanitize=signed-integer-overflow`.

> * What coverage do these tests buy us compared to integer_ops? There seems to be a lot of overlap. Also these tests only cover a subset of the types and vector sizes.

My understanding was that these tests (and many others in the `basic` suite) just act as smoke tests to check that the OpenCL implementation is not completely broken. I don't expect that they cover anything that isn't covered in `integer_ops`, just like the image tests in `basic` don't (or shouldn't) cover anything that isn't covered in `images/*`. I wouldn't object to just removing these tests from `basic` though, if that's your preference.

kpet:
Thanks for clarifying!

>My understanding was that these tests (and many others in the basic suite) just act as smoke tests to check that the OpenCL implementation is not completely broken

That might have been the initial spirit but it's not true in practice. I think we'd be better served by a smoke mode on the full tests than by duplicated tests that do similar things. integer_ops has such a mode (all the `quick` variants). I don't know how good those are but I think this is where we should go.

In cases where basic does provide additional coverage we probably should move the tests in question to the dedicated suite.

>I don't expect that they cover anything that isn't covered in integer_ops,

I think the only difference is that these tests have coverage for multiply-accumulate but those in integer_ops don't. 

>just like the image tests in basic don't (or shouldn't) cover anything that isn't covered in images/*.

That's a nice counter example :). image tests in basic **do** cover things that are not covered by the dedicated suites.

>I wouldn't object to just removing these tests from basic though, if that's your preference.

I think over time we should aim to remove duplicated tests and make tests that exercise a given area more easily discoverable/auditable and relateable to a given section of the specification where possible.

Given the uncertainty around coverage and the likely need to clean up more of integer_ops to get a clearer picture, I suggest we get this change in as it's a great step forward and create an issue to look into deleting/folding these into integer_ops.

How does that sound?

jrprice:
That all sounds good to me. I agree that there's lots of stuff in basic that would be better placed in existing or new dedicated suites.

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,295 @@
...
+                         __global TYPE *srcC,
+                         __global TYPE *dst)
+  {
+      int  tid = get_global_id(0);
+      dst[tid] = srcA[tid] * srcB[tid] + srcC[tid];
+  };
+)";
+
+template <typename T>
+int verify_add(const std::vector<T> &inputA, const std::vector<T> &inputB,
```
kpet, line 64: You could probably unify a lot of the verification functions with https://en.cppreference.com/w/cpp/header/functional but I don't think it's worth spending too much time on this.

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,295 @@
...
+    std::vector<T> inputA(num_elements * N);
+    std::vector<T> inputB(num_elements * N);
+    std::vector<T> inputC(num_elements * N);
+    std::vector<T> output(num_elements * N);
+    MTdataHolder d(gRandomSeed);
+    for (int i = 0; i < num_elements; i++)
+    {
+        inputA[0] = (T)genrand_int64(d);
+        inputB[1] = (T)genrand_int64(d);
+        inputC[2] = (T)genrand_int64(d);
```
kpet, line 161: I'm guessing you intended `inputX[i]`.

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,295 @@
...
+    test_error(err, "clCreateBuffer failed");
+    streams[1] =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, datasize, NULL, &err);
+    test_error(err, "clCreateBuffer failed");
+    streams[2] =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, datasize, NULL, &err);
+    test_error(err, "clCreateBuffer failed");
+    streams[3] =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, datasize, NULL, &err);
+    test_error(err, "clCreateBuffer failed");
```
kpet, line 178: Maybe this could be a loop.

test_conformance/basic/test_intmath.cpp
jrprice: Whoops 😳

test_conformance/basic/test_intmath.cpp
jrprice: Done. Also refactored things a bit to use loops for kernel creation/args.

test_conformance/basic/test_intmath.cpp
jrprice: I had started doing this but thought that might have been changing too much. I've refactored things further now to include the use of lambda for results verification.

jrprice:
I've refactored things further to address the review comments and generally make things more concise. PTAL.

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,233 @@
...
+    };
+
+    clMemWrapper streams[4];
+    cl_int err;
+
+    if (std::is_same<T, cl_ulong>::value && !gHasLong)
+    {
+        log_info("64-bit integers are not supported on this device. Skipping "
+                 "test.\n");
+        return 0;
```
gwawiork, line 107: TEST_SKIP ?

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,233 @@
...
+        test_error(err, "clEnqueueReadBuffer failed\n");
+
+        // Verify results
+        for (int i = 0; i < num_elements * N; i++)
+        {
+            T r = test.ref(inputA[i], inputB[i], inputC[i]);
+            if (r != output[i])
+            {
+                log_error("failed\n");
+                return -1;
```
gwawiork, line 184: TEST_FAIL ?

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,233 @@
...
+                                  output.data(), 0, NULL, NULL);
+        test_error(err, "clEnqueueReadBuffer failed\n");
+
+        // Verify results
+        for (int i = 0; i < num_elements * N; i++)
+        {
+            T r = test.ref(inputA[i], inputB[i], inputC[i]);
+            if (r != output[i])
+            {
+                log_error("failed\n");
```
gwawiork, line 183: Can you add more details: ref value, obtained value, i ?

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,233 @@
...
+        log_info("passed\n");
+    }
+
+    return err;
+}
+
+int test_intmath_int(cl_device_id device, cl_context context,
+                     cl_command_queue queue, int num_elements)
+{
+    return test_intmath<cl_uint, 1>(device, context, queue, num_elements,
```
gwawiork, line 196: Previously cl_int types was tested not cl_uint. Maybe both should be tested? The same remark to all other cases

jrprice: This depends on whether we want these tests to be more than just a basic smoke test, as discussed above. I switched to `uint` as there was a signed overflow bug in the previous tests. I can add signed `int` back if we really want that coverage here (just need to do some range reduction on the input values to avoid overflow).

gwawiork: I am ok with this but then little confusion with function names (test_intmath_int2).  Consider changing names to test_intmath_uint2?

kpet: I think we shouldn't do that in this PR but as part of the follow up issue where we look into folding these tests into integer_ops.

gwawiork: then LGTM

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,240 @@
...
+        test_error(err, "clEnqueueReadBuffer failed\n");
+
+        // Verify results
+        for (int i = 0; i < num_elements * N; i++)
+        {
+            T r = test.ref(inputA[i], inputB[i], inputC[i]);
+            if (r != output[i])
+            {
+                log_error("\n\nverification failed at index %d\n", i);
+                log_error("-> inputs: %llu, %llu, %llu\n",
```
kpet, line 184: This will trigger `-Wformat` on some platforms but I wouldn't lose sleep over that.

test_conformance/basic/test_intmath.cpp
```diff
@@ -0,0 +1,240 @@
...
+    };
+
+    clMemWrapper streams[4];
+    cl_int err;
+
+    if (std::is_same<T, cl_ulong>::value && !gHasLong)
+    {
+        log_info("64-bit integers are not supported on this device. Skipping "
+                 "test.\n");
+        return TEST_SKIP;
```
kpet, line 107: This is quirky and we should fix the harness but the correct way of doing this at the moment is to return `TEST_SKIPPED_ITSELF`.

test_conformance/basic/test_intmath.cpp
jrprice: Yeah, it struck me while writing this code that we should probably provide C++ stream oriented logging functions to avoid this sort of issue.

kpet: Yes, we should definitely rework logging as a whole at some point.

kpet:
@jrprice Please create an issue to cover looking at folding this into integer_ops.

PR #762, opened 2020-05-05
Test feature macro OpenCL 3.0

### LINKED ISSUE #755: OpenCL 3.0 feature macros CTS test
I am going to prepare CTS test  to validate feature macros in OpenCL 3.0
**Where:** Inside of test_compiler suite
**How:** Similar to test_image_macro.cpp . The test will verify if particular param_names values returned by clDeviceInfo are equal to values that corresponds to turn on/off feature. Then appropriate kernel will be executed to validate feature turn on/off
**Test scope:**
- [x] __opencl_c_3d_image_writes
- [x] __opencl_c_atomic_order_acq_rel
- [x] __opencl_c_atomic_order_seq_cst
- [x] __opencl_c_atomic_scope_device
- [x] __opencl_c_atomic_scope_all_devices
- [x] __opencl_c_device_enqueue
- [x] __opencl_c_generic_address_space
- [x] __opencl_c_pipes
- [x] __opencl_c_program_scope_global_variables
- [x] __opencl_c_read_write_images
- [x] __opencl_c_subgroups
- [x] __opencl_c_work_group_collective_functions

gwawiork:
Fixes #755

### DIFF
--- a/test_conformance/compiler/main.cpp
+++ b/test_conformance/compiler/main.cpp
@@ -98,6 +98,21 @@ test_definition test_list[] = {
     ADD_TEST(unload_build_threaded),
     ADD_TEST(unload_build_info),
     ADD_TEST(unload_program_binaries),
+    ADD_TEST_VERSION(feature_macro_atomic_order_acq_rel, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_atomic_order_seq_cst, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_atomic_scope_device, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_atomic_scope_all_devices, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_3d_image_writes, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_device_enqueue, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_generic_adress_space, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_pipes, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_program_scope_global_variables,
+                     Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_read_write_images, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_subgroups, Version(3, 0)),
+    ADD_TEST_VERSION(feature_macro_work_group_collective_functions,
+                     Version(3, 0))
+
 };
 
 const int test_num = ARRAY_SIZE(test_list);
--- a/test_conformance/compiler/main.cpp
+++ b/test_conformance/compiler/main.cpp
@@ -89,7 +89,7 @@ test_definition test_list[] = {
 
     ADD_TEST_VERSION(pragma_unroll, Version(2, 0)),
 
-    ADD_TEST_VERSION(features_macro, Version(2, 0)),
+    ADD_TEST_VERSION(features_macro, Version(3, 0))
     ADD_TEST(unload_valid),
     ADD_TEST(unload_invalid),
     ADD_TEST(unload_repeated),
--- a/test_conformance/compiler/main.cpp
+++ b/test_conformance/compiler/main.cpp
@@ -89,7 +89,7 @@ test_definition test_list[] = {
 
     ADD_TEST_VERSION(pragma_unroll, Version(2, 0)),
 
-    ADD_TEST_VERSION(features_macro, Version(2, 0)),
+    ADD_TEST_VERSION(features_macro, Version(3, 0)),
     ADD_TEST(unload_valid),
     ADD_TEST(unload_invalid),
     ADD_TEST(unload_repeated),
--- a/test_conformance/compiler/main.cpp
+++ b/test_conformance/compiler/main.cpp
@@ -89,7 +89,7 @@ test_definition test_list[] = {
 
     ADD_TEST_VERSION(pragma_unroll, Version(2, 0)),
 
-    ADD_TEST_VERSION(features_macro, Version(3, 0)),
+    ADD_TEST_VERSION(features_macro, Version(2, 0)),
     ADD_TEST(unload_valid),
     ADD_TEST(unload_invalid),
     ADD_TEST(unload_repeated),
@@ -97,7 +97,8 @@ test_definition test_list[] = {
     ADD_TEST(unload_build_unload_create_kernel),
     ADD_TEST(unload_link_different),
     ADD_TEST(unload_build_threaded),
-
+    ADD_TEST(unload_build_info),
+    ADD_TEST(unload_program_binaries),
 
 };
 
--- a/test_conformance/compiler/main.cpp
+++ b/test_conformance/compiler/main.cpp
@@ -96,8 +96,7 @@ test_definition test_list[] = {
     ADD_TEST(unload_build_unload_create_kernel),
     ADD_TEST(unload_link_different),
     ADD_TEST(unload_build_threaded),
-    ADD_TEST(unload_build_info),
-    ADD_TEST(unload_program_binaries),
+
     ADD_TEST_VERSION(feature_macro_atomic_order_acq_rel, Version(3, 0)),
     ADD_TEST_VERSION(feature_macro_atomic_order_seq_cst, Version(3, 0)),
     ADD_TEST_VERSION(feature_macro_atomic_scope_device, Version(3, 0)),
--- a/test_conformance/compiler/main.cpp
+++ b/test_conformance/compiler/main.cpp
@@ -89,6 +89,7 @@ test_definition test_list[] = {
 
     ADD_TEST_VERSION(pragma_unroll, Version(2, 0)),
 
+    ADD_TEST_VERSION(features_macro, Version(2, 0)),
     ADD_TEST(unload_valid),
     ADD_TEST(unload_invalid),
     ADD_TEST(unload_repeated),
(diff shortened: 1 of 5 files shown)


test_conformance/compiler/main.cpp
```diff
@@ -87,7 +87,22 @@ test_definition test_list[] = {
     ADD_TEST( program_binary_type ),
     ADD_TEST( compile_and_link_status_options_log ),
 
-    ADD_TEST_VERSION( pragma_unroll, Version(2, 0) ),
+    ADD_TEST_VERSION( pragma_unroll, Version(3, 0) ),
```
kpet, line 90: This looks incorrect to me. Isn't the min required version 2.0?

gwawiork:
> I suggest using tables and a loop to reduce code duplication and improve maintainability.

Do you suggest have one command line with multiple subcases ? I would stay with separate list of subcases

kpet:
I think the test should loop over {feature macro, device info checks} pairs instead of having the code duplicated many times over. This could be a single test from the framework's PoV. That being said, I agree it would be nice to report individual results but the harness doesn't give us that ability at the moment. If we have to choose, my vote goes to making the code shorter and more generic.

gwawiork:
Refactored - use template function. Test cases are separated.

gwawiork:
I will refactor test to have one test case. 
The step with comparing content of CL_DEVICE_OPENCL_C_FEATURES will be added

kpet:
@gwawiork This now needs rebasing.

gwawiork:
Rebased,

test_conformance/compiler/main.cpp
```diff
@@ -89,15 +89,16 @@ test_definition test_list[] = {
...
+    ADD_TEST_VERSION(features_macro, Version(3, 0)),
     ADD_TEST(unload_valid),
     ADD_TEST(unload_invalid),
     ADD_TEST(unload_repeated),
     ADD_TEST(unload_compile_unload_link),
     ADD_TEST(unload_build_unload_create_kernel),
     ADD_TEST(unload_link_different),
     ADD_TEST(unload_build_threaded),
-    ADD_TEST(unload_build_info),
-    ADD_TEST(unload_program_binaries),
```
kpet, line 100: Please add these two tests back :).

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -0,0 +1,398 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "testBase.h"
+#include <vector>
+#include <algorithm>
+
+const char* macro_supported_source =
```
kpet, line 20: Suggest using C++11 raw literals to avoid having to explicitly manage line endings.

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -0,0 +1,398 @@
...
+    "int n = get_global_id(0); \r\n"
+    "buf[n] = 0; \r\n "
+    "#ifdef %s \r\n"
+    "ERROR; \r\n"
+    "#endif \r\n"
+    "\r\n } \r\n";
+
+
+template <typename T>
+cl_int check_api_feature_info(cl_device_id deviceID, cl_context context,
```
kpet, line 41: This should ideally be split into separate functions to deal with the different kinds of queries. You could have several specialised `test_feature_macro` functions (overloads or different names) that share the compiler and comparison part. This is not strictly required but would make the code a bit easier to read IMHO.

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -0,0 +1,398 @@
...
+    status_not_supported =
+        clBuildProgram(program_not_supported, 1, &deviceID, NULL, NULL, NULL);
+    if (status_supported != status_not_supported)
+    {
+        if (status_not_supported == CL_SUCCESS)
+        {
+            // kernel which verifies not supporting return passed
+            status = false;
+        }
+        if (status_supported == CL_SUCCESS)
```
kpet, line 128: Maybe add an `else` here for readability.

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -0,0 +1,398 @@
...
+    NEW_FEATURE_MACRO_TEST(atomic_order_acq_rel);
+    NEW_FEATURE_MACRO_TEST(atomic_order_seq_cst);
+    NEW_FEATURE_MACRO_TEST(atomic_scope_device);
+    NEW_FEATURE_MACRO_TEST(atomic_scope_all_devices);
+    NEW_FEATURE_MACRO_TEST(device_enqueue);
+    NEW_FEATURE_MACRO_TEST(generic_address_space);
+    NEW_FEATURE_MACRO_TEST(pipes);
+    NEW_FEATURE_MACRO_TEST(read_write_images);
+    NEW_FEATURE_MACRO_TEST(subgroups);
+    NEW_FEATURE_MACRO_TEST(work_group_collective_functions);
```
kpet, line 393: Need to add the following features:
- `__opencl_c_images`
- `__opencl_c_fp64`
- `__opencl_c_int64`

kpet (changes requested):
Looks generally good. The code could be made a bit easier to read and few macros added after the initial spec need to be covered as well.

gwawiork:
@kpet I think all you requests have been addressed in last commits

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -0,0 +1,740 @@
...
+    cl_bool compiler_status;
+    log_info("\n%s ...\n", test_macro_name.c_str());
+    error = check_api_feature_info_supported_image_formats(
+        deviceID, context, api_status,
+        CL_MEM_WRITE_ONLY | CL_MEM_READ_WRITE | CL_MEM_KERNEL_READ_AND_WRITE,
+        CL_MEM_OBJECT_IMAGE3D);
+    if (error != CL_SUCCESS)
+    {
+        return error;
+    }
```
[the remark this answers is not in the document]
gwawiork: ok thanks, fixed.

bashbaug:
Waiting for one final review but aiming to merge soon.  Ping @kpet

kpet (dismissed):
Not a thorough review but looks good. As discussed on the 2020/09/22 teleconference, it's now more valuable to merge this as a base change and iterate on top if need be.

kpet:
@lakshmih @bcalidas Any objections to this getting merged? AFAICT your comments have been addressed.

test_conformance/compiler/main.cpp
```diff
@@ -89,6 +89,7 @@ test_definition test_list[] = {
 
     ADD_TEST_VERSION(pragma_unroll, Version(2, 0)),
 
+    ADD_TEST_VERSION(features_macro, Version(2, 0)),
```
alycm, line 92: This should be 3.0, right?

```suggestion
    ADD_TEST_VERSION(features_macro, Version(3, 0)),
```

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -0,0 +1,750 @@
...
+#include "testBase.h"
+#include <vector>
+#include <algorithm>
+#include "errorHelpers.h"
+
+const char* macro_supported_source = R"(kernel void enabled(global int * buf) {
+        int n = get_global_id(0);
+        buf[n] = 0;
+        #ifndef %s
+            ERROR;
```
alycm, line 25: The point of `ERROR;` is to cause a compiler failure?  Using an error directive seems more explicit.

```suggestion
            #error Expected feature macro was not defined
```

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -0,0 +1,750 @@
...
+        #endif
+})";
+
+const char* macro_not_supported_source =
+    R"(kernel void not_enabled(global int * buf) {
+        int n = get_global_id(0);
+        buf[n] = 0;
+        #ifdef %s
+            ERROR;
```
alycm, line 34:
```suggestion
            #error Feature macro was defined when it was not expected
```

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -0,0 +1,750 @@
...
+    for (auto each_f : vec_device_feature)
+    {
+        vec_device_feature_names.push_back(each_f.name);
+    }
+    sort(vec_to_cmp.begin(), vec_to_cmp.end());
+    sort(vec_device_feature_names.begin(), vec_device_feature_names.end());
+
+    cl_bool result =
+        std::equal(vec_device_feature_names.begin(),
+                   vec_device_feature_names.end(), vec_to_cmp.begin());
```
alycm, line 691: Because this checks the entire range of `vec_device_feature_names` and compares each element of `vec_to_cmp` against it, if `vec_to_cmp` is the longer list then then will pass even though the lists clearly aren't equal.

Swapping the names doesn't fix that because what if `vec_device_feature_names` is longer.

An example of genuine output that clearly contains a problem, but still PASSED:

```
Comparison list of features: CL_DEVICE_OPENCL_C_FEATURES vs API/compiler queries.
Empty list of CL_DEVICE_OPENCL_C_FEATURES returned by clGetDeviceInfo on this device.
Comparison list of features - passed
Supported features based on CL_DEVICE_OPENCL_C_FEATURES API query:

Supported features based on queries to API/compiler :
__opencl_c_fp64
__opencl_c_int64
features_macro passed
PASSED sub-test.
PASSED test.
```

If I switch this to

```suggestion
    cl_bool result = vec_device_feature_names == vec_to_cmp;
```

the test then fails as you would expect.

alycm (commented):
LGTM, I think the 2.0 vs 3.0 point is the only point that I think is wrong, everything else is a quality point that could be addressed in a new PR.

gwawiork:
@alycm Thanks . Version fixed other issues I will open new github Issue

alycm (approved):
Thanks, LGTM, remaining comments moved to https://github.com/KhronosGroup/OpenCL-CTS/issues/979.

alycm:
I'm going to merge this, multiple people have looked at this and I believe all issues have been addressed.  If I'm wrong please file issues for any outstanding point.

PR #763, opened 2020-05-05
Remove duplicated random_in_ranges function

radszy:
This function is duplicated across clCopyImages files, so just keep one
definition in common file, and get rid of the duplicated ones.

### DIFF
--- a/test_conformance/images/common.cpp
+++ b/test_conformance/images/common.cpp
@@ -158,3 +158,10 @@ int get_format_list(cl_context context, cl_mem_object_type imageType,
     test_error(error, "Unable to get list of supported image formats");
     return 0;
 }
+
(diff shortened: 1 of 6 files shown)


test_conformance/images/common.cpp
```diff
@@ -158,3 +158,10 @@ int get_format_list(cl_context context, cl_mem_object_type imageType,
     test_error(error, "Unable to get list of supported image formats");
     return 0;
 }
+
+size_t random_in_ranges(size_t minimum, size_t rangeA, size_t rangeB, MTdata d)
+{
+    if (rangeB < rangeA) rangeA = rangeB;
+    if (rangeA < minimum) return rangeA;
```
kpet, line 165: Not very random :). This suggests the function is used in weird ways. Not worth thinking about in this PR.

PR #764, opened 2020-05-06
Fix test_gl error reporting

DziubanMaciejIntel:
Overwriting 'error' variable after check for Msaa/Depth support was clearing
the error counter incremented after failure in test_image_format_write. In
effect the test might return 0 even if there were errors.

### DIFF
--- a/test_conformance/gl/test_images_write_common.cpp
+++ b/test_conformance/gl/test_images_write_common.cpp
@@ -813,7 +813,7 @@ int test_images_write_common(cl_device_id device, cl_context context,
             get_base_gl_target(targets[ tidx ]) == GL_TEXTURE_2D_MULTISAMPLE_ARRAY)
         {
             bool supports_msaa;
-            error = supportsMsaa(context, &supports_msaa);
+            bool error = supportsMsaa(context, &supports_msaa);
             if( error != 0 ) return error;
             if (!supports_msaa) return 0;
         }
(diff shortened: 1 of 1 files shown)


test_conformance/gl/test_images_write_common.cpp
```diff
@@ -813,15 +813,15 @@ int test_images_write_common(cl_device_id device, cl_context context,
             get_base_gl_target(targets[ tidx ]) == GL_TEXTURE_2D_MULTISAMPLE_ARRAY)
         {
             bool supports_msaa;
-            error = supportsMsaa(context, &supports_msaa);
+            bool error = supportsMsaa(context, &supports_msaa);
```
kpet, line 816: Declaring another variable with the same name will shadow the previous declaration so previously recorded errors won't be tested in the condition below. I'm not sure I understand what's happening and how this fixes the issue.

DziubanMaciejIntel: Yes, I meant to shadow it. IMO the test in line 825 should only test the result of supportsMsaa function. Outer scope `error` variable is used for counting how many formats failed (line 850). Suppose the scenario:
- check msaa support for first format - success, error=0
- first format validation fails - error=1
- check msaa support for second format - success, error=0
- second format validation succeds - error=0
- rest of the formats also pass

So the whole test returns 0, even though the first format validation failed.

kpet: Right, the test isn't designed to stop on the first outer scope error, makes sense. I suggest you rename the inner variable to make this easier to follow and remove the unnecessary bool/int implicit conversions.

kpet:
@DziubanMaciejIntel This needs formating fixes to go in.

kpet:
@DziubanMaciejIntel Thanks but you missed the spaces after the if :). I suggest you use `git-clang-format` to do the formating (see https://github.com/KhronosGroup/OpenCL-CTS/blob/master/check-format.sh for an example), much easier than doing it manually.

DziubanMaciejIntel:
@kpet right, I put too much faith in myself :D Thanks and sorry for notification spam.

PR #765, opened 2020-05-06
Add 3.0 as valid device version to version parsing logic

FranklandJack:


kpet:
This will need formating changes though.

PR #767, opened 2020-05-06
Remove duplicated function declaration in clCopyImage

radszy:
There's no need to redeclare test_copy_image_generic in each file, you
can just do it once in a common file, which is already included where
it's needed.


radszy:
I think we'll have to reject this change then.

PR #768, opened 2020-05-07
Update computeinfo tests

pierremoreau:
Besides the code reformatting and addressing compilation warnings, the main contribution is to ensure that computeinfo does not test for image types and flags that are not supported by the OpenCL version being tested.

### DIFF
--- a/test_conformance/computeinfo/main.cpp
+++ b/test_conformance/computeinfo/main.cpp
@@ -250,33 +250,40 @@ config_info config_infos[] = {
                 cl_uint),
 };
 
-#define ENTRY(T)                                                               \
+#define ENTRY(major, minor, T)                                                 \
     {                                                                          \
-        T, #T                                                                  \
+        { major, minor }, T, #T                                                \
     }
 struct image_type_entry
 {
+    version_t
+        version; // Image type is introduced in this version of OpenCL spec.
     cl_mem_object_type val;
     const char* str;
 };
 static const struct image_type_entry image_types[] = {
-    ENTRY(CL_MEM_OBJECT_IMAGE1D),       ENTRY(CL_MEM_OBJECT_IMAGE1D_BUFFER),
-    ENTRY(CL_MEM_OBJECT_IMAGE2D),       ENTRY(CL_MEM_OBJECT_IMAGE3D),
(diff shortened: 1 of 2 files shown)


pierremoreau:
I kept the reformatting in a separate patch, which can be easily dropped if you want to wait for everyone to reply on #625 first.

test_conformance/computeinfo/main.cpp
```diff
@@ -655,7 +655,7 @@ void dumpConfigInfo(cl_device_id device, config_info* info)
             log_info("\t%s == %zu\n", info->opcode_name, info->config.sizet);
             break;
         case type_cl_ulong:
-            log_info("\t%s == %lld\n", info->opcode_name, info->config.ull);
+            log_info("\t%s == %lu\n", info->opcode_name, info->config.ull);
```
gwawiork, line 674: Should not be `%llu` ?

pierremoreau: I am getting “Format specifies type 'unsigned long long' but the argument has type 'cl_ulong' (aka 'unsigned long')” if I do that. But using `%lu` will be a problem on Win64.
Looking around, there are some standard macros to handle that: https://en.cppreference.com/w/cpp/header/cinttypes. I will switch to using those macros instead.

pierremoreau: This should be addressed in the updated version.

gwawiork: I get hint in MSVC that ull variable has type `typedef unsigned long long cl_ulong` Then for MSVC cl_ulong is long long int

pierremoreau: Right, because on Win64 `unsigned long` is only 32 bits so `unsigned long long` is needed in that scenario, but on Unix `unsigned long` is already 64 bits so `cl_ulong` is just an `unsigned long` there. I had forgotten about that difference when writing the initial version.

Could you please check if the updated pull request works fine with MSCV? The macro `PRIX64` should handle both cases automatically (by changing its definition based on the platform).

gwawiork: Yes I've verified looks ok. Then LGTM.

pierremoreau: Thank you for checking and reviewing.

PR #769, opened 2020-05-09
SPIR-V: BuiltIn variables use the Input storage class

pierremoreau:
The OpenCL SPIR-V Environment Specification was updated to mandate that
all variables decorated with `BuiltIn` must be in the `Input` storage
class, see https://github.com/KhronosGroup/OpenCL-Docs/pull/278.

Global variables with the Input storage class have to be listed in the
entry points’ interface, which was also taken care of in this commit.


pierremoreau:
FYI, I ran spirv-val on all the updated binaries and didn’t get any validation errors (besides the one tracked in https://github.com/KhronosGroup/SPIRV-Tools/issues/3322 about NonWritable), and tested a few binaries on an OpenCL implementation but not all of them.

alycm:
This change made me think a bit more about the spec change (https://github.com/KhronosGroup/OpenCL-Docs/issues/269#issuecomment-626403247), but regardless of the thoughts there I think this change makes sense.  It brings the `spirv_new` tests closer to what current tooling produces.

gwawiork:
LGTM

kpet:
@StuartDBrady You may want to take a look at this change.

pierremoreau:
Friendly ping

pierremoreau:
Thank you for approving the changes and merging.

PR #770, opened 2020-05-12
Mem-leaks from conformance mem_host_flags

JohnKesapidesARM:
Fix various resource leaks in the tests for mem_host_flags.

### DIFF
--- a/test_conformance/mem_host_flags/checker.h
+++ b/test_conformance/mem_host_flags/checker.h
@@ -132,6 +132,7 @@ cBuffer_checker<T>::~cBuffer_checker()
 {
 }
 
+
 template < class T >
 cl_int cBuffer_checker<T>::SetupBuffer()
 {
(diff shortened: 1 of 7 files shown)


test_conformance/mem_host_flags/checker.h
```diff
@@ -130,6 +130,11 @@ cBuffer_checker<T>::cBuffer_checker(cl_device_id deviceID, cl_context context,
 template < class T >
 cBuffer_checker<T>::~cBuffer_checker()
 {
+    clReleaseMemObject(m_buffer);
+    if (m_buffer_parent)
+    {
+        clReleaseMemObject(m_buffer_parent);
+    }
```
kpet, line 137: Hmm, just spotted this: both of these are `clMemWrapper`. You shouldn't have to call `clReleaseMemObject` on them. This looks suspicirous.

JohnKesapidesARM: yeap, I though I reverted this bit. Let me retest to make sure, everything still works

PR #772, opened 2020-05-12
Mem-leaks from conformance pipes

JohnKesapidesARM:
Fix various memory leaks around events.
Convert test to use supplied typewrappers
to avoid memory leaks. Also use error helper
functions to reduce code size.
Use stringstreams to synthesize kernel sources, and
raw c+11 string literals.

### DIFF
--- a/test_conformance/pipes/test_pipe_readwrite_errors.cpp
+++ b/test_conformance/pipes/test_pipe_readwrite_errors.cpp
@@ -64,23 +64,26 @@ const char* pipe_readwrite_errors_kernel_code = {
 
 int test_pipe_readwrite_errors(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
 {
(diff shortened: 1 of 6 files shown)


test_conformance/pipes/test_pipe_readwrite_errors.cpp
```diff
@@ -64,143 +64,79 @@ const char* pipe_readwrite_errors_kernel_code = {
...
+    size_t min_alignment = get_min_alignment(context);
 
     global_work_size[0] = num_elements;
 
     size = num_elements * sizeof(cl_int);
 
     inptr = (cl_int *)align_malloc(size, min_alignment);
 
-    for(i = 0; i < (cl_int)(size / sizeof(int)); i++){
+    for (i = 0; i < (cl_int)(size / sizeof(int)); i++)
```
kpet, line 94: Could just use `num_elements`.

Issue #698, opened 2020-03-20
CSV files for offline/spirv mode need updated


kbenzie:
Splitting out from #689 following discussions with @gwawiork and @kpet 

The `opencl_conformance_tests_21_full_spirv.csv` and `opencl_conformance_tests_generate_spirv.csv` files still reference the now defunct command line arguments `-offlineCompiler spir_v`, `cache .`, and `generate .` which are no longer supported by the conformance tests after switching to unified offline/spirv compilation. The CSV files should be updated to use the new `--compilation-{mode,cache-mode,cache-path,program}` set of command line arguments.

kpet:
Fixed by #774, closing.

PR #774, opened 2020-05-14
Update SPIR-V CSV Options

FranklandJack:
* Update `opencl_conformance_tests_21_full_spirv.csv` with
`/-offlineCompiler/--compilation-mode` and
`/cache/--compilation-cache-path` in order to match the offline
compilation options in the cts executables. Do not add the
`--compilation-cache-mode` and `--compilation-program` options since
they are user specific and the default values are generic.

* Delete `opencl_conformance_tests_generate_spirv.csv` since there is no
longer a `generate` option for offline compilation in the cts
executables.


kpet:
It seems this fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/698 entirely.

PR #775, opened 2020-05-15
Remove `clCreatePipe` call from test_clone_kernel

FranklandJack:
* Remove the call to `clCreatePipe` in the `test_clone_kernel` test as
the result of this call (the pipe object) is unused. Further more
`clCreatePipe` is an optional API in OpenCL-3.0, so even if the pipe
object did have a use, support for this API would need to be queried in
the case of a OpenCL-3.0 driver where returning `CL_INVALID_OPERATION`
would be conformant behaviour. As it stands this test will incorrectly
fail for a 3.0 driver which does not support pipes.

* Correct comment regarding pipes and SVM since these features were not
actually tested in the source.


FranklandJack:
> Fine with this change but we probably ought to have exhaustive coverage of the argument types. Could you create an issue please?

Have created an issue here: https://github.com/KhronosGroup/OpenCL-CTS/issues/777, is this what you had in mind?

PR #776, opened 2020-05-15
Fix return-type warnings with harness library and basic test

layog:
Reproduction steps - Compile with flags `-Werror -Wreturn-type`

### DIFF
--- a/test_common/harness/deviceInfo.h
+++ b/test_common/harness/deviceInfo.h
@@ -30,6 +30,10 @@ extern "C" {
 /* Determines if an extension is supported by a device. */
 int is_extension_available(cl_device_id device, const char *extensionName);
 
+#ifdef __cplusplus
+}
+#endif // __cplusplus
+
 /* Returns a string containing the supported extensions list for a device. */
 std::string get_device_extensions_string(cl_device_id device);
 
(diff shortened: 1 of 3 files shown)


layog:
@kpet Agreed. I wanted to do that but since it's my first commit to this repo, I kept it small.

There are no C dependencies in the code. I'll confirm that and will push upstream.

kpet:
@layog Makes sense, thanks!

keryell:
Just curious: why is there C++ code at the first place in a CTS checking for a C API?

layog:
@keryell I am not sure whether this question is related to this PR.
In general, OpenCL's C-style API is fully accessible through C++ and there is C++ style API as well.

keryell:
I found just curious to have C++ code in the CTS for a C API. How a host without a C++ compiler can pass the CTS? It is just a question out of the blue...

jrprice:
> How a host without a C++ compiler can pass the CTS?

Cross-compile static binaries from a system that does have a C++ compiler?

keryell:
It does not really change the problem: you need a a C++ cross-compiler at the first place. Then you hope the target has enough memory to fit the C++ runtime...

PR #778, opened 2020-05-15
Correct khr Extension Checks for Device Versions

FranklandJack:
* Correct khronos extension device version checks so that 2.0, 2.1 and
2.2 extensions are only require for devices versions < 3.0
(`cl_khr_3d_image_writes`, `cl_khr_image2d_from_buffer` and
`cl_khr_depth_images` are all optional for devices supporting
OpenCL-3.0).

* Remove `cl_khr_fp64` from list or required extensions for 2.0,
`cl_khr_fp64` is a required extension for OpenCL-1.2 devices if they
suport doubles. Check for support of `cl_khr_fp64` on 1.2 devices.

* Remove duplicated `cl_khr_byte_addressable_store` from list of 2.0
extensions. Since this is a 1.1 extension and all versions greater than
1.1 are backwards compatible with 1.1 this will be checked anyway.

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -1872,145 +1872,97 @@ int test_min_max_queue_properties(cl_device_id deviceID, cl_context context, cl_
 
 int test_min_max_device_version(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
 {
-    cl_int error, i;
-    cl_char buffer[ 4098 ];
-    size_t length;
+    // Query for the device version.
+    Version device_cl_version = get_device_cl_version(deviceID);
+    log_info("Returned version %s.\n", device_cl_version.to_string().c_str());
 
-
-    // Device version should fit the regex "OpenCL [0-9]+\.[0-9]+ *.*"
-    error = clGetDeviceInfo( deviceID, CL_DEVICE_VERSION, sizeof( buffer ), buffer, &length );
-    test_error( error, "Unable to get device version string" );
-    if( memcmp( buffer, "OpenCL ", strlen( "OpenCL " ) ) != 0 )
-    {
-        log_error( "ERROR: Initial part of device version string does not match required format! (returned: %s)\n", (char *)buffer );
-        return -1;
-    }
-
-    log_info("Returned version %s.\n", buffer);
-
-    char *p1 = (char *)buffer + strlen( "OpenCL " );
-    while( *p1 == ' ' )
-        p1++;
-    char *p2 = p1;
-    if( ! isdigit(*p2) )
-    {
-        log_error( "ERROR: Major revision number must follow space behind OpenCL! (returned %s)\n", (char*) buffer );
-        return -1;
-    }
-    while( isdigit( *p2 ) )
-        p2++;
-    if( *p2 != '.' )
-    {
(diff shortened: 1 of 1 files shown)


test_conformance/api/test_api_min_max.cpp
```diff
@@ -1946,64 +1946,90 @@ int test_min_max_device_version(cl_device_id deviceID, cl_context context, cl_co
         return -1;
     }
```
kpet, line 1948: Please replace all the version parsing code above with `get_device_cl_version`.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1946,64 +1946,90 @@ int test_min_max_device_version(cl_device_id deviceID, cl_context context, cl_co
...
 
-    if( major * 10 + minor >= 11 )
+
+    int version_as_integer = major * 10 + minor;
+    if (version_as_integer >= 11)
     {
         log_info( "Checking for required extensions for OpenCL 1.1 and later devices...\n" );
-        for( i = 0; NULL != requiredExtensions[i]; i++ )
+        for (i = 0;
+             i < sizeof(requiredExtensions11) / sizeof(requiredExtensions11[0]);
```
kpet, line 1973: Please use `ARRAY_SIZE`

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1946,64 +1946,90 @@ int test_min_max_device_version(cl_device_id deviceID, cl_context context, cl_co
...
-            // if double precision is not supported, then don't check extension "cl_khr_fp64"
-            numRequiredExtension20 -= (value == 0);
-
-            for( i = 0; i<numRequiredExtension20; i++ )
+        if (version_as_integer >= 20 && version_as_integer < 30)
+        {
+            log_info("Checking for required extensions for OpenCL 2.0, 2.1 and "
+                     "2.2 devices...\n");
+            for (i = 0; i < sizeof(requiredExtensions2x)
+                     / sizeof(requiredExtensions2x[0]);
```
kpet, line 2019: `ARRAY_SIZE`

test_conformance/api/test_api_min_max.cpp
FranklandJack: Thanks for your feedback, I have updated to use `ARRAY_SIZE`.

test_conformance/api/test_api_min_max.cpp
FranklandJack: Thanks for your feedback, I have updated to use `ARRAY_SIZE`.

test_conformance/api/test_api_min_max.cpp
FranklandJack: Thanks for your feedback. I have update te parsing code, I wasn't sure about the checking of the `OPENCL_1_0_DEVICE` and `OPENCL_1_1_DEVICE`, I assume we still want to keep these checks so I've left them in.

kpet: I'd get rid of these, I think they were used to adapt the CTS behaviour to 1.1 and 1.0 implementations. We're now using queries for this.

FranklandJack: Thanks, okay that makes sense, I have removed them and reduced the scope of the loop counters and the `error` variable.

PR #781, opened 2020-05-18
Remove "C" linkages

layog:
@kpet This PR is to remove all the "C" linkages as discussed in #776


test_conformance/api/test_kernel_arg_changes.cpp
```diff
@@ -15,7 +15,7 @@
 //
 #include "testBase.h"
 
-extern "C" { extern cl_uint gRandomSeed;}
+extern cl_uint gRandomSeed;
```
kpet, line 18: `gRandomSeed` is declared in `testHarness.h`. All other declarations should be removed.

test_conformance/spir/datagen.cpp
```diff
@@ -93,38 +93,6 @@ void DataGenerator::setArgGenerator(const KernelArgInfo& argInfo,
     m_argGenerators[argInfo.getTypeName()] = pGen;
 }
 
-float get_random_float(float low, float high, MTdata d)
```
kpet, line 96: Nice :).

test_conformance/api/test_kernel_arg_changes.cpp
layog: Done. I have removed the declarations from other places as well (where `extern "C"` wasn't used). Let me know if this violates some comitting standards.

kpet: Nope, makes perfect sense. Thanks for doing this!

PR #782, opened 2020-05-19
Fix warnings in basic suite

jrprice:
The "returning reference to temporary" one was introduced recently by #776. The other one showed up on my machine when enabling `-Werror`.

### DIFF
--- a/test_conformance/basic/test_progvar.cpp
+++ b/test_conformance/basic/test_progvar.cpp
@@ -37,6 +37,7 @@
 // TODO: pointer-to-half (and its vectors)
 // TODO: union of...
 
+#include <algorithm>
 #include <cstdio>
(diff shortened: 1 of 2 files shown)


jrprice:
> We really ought to enable Werror soon in CI builds.

SGTM - I'll open a PR for this. Do we want to limit it to CI, or make it part of all CMake builds (so developers see the issues on their own machines first)? The downside of the latter is that different build environments will catch different things, but it'd hopefully only be short-term pain for a few people until we have disabled/fixed all the problematic warnings.

kpet:
I'm leaning toward the latter so that we get wider compiler coverage. It's fairly easy to remove Werror if it gets in the way and it will encourage contributions.

PR #786, opened 2020-05-20
Enable -Werror for GCC/Clang builds

jrprice:
Fixes many of the errors this produces, and disables a handful that
didn't have solutions that were obvious (to me).

### DIFF
--- a/test_conformance/half/cl_utils.h
+++ b/test_conformance/half/cl_utils.h
@@ -93,11 +93,6 @@ void ReleaseCL( void );
 int RunKernel( cl_device_id device, cl_kernel kernel, void *inBuf, void *outBuf, uint32_t blockCount , int extraArg);
 cl_program MakeProgram( cl_device_id device, const char *source[], int count );
 
-#define STRING( _x )    STRINGIFY( _x )
-#ifndef STRINGIFY
-#define STRINGIFY(x)    #x
-#endif
-
 static inline float as_float(cl_uint u) { union { cl_uint u; float f; }v; v.u = u; return v.f; }
 static inline double as_double(cl_ulong u) { union { cl_ulong u; double d; }v; v.u = u; return v.d; }
 
--- a/test_conformance/half/cl_utils.h
+++ b/test_conformance/half/cl_utils.h
@@ -94,7 +94,9 @@ int RunKernel( cl_device_id device, cl_kernel kernel, void *inBuf, void *outBuf,
 cl_program MakeProgram( cl_device_id device, const char *source[], int count );
 
 #define STRING( _x )    STRINGIFY( _x )
+#ifndef STRINGIFY
 #define STRINGIFY(x)    #x
+#endif
 
 static inline float as_float(cl_uint u) { union { cl_uint u; float f; }v; v.u = u; return v.f; }
 static inline double as_double(cl_ulong u) { union { cl_ulong u; double d; }v; v.u = u; return v.d; }
--- a/test_conformance/images/kernel_read_write/test_iterations.cpp
+++ b/test_conformance/images/kernel_read_write/test_iterations.cpp
@@ -85,7 +85,7 @@ static const char *lodOffsetSource =
 static const char *offsetSource =
 "   int offset = tidY*get_image_width(input) + tidX;\n";
 
-#define ABS_ERROR( result, expected ) ( fabsf( (float)expected - (float)result ) )
+#define ABS_ERROR(result, expected) (std::abs(expected - result))
 
 extern void read_image_pixel_float( void *imageData, image_descriptor *imageInfo,
                             int x, int y, int z, float *outData );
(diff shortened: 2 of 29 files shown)


test_conformance/half/cl_utils.h
```diff
@@ -94,7 +94,9 @@ int RunKernel( cl_device_id device, cl_kernel kernel, void *inBuf, void *outBuf,
 cl_program MakeProgram( cl_device_id device, const char *source[], int count );
 
 #define STRING( _x )    STRINGIFY( _x )
+#ifndef STRINGIFY
```
kpet, line 97: This rings a bell. IIRC, the other definition can just be deleted.

test_conformance/images/kernel_read_write/test_iterations.cpp
```diff
@@ -1008,9 +1008,12 @@ int validate_image_2D_sRGB_results(void *imageValues, void *resultValues, double
...
                                                                         imageSampler, expected, 0, &containsDenormals );
-                        float err1 = fabsf( sRGBmap( resultPtr[0] ) - sRGBmap( expected[0] ) );
-                        float err2 = fabsf( sRGBmap( resultPtr[1] ) - sRGBmap( expected[1] ) );
-                        float err3 = fabsf( sRGBmap( resultPtr[2] ) - sRGBmap( expected[2] ) );
+                        float err1 =
+                            fabs(sRGBmap(resultPtr[0]) - sRGBmap(expected[0]));
+                        float err2 =
+                            fabs(sRGBmap(resultPtr[1]) - sRGBmap(expected[1]));
+                        float err3 =
+                            fabs(sRGBmap(resultPtr[2]) - sRGBmap(expected[2]));
```
kpet, line 1016: Hmm, given how many times you've had to make the same change, I'd guess there's a common function to be extracted there :).

test_conformance/math_brute_force/unary_two_results.cpp
```diff
@@ -253,8 +252,7 @@ int TestFunc_Float2_Float(const Func *f, MTdata d)
             if ( gTestFastRelaxed && strcmp(f->name,"sincos") == 0 )
             {
               float pj = *(float *)&p[j];
-              if(fabs(pj) > M_PI)
-                p[j] = NAN;
+              if (fabs(pj) > M_PI) p[j] = 0;
```
kpet, line 255: I think I'd rather we don't try to fix these in this PR. See https://github.com/KhronosGroup/OpenCL-CTS/pull/699.

jrprice:
Addressed feedback and fixed additional issues found in CI. I've also tested locally on Linux (GCC + Clang), macOS, and an Android build.

test_conformance/images/kernel_read_write/test_iterations.cpp
```diff
@@ -85,7 +85,7 @@ static const char *lodOffsetSource =
 static const char *offsetSource =
 "   int offset = tidY*get_image_width(input) + tidX;\n";
 
-#define ABS_ERROR( result, expected ) ( fabsf( (float)expected - (float)result ) )
+#define ABS_ERROR(result, expected) (fabs(expected - result))
```
kpet, line 88: I was thinking of something higher level but that goes beyond the scope of this PR. Fine with the macro for now.

test_conformance/images/kernel_read_write/test_read_1D.cpp
```diff
@@ -66,7 +66,7 @@ const char *float1DKernelSource =
 
 static const char *samplerKernelArg = " sampler_t imageSampler,";
 
-#define ABS_ERROR( result, expected ) ( fabsf( (float)expected - (float)result ) )
+#define ABS_ERROR(result, expected) (fabs((float)expected - (float)result))
```
kpet, line 69: Hmm, we have several definitions of it. Is it worth pushing it to the common image code or the harness?

test_conformance/math_brute_force/main.cpp
```diff
@@ -1201,43 +1144,6 @@ test_status InitCL( cl_device_id device )
         return TEST_FAIL;
 #endif
     }
-#if defined( __APPLE__ )
-    else if(is_extension_available(gDevice, "cl_APPLE_fp64_basic_ops" ))
```
kpet, line 1205: Do recent Apple devices support cl_khr_fp64? If they don't we may want to raise an issue to modify the double path to support this as well but I agree the current implementation needs to go.

jrprice: The ones in my system do. But I don't think we should be testing vendor extensions in CTS anyway?

test_conformance/images/kernel_read_write/test_read_1D.cpp
jrprice: Done.

test_conformance/math_brute_force/main.cpp
kpet: Ok, I think that settles it for me. I suggest we don't create an issue for now. It's likely an old extension that pre-dates cl_khr_fp64.

On vendor extensions, we did discuss it at a previous F2F (I think it was the last one) and concluded that we'd make decisions on a case-by-case basis for a number of reasons:
- This came up as a requirement to allowing negative testing (a vendor might have an extension to relax an error case, they would need to add support to the CTS).
- There is value in having testing for vendor extensions when it's not intrusive as it makes it possible for other vendors to evaluate the feature and users to check conformance of their platform and could encourage promotion to ext/khr.
- There are cases where the CTS relies on implementation-defined behaviour. This is the reason we need support for cl_arm_printf for example.

jrprice: Makes sense, thanks for the info.

Issue #773, opened 2020-05-12
OpenCL 3.0 issues to create


kpet:
This issue covers creating issues for all changes in OpenCL 3.0 that require CTS work that hasn't been done yet.

# Feature Optionality

### Shared Virtual Memory

See existing [SVM test suite](https://github.com/KhronosGroup/OpenCL-CTS/tree/master/test_conformance/SVM).

* [x] Need to add checks for lack of SVM support.
* [x] Evaluate how much can be tested without the OpenCL 2.0 memory consistency model / atomic support. (#887)

### Memory Consistency Model

See existing [C11 Atomics test suite](https://github.com/KhronosGroup/OpenCL-CTS/tree/master/test_conformance/c11_atomics).

This is likely to interact with many other OpenCL 2.x tests.

* [x] Need to update to cover at least OpenCL 1.2 (minimum) and OpenCL 2.0 level of support.

### Device Enqueue

See existing [Device Execution test suite](https://github.com/KhronosGroup/OpenCL-CTS/tree/master/test_conformance/device_execution).

This test suite may also be using program scope global variables.

* [x] Need to update to check for lack of device enqueue support.
* [x] Confirm: nothing needed for [`cl_khr_device_enqueue_local_arg_types`](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_Ext.html#cl_khr_device_enqueue_local_arg_types)? (#888)
    * It doesn't appear that this extension is supported by Clang, see: https://godbolt.org/z/rha8fy
    * AMD is supporting the extension and can contribute tests.  Work is still needed to upstream to Clang.

### Program Scope Global Variables

See existing [basic progvar test](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/basic/test_progvar.cpp).

Recording medium due to likely interactions with other tests.

* [x] Need to add checks for lack of Program Scope Global Variable support. (#878)

### Non-Uniform Work Groups

See existing [Non-Uniform Work Groups test suite](https://github.com/KhronosGroup/OpenCL-CTS/tree/master/test_conformance/non_uniform_work_group).

* [x] Need to add checks for lack of Non-Uniform Work Groups support (https://github.com/KhronosGroup/OpenCL-CTS/issues/791).
* [x] Need to test for `get_enqueued_local_size`, etc. even when Non-Uniform Work Groups are not supported? (#877)

### Read-Write Images

[...]

PR #788, opened 2020-05-21
Skip test_sub_group_dispatch if not Supported

FranklandJack:
* `test_sub_group_dispatch` is testing the `cl_khr_subgroups` extension.
Skip this test in the case the device does not support
`cl_khr_subgroups`.

### DIFF
--- a/test_conformance/api/test_sub_group_dispatch.cpp
+++ b/test_conformance/api/test_sub_group_dispatch.cpp
@@ -75,6 +75,14 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
     size_t ret_ndrange2d_flattened;
     size_t ret_ndrange3d_flattened;
 
+    // This test should be skipped in the case that the cl_khr_subgroups
+    // extension is not supported.
+    if (!is_extension_available(deviceID, "cl_khr_subgroups"))
+    {
+        log_info("Skipping since cl_khr_subgroups is not supported on device\n");
+        return TEST_SKIP;
+    }
+
     error = create_single_kernel_helper_with_build_options(context, &program, &kernel, 1, subgroup_dispatch_kernel, "subgroup_dispatch_kernel", "-cl-std=CL2.0");
     if (error != 0)
         return error;


test_conformance/api/test_sub_group_dispatch.cpp
```diff
@@ -75,6 +75,14 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
     size_t ret_ndrange2d_flattened;
     size_t ret_ndrange3d_flattened;
 
+    // This test should be skipped in the case that the cl_khr_subgroups
+    // extension is not supported.
+    if (!is_extension_available(deviceID, "cl_khr_subgroups"))
+    {
+        log_info("Skipping since cl_khr_subgroups is not supported on device\n");
```
alycm, line 82: The comment and log message could just refer to subgroups being missing rather than `cl_khr_subgroups` given this is a core feature in 2.1+.

But I think actually using the extension for the check is fine, as that is backwards compatible to older versions.

gwawiork: I agree. You cannot skip this test in case extension not found on devices 2.1+
In my opinion it should be similar to [this](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/subgroups/main.cpp#L66). Shortly:
devices 2.1 and 2.2 - always test (even no extension)
device 3.X - check if subgroups feature is on CL_DEVICE_MAX_NUM_SUB_GROUPS > 0

alycm:
Given that subgroups are listed on the set of things to break down into issues (https://github.com/KhronosGroup/OpenCL-CTS/issues/773) it would helpful to file an issue for the "Is anything needed beyond checks for cl_khr_subgroups?" item in the OpenCL 3.0 project.  And then tick the box.

That will help with tracking progress being made on the 3.0 CTS.

test_conformance/api/test_sub_group_dispatch.cpp
alycm: Actually @gwawiork is correct, the specification does not require reporting `cl_khr_subgroups` even when subgroups are supported.  Thinking about it, an implementation **must not** report it if subgroups are implemented without IFP.

gwawiork: If you are interested more in subgroups topic I am preparing changes #682  that maybe add some more clarification in subgroups
@alycm When this will be ready and merged then we could I think thick the box in #773

kpet: We really need more framework support to make these checks easy and reusable. I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/792 to cover this.

test_conformance/api/test_sub_group_dispatch.cpp
```diff
@@ -75,6 +75,14 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
...
+        return TEST_SKIP;
```
kpet, line 83: This should be TEST_SKIPPED_ITSELF. I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/793 to fix this.

test_conformance/api/test_sub_group_dispatch.cpp
kpet: Agree this won't work for the core feature BTW.

kpet:
This is no longer needed if/when #816 goes in.

FranklandJack:
> This is no longer needed if/when #816 goes in.

Thanks for your feedback. In that case I will wait till #816 is merged then close this PR.

FranklandJack:
I can confirm that in local testing  #816 fixes the issue that this PR attempted to solve.

test_conformance/api/test_sub_group_dispatch.cpp
```diff
@@ -75,6 +75,14 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
     size_t ret_ndrange2d_flattened;
     size_t ret_ndrange3d_flattened;
 
+    // This test should be skipped in the case that the cl_khr_subgroups
+    // extension is not supported.
+    if (!is_extension_available(deviceID, "cl_khr_subgroups"))
```
Jeremy-Kemp, line 80: Could you add the logic for checking device version from #864 please?

I created that PR without realising that this one already existed.

FranklandJack: Thanks for your feedback @Jeremy-Kemp. I think this PR has been superceded by #816 which handles optional features in a more generic and fashion. I was waiting for #816 to be merged before closing this, but I'm also happy to close this in favour of #816 since you have already implemented the check for the device version and the core feature.

kpet: @FranklandJack It will be eventually but I've been struggling to find the time to spend on #816 and this is starting to get in the way of progress so I suggested we merge your PRs / Jeremy's and I'll rework this when I work on #816 again.

FranklandJack: @kpet Okay that sounds like a good plan to me. I've updated #789 this morning and am happy to close this PR in favor of @Jeremy-Kemp 's #864 since the work is already done there.

alycm:
Based on the discussions above I am closing this PR at https://github.com/KhronosGroup/OpenCL-CTS/pull/864 is now merged and achieves the same thing but more completely.

PR #789, opened 2020-05-21
Skip test_enqueued_local_size if non-uniform work-groups not supported

FranklandJack:
* `test_enqueued_local_size` will be skipped for OpenCL < 2.0. However,
non-uniform work-groups became optional in 3.0 and so check that they
are supported before running `test_enqueued_local_size` and otherwise
skip.

### DIFF
--- a/test_conformance/basic/test_enqueued_local_size.cpp
+++ b/test_conformance/basic/test_enqueued_local_size.cpp
@@ -74,14 +74,41 @@ test_enqueued_local_size(cl_device_id device, cl_context context, cl_command_que
     size_t localsize[2];
     int err;
 
+    // For an OpenCL-3.0 device that does not support non-uniform work-groups
+    // we cannot enqueue local sizes which do not divide the global dimensions
+    // but we can still run the test checking that get_enqueued_local_size ==
+    // get_local_size.
+    bool use_uniform_work_groups{ false };
+    if (get_device_cl_version(device) >= Version(3, 0))
+    {
+        cl_bool areNonUniformWorkGroupsSupported = false;
+        err = clGetDeviceInfo(device, CL_DEVICE_NON_UNIFORM_WORK_GROUP_SUPPORT,
+                              sizeof(areNonUniformWorkGroupsSupported),
+                              &areNonUniformWorkGroupsSupported, nullptr);
+        test_error_ret(err, "clGetDeviceInfo failed.", TEST_FAIL);
+
+        if (CL_FALSE == areNonUniformWorkGroupsSupported)
+        {
+            log_info("Non-uniform work group sizes are not supported, "
+                     "enqueuing with uniform workgroups\n");
+            use_uniform_work_groups = true;
+        }
+    }
+
     output_ptr   = (int*)malloc(2 * sizeof(int));
 
     streams = clCreateBuffer(context, (cl_mem_flags)(CL_MEM_READ_WRITE), 2*sizeof(int), NULL, &err);
     test_error( err, "clCreateBuffer failed.");
 
-    err = create_single_kernel_helper_with_build_options(context, &program[0], &kernel[0], 1, &enqueued_local_size_1d_code, "test_enqueued_local_size_1d", "-cl-std=CL2.0");
+    std::string cl_std = "-cl-std=CL";
+    cl_std += (get_device_cl_version(device) == Version(3, 0)) ? "3.0" : "2.0";
+    err = create_single_kernel_helper_with_build_options(
+        context, &program[0], &kernel[0], 1, &enqueued_local_size_1d_code,
+        "test_enqueued_local_size_1d", cl_std.c_str());
     test_error( err, "create_single_kernel_helper failed");
-    err = create_single_kernel_helper_with_build_options(context, &program[1], &kernel[1], 1, &enqueued_local_size_2d_code, "test_enqueued_local_size_2d", "-cl-std=CL2.0");
+    err = create_single_kernel_helper_with_build_options(
+        context, &program[1], &kernel[1], 1, &enqueued_local_size_2d_code,
+        "test_enqueued_local_size_2d", cl_std.c_str());
     test_error( err, "create_single_kernel_helper failed");
 
     err  = clSetKernelArg(kernel[0], 0, sizeof streams, &streams);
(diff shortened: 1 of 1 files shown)


FranklandJack:
I have if defd out this code since it will cause compilation errors if `CL_VERSION_3_0` is not defined. I'm not sure whether the CTS is intended to work this way, or if it should always be compiled with the most recent OpenCL headers anyway since it used with the ICD loader?

alycm:
Given that non-uniform workgroups are listed on the set of things to break down into issues (https://github.com/KhronosGroup/OpenCL-CTS/issues/773) it would helpful to file an issue for the "Need to add checks for lack of Non-Uniform Work Groups support." item in the OpenCL 3.0 project.  And then tick the box.

That will help with tracking progress being made on the 3.0 CTS.

FranklandJack:
> Given that non-uniform workgroups are listed on the set of things to break down into issues (#773) it would helpful to file an issue for the "Need to add checks for lack of Non-Uniform Work Groups support." item in the OpenCL 3.0 project. And then tick the box.
> 
> That will help with tracking progress being made on the 3.0 CTS.

@alycm I have created an issue https://github.com/KhronosGroup/OpenCL-CTS/issues/791 and commened on #773, but I don't think I have edit permissions to tick the box.

gwawiork:
LGTM  but
I am not sure if these defines with CL_VERSION_3_0 is needed rather we check device OpenCL version reported like [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/generic_address_space/main.cpp#L82)
I would be nice added the same code [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/non_uniform_work_group/main.cpp#L41)

FranklandJack:
> LGTM but
> I am not sure if these defines with CL_VERSION_3_0 is needed rather we check device OpenCL version reported like [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/generic_address_space/main.cpp#L82)
> It would be nice to have the same code [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/non_uniform_work_group/main.cpp#L41)

Thank you for your feedback, I wasn't sure myself, but is it possible that the CTS could be compiled without `CL_VERSION_3_0` say, `CL_VERSION_1_2`, in which case `CL_DEVICE_NON_UNIFORM_WORK_GROUP_SUPPORT` will not be defined since it is conditionally defined [here](https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl.h#L404)

kpet:
The CTS must always be compiled with the latest headers and all features enabled so this is a non-issue.

A revised version of #816 is likely to supersede this PR.

FranklandJack:
> The CTS must always be compiled with the latest headers and all features enabled so this is a non-issue.
> 
> A revised version of #816 is likely to supersede this PR.

Thanks for your feedback. In that case I will wait until #816 is merged and then close this PR.

FranklandJack:
I can confirm that in local testing #816 fixes the issue that this PR attempted to solve.

test_conformance/basic/test_enqueued_local_size.cpp
```diff
@@ -74,6 +74,27 @@ test_enqueued_local_size(cl_device_id device, cl_context context, cl_command_que
...
+        cl_bool areNonUniformWorkGroupsSupported = false;
+        err = clGetDeviceInfo(device, CL_DEVICE_NON_UNIFORM_WORK_GROUP_SUPPORT,
+                              sizeof(areNonUniformWorkGroupsSupported),
+                              &areNonUniformWorkGroupsSupported, nullptr);
+        if (CL_SUCCESS != err)
+        {
+            test_error(err, "clGetDeviceInfo failed.");
+        }
+
+        if (false == areNonUniformWorkGroupsSupported)
```
alycm, line 90:
```suggestion
        if (CL_FALSE == areNonUniformWorkGroupsSupported)
```

FranklandJack: Thanks @alycm, have updated.

test_conformance/basic/test_enqueued_local_size.cpp
```diff
@@ -74,6 +74,27 @@ test_enqueued_local_size(cl_device_id device, cl_context context, cl_command_que
...
+    // non-uniform work-groups are not supported.
+    if (get_device_cl_version(device) >= Version(3, 0))
+    {
+        cl_bool areNonUniformWorkGroupsSupported = false;
+        err = clGetDeviceInfo(device, CL_DEVICE_NON_UNIFORM_WORK_GROUP_SUPPORT,
+                              sizeof(areNonUniformWorkGroupsSupported),
+                              &areNonUniformWorkGroupsSupported, nullptr);
+        if (CL_SUCCESS != err)
+        {
+            test_error(err, "clGetDeviceInfo failed.");
```
Jeremy-Kemp, line 87: Should return `TEST_FAIL` here.

test_conformance/basic/test_enqueued_local_size.cpp
```diff
@@ -74,6 +74,27 @@ test_enqueued_local_size(cl_device_id device, cl_context context, cl_command_que
...
+        }
+
+        if (CL_FALSE == areNonUniformWorkGroupsSupported)
+        {
+            log_info("Skipping since non-uniform work group sizes are not "
+                     "supported\n");
+            return TEST_SKIPPED_ITSELF;
```
bashbaug, line 94: This is OK but it's a pretty big hammer since it means that none of this test will be executed without support for non-uniform work groups.  Would it instead be preferable to execute the test with uniform work groups on an OpenCL 3.0 device without non-uniform work group support so there is at least some coverage for `get_enqueued_local_size`?

test_conformance/basic/test_enqueued_local_size.cpp
FranklandJack: Good catch, I've replace this whole block with `test_error_ret(err, "clGetDeviceInfo failed.", TEST_FAILED)". Thanks for the feedback!

FranklandJack:
>     * This test currently compiles kernels with `-cl-std=CL2.0`, but this may not be supported by all OpenCL 3.0 devices.  For now, I think we can pass `-cl-std=CL3.0` for OpenCL 3.0 devices, but eventually we may need more complicated logic here.

@bashbaug Thanks for your feedback! This is a good point, I created an issue to track this a while back #823, although that issue kinda got conflated with the fact that some tests which were optional for 3.0 aren't getting skipped.

I've updated the test to dynamically select `-cl-std=CL3.0` for a OpenCL-3.0 device and otherwise fall back on `-cl-std=CL2.0`.

test_conformance/basic/test_enqueued_local_size.cpp
FranklandJack: Thanks for the feedback @bashbaug. This is a really good point, I've updated the test slightly so that instead of skipping it clamps the global work dimensions such that the local work-group size divides the global size in the case that non-uniform work-groups are not supported. Let me know if this is the type of thing you had in mind?

FranklandJack:
> Latest changes look good, thanks!
> 
> When we're enqueuing uniform work groups it's probably slightly safer to align up than to align down, just to guarantee we launch at least one work group, but in practice this shouldn't matter.

Oops, didn't think about the case we could end up with a global size less than the local size, good catch! I've updated the rounding logic to round up instead, and updated the commit message since it had gone a bit stale.

bashbaug:
Merging - this has two approvals and I'm guessing @alycm is OK with these changes.

PR #797, opened 2020-05-26
Improve async build callback testing

jrprice:
Check that the program build status can be queried from inside the
program completion callback, and also add a test to ensure that the
completion callback is called when the build fails.

This was clarified in the specification in [a recent change](https://github.com/KhronosGroup/OpenCL-Docs/pull/285).

### DIFF
--- a/test_conformance/compiler/test_async_build.cpp
+++ b/test_conformance/compiler/test_async_build.cpp
@@ -19,75 +19,139 @@
 #include <unistd.h>
 #endif
 
+#include <atomic>
+#include <string>
 
 const char *sample_async_kernel[] = {
-"__kernel void sample_test(__global float *src, __global int *dst)\n"
-"{\n"
-"    int  tid = get_global_id(0);\n"
-"\n"
-"    dst[tid] = (int)src[tid];\n"
-"\n"
-"}\n" };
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_async_build.cpp
```diff
@@ -19,75 +19,139 @@
...
+
+std::atomic<int> callbackResult;
 
 void CL_CALLBACK test_notify_build_complete( cl_program program, void *userData )
 {
-    if( userData == NULL || strcmp( (char *)userData, "userData" ) != 0 )
+    TestData *data = reinterpret_cast<TestData *>(userData);
+
+    // Check string in user data is correct
+    if (userData == NULL || data->str != "userData")
```
kpet, line 60: Could use `data` as well here.

test_conformance/compiler/test_async_build.cpp
```diff
@@ -19,75 +19,139 @@
...
+    "\n"
+    "    dst[tid] = badcodehere;\n"
+    "\n"
+    "}\n"
+};
+
+// Data passed to a program completion callback
+struct TestData
+{
+    std::string str;
```
kpet, line 48: Not sure this `str` is really useful.

Issue #800, opened 2020-05-28
Linux build errors


gwawiork:
I see Linux build errors after changes #786 

- test_gl

`test_fence_sync.cpp:288:24: error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]
09:55:00,411 WARN  - return (void *)error;`

`test_image_methods.cpp:35:1: error: ‘typedef’ was ignored in this declaration [-Werror]
08:41:27,333 WARN  - };`

- test_printf

`test_printf.cpp:147:34: error: ignoring return value of ‘FILE* freopen(const char*, const char*, FILE*)’, declared with attribute warn_unused_result [-Werror=unused-result]
10:19:28,655 WARN  - freopen(gFileName,"w",stdout);`

gwawiork:
fixed

PR #801, opened 2020-05-28
Fix test_printf build errors

gwawiork:
Fix test_printf build errors Issue #800

### DIFF
--- a/test_conformance/printf/test_printf.cpp
+++ b/test_conformance/printf/test_printf.cpp
@@ -144,7 +144,10 @@ static int getTempFileName()
 static int acquireOutputStream()
 {
     int fd = streamDup(fileno(stdout));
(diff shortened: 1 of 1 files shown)


test_conformance/printf/test_printf.cpp
```diff
@@ -144,7 +144,10 @@ static int getTempFileName()
 static int acquireOutputStream()
 {
     int fd = streamDup(fileno(stdout));
-    freopen(gFileName,"w",stdout);
+    if (!freopen(gFileName, "w", stdout))
+    {
+        log_error("Error while redirection stdout to file");
```
kpet, line 149: Probably worth throwing an exception or returning a negative value and updating all using code to check the error.

gwawiork:
Fixed. Added error code

PR #802, opened 2020-05-28
Fix test_gl build errors

gwawiork:
Fix test_gl build errors Issue #800


test_conformance/gl/test_fence_sync.cpp
```diff
@@ -285,7 +285,7 @@ class RunThread : public genericThread
     virtual void * IRun( void )
     {
         cl_int error = run_cl_kernel( mKernel, mQueue, mStream0, mStream1, mRowIdx, mFenceEvent, mNumThreads );
-        return (void *)error;
+        return reinterpret_cast<void *>(error);
```
kpet, line 288: I doubt that fixes the issue, you probably need a first cast to `uintptr_t`. What was the error message?

kpet:
Now folded into #804, closing.

PR #803, opened 2020-05-28
Turn on building test_gl in travis

gwawiork:
Add building test_gl binaries on Ubuntu

PR #804, opened 2020-05-28
Turn on building test_gl in travis

gwawiork:
Add building test_gl binaries on Ubuntu - reopen #803

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,18 +15,8 @@ jobs:
             - clang-format-9
   - os: linux
     env: JOB_ARCHITECTURE=arm
-    addons:
-        apt:
-          packages:
-            - libqt4-dev 
-            - libqt4-opengl-dev
   - os: linux
     env: JOB_ARCHITECTURE=aarch64
-    addons:
-        apt:
-          packages:
-            - libqt4-dev 
-            - libqt4-opengl-dev
 
 script:
   - ./travis.sh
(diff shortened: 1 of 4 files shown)


gwawiork:
Could somebody help? :
- build job 1 (Ubuntu) it will work (fix in PR #802 )
- build job 2 (OSx).  Does somebody know how to add GL headers on OSx ?
- build job 4 and build job 5 (Ubuntu) - at first look I do not see reason why this do not work as for job 1. What are the differences in system configuration between them ?

kpet:
The Arm and AArch64 jobs are using cross-compilation so the native GL libs won't help. Either we need Arm or AArch64 GL libs or maybe we can disable GL for these jobs.

For Mac, I don't really know. @jrprice Any insights?

jrprice:
> For Mac, I don't really know. @jrprice Any insights?

OpenGL is deprecated on Apple, so I'm not sure it's useful to test this. You can add `-DGL_SILENCE_DEPRECATION` to the compiler options, but I get various other errors when I do this on my machine.

kpet:
Right, we should probably only enable GL in native Linux builds then. Some coverage is still infinitely better than none.

gwawiork:
Thank you for your suggestions. I've turned on GL only on native linux builds. Having merged #803 this build should pass

kpet:
@gwawiork The CI changes look good and the Linux job now fails as expected. Could you fold https://github.com/KhronosGroup/OpenCL-CTS/pull/802 into this PR so we can see that it fixes the issues?

gwawiork:
Added commits from #802

kpet:
@gwawiork Looks like it isn't building: https://travis-ci.org/github/KhronosGroup/OpenCL-CTS/builds/699338622

PR #807, opened 2020-06-01
Remove `gTestFastRelaxed` Dependencies in Brute Force

FranklandJack:
* The global variable `gTestFastRelaxed` has state which is used to
control the behaviour of the compiler flag `-cl-fast-relaxed-math` and
the precision testing of relaxed, fp32 and fp64 types. This is confusing
since the global variable is being set and read in different translation
units, making it very difficult to reason about the logic of the brute
force framework. It is particular difficult to follow since the global
variables is cached and then turned off in the case of fp32 and f64 in
order to use the same code path as relaxed testing, after it is then
turned back on.

* Remove uses of the global variable outside of `main.cpp` (the global
variable remains in use within `main.cpp` since it is a command line
option and used to turn of relaxed testing completely). Replace all uses
of the global variable with boolean `relaxedMode` which is passed as a
function paramter but replaces `gTestFastRelaxed` semantically.

### DIFF
--- a/test_conformance/math_brute_force/main.cpp
+++ b/test_conformance/math_brute_force/main.cpp
@@ -68,8 +68,9 @@ int             gForceFTZ = 0;
 int             gWimpyMode = 0;
 int             gHasDouble = 0;
 int             gTestFloat = 1;
-//This flag should be 'ON' by default and it can be changed through the command line arguments.
-volatile int             gTestFastRelaxed = 1;
+// This flag should be 'ON' by default and it can be changed through the command
+// line arguments.
+static int gTestFastRelaxed = 1;
 /*This flag corresponds to defining if the implementation has Derived Fast Relaxed functions.
   The spec does not specify ULP for derived function.  The derived functions are composed of base functions which are tested for ULP, thus when this flag is enabled,
   Derived functions will not be tested for ULP, as per table 7.1 of OpenCL 2.0 spec.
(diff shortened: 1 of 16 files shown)


test_conformance/math_brute_force/main.cpp
```diff
@@ -1500,7 +1495,7 @@ int MakeKernel( const char **c, cl_uint count, const char *name, cl_kernel *k, c
       strcat(options," -cl-denorms-are-zero");
     }
 
-    if( gTestFastRelaxed )
+    if (true)
```
kpet, line 1498: relaxedMode

test_conformance/math_brute_force/main.cpp
```diff
@@ -1543,7 +1540,7 @@ int MakeKernels( const char **c, cl_uint count, const char *name, cl_uint kernel
       strcat(options," -cl-fp32-correctly-rounded-divide-sqrt ");
     }
 
-    if( gTestFastRelaxed )
+    if (true)
```
kpet, line 1543: relaxedMode

FranklandJack:
> Thanks for doing this! Please fix the hard-coding of `-cl-fast-relaxed-math`.

No problem, thanks for your feed back. Good catch, I must have hard coded the condition to `true` for testing purposes and forgotten to change it back. I have updated the two conditional statements.

PR #808, opened 2020-06-01
Subgroups - refactor code subgroup functions registration for better …

gwawiork:
…maintainability

This work is related to subgroups next test code and comment in one of the previous review from @kpet https://github.com/KhronosGroup/OpenCL-CTS/pull/533#pullrequestreview-338602913

### DIFF
--- a/test_conformance/subgroups/test_workgroup.cpp
+++ b/test_conformance/subgroups/test_workgroup.cpp
@@ -623,8 +623,8 @@ template <typename Ty> struct BC
 
 #define G 2000
 #define L 200
-class run_for_type {
-public:
+struct run_for_type
+{
     run_for_type(cl_device_id device, cl_context context,
                  cl_command_queue queue, int num_elements,
                  bool useCoreSubgroups)
@@ -672,6 +672,7 @@ public:
         return error;
     }
 
+private:
     cl_device_id device_;
     cl_context context_;
     cl_command_queue queue_;
--- a/test_conformance/subgroups/test_workgroup.cpp
+++ b/test_conformance/subgroups/test_workgroup.cpp
@@ -621,6 +621,63 @@ template <typename Ty> struct BC
     }
 };
 
+#define G 2000
+#define L 200
+class run_for_type {
+public:
+    run_for_type(cl_device_id device, cl_context context,
+                 cl_command_queue queue, int num_elements,
+                 bool useCoreSubgroups)
+    {
+        device_ = device;
+        context_ = context;
+        queue_ = queue;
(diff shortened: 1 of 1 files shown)


test_conformance/subgroups/test_workgroup.cpp
```diff
@@ -621,234 +621,85 @@ template <typename Ty> struct BC
     }
 };
 
+#define G 2000
+#define L 200
+class run_for_type {
```
kpet, line 626: With `struct` you could avoid having to explicitly add `public`.

test_conformance/subgroups/test_workgroup.cpp
```diff
@@ -621,234 +621,85 @@ template <typename Ty> struct BC
...
+            device_, context_, queue_, num_elements_, "test_scexmin",
+            scexmin_source, 0, useCoreSubgroups_);
+        return error;
+    }
+
+    cl_device_id device_;
+    cl_context context_;
+    cl_command_queue queue_;
+    int num_elements_;
+    bool useCoreSubgroups_;
```
kpet, line 679: Don't you want the above to be private?

test_conformance/subgroups/test_workgroup.cpp
```diff
@@ -621,234 +621,85 @@ template <typename Ty> struct BC
...
                                          "test_all", all_source, 0,
                                          useCoreSubgroups);
+    run_for_type rft(device, context, queue, num_elements, useCoreSubgroups);
+    error |= rft.run<cl_uint>();
+    error |= rft.run<cl_int>();
+    error |= rft.run<cl_ulong>();
+    error |= rft.run<cl_long>();
+    error |= rft.run<float>();
+    error |= rft.run<double>();
+    // error |= rft.run<cl_half>();
```
kpet, line 701: We should create an issue to look into this. Why was half disabled?

gwawiork: Regarding half type - I do not know why this is commented (I just leave this as is for today) However I have implemented/fixed this in code related with subgroups next.

test_conformance/subgroups/test_workgroup.cpp
gwawiork: fixed

test_conformance/subgroups/test_workgroup.cpp
kpet: I think it's worth creating an issue to make sure we don't forget about this.

gwawiork: Done #812

PR #813, opened 2020-06-03
Tests for cl_khr_device_uuid

einhov:
Basic conformance testing for the cl_khr_device_uuid extension.  Checks
that the ID queries return the right sizes and that the returned IDs are
stable across two calls to clGetDeviceInfo. If device LUID validity is
true, also checks that one and only one bit is set in the device node
mask.

### DIFF
--- a/test_conformance/computeinfo/CMakeLists.txt
+++ b/test_conformance/computeinfo/CMakeLists.txt
@@ -2,6 +2,7 @@ set(MODULE_NAME computeinfo)
 
 set(${MODULE_NAME}_SOURCES
         main.cpp
+        device_uuid.cpp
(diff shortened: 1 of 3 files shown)


Jeremy-Kemp:
Thanks @einhov . I can't tell which commit was used for `OpenCL-Headers` in the Windows build, but the new defines are there now so would think that the build should pass if it were triggered again.

kpet:
I did retrigger Travis after the headers change went in but couldn't retrigger Appveyor, hence the failure.

kpet:
@einhov  Could you rebase this please so we get a new Appveyor run?

gwawiork:
LGTM

Issue #812, opened 2020-06-03
Subgroups - enable half type testing 


gwawiork:
Now the half type is commented. Add support for half type in test_workgroup.cpp

rjodinchr:
Fixed by PR #1074 (commit `71bef856`), which overhauled the subgroup test suite and enabled `half` type testing for reduction, scan, and broadcast operations (further refined in PR #1278 and PR #1346).

PR #814, opened 2020-06-03
New extended subgroups - test development

gwawiork:
Contributes to #532
Initial version. Before adding tests for cl_khr_subgroups_extended_types extension, set of modifications need to be done in existing code. I've spited previous #533 to more commits without new test cases (please review commit by commit). The reset of code from #533 will be published after this one is merged.
It helps solving partially issue #812. It needs some other code (I will push it in next commits)


gwawiork:
New will be opened with refactored code.

PR #816, opened 2020-06-03
Framework to manage features

### LINKED ISSUE #792: Introduce a generic framework to skip tests based on supported features
There are many tests that should only be run when the features they exercise are supported. These feature checks are currently duplicated in a number of places throughout the code base. We should centralise them and provide a mechanism as part of the harness to make feature checks easy and reusable.

kpet:
This change introduces a basic framework that makes it easier
to predicate the execution of a given test on a pre-defined feature.

Features are objects and reusable features can be defined as singleton
global objects (hence the need to initialise the test lists within the
main function of individual suites to guarantee that features are
already initialised).
Each feature comprises a description and a predicate function defined
as a member lambda. The predicate function returns true when the device
passed as argument supports the feature described by the containing
feature object.

Features can be used within expressions, either to define other features
or as part of test registration. Here are a couple of examples:

    const feature& F_complex_feature = F_featA && F_featB

    ADD_TEST_FEATURE(mytest, F_extension("cl_my_extension") || F_featA)

AND expressions stop evaluation as soon as one of the required features
is missing. This makes it possible to describe expressions that rely
on queries that are not supported by all versions of OpenCL.

The harness now skips tests based on whether they support the features
declared at registration time. It prints a message that describes the
feature expression that was evaluated when skipping a test. Here's an
example:

    progvar_prog_scope_init skipped (requires feature
    '((Version >= 2.0 && version < 3.0) ||
     (Version >= 3.0 && Program scope variables core optional))'
    that the device doesn't support

The uniqueness of the reusable objects make it possible to introduce a
caching mechanism to avoid repeating complex sequences of devices queries
but this change does not introduce such a mechanism.

This change uses the new framework to manage optionality in the api
and basic suites.

Fixes #792


test_common/harness/testHarness.h
```diff
@@ -18,58 +18,36 @@
...
 
-typedef struct test_definition
+struct test_definition
 {
     basefn func;
     const char* name;
     Version min_version;
-} test_definition;
-
+    const feature &required_feature;
```
gwawiork, line 49: I have huge amount of warnings in MSVC here:
class "test_definition" defines no constructor to initialize the following:
            reference member "test_definition::required_feature"
  struct test_definition
         ^

test_common/harness/feature.cpp
```diff
@@ -0,0 +1,81 @@
...
+            ASSERT_SUCCESS(err, "clGetDeviceInfo");
+            return max_sub_groups != 0;
+        };
+    }
+} F_subgroups_core_optional;
+
+const feature& F_subgroups_extension = F_extension("cl_khr_subgroups");
+
+const feature& F_subgroups = (F_version_ge(2, 0) && F_subgroups_extension)
+    || F_version_eq(2, 1) || (F_version_gt(2, 2) && F_subgroups_core_optional);
```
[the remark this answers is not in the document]
kpet: Oh, nice catch. I'd used a complex expression for testing. This can be simplified a lot.

test_common/harness/testHarness.h
kpet: Yes, best to use a constructor here. I'll do that.

kpet:
Thanks for the feedback!

>Is this possible to get the feature name or extension name from the test level that was set in registration phase ?. It would be nice to get this for example for adding automatically pragma to the kernel.

I'll have a think about that but I think it should be a separate PR.

test_conformance/basic/main.cpp
```diff
@@ -25,137 +25,137 @@
...
+        ADD_TEST_FEATURE(progvar_prog_scope_uninit, F_program_scope_variables),
+        ADD_TEST_FEATURE(progvar_prog_scope_init, F_program_scope_variables),
+        ADD_TEST_FEATURE(progvar_func_scope, F_program_scope_variables),
+
+        ADD_TEST(global_work_offsets),
+        ADD_TEST(get_global_offset),
+
+        ADD_TEST_FEATURE(global_linear_id, F_version_ge(2, 0)),
+        ADD_TEST_FEATURE(local_linear_id, F_version_ge(2, 0)),
+        ADD_TEST_FEATURE(enqueued_local_size, F_version_ge(2, 0)),
```
kpet, line 145: Need to handle optionality in 3.0.

test_conformance/basic/main.cpp
```diff
@@ -25,137 +25,137 @@
...
 #if defined( __APPLE__ )
-    ADD_TEST( queue_priority ),
+        ADD_TEST(queue_priority),
 #endif
 
-    ADD_TEST_VERSION( get_linear_ids, Version(2, 0) ),
-    ADD_TEST_VERSION( rw_image_access_qualifier, Version(2, 0) ),
-};
+        ADD_TEST_FEATURE(get_linear_ids, F_version_ge(2, 0)),
+        ADD_TEST_FEATURE(rw_image_access_qualifier, F_version_ge(2, 0)),
```
FranklandJack, line 155: Since read-write images are optional in OpenCL-3.0, should we have a feature to check that clGetDeviceInfo returns a non-zero value for CL_​DEVICE_​MAX_​READ_​WRITE_​IMAGE_​ARGS?

FranklandJack:
I can confirm that in local testing this fixes the issue #788 however #789 would need a feature to test for non uniform work groups to fix that issue.

test_conformance/basic/main.cpp
kpet: Yup, I was just focusing on migrating the initial checks and open PRs with this first change but this is simple enough to be done here.

FranklandJack: Thank you. I can confirm that with the latest changes #789 is now fixed. `test_rw_image_access_qualifier` is also successfully skipped.

kpet:
@FranklandJack @gwawiork Thanks for the testing! I think I've now addressed all your comments. Unfortunately it seems this change triggers a bug in the version of MSVC we're using in the CI. I would appreciate if someone with MSVC on their local computer could give this a spin and let me know whether they are facing issues.

FranklandJack:
> @FranklandJack @gwawiork Thanks for the testing! I think I've now addressed all your comments. Unfortunately it seems this change triggers a bug in the version of MSVC we're using in the CI. I would appreciate if someone with MSVC on their local computer could give this a spin and let me know whether they are facing issues.

Thank you. Unfortunetly I'm on linux and haven't got a windows VM set up, although I could look into setting one up if @gwawiork is also not using MSVC.

gwawiork:
I will take a look on this issue. I am using MSVC but did not noticed any issues with Intel compiler. Maybe native msbuild has some issues.

gwawiork:
@kpet I've found rootcause of this issue.  MSVC has problem because of doubled test_progvar.cpp entry in CMakeLists.txt. Remove one please:
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/basic/CMakeLists.txt#L64
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/basic/CMakeLists.txt#L52

kpet:
@gwawiork You rock! Thanks for taking the time to look into it. I've updated the PR.

gwawiork:
Hmmm. Something else probably some headers dependencies ... I will check this.

There is more duplication. Maybe worth to fix it:
test_progvar.cpp
test_wg_barrier.cpp
test_global_linear_id.cpp
test_local_linear_id.cpp
test_enqueued_local_size.cpp

gwawiork:
Looks like collision in namespace with functional lib. I suspect `class TypeInfo {` and its object [type_info](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/basic/test_progvar.cpp#L270)
Changing the variable name to any other solve issue (I hope :))

kpet:
Oh, I see, thanks. I've pluralised the variable name. That should hopefully do it.

gwawiork:
@kpet Could you verify if spirv_new tests works. On my side I see program is terminated unexpectedly. Looks like issue with test case registration.

kpet:
@gwawiork There was a global constructor ordering issue. I've pushed a workaround for now but we'll need a bit more work to make features work with the static registration that spirv_new uses.

gwawiork:
@kpet It does not help. I see termination again. New huge amount of  warning messages appeared  in build log:
`2>test_op_vector_times_scalar.cpp
2>test_common\harness/feature.h(34): message #2407: the initialization of member "feature::m_predicate" will be done before that of member "feature::m_name"
2>      feature(): m_name("true"), m_predicate(nullptr) {}`

gwawiork:
@kpet Warning disappear , so the fix works here. I see still uncaught exception in code [line](https://github.com/kpet/OpenCL-CTS/blob/features-deliver/test_common/harness/testHarness.cpp#L693) Looking into lower layer it interrupts in code from functional MSVC library `c:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\include\functional`
line 1281 and code looks:
`	bool _Empty() const noexcept
		{	// return true if no stored object
		return (_Getimpl() == nullptr);
		}`

kpet:
@gwawiork Thanks. I've spent a bit more time on this over the weekend and found that two of the changes I did (adding the constructor and getting rid of `F_true`) to address review comments actually broke the whole thing in a fairly subtle way (C++ rules for lifetime extension of temporaries are tricky). I've reworked the change to a) no longer rely on obscure corners of the C++ spec b) make it possible to use features with static test registration. I believe I've got all of that now working and just need to spend a bit more time on it to tidy it up and update the PR.

test_common/harness/testHarness.h
```diff
@@ -18,58 +18,40 @@
...
 {
+    test_definition(basefn fn, const char *n, Version v, const feature &feat)
+        : func(fn), name(n), min_version(v), required_feature(feat)
+    {}
     basefn func;
     const char* name;
     Version min_version;
-} test_definition;
-
+    const feature &required_feature;
```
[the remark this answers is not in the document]
kpet: This is known, see https://github.com/KhronosGroup/OpenCL-CTS/pull/816#issuecomment-644042693

test_conformance/api/main.cpp
```diff
@@ -29,104 +29,103 @@
 // (for example, generate_random_image_data()), the tests are required to declare
 // the following variables (<rdar://problem/11111245>):
 
-test_definition test_list[] = {
```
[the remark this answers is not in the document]
kpet: I did consider this but thought the syntax would be a bit clunky. I did end up solving this in a different way (see https://github.com/KhronosGroup/OpenCL-CTS/pull/816#issuecomment-644042693).

PR #818, opened 2020-06-05
Test bruteforce - fix relaxedMode in unary.cpp

### LINKED ISSUE #817: test burteforce - broken mode -cl-fast-relaxed-math in some tests
After recent changes #807 relaxed mode has been broken. The releaxedMode variable is false while should be true in unary.cpp

gwawiork:
Fixes #817


gwawiork:
Travis builds are passed . Looks like status cannot be refreshed.

PR #819, opened 2020-06-09
Skip Reduced Precision Testing

FranklandJack:
* OpenCL versions before 2.0 do not have precision requirements for
reduced precision math.

* Skip reduced precision testing for devices with
versions < 2.0.

### DIFF
--- a/test_conformance/math_brute_force/main.cpp
+++ b/test_conformance/math_brute_force/main.cpp
@@ -174,24 +174,29 @@ int doTest( const char* name )
             InitILogbConstants();
         }
 
-        if ( gTestFastRelaxed )
+        if (gTestFastRelaxed && func_data->relaxed)
         {
-            if( func_data->relaxed )
+            if (get_device_cl_version(gDevice) > Version(1, 2))
             {
                 gTestCount++;
-                vlog( "%3d: ", gTestCount );
+                vlog("%3d: ", gTestCount);
                 // Test with relaxed requirements here.
                 if (func_data->vtbl_ptr->TestFunc(func_data, gMTdata,
                                                   true /* relaxed mode */))
                 {
                     gFailCount++;
                     error++;
-                    if( gStopOnError )
+                    if (gStopOnError)
                     {
                         gSkipRestOfTests = true;
                         return error;
                     }
                 }
+            }
+            else
+            {
+                vlog("Skipping reduced precision testing for device with "
+                     "version 1.2 or less\n");
             }
         }
 


test_conformance/math_brute_force/main.cpp
```diff
@@ -194,6 +194,11 @@ int doTest( const char* name )
                 }
             }
         }
+        else
```
alycm, line 197: This will print the message even if `func_data->relaxed` is false anyway.   I.e. it will print "skip" for tests that would never run anyway.

test_conformance/math_brute_force/main.cpp
```diff
@@ -194,6 +194,11 @@ int doTest( const char* name )
...
+        {
+            vlog("Skipping reduced precision testing for device with "
```
alycm, line 199: I've seen discussion recently about having a more standardized way for recording tests that skip, do you happen to know if something happened about that?

FranklandJack: Yes, support for that will be introduced in #816. I thought about using this framework, however I'm not sure how it would work since the `feature` object would need to skip the test [here](https://github.com/KhronosGroup/OpenCL-CTS/pull/816/files#diff-1c97642f8bb6a546daaa3815c4b8c8e3R693) based on whether it was running in reduced mode or not.

test_conformance/math_brute_force/main.cpp
FranklandJack: Good point, I have updated this and refactored some of the control flow to make it cleaner.

alycm: Thanks.

FranklandJack:
Do you have any thoughts on this PR @kpet, it'd be good to get it in?

kpet:
The change looks good. A few comments:
- This behaviour is not documented in the unified spec. I remember we discussed creating an issue for this but I couldn't find it. @alycm Have I missed something?
- Technically the relaxed requirements are linked to `-cl-unsafe-math-optimizations` and not `-cl-fast-relaxed-math`. The latter implies the former, so the code is correct but doesn't represent the spec as closely as possible. I don't think we need to change it here though.
- We could introduce a feature later and plug this into the framework if we wanted but we'd need substantial changes to how the test is structured (so we can skip parts of the test). Not worth the time with the current state of things I think.

alycm:
> This behaviour is not documented in the unified spec. I remember we discussed creating an issue for this but I couldn't find it. @alycm Have I missed something?

That's because the ULP tables are in the OpenCL C spec, but I've still filed an issue about it (https://github.com/KhronosGroup/OpenCL-Docs/issues/340) as I'd like to get that sorted out soon as well.  So currently in terms of documentation it's their absence from the description of OpenCL C 1.2 (in the OpenCL 1.2 spec) that is important -- but obviously we can do better than that for documenting it.

The fact that these kernels are not compiled with `-cl-std=CL2.0` did set off a thought process that is covered in that ticket, but it's a bit of a rabbit hole that I don't think need to affect this PR.

kpet:
> That's because the ULP tables are in the OpenCL C spec [...]

I know that ;). I did check the 1.2 spec and latest unified spec as part of reviewing this change. I made the comment to make sure that we capture the need to document the 1.2/post-1.2 difference in the unified spec since we decided that we want to cover all versions of OpenCL there. Having https://github.com/KhronosGroup/OpenCL-Docs/issues/340 is enough for me, thanks. Since this change is simple and has been in review for a long time, I'll merge it straight away.

FranklandJack:
Thanks :)

PR #821, opened 2020-06-10
Widen 3D images for array-image copy

jenatali:
According to the OpenCL spec for clEnqueueCopyImageToBuffer (and similar for the reverse), the driver is allowed to fail requests that would result in using resource dimensions or pitches that are not supported:

> CL_​INVALID_​IMAGE_​SIZE if image dimensions (image width, height, specified or compute row and/or slice pitch) for src_image are not supported by device associated with queue.

For OpenCLOn12, we would like to require 256-byte row pitches and 512-byte slice pitches, because are the limits enforced by the [ID3D12GraphicsCommandList::CopyTextureRegion](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-copytextureregion) API. When the CTS gets down to one-byte-per-pixel image formats with a 128 texel wide image, that results in row pitches of 128 bytes. The simplest way to allow OpenCLOn12 to pass would be to just use wider textures.

### DIFF
--- a/test_conformance/basic/test_arrayimagecopy3d.cpp
+++ b/test_conformance/basic/test_arrayimagecopy3d.cpp
@@ -27,7 +27,7 @@ int test_arrayimagecopy3d_single_format(cl_device_id device, cl_context context,
 {
   cl_uchar    *bufptr, *imgptr;
   clMemWrapper      buffer, image;
-  int        img_width = 128;
+  int        img_width = 256;
   int        img_height = 128;
   int        img_depth = 32;
   size_t    elem_size;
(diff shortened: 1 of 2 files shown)


kpet:
FWIW, I think I wouldn't accept a patch doing this as there is nothing that constrains the minimal size of images in the specification (and the conformance process requires that some test be run with `small_images` as per https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/opencl_conformance_tests_12_full.csv#L49). The error you quoted is only there to catch the case where an image has been created with valid dimentions for at least one device in the context but not the device the copy command is being submitted to. Granted, it's not very clear. I'll raise a spec issue to look into clarifying this.

Two ways forward that come to mind:
1. Revert to a less efficient copy method when images don't meet D3D constraints.
1. Propose spec changes for a future version of OpenCL (that wouldn't be backwards compatible though)

jenatali:
Thanks @kpet, I've already done your 1st proposed suggestion. I'll go ahead and close this.

PR #824, opened 2020-06-11
Use std::thread for userevents_multithreaded test

jrprice:
Also fixes a thread leak reported by TSan.


jrprice:
> we should probably use plain C++ threads instead. I'm not sure this abstraction buys us much now that C++11 is required.

Agree, I've made this change since it's simple enough.

PR #827, opened 2020-06-14
Use power-of-two alignment values for allocating pixel data

### LINKED ISSUE #822: align_malloc fails to alloc memory for RGB float textures on Windows
RGB textures would request 12-byte aligned. The _aligned_malloc function used requires power-of-two alignment.

jenatali:
Using RGB 32bpc textures causes the CTS to request 12-byte-aligned data, but aligned allocation methods require power-of-two alignments. Fixes #822.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -75,12 +75,12 @@ sRGBunmap(float fc)
 }
 
 
-size_t get_format_type_size( const cl_image_format *format )
+uint32_t get_format_type_size(const cl_image_format *format)
 {
     return get_channel_data_type_size( format->image_channel_data_type );
 }
 
-size_t get_channel_data_type_size( cl_channel_type channelType )
+uint32_t get_channel_data_type_size(cl_channel_type channelType)
 {
     switch( channelType )
     {
@@ -132,12 +132,12 @@ size_t get_channel_data_type_size( cl_channel_type channelType )
     }
 }
 
-size_t get_format_channel_count( const cl_image_format *format )
+uint32_t get_format_channel_count(const cl_image_format *format)
 {
     return get_channel_order_channel_count( format->image_channel_order );
 }
 
-size_t get_channel_order_channel_count( cl_channel_order order )
+uint32_t get_channel_order_channel_count(cl_channel_order order)
 {
     switch( order )
     {
@@ -284,7 +284,7 @@ int is_format_signed( const cl_image_format *format )
     }
 }
 
-size_t get_pixel_size( cl_image_format *format )
+uint32_t get_pixel_size(cl_image_format *format)
 {
   switch( format->image_channel_data_type )
   {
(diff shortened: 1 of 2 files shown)


fooishbar:
This fixes the problem for me too. It looks like the Bionic build is also running clang-format (?) which wants to reformat extant code. I'm not sure whether modifying it in this PR is good style or not, @kpet?

EwanC (dismissed):
Looks like clang-format is failing :
```
diff --git a/test_common/harness/imageHelpers.cpp b/test_common/harness/imageHelpers.cpp

index 76853e9..d88d8c3 100644

--- a/test_common/harness/imageHelpers.cpp

+++ b/test_common/harness/imageHelpers.cpp

@@ -1261,7 +1261,8 @@ char * generate_random_image_data( image_descriptor *imageInfo, BufferOwningPtr<

     }

 #else

     P.reset( NULL ); // Free already allocated memory first, then try to allocate new block.

-    char *data = (char *)align_malloc(allocSize, get_pixel_alignment(imageInfo->format));

+    char *data =

+        (char *)align_malloc(allocSize, get_pixel_alignment(imageInfo->format));

     P.reset(data,NULL,0,allocSize, true);

 #endif

 

@@ -3184,8 +3185,9 @@ char *create_random_image_data( ExplicitType dataType, image_descriptor *imageIn

       P.reset(data);

     }

 #else

-    char *data = (char *)align_malloc(allocSize, get_pixel_alignment(imageInfo->format));

-    P.reset(data,NULL,0,allocSize,true);

+  char *data =

+      (char *)align_malloc(allocSize, get_pixel_alignment(imageInfo->format));

+  P.reset(data, NULL, 0, allocSize, true);

 #endif
```

jenatali:
Thanks for reminding me that I actually needed to get back to this. I saw @kpet's comment and put it on a mental to-do list, and then promptly forgot about it. I'll see if I can address both of these soon.

alycm (dismissed):
LGTM.

alycm (commented):
Actually, second thought: `next_power_of_two` takes and returns an `int`, and `get_pixel_alignment` returns an `int`.   But `get_pixel_size` returns a `size_t` and `align_malloc` expects a `size_t`.

So it would be more consistent to just use size_t everywhere?

jenatali:
> Actually, second thought: `next_power_of_two` takes and returns an `int`, and `get_pixel_alignment` returns an `int`. But `get_pixel_size` returns a `size_t` and `align_malloc` expects a `size_t`.
> 
> So it would be more consistent to just use size_t everywhere?

Since the `next_power_of_two` uses bit tricks for rounding, I'd prefer to stick with something that's a consistent bit-size, rather than `size_t`. Perhaps `get_pixel_alignment` should return `unsigned` (or `uint32_t`) instead, so everything stays 32-bit until being promoted to `size_t` for `aligned_malloc`. Opinion?

alycm:
> Since the `next_power_of_two` uses bit tricks for rounding, I'd prefer to stick with something that's a consistent bit-size, rather than `size_t`. Perhaps `get_pixel_alignment` should return `unsigned` (or `uint32_t`) instead, so everything stays 32-bit until being promoted to `size_t` for `aligned_malloc`. Opinion?

Ah, of course, 64-bits would require an extra shift.  `uint32_t` makes sense to me, as it makes the specific bit-width requirement explicit.

alycm (approved):
Thanks!  This makes sense to me, using `uint32_t` consistently.

alycm:
Ah, this has hit the clang-format check again.

jenatali:
> Ah, this has hit the clang-format check again.

Not sure why it only triggers on changed lines... that seems odd. But oh well, this should work now.

alycm:
> > Ah, this has hit the clang-format check again.
> 
> Not sure why it only triggers on changed lines... that seems odd. But oh well, this should work now.

It is working as intended.  The aim being that all new code or touched code should be formatted.  But we really need to do a mass-formatting to avoid these line-by-line situations (https://github.com/KhronosGroup/OpenCL-CTS/issues/625).

But not in this PR.

alycm (approved):

PR #830, opened 2020-06-18
Add `CL_DEVICE_IMAGE_SUPPORT` to clDeviceInfo File

FranklandJack:
- [x] Add `CL_DEVICE_IMAGE_SUPPORT` to the clDeviceInfo file
which contains
device specific information for the purposes of offline compilation. For
devices with images support `CL_DEVICE_IMAGE_SUPPORT=1` is printed,
otherwise `CL_DEVICE_IMAGE_SUPPORT=0`.

### DIFF
--- a/test_common/harness/kernelHelpers.cpp
+++ b/test_common/harness/kernelHelpers.cpp
@@ -372,6 +372,8 @@ static cl_int get_cl_device_info_str(const cl_device_id device, const cl_uint de
         clDeviceInfoStream << "CL_DEVICE_IL_VERSION=\"" << ilVersionString << "\"" << std::endl;
     }
     clDeviceInfoStream << "CL_DEVICE_VERSION=\"" << versionString << "\"" << std::endl;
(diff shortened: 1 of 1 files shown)


FranklandJack:
Should we ideally be printing all the device queries here, or at least the set of queries which contain information relevant for offline compilation? At the moment only the following queries are written to the file:
```
CL_DEVICE_ADDRESS_BITS
CL_DEVICE_EXTENSIONS
CL_DEVICE_IL_VERSION
CL_DEVICE_VERSION
CL_DEVICE_IMAGE_SUPPORT
```

kpet:
I think emitting only the queries that we really need makes things easier to manage and reason about (we know what queries are actually used/needed).

FranklandJack:
> I think emitting only the queries that we really need makes things easier to manage and reason about (we know what queries are actually used/needed).

Thanks for your feedback. Sounds good, it should be straightforward to add anymore queries this way as/when they are needed.

FranklandJack:
@kpet Since this has been approved is it good to merge?

PR #834, opened 2020-06-25
New subgroups - extended types cases

gwawiork:
Contributes to #532
 I've spited previous #533 to more commits without new test cases (please review commit by commit). It helps solving partially issue #812. It needs some other code (I will push it in next commits).

Some remarks and explanation of implementation:
Need to defined custom type for vector size = 3 and half type. This is because code with templates could not compile because of doubled definition (cl_half is the same as cl_short , vector size 3 is defined as vector size 4 in headers). That is why I need to redefine this types using structs in namespace subgroups. 
General solution base on type_traits library support with adding support for custom types. Specialization templates have been defined to define each type property and parameters required by particular test case (like numeric limits).
The half arithmetic executed on host side was done with usage existing functions which coverts half to float and back. These function were found in 2 different places in repository so I'have decided to [move it](https://github.com/KhronosGroup/OpenCL-CTS/commit/0ed70e0460ed5ec008e4cd9c582bb46f613b9b14) to one place to harness/conversion.h

Once this pull request is accepted next code for the rest of functions will be pushed.


gwawiork:
@kpet Could you look at current changes? I have added some enhancements and code reduction refactoring. Please read above explanation as well. Thanks !

test_conformance/subgroups/workgroup_common_kernels.cpp
```diff
@@ -0,0 +1,104 @@
...
+                           "    out[gid] = sub_group_broadcast(x, xy[gid].z);\n"
+
+                           "}\n";
+
+const char* redadd_source = "__kernel void test_redadd(const __global Type "
+                            "*in, __global int4 *xy, __global Type *out)\n"
+                            "{\n"
+                            "    int gid = get_global_id(0);\n"
+                            "    XY(xy,gid);\n"
+                            "    out[gid] = sub_group_reduce_add(in[gid]);\n"
```
StuartDBrady, line 33: Do we / should we check the type of the result of the reduction, anywhere?  Also, do we / should we check the type of the operand, too?

test_conformance/subgroups/subhelpers.h
```diff
@@ -56,219 +66,943 @@ class subgroupsAPI {
...
+template <typename Ty> inline bool compare_ordered(const Ty &lhs, const Ty &rhs)
+{
+    return lhs == rhs;
+}
+
+template <>
+inline bool compare_ordered(const subgroups::cl_half &lhs,
+                            const subgroups::cl_half &rhs)
+{
+    return lhs.data == rhs.data && !isnan_half(lhs);
```
bashbaug, line 1004: Is this safe?  Consider comparing positive zero with negative zero.

bashbaug (changes requested):
Mostly looks good but I have a few comments.  Thanks!

test_conformance/subgroups/subhelpers.h
```diff
@@ -56,219 +66,943 @@ class subgroupsAPI {
     clGetKernelSubGroupInfoKHR_fn _clGetKernelSubGroupInfo_ptr;
 };
 
-// Some template helpers
-template <typename Ty> struct TypeName;
-template <> struct TypeName<cl_half>
+namespace subgroups {
+struct cl_char3
```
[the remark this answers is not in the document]
gwawiork: Yes

test_conformance/subgroups/subhelpers.h
```diff
@@ -354,25 +1093,55 @@ struct test
...
         clKernelWrapper kernel;
         cl_platform_id platform;
-        cl_int sgmap[2 * GSIZE];
+        cl_int sgmap[4 * GSIZE];
         Ty mapin[LSIZE];
         Ty mapout[LSIZE];
+        std::stringstream kernel_sstr;
 
+        kernel_sstr << "#define NON_UNIFORM " + std::to_string(NON_UNIFORM)
+                + " \n";
```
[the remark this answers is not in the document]
gwawiork: It is for future tests

test_conformance/subgroups/subhelpers.h
```diff
@@ -22,6 +22,16 @@
...
 #include <limits>
 #include <vector>
+#include <type_traits>
+
+#undef min
+#undef max
+
+// Adjust these individually below if desired/needed
+#define G 2000
+#define L 200
```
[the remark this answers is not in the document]
gwawiork: see template definition in subhelpers.h file
`template <typename Ty, typename Fns, size_t GSIZE, size_t LSIZE,
          size_t TSIZE = 0>`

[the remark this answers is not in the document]
gwawiork: GSIZE is G LSIZE is L

test_conformance/subgroups/workgroup_common_kernels.cpp
gwawiork: Each kernel defines which data type will be used using [define](https://github.com/gwawiork/OpenCL-CTS/blob/extsub_2711_2020_all_changes/test_conformance/subgroups/subhelpers.h#L1285)
Validation is done on host side (at least results). 
I do not know if on device side there is a such need. Have we ever done such checks on device side ?

test_conformance/subgroups/subhelpers.h
gwawiork: Ben
Reading in the Internet [info ](https://en.wikipedia.org/wiki/Signed_zero)and comparing compilation VS and gcc the results is always true (equal)
https://godbolt.org/z/TMzEET

mantognini: Your example is testing float. Half being a non-standard type in C/C++, the host code probably cannot rely on the same assumption as for `float` or `double`.

mantognini:
I wonder, should this PR be closed in favour of #1074? If not, what is the main difference? Thanks.

test_conformance/subgroups/subhelpers.h
bashbaug: Right - in `cl_platform.h` we've defined `cl_half` to be an unsigned 16-bit integer that can encode the bit patterns for a half but this means that it does not necessarily obey the floating-point behavior since there are no overloaded operators.  I believe this demonstrates the problem: https://godbolt.org/z/d17rd6

One solution (untested) might be:

```c
return cl_half_to_float(lhs.data) == cl_half_to_float(rhs.data);
```

(This comment applies equally well to #1074, if this PR gets closed.)

test_common/harness/conversions.h
```diff
@@ -119,6 +119,133 @@ static inline int IsHalfSubnormal( cl_half x )
     return ( ( x & 0x7fffU ) - 1U ) < 0x03ffU; 
 }
 
+static inline float half2float(cl_ushort us)
```
jlewis-austin, line 122: since we're moving these, it might be a good time to change them to use cl_half throughout.

test_conformance/subgroups/subhelpers.h
```diff
@@ -56,219 +66,943 @@ class subgroupsAPI {
...
+// compare for scalar halfs
+template <typename Ty>
+typename std::enable_if<TypeManager<Ty>::is_sb_scalar_type::value, bool>::type
+compare(const Ty &lhs, const Ty &rhs)
+{
+    return lhs.data == rhs.data;
+}
+
+
+inline bool isnan_half(const subgroups::cl_half &x)
```
jlewis-austin, line 990: This and a few others are in imageHelpers.h: https://github.com/KhronosGroup/OpenCL-CTS/blob/24e6a9125c44ef9fe42603b605a685af9acb4bdf/test_common/harness/imageHelpers.h#L712
Might consider moving those to the harness as well.

test_conformance/subgroups/subhelpers.h
jlewis-austin: I haven't found the usages of this yet, but we should define and document the semantics we're trying to achieve. Most tests don't want to do a straight IEEE-754 comparison because NaNs will fail. Typical options are:

1. Bitwise exact (just compare as integers)
2. Bitwise exact, excluding NaNs (allows the NaN payload to vary from the reference)
3. Bitwise exact, excluding NaNs and subnormals (if flush to zero is allowed)

There are examples of #3 in the image tests (e.g. https://github.com/KhronosGroup/OpenCL-CTS/blob/24e6a9125c44ef9fe42603b605a685af9acb4bdf/test_conformance/images/kernel_read_write/test_iterations.cpp#L1173)

test_common/harness/conversions.h
gwawiork: Sorry for confusion. This PR has to be closed. This is the old version of code. The new https://github.com/KhronosGroup/OpenCL-CTS/pull/1074 do not uses this function and uses from OpenCL headers  https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl_half.h#L361

test_conformance/subgroups/subhelpers.h
gwawiork: fixed in https://github.com/KhronosGroup/OpenCL-CTS/pull/1074

test_conformance/subgroups/subhelpers.h
gwawiork: Then I am not sure for now which way to go :) I will use Ben's proposal. Fixed https://github.com/KhronosGroup/OpenCL-CTS/pull/1074

gwawiork:
This PR is a susbset of https://github.com/KhronosGroup/OpenCL-CTS/pull/1074. If any comment please use this new PR. 
Thanks !

PR #836, opened 2020-07-01
Add `CL_DEVICE_NAME` to clDeviceInfo File

FranklandJack:
- [x] Add `CL_DEVICE_NAME` to list of device properties printed into the
clDeviceInfo file for offline compilation testing.
- [x] Add `get_device_name` helper function.
- [x] Update offline compiler interface explanation file with
`CL_DEVICE_NAME` and `CL_DEVICE_IMAGE_SUPPORT` which was missed from
this file when added.


kpet:
Why is the device name required?

FranklandJack:
> Why is the device name required?

We have an offline compiler supporting multiple devices, this enables us to dynamically select the device to compile for.

kpet:
Ok, makes sense. I'm guessing this is for use with binaries and not SPIR-V, is it?

FranklandJack:
> Ok, makes sense. I'm guessing this is for use with binaries and not SPIR-V, isn't it?

Yep exactly.

PR #838, opened 2020-07-03
Print hex values when reporting a failure

grey-eminence:
This fixes the problem with reporting not precise enough information in case of verification failure.

### DIFF
--- a/test_conformance/math_brute_force/binary_i.cpp
+++ b/test_conformance/math_brute_force/binary_i.cpp
@@ -694,7 +694,12 @@ static cl_int TestFloat( cl_uint job_id, cl_uint thread_id, void *data  )
                 }
                 if( fail )
                 {
-                    vlog_error( "\nERROR: %s%s: %f ulp error at {%a, %d}: *%a vs. %a (0x%8.8x) at index: %d\n", name, sizeNames[k], err, s[j], s2[j], r[j], test, ((cl_uint*)&test)[0], j );
+                    vlog_error(
+                        "\nERROR: %s%s: %f ulp error at {%a (0x%8.8x), %d}: "
+                        "*%a (0x%8.8x) vs. %a (0x%8.8x) at index: %d\n",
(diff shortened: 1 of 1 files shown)


test_conformance/math_brute_force/binary_i.cpp
```diff
@@ -695,6 +695,7 @@ static cl_int TestFloat( cl_uint job_id, cl_uint thread_id, void *data  )
                 if( fail )
                 {
                     vlog_error( "\nERROR: %s%s: %f ulp error at {%a, %d}: *%a vs. %a (0x%8.8x) at index: %d\n", name, sizeNames[k], err, s[j], s2[j], r[j], test, ((cl_uint*)&test)[0], j );
+                    vlog_error( "\nERROR: %s%s: %f ulp error at {%a (0x%8.8x), %d}: *%a (0x%8.8x) vs. %a (0x%8.8x) at index: %d\n", name, sizeNames[k], err, s[j], ((uint32_t*)s)[j], s2[j], r[j], ((uint32_t*)r)[j], test, ((cl_uint*)&test)[0], j );
```
b-sumner, line 698: Isn't the real problem simply that %a is used instead of %.6a?

grey-eminence: I need the exact bit representation of the values that make the test fail. We do the same thing for unary functions..

b-sumner: But %.6a gives you the exact bit representation, formatted in an easily understandable way.  You don't really need both.  But I am not opposing this, especially since it is (unfortunately) done elsewhere.

grey-eminence: @b-sumner, I agree that %.6a gives enough information to _deduct_ the exact bit representation, but very often I need to directly edit/override registers in simulation/emulation of upcoming hardware, edit ISA, etc. I need a machine understandable format. I wouldn't call that _unfortunate_, I would call it _helpful_..

PR #839, opened 2020-07-06
Bruteforce - embedded, relaxed ulp requirements

### LINKED ISSUE #760: Relaxed math testing uses Full Profile ULP requirements for Embedded Profile
The embedded profile ULP requirements for fast relaxed math are not considered.

Jeremy-Kemp:
The crux of this issue revolved specifically around `exp`, which has differing base ULP requirements between `FULL_PROFILE` and `EMBEDDED_PROFILE`

This change adds an additional field to the `Func` struct, which allows the specification of an embedded, relaxed ULP requirement.

Given that this issue specifically revolves around `exp` the change to `Func` could be technically be dropped, but this approach allows us some flexibility in future if required and gives good separation between the profiles.

Fixes #760

### DIFF
--- a/test_conformance/math_brute_force/FunctionList.h
+++ b/test_conformance/math_brute_force/FunctionList.h
@@ -88,6 +88,7 @@ typedef struct Func
   int             ftz;
   int             relaxed;
   const vtbl      *vtbl_ptr;
+  float float_embedded_relaxed_ulps;
 }Func;
 
 
--- a/test_conformance/math_brute_force/FunctionList.h
+++ b/test_conformance/math_brute_force/FunctionList.h
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -85,10 +85,10 @@ typedef struct Func
   float           double_ulps;
   float           float_embedded_ulps;
   float           relaxed_error;
+  float relaxed_embedded_error;
   int             ftz;
   int             relaxed;
   const vtbl      *vtbl_ptr;
-  float float_embedded_relaxed_ulps;
 }Func;
 
 
--- a/test_conformance/math_brute_force/Utility.h
+++ b/test_conformance/math_brute_force/Utility.h
@@ -29,6 +29,7 @@
 #include "harness/testHarness.h"
 #include "harness/ThreadPool.h"
 #include "harness/conversions.h"
+#include "FunctionList.h"
 #define BUFFER_SIZE         (1024*1024*2)
 
 #if defined( __GNUC__ )
@@ -223,6 +224,8 @@ int compareFloats(float x, float y);
 int compareDoubles(double x, double y);
 
 void logFunctionInfo(const char *fname, unsigned int float_size, unsigned int isFastRelaxed);
+
+float getAllowedUlpError(const Func *f, const bool relaxed);
 
 #endif /* UTILITY_H */
 
--- a/test_conformance/math_brute_force/Utility.h
+++ b/test_conformance/math_brute_force/Utility.h
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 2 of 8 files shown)


Jeremy-Kemp:
Thanks for the feedback @gwawiork . I've fixed the compilation issue, and also the formatting problems.

test_conformance/math_brute_force/unary_two_results.cpp
```diff
@@ -220,7 +220,17 @@ int TestFunc_Float2_Float(const Func *f, MTdata d, bool relaxedMode)
...
+    {
+        if (gIsEmbedded)
+        {
+            float_ulps = f->float_embedded_relaxed_ulps;
+        }
+        else
+        {
+            float_ulps = f->relaxed_error;
+        }
+    }
```
kpet, line 233: The overall change looks good to me. It's a bit unfortunate that all the code formatting makes this harder to review than it should be.

I think we should introduce a utility function to get the allowed relative error for a given function. How about
```
float getAllowedUlpError(Func* f, bool relaxed);
```

Jeremy-Kemp: I agree, a utility function would be nice here. I'll put something together.

Jeremy-Kemp: I've added a function to do this but kept its use restricted to the necessary locations for this particular issue.

I will raise a new issue to track using it throughout the rest of the test, as well as supporting double ULPs.

test_conformance/math_brute_force/FunctionList.h
```diff
@@ -88,6 +88,7 @@ typedef struct Func
   int             ftz;
   int             relaxed;
   const vtbl      *vtbl_ptr;
+  float float_embedded_relaxed_ulps;
```
bashbaug, line 91: Somewhat of a cosmetic comment, but I would find these changes easier to understand if this field was listed in the structure after `relaxed_error`, and if it were renamed slightly - maybe `relaxed_embedded_error`?  This would make it more obvious that these two values are related.

test_conformance/math_brute_force/Utility.h
```diff
@@ -29,6 +29,7 @@
 #include "harness/testHarness.h"
 #include "harness/ThreadPool.h"
 #include "harness/conversions.h"
+#include "FunctionList.h"
```
bashbaug, line 32: I'm not going to lose sleep over this, but it does feel a little awkward to include this file here.  Consider forward declaring `Func` instead?

test_conformance/math_brute_force/FunctionList.h
Jeremy-Kemp: No problem in renaming, but I'm a bit reluctant to re-order.

Doing so would also involve changing the macros in `FunctionList.cpp`. I've managed to avoid that thus far as the new member is at the end of the struct, so does not break the implicit ordering of the existing macros.

test_conformance/math_brute_force/Utility.h
Jeremy-Kemp: The implementation still needs to be able to de-reference `Func`.

Jeremy-Kemp: I could also just move it into 'FunctionList.cpp' which might be a more logical place anyway?

test_conformance/math_brute_force/FunctionList.h
bashbaug: Would this though?  I agree we'd need to modify `ENTRY_EXT`, but we're doing this anyways.  I don't think we'd need to modify any of the others though, we'd just need to insert `INFINITY` into the middle of the structure vs. the end.

test_conformance/math_brute_force/Utility.h
bashbaug: Right, the implementation needs to dereference `Func`, but that just means that `Utility.cpp` needs to include `FunctionList.h`, not `Utility.h`.

I thought about moving it to `FunctionList.cpp` also, but because the implementation accesses `gIsEmbedded`, it'd mean that `FunctionList.cpp` would need to include `Utility.h`, which also feels a bit awkward.

I'm honestly fine leaving this as-is.  It may be that it's awkward regardless, and I agree that having a utility function is a nice improvement.

test_conformance/math_brute_force/FunctionList.h
Jeremy-Kemp: It was the addition of `INFINITY` that I was getting at as I usually prefer to limit the scope of changes for bug fixes.

Regardless, I've made the change now.

PR #840, opened 2020-07-07
Update List of Skipped Tests in Offline Mode

FranklandJack:
Fix Tests Assuming Online Compilation in Offline MOde

* Update the list of tests to skip in offline mode. These tests need
to be skipped for offline-binary mode since they make API calls that
rely on a compiler being present in the runtime.

- [x] Skip `get_kernel_arg_info_compatibility`  since it makes calls to
`clBuildProgram`. (these tests cannot be run in offline mode since:
*Kernel argument information is only available if the program object
associated with kernel is created with `clCreateProgramWithSource` and
the program executable was built with the `-cl-kernel-arg-info` option
specified in options argument to `clBuildProgram` or `clCompileProgram`
.*)

- [x] Skip `compiler` tests that make calls to `clCompileProgram`. These
tests could still be run in offline spirv mode if there is a compiler in
the driver.

- [x] Use offline compilation path in `contractions` in the case that
CTS is run in offline.


test_common/harness/errorHelpers.cpp
```diff
@@ -748,10 +749,46 @@ const char * subtests_to_skip_with_offline_compiler[] = {
...
+                  "failed: %s\n",
+                  error);
+        return 1;
+    }
+    if (!is_device_compiler_available) {
+        auto device_name = get_device_name(device);
+        log_info(
+            "Skiping subtest %s because there is no compiler for device: %s\n",
+            subtestname, device_name.c_str());
+        return 1;
```
FranklandJack, line 789: I'm not sure whether this should be considered a skip or a failure in the case that there is no compiler support?

test_conformance/contractions/contractions.cpp
```diff
@@ -576,56 +576,39 @@ test_status InitCL( cl_device_id device )
...
+        size_t strCount = sizeof(kernels) / sizeof(kernels[0]);
         kernels[0] = "";
 
-        for( j = 2; j < strCount; j += 2 )
-            kernels[j] = sizeNames[i];
-
-        gProgram[i] = clCreateProgramWithSource(gContext, strCount, kernels, NULL, &error);
-        if( NULL == gProgram[i] )
+        for (j = 2; j < strCount; j += 2) kernels[j] = sizeNames[i];
+        const auto error = create_single_kernel_helper(
```
bashbaug, line 585: Very minor: `error` is already declared on Line 463.  I think it would be better not to add this shadow declaration.  See also Line 604.

```suggestion
        error = create_single_kernel_helper(
```

test_conformance/contractions/contractions.cpp
```diff
@@ -576,56 +576,39 @@ test_status InitCL( cl_device_id device )
...
-            gProgram_double[i] = clCreateProgramWithSource(gContext, strCount, kernels, NULL, &error);
-            if( NULL == gProgram_double[i] )
-            {
-                vlog_error( "clCreateProgramWithSource failed\n" );
-                return TEST_FAIL;
-            }
-
-            if(( error = clBuildProgram(gProgram_double[i], 1, &device, NULL, NULL, NULL) ))
+            for (j = 2; j < strCount; j += 2) kernels[j] = sizeNames_double[i];
+            const auto error = create_single_kernel_helper(
```
bashbaug, line 604:
```suggestion
            error = create_single_kernel_helper(
```

test_conformance/contractions/contractions.cpp
EwanC: Pushed a new commit with this, was a slight clang-format modification too as a result

PR #841, opened 2020-07-08
Add test for CL_PROFILING_COMMAND_COMPLETE

ouakheli:
The test checks profiling value CL_PROFILING_COMMAND_COMPLETE
compared to CL_PROFILING_COMMAND_END for two cases:
1) kernel with children with different levels of nesting
2) kernel without children

### DIFF
--- a/test_conformance/device_execution/enqueue_profiling.cpp
+++ b/test_conformance/device_execution/enqueue_profiling.cpp
@@ -1,0 +1,171 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include <stdio.h>
+#include <string.h>
+#include "harness/testHarness.h"
+#include "harness/typeWrappers.h"
+
+#include <vector>
+
+#include "procs.h"
+#include "utils.h"
+#include <time.h>
+
+static int max_nestingLevel = 10;
+
+static const char* enqueue_multi_level = R"(
+    void block_fn(__global int* res, int level)
+    {
+      queue_t def_q = get_default_queue();
+      if(--level < 0) return;
+      void (^kernelBlock)(void) = ^{ block_fn(res, level); };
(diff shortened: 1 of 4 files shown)


test_conformance/device_execution/main.cpp
```diff
@@ -54,17 +54,12 @@ test_status InitCL(cl_device_id device) {
...
-    ADD_TEST( enqueue_multi_queue ),
-    ADD_TEST( host_multi_queue ),
-    ADD_TEST( enqueue_ndrange ),
-    ADD_TEST( host_queue_order ),
+    ADD_TEST(device_info),           ADD_TEST(device_queue),
+    ADD_TEST(execute_block),         ADD_TEST(enqueue_block),
+    ADD_TEST(enqueue_nested_blocks), ADD_TEST(enqueue_wg_size),
+    ADD_TEST(enqueue_flags),         ADD_TEST(enqueue_multi_queue),
+    ADD_TEST(host_multi_queue),      ADD_TEST(enqueue_ndrange),
+    ADD_TEST(host_queue_order),      ADD_TEST(enqueue_profiling),
```
kpet, line 62: Did clang-format do that? I think I'd rather we stick to one per line.

test_conformance/device_execution/enqueue_profiling.cpp
```diff
@@ -0,0 +1,169 @@
...
+        test_error(err_ret, "clGetEventProfilingInfo() failed");
+
+        cl_ulong complete;
+        err_ret =
+            clGetEventProfilingInfo(kernel_event, CL_PROFILING_COMMAND_COMPLETE,
+                                    sizeof(complete), &complete, NULL);
+        test_error(err_ret, "clGetEventProfilingInfo() failed");
+
+        if (check_error(status, "Kernel execution status %d", status))
+            return status;
```
kpet, line 135: Might be better to do that before the calls to `clGetEventProfilingInfo` since these would fail if the status was not `CL_COMPLETE`.

test_conformance/device_execution/enqueue_profiling.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+        if (result != level)
+        {
+            log_error("Kernel execution failed %d", level);
```
kpet, line 139: The error message could be more explicit. Something along the lines of "expected X but got Y" would be useful.

kpet (approved):
The code is correct as is but should ideally use `TEST_PASS` and `TEST_FAIL` to return status.

bcalidas:
We have concerns about this patch.  Our understanding of the spec is that CL_PROFILING_COMMAND_COMPLETE is guaranteed to be equal to CL_PROFILING_COMMAND_END only when the device does not support on-device enqueue. For devices that do support on-device enqueue there could be some additional processing required irrespective of whether a child kernel gets enqueued. This would mean that CL_PROFILING_COMMAND_COMPLETE would be later than CL_PROFILING_COMMAND_END.

PR #851, opened 2020-07-22
Arbitrary output directory support for compiler test build copies (#848)

### LINKED ISSUE #848: Generalize test_compiler header file copy support
Some tests from test_compiler rely on accessing header files that need to be made available by the build:
```
options_include_directory
preprocessor_include
execute_after_included_header_link
```
The top-level CMakeLists.txt copies the required headers into the binary directory, but at a fixed relative path that relies on CMAKE_RUNTIME_OUTPUT_DIRECTORY being set as the current build directory for the module (via -DCMAKE_RUNTIME_OUTPUT_DIRECTORY=. in the build script).

Need to make this more generic to work with any specified CMAKE_RUNTIME_OUTPUT_DIRECTORY path.

jlewis-austin:
* Move headers copy from top level CMakeLists.txt down to test_compile module as post-build copy
* Fix windows VS2015 build break

resolves #848

### DIFF
--- a/test_conformance/compiler/CMakeLists.txt
+++ b/test_conformance/compiler/CMakeLists.txt
@@ -13,3 +13,20 @@ set(${MODULE_NAME}_SOURCES
 )
 
 include(../CMakeCommon.txt)
+
+# Copy required CL include directories into the binary directory
+
+set(COMPILER_SOURCE_DIR ${CLConform_SOURCE_DIR}/test_conformance/compiler)
+set(COMPILER_TARGET ${${MODULE_NAME}_OUT})
+
+add_custom_command(
+    TARGET  ${COMPILER_TARGET}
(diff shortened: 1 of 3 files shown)


jlewis-austin:
Change tested on linux and windows with VS 2015/2017/2019

test_conformance/compiler/CMakeLists.txt
```diff
@@ -13,3 +13,20 @@ set(${MODULE_NAME}_SOURCES
 )
 
 include(../CMakeCommon.txt)
+
+# Copy required CL include directories into the binary directory
+
+set(COMPILER_SOURCE_DIR ${CLConform_SOURCE_DIR}/test_conformance/compiler)
```
kpet, line 19: Could use `PROJECT_SOURCE_DIR`.

PR #857, opened 2020-07-28
Test against clCreateBufferWithProperties and clCreateImageWithProperties OpenCL 3.0 API

### LINKED ISSUE #853: OpenCL 3.0 tests against clCreateBufferWithProperties and clCreateImageWithProperties
This issue is to create simple test for two new OpenCL api calls:
- [x] clCreateBufferWithProperties 
- [x] clCreateImageWithProperties

The plan is to add new test cases to test_api suite:
test_buffer_properties.cpp
test_image_properties.cpp

gwawiork:
Fixes #853

### DIFF
--- a/test_conformance/api/test_mem_object_properties_queries.cpp
+++ b/test_conformance/api/test_mem_object_properties_queries.cpp
@@ -18,6 +18,7 @@
 #include "testBase.h"
 #include "harness/typeWrappers.h"
 #include <vector>
+#include <algorithm>
 
 using namespace std;
 
@@ -27,16 +28,9 @@ typedef enum
     buffer
 } mem_obj_type;
 
-typedef enum
-{
-    NON_NULL_PATH,
-    NULL_PATH
-} query_type;
-
 struct test_data
 {
     mem_obj_type obj_t;
-    query_type query_t;
     std::vector<cl_mem_properties> properties;
     std::string description;
     std::string src;
@@ -64,7 +58,7 @@ int create_object_and_check_properties(cl_context context,
 
     if (test_case.obj_t == image)
     {
-        if (test_case.query_t == NULL_PATH)
+        if (test_case.properties.size() == 0)
         {
             test_object =
                 clCreateImageWithProperties(context, NULL, flags, &format,
@@ -80,7 +74,7 @@ int create_object_and_check_properties(cl_context context,
     }
     if (test_case.obj_t == buffer)
     {
-        if (test_case.query_t == NULL_PATH)
+        if (test_case.properties.size() == 0)
         {
             test_object = clCreateBufferWithProperties(
                 context, NULL, flags, local_data.size() * sizeof(cl_uint),
@@ -98,15 +92,25 @@ int create_object_and_check_properties(cl_context context,
     clGetMemObjectInfo(test_object, CL_MEM_PROPERTIES, 0, NULL, &set_size);
     test_error(error,
                "clGetMemObjectInfo failed asking for CL_MEM_PROPERTIES.");
-    if (test_case.query_t == NULL_PATH && set_size == 0)
+
+    // verify set_size 0 returned
+    if ((test_case.properties.size() == 0
+         || test_case.properties.size() == 1 && test_case.properties[0] == 0)
+        && set_size == 0)
     {
         return TEST_PASS;
     }
-    if (test_case.query_t == NON_NULL_PATH && set_size == 0)
+
+    if (test_case.properties.size() > 1)
     {
-        log_error("clGetMemObjectInfo - invalid value returned");
-        return TEST_FAIL;
+        if (test_case.properties.size() * sizeof(cl_mem_properties) > set_size)
+        {
+            log_error(
+                "Incorrect properties size returned by clGetMemObjectInfo\n");
+            return TEST_FAIL;
+        }
     }
+
     cl_uint number_of_props = set_size / sizeof(cl_mem_properties);
     object_properties_check.resize(number_of_props);
     clGetMemObjectInfo(test_object, CL_MEM_PROPERTIES, set_size,
@@ -114,12 +118,31 @@ int create_object_and_check_properties(cl_context context,
     test_error(error,
                "clGetMemObjectInfo failed asking for CL_MEM_PROPERTIES.");
 
-    if (object_properties_check != test_case.properties)
+    // check list with 0 terminator is returned
+    if (object_properties_check.size() == 1 && object_properties_check[0] == 0
+        && test_case.properties.size() == 0)
     {
-        log_error("clGetMemObjectInfo returned different properties than was "
-                  "set initially.\n");
-        error = TEST_FAIL;
+        return TEST_PASS;
     }
+    if (object_properties_check.size() == 1 && object_properties_check[0] == 0
+        && test_case.properties.size() == 1 && test_case.properties[0] == 0)
+    {
+        return TEST_PASS;
+    }
+
+
+    for (auto set_property : test_case.properties)
+    {
+        std::vector<cl_mem_properties>::iterator it =
+            std::find(object_properties_check.begin(),
+                      object_properties_check.end(), set_property);
+        if (it == object_properties_check.end())
+        {
+            log_error("ERROR: Property not found ...");
+            return TEST_FAIL;
+        }
+    }
+
     return error;
 }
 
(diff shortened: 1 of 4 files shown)


test_conformance/api/test_mem_object_properties_queries.cpp
```diff
@@ -0,0 +1,294 @@
...
+                       object_properties_check.data(), NULL);
+    test_error(error,
+               "clGetMemObjectInfo failed asking for CL_MEM_PROPERTIES.");
+
+    if (object_properties_check != test_case.properties)
+    {
+        log_error("clGetMemObjectInfo returned different properties than was "
+                  "set initially.\n");
+        error = TEST_FAIL;
+    }
```
bashbaug, line 122: The behavior being tested here is valid behavior but I don't think it's the _only_ valid behavior, see https://github.com/KhronosGroup/OpenCL-Docs/issues/317.

Since we haven't defined any new properties in OpenCL 3.0, I think we can only test that the query executed successfully:

* It's valid to return "0" for the size of the queried properties, since we have not defined any properties.  This is valid regardless whether the passed-in properties are `NULL` or an empty property list consisting only of the "0" terminator.
* It's valid to return a single property value of "0", indicating an empty property list.  Again, this is valid regardless whether the passed-in properties are `NULL` or an empty property list consisting only of the "0" terminator.
* It's valid to return additional default properties, say if an extension adds new properties with default values.

I suppose we could test that the property list consists of property pairs and is terminated by zero?  We can't check for equality with the passed-in properties, though.

test_conformance/api/test_mem_object_properties_queries.cpp
```diff
@@ -0,0 +1,294 @@
...
+    d = init_genrand(gRandomSeed);
+
+    generate_random_data(kUInt, size, d, src_data.data());
+    generate_random_data(kUInt, size, d, dst_data.data());
+    free_mtdata(d);
+    d = NULL;
+    const char* kernel_src = test_case.src.c_str();
+    error = create_single_kernel_helper_with_build_options(
+        context, &program, &kernel, 1, &kernel_src,
+        test_case.kernel_name.c_str(), "-cl-std=CL3.0");
```
bashbaug, line 165: Minor: this test is testing OpenCL 3.0 APIs, so it's probably safe to pass `-cl-std=CL3.0` here,  But, there's nothing in either kernel that actually requires OpenCL 3.0, so I'd consider omitting any build options to reduce complexity.

gwawiork:
@bashbaug I think all issues has been resolved in last commit. Please verify

test_conformance/api/test_mem_object_properties_queries.cpp
```diff
@@ -0,0 +1,301 @@
...
+    {
+        std::vector<cl_mem_properties>::iterator it =
+            std::find(object_properties_check.begin(),
+                      object_properties_check.end(), set_property);
+        if (it == object_properties_check.end())
+        {
+            log_error("ERROR: Property not found ...");
+            return TEST_FAIL;
+        }
+    }
```
bashbaug, line 144: This is getting very close - thank you! - but I think it's still not quite right.  In the general case, there are up to three conditions to check:

1. If the passed-in properties (`test_case.properties`) are `NULL` or consist only of the property terminator `0`, and if the queried property size is zero, then the test passes.  This check is on Line 97 - Line 102 and looks correct.

Otherwise, when the queried property size is non-zero:

2. For each passed-in property name-value pair, the same property name-value pair must be in the queried properties (`object_properties_check`) or the test fails.  This check is partially implemented on Line 139, but not completely, because it is looking at the individual properties and values and not property name-value pairs.  To check this condition, we should step through the passed-in properties and the queried properties by two, since we're checking for property name-value pairs.

    a. Note that this check is only needed in the general case, and not in this specific case, since we do not currently pass in any property name-value pairs for this test (yet).

3. The list of queried properties must end with the property terminator `0` or the test fails.  This check is partially implemented on Line 122 - Line 131, but we should check for the property terminator regardless of the size of the passed-in properties and the size of the queried properties.

I can provide the code to implement these checks if it would be helpful, just let me know.  Thanks!

test_conformance/api/test_mem_object_properties_queries.cpp
bashbaug: This thread has been superseded by my comment below (https://github.com/KhronosGroup/OpenCL-CTS/pull/857#discussion_r463286622).

test_conformance/api/test_mem_object_properties_queries.cpp
bashbaug: Minor update for (1): it's possible that this condition is only valid when the passed-in properties are `NULL`, and not when the passed-in properties consist only of the property terminator `0`.  See this spec question:

https://github.com/KhronosGroup/OpenCL-Docs/issues/317#issuecomment-666740492

gwawiork: Please, check now, I hope now all is addressed

gwawiork: Similar code will be entered to https://github.com/KhronosGroup/OpenCL-CTS/pull/858

bashbaug: Thanks!  I think this has all of the right checks now.

Can you please verify that the checks on Line 102 and Line 126 are necessary?  I think these conditions are already checked by other code paths but I could be mistaken.  If they aren't needed it would be nice to remove them to simplify the test.

gwawiork: I agree these conditions are not needed

test_conformance/api/test_mem_object_properties_queries.cpp
```diff
@@ -0,0 +1,302 @@
...
+            log_error("ERROR: Output results mismatch.");
+            return TEST_FAIL;
+        }
+    }
+
+    return 0;
+}
+
+
+int test_image_properties_queries(cl_device_id deviceID, cl_context context,
```
Jeremy-Kemp, line 249: The test should check that `CL_DEVICE_IMAGE_SUPPORT` is `CL_TRUE` before proceeding with the rest of the test case.

test_conformance/api/test_mem_object_properties_queries.cpp
bashbaug: See related comment here: https://github.com/KhronosGroup/OpenCL-CTS/pull/858#discussion_r465470110

test_conformance/api/test_mem_object_properties_queries.cpp
gwawiork: fixed

test_conformance/api/test_mem_object_properties_queries.cpp
gwawiork: > See related comment here: [#858 (comment)](https://github.com/KhronosGroup/OpenCL-CTS/pull/858#discussion_r465470110)

fixed

test_conformance/api/test_mem_object_properties_queries.cpp
```diff
@@ -0,0 +1,317 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "testBase.h"
+#include "harness/typeWrappers.h"
+#include <vector>
+#include <algorithm>
+
+using namespace std;
```
FranklandJack, line 22: Very minor thing, but since you have made use of the namespace resolution `std::` for `std::vector` and `std::string` in this file, would it make sense to remove this using statement?

test_conformance/api/test_mem_object_properties_queries.cpp
```diff
@@ -0,0 +1,317 @@
...
+{
+    int error = CL_SUCCESS;
+    log_info("\nTC description: %s\n", test_case.description.c_str());
+
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+    clMemWrapper obj_src;
+    clMemWrapper obj_dst;
+    clEventWrapper event;
+    MTdata d;
```
FranklandJack, line 165: Another minor point, but would it might be helpful for future developers to give this variable a slightly more descriptive name? It could also be initialized when it is defined.

gwawiork:
@EwanC and @FranklandJack Thanks for review. I think latest changes addressed all issues.

FranklandJack:
> @EwanC and @FranklandJack Thanks for review. I think latest changes addressed all issues.

Great thank you! @EwanC is away on holiday this week, but I'm sure he'd be happy for someone with write access to approve and merge this since you have addressed his comments :)

bashbaug:
This test will need to be updated slightly to reflect the latest spec clarification regarding `NULL` properties (see https://github.com/KhronosGroup/OpenCL-Docs/pull/398#issuecomment-672211863), but I think we should still merge these changes as-is.

I'll file a separate issue to ensure any tests for property queries are properly updated for the `NULL` property case.  We'll also need to update the test for the sampler property query, for example (added by https://github.com/KhronosGroup/OpenCL-CTS/pull/858).

bashbaug:
I've added the issue to check the property queries for the `NULL` property case here: https://github.com/KhronosGroup/OpenCL-CTS/issues/880.

PR #858, opened 2020-07-28
OpenCL 3.0 CL_SAMPLER_PROPERTIES tests.

gwawiork:


test_conformance/api/test_queries.cpp
```diff
@@ -209,6 +209,31 @@ int test_get_sampler_info(cl_device_id deviceID, cl_context context, cl_command_
...
+        numer_of_props = set_size / sizeof(cl_sampler_properties);
+        get_properties.resize(numer_of_props);
+        error = clGetSamplerInfo(sampler, CL_SAMPLER_PROPERTIES, set_size,
+                                 get_properties.data(), 0);
+        test_error(error, "clGetSamplerInfo failed.");
+        if (set_properties != get_properties)
+        {
+            log_error("Data mismatch while comparing CL_SAMPLER_PROPERTIES\n");
+            return TEST_FAIL;
+        }
```
bashbaug, line 234: I have a similar comment here as with https://github.com/KhronosGroup/OpenCL-CTS/pull/857#discussion_r462059337 - this is checking valid behavior, but it's not the only valid behavior.

For these sampler properties I think we should check:

* The query must execute successfully (we're already checking this).
* If the property for `CL_SAMPLER_NORMALIZED_COORDS` is present, it should have value `CL_TRUE`.
* If the property for `CL_SAMPLER_ADDRESSING_MODE` is present, it should have value `CL_ADDRESS_CLAMP`.
* The property for `CL_SAMPLER_FILTER_MODE` must be present and must have value `CL_FILTER_LINEAR`.  It cannot be omitted, since the default filter mode is `CL_FILTER_NEAREST`.
* Other properties are OK and should be ignored, since they could have been added via an extension.
* The returned properties list must be terminated with "0".

test_conformance/api/test_queries.cpp
```diff
@@ -209,6 +209,31 @@ int test_get_sampler_info(cl_device_id deviceID, cl_context context, cl_command_
...
     }
 
+    Version version = get_device_cl_version(deviceID);
+    if (version >= Version(3, 0))
+    {
+        std::vector<cl_sampler_properties> get_properties;
+        std::vector<cl_sampler_properties> set_properties(
+            properties, properties + ARRAY_SIZE(properties));
+        size_t set_size;
+        cl_uint numer_of_props = 0;
```
bashbaug, line 219: Minor: Possible misspelling?

test_conformance/api/test_queries.cpp
gwawiork: @bashbaug The test [test_get_sampler_info](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/api/test_queries.cpp#L144) at the beginning creates sampler with properties and then one by one checks if properties exists and have proper value. I am not sure if there is need to verify this second time while `set_properties.data()` is a copy of `properties` array. 
Regarding "0" -  `properties` array has `0` terminator set as a last element and because `set_properties` is a copy of this array then comparison would fail if `0` will not exist in `get_properties.data()` array

bashbaug: Linking with https://github.com/KhronosGroup/OpenCL-Docs/issues/317, since I think the resolution could affect this test as well.

For now, how about we test the basics:

* The query must execute successfully (we're already checking this).
* If the returned properties list is not empty, it must be terminated with "0".

We can add more checks pending resolution of the spec issue.

bashbaug: Sorry for the thrash - given the resolution for https://github.com/KhronosGroup/OpenCL-Docs/issues/317 I believe this test needs to be modified one (hopefully!) final time.  Luckily, I think the test can now be simpler:

1. If the _properties_ argument is `NULL` then the queried properties must either have zero size (there are no properties to return), or must consist of a single property value `0` to indicate a list with no properties.  This condition doesn't affect this test since the properties are non-`NULL`, but it could affect other similar tests.

2. Otherwise, when the _properties_ argument is non-`NULL`, the queried properties must exactly match the passed-in properties, in the exact same order and with no additional properties.

Thanks!

gwawiork: Fixed. Just added condition to check if vectors differs. I leave code to indicate where is the problem (property not found our out of order). Please verify if correct I will apply fix to https://github.com/KhronosGroup/OpenCL-CTS/pull/857

bashbaug: Thanks!

I don't have a strong opinion whether we just fail if the properties are different or provide additional information about missing properties or extra properties, so I'll defer to other reviewers about it.  These latest changes look correct to me though.  👍

PR #861, opened 2020-07-29
add a test for clSetContextDestructorCallback

bashbaug:
Adds a new test for the OpenCL 3.0 `clSetContextDestructorCallback` API.  This test is intentionally very similar to the existing test for `clSetMemObjectDestructorCallback`.


bashbaug:
Note, this PR requires changes to the headers (https://github.com/KhronosGroup/OpenCL-Headers/pull/101), the ICD loader (https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/113), and we need to update these CTS CI builds to build an OpenCL 3.0 ICD loader (https://github.com/KhronosGroup/OpenCL-CTS/pull/860), otherwise automated builds will fail.

test_conformance/api/test_context_destructor_callback.cpp
```diff
@@ -0,0 +1,79 @@
...
+}
+
+#ifndef ABS
+#define ABS( x ) ( ( x < 0 ) ? -x : x )
+#endif
+
+int test_context_destructor_callback(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
+{
+    cl_int error;
+    clContextWrapper localContext = clCreateContext(NULL, 1, &deviceID, NULL, NULL, &error );
```
Jeremy-Kemp, line 35: If we don't use the wrapper here, we could remove the assignment of `NULL` to `localContext` later in the test.

test_conformance/api/test_context_destructor_callback.cpp
```diff
@@ -0,0 +1,79 @@
...
+    // Note: since we manually released the context, we need to set it to NULL to prevent a double-release
+    localContext = NULL;
+
+    // At this point, all three callbacks should have already been called
+    int numErrors = 0;
+    for(  i = 0; i < 3; i++ )
+    {
+        // Spin waiting for the release to finish.  If you don't call the context_destructor_callback, you will not
+        // pass the test.
+        while( 0 == callbackOrders[i] )
```
Jeremy-Kemp, line 67: Could we add a "reasonable" timeout here to prevent bugged implementations from hanging?

The events callbacks test has a 10 second tolerance. I realise that the `clSetMemObjectDestructorCallback` test on which this is modelled has no timeout either so perhaps this is a wider question about how we should deal with callbacks never being triggered across all CTS.

Not a major issue but might cause headaches in automation.

bashbaug: I'll see if I can port a similar timeout mechanism from the event callbacks test to this test and the similar `clSetMemObjectDestructorCallback` test.  I agree this would be a nice upgrade.

bashbaug: I've added a 10 second timeout similar to the event callback test.  Since there are three callbacks, it actually ends up being an up-to-30 second timeout, but it still means the test will properly fail and exit if the callbacks are never called.  Thanks for the suggestion!

test_conformance/api/test_context_destructor_callback.cpp
```diff
@@ -0,0 +1,99 @@
...
+    int *userPtr = (int *)userData;
+
+    // ordering of callbacks is guaranteed, meaning we don't need to do atomic
+    // operation here
+    *userPtr = ++sDestructorIndex;
+}
+
+#ifndef ABS
+#define ABS(x) ((x < 0) ? -x : x)
+#endif
```
kpet, line 31: `std::abs`?

test_conformance/api/test_context_destructor_callback.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+int test_context_destructor_callback(cl_device_id deviceID, cl_context context,
+                                     cl_command_queue queue, int num_elements)
+{
+    cl_int error;
+    clContextWrapper localContext =
+        clCreateContext(NULL, 1, &deviceID, NULL, NULL, &error);
+    test_error(error, "Unable to create local context");
+
+    int i;
```
kpet, line 41: Maybe better to declare it as part of the for loop below.

test_conformance/api/test_context_destructor_callback.cpp
```diff
@@ -0,0 +1,99 @@
...
+        if (ABS(callbackOrders[i]) != 3 - i)
+        {
+            log_error("\tERROR: Callback %d was called in the wrong order! "
+                      "(Was called order %d, should have been order %d)\n",
+                      i + 1, ABS(callbackOrders[i]), i);
+            numErrors++;
+        }
+    }
+
+    return (numErrors > 0) ? -1 : 0;
```
kpet, line 98: `TEST_PASS` / `TEST_FAIL`?

test_conformance/api/test_context_destructor_callback.cpp
```diff
@@ -0,0 +1,99 @@
...
+    error = clReleaseContext(localContext);
+    test_error(error, "Unable to release local context");
+
+    // Note: since we manually released the context, we need to set it to NULL
+    // to prevent a double-release
+    localContext = NULL;
+
+    // At this point, all three callbacks should have already been called
+    int numErrors = 0;
+    for (i = 0; i < 3; i++)
```
kpet, line 70: The 3 is hard-coded in a few places but nothing too bad I think.

test_conformance/api/test_context_destructor_callback.cpp
bashbaug: Fixed - I just removed ABS since it didn't seem to be required.

test_conformance/api/test_context_destructor_callback.cpp
bashbaug: Done!

test_conformance/api/test_context_destructor_callback.cpp
bashbaug: Done!

test_conformance/api/test_context_destructor_callback.cpp
bashbaug: I've left this as-is.

PR #862, opened 2020-07-30
Test for IL support in unload_platform_compiler

Jeremy-Kemp:
Test assumes all 3.0 devices support IL.

### DIFF
--- a/test_conformance/compiler/test_unload_platform_compiler.cpp
+++ b/test_conformance/compiler/test_unload_platform_compiler.cpp
@@ -262,7 +262,10 @@ public:
                   const cl_device_id device)
         : build_base{ context, device }
     {
-        if (get_device_cl_version(device) >= Version(2, 1))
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_unload_platform_compiler.cpp
```diff
@@ -262,7 +262,10 @@ class build_with_il : public build_base {
                   const cl_device_id device)
         : build_base{ context, device }
     {
-        if (get_device_cl_version(device) >= Version(2, 1))
+        Version version = get_device_cl_version(device);
+        std::string sILVersion = get_device_il_version_string(device);
+        if ((version >= Version(2, 1) && version < Version(3, 0)) ||
+            (version >= Version(3, 0) && sILVersion.length() != 1))
```
kpet, line 268: In theory this should test that SPIR-V is in the list of ILs. In practise this test is more than good enough.

PR #864, opened 2020-07-31
Skip sub_group_dispatch.

Jeremy-Kemp:
If sub groups aren't supported, skip the test.

### DIFF
--- a/test_conformance/api/test_sub_group_dispatch.cpp
+++ b/test_conformance/api/test_sub_group_dispatch.cpp
@@ -89,8 +89,7 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
             return error;
         }
 
-        if (max_num_sub_groups == 0
-            || !is_extension_available(deviceID, "cl_khr_subgroups"))
+        if (max_num_sub_groups == 0)
         {
             return TEST_SKIPPED_ITSELF;
         }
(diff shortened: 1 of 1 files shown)


Jeremy-Kemp:
Just realised this is a partial duplicate of #788

FranklandJack:
LGTM, I've tested this locally and it works. I think this should replace #788 since it is more complete.

Jeremy-Kemp:
Apologies @FranklandJack , it certainly wasn't my intention to re-do your work. Regardless, thank you for the comment.

FranklandJack:
> Apologies @FranklandJack , it certainly wasn't my intention to re-do your work. Regardless, thank you for the comment.

No worries, don't apologize! It's easily done, my PR was incomplete and had been open for over 2 months.

test_conformance/api/test_sub_group_dispatch.cpp
```diff
@@ -75,6 +75,27 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
...
+                                sizeof(max_num_sub_groups), &max_num_sub_groups,
+                                NULL);
+        if (error != CL_SUCCESS)
+        {
+            print_error(error, "Unable to get max num subgroups");
+            return error;
+        }
+
+        if (max_num_sub_groups == 0
+            || !is_extension_available(deviceID, "cl_khr_subgroups"))
```
alycm, line 93: I think it is valid for a 3.0 device to report `CL_DEVICE_MAX_NUM_SUB_GROUPS` as non-zero, but to also *not* report the `cl_khr_subgroups` extension because the extension requires IFP whereas the core feature does not.

So the `|| extension` part can be dropped?

If the test actually does require IFP then either a comment or checking `CL_​DEVICE_​SUB_​GROUP_​INDEPENDENT_​FORWARD_​PROGRESS` would be clearer to me.

Jeremy-Kemp: Good spot, I'll update this and check whether the test needs IFP.

PR #868, opened 2020-08-05
Include Hash of Thread ID in File Name

FranklandJack:
* Data races can occur in the file system due to multiple threads in the thread
pool writing to the same file during offline compilation.

- [x] Include the thread ID in the hash so that the file name is unique
for each thread.

### DIFF
--- a/test_common/harness/kernelHelpers.cpp
+++ b/test_common/harness/kernelHelpers.cpp
@@ -28,6 +28,7 @@
 #include <fstream>
 #include <sstream>
 #include <iomanip>
+#include <thread>
 
 #if defined(_WIN32)
 std::string slash = "\\";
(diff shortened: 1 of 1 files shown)


FranklandJack:
> Doesn't that completely defeat the purpose of the cache? Am I missing something?

Thanks for your feedback. I'm not sure, it means concurrent threads can't reuse the same file, but on subsequent runs of the executable, each thread can reuse its file. Maybe the intention of the cache was to reuse the file among threads, but currently I don't think this is done in a thread safe way since each thread using the same kernel writes into the same file. For example the output looks like:

> Saved kernel source to file: ./test.43b218dd.cl
Saved kernel source to file: ./test.43b218dd.cl
Saved kernel source to file: ./test.43b218dd.cl
Saved kernel source to file: ./test.43b218dd.cl
Saved kernel source to file: ./test.43b218dd.cl
Saved kernel source to file: ./test.43b218dd.cl

kpet:
That would make sense for source files only but this change AFAICT will affect the name of binaries as well which means binaries can only be loaded from the cache by a thread with the same ID as the one who created the entry. We actually have a downstream change that fixes the issue by adding a lock around the offline compilation logic. This is just 3 lines of code and the performance impact was neglibile IIRC. I'm asking the engineer who did that to upstream the change.

FranklandJack:
> That would make sense for source files only but this change AFAICT will affect the name of binaries as well which means binaries can only be loaded from the cache by a thread with the same ID as the one who created the entry. We actually have a downstream change that fixes the issue by adding a lock around the offline compilation logic. This is just 3 lines of code and the performance impact was neglibile IIRC. I'm asking the engineer who did that to upstream the change.

Ah okay thanks, adding a lock around the offline compilation logic sounds like a better solution. Let us know when the PR is open and Codeplay can review it :)

kpet:
@FranklandJack You may want to have a look at #871

FranklandJack:
> @FranklandJack You may want to have a look at #871

@kpet Thanks! We've tested that locallly and it works for us!

FranklandJack:
Closing in favor of #871.

PR #869, opened 2020-08-05
OpenCL 3.0 test CL_QUEUE_PROPERTIES_ARRAY

gwawiork:
- Added new function which helps display `cl_command_queue_properties` property
- Testing different properties sets depending on device support (including NULL properties)

### DIFF
--- a/test_conformance/api/test_queue_array_properties.cpp
+++ b/test_conformance/api/test_queue_array_properties.cpp
@@ -33,73 +33,82 @@ int verify_if_properties_supported(
     cl_uint requested_size)
 {
     int error = CL_SUCCESS;
-    cl_command_queue_properties on_host_properties = 0;
-    cl_device_device_enqueue_capabilities device_enqueue_caps;
-    std::vector<cl_command_queue_properties> all_properties = {
-        CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, CL_QUEUE_PROFILING_ENABLE,
-        CL_QUEUE_ON_DEVICE, CL_QUEUE_ON_DEVICE_DEFAULT
-    };
+    bool on_host_queue = true;
 
-    error =
-        clGetDeviceInfo(deviceID, CL_DEVICE_QUEUE_ON_HOST_PROPERTIES,
-                        sizeof(on_host_properties), &on_host_properties, NULL);
-    test_error(error,
-               "clGetDeviceInfo for CL_DEVICE_QUEUE_ON_HOST_PROPERTIES failed");
-
-    error = clGetDeviceInfo(deviceID, CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES,
-                            sizeof(device_enqueue_caps), &device_enqueue_caps,
-                            NULL);
-    test_error(
-        error,
-        "clGetDeviceInfo for CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES failed");
-    if (!(device_enqueue_caps & CL_DEVICE_QUEUE_SUPPORTED)
-        && requested_size > 0)
+    if ((requested_bitfield & CL_QUEUE_ON_DEVICE)
+        || (requested_bitfield & CL_QUEUE_ON_DEVICE_DEFAULT))
     {
-        log_info("\tCL_DEVICE_QUEUE_SUPPORTED false, skipped\n");
-        return TEST_SKIPPED_ITSELF;
+        on_host_queue = false;
     }
-    else if ((device_enqueue_caps & CL_DEVICE_QUEUE_SUPPORTED)
-             && requested_size > 0)
+    cl_command_queue_properties supported_properties = 0;
+    cl_command_queue_properties all_properties = 0;
+
+    std::vector<cl_command_queue_properties> all_properties_vector{
+        CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, CL_QUEUE_PROFILING_ENABLE
+    };
+    for (auto each_property : all_properties_vector)
     {
-        cl_uint max_packet_size = 0;
-        error =
-            clGetDeviceInfo(deviceID, CL_DEVICE_QUEUE_ON_DEVICE_MAX_SIZE,
-                            sizeof(max_packet_size), &max_packet_size, NULL);
-        test_error(
-            error,
-            "clGetDeviceInfo for CL_DEVICE_QUEUE_ON_DEVICE_MAX_SIZE failed");
-        if (requested_size > max_packet_size)
+        all_properties |= each_property;
+    }
+    cl_command_queue_properties requested_properties =
+        all_properties & requested_bitfield;
+    cl_device_device_enqueue_capabilities device_enqueue_caps;
+
+    if (on_host_queue)
+    {
+        error = clGetDeviceInfo(deviceID, CL_DEVICE_QUEUE_ON_HOST_PROPERTIES,
+                                sizeof(supported_properties),
+                                &supported_properties, NULL);
+        test_error(error,
+                   "clGetDeviceInfo asking for "
+                   "CL_DEVICE_QUEUE_ON_HOST_PROPERTIES failed");
+    }
+    else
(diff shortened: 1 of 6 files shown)


test_conformance/api/test_queue_array_properties.cpp
```diff
@@ -0,0 +1,352 @@
...
+    std::string test_kernel = {
+        "__kernel void vec_cpy(__global int *src, __global int *dst)\n"
+        "{\n"
+        "    int  tid = get_global_id(0);\n"
+        "\n"
+        "    dst[tid] = src[tid];\n"
+        "\n"
+        "}\n"
+    };
+
```
bashbaug, line 295: Note: This is a question/suggestion and not a requirement!

As-written, this test will compile a program and create new buffers for each test case.  Since each test cases uses the same program source and the same size buffers, would it make sense to create them here instead and reuse them for each test case?  I suspect most of the execution time for this test - which admittedly is probably pretty short - will be consumed by compiling kernels currently.

test_conformance/api/test_queue_array_properties.cpp
```diff
@@ -0,0 +1,352 @@
...
+                       num_elements * sizeof(int), NULL, &error);
+    test_error(error, "clCreateBuffer failed.");
+
+    error = clSetKernelArg(kernel, 0, sizeof(streams[0]), &streams[0]);
+    test_error(error, "clSetKernelArg failed.");
+
+    error = clSetKernelArg(kernel, 1, sizeof(streams[1]), &streams[1]);
+    test_error(error, "clSetKernelArg failed.");
+
+    error = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, &num_elements, NULL,
```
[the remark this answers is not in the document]
gwawiork: I am thinking more about option 2

test_conformance/api/test_queue_array_properties.cpp
gwawiork: I think that removing execution part will solve this issue as well

test_conformance/api/test_queue_array_properties.cpp
```diff
@@ -0,0 +1,273 @@
...
+                if ((check_property & CL_QUEUE_ON_DEVICE)
+                    || (check_property & CL_QUEUE_ON_DEVICE_DEFAULT))
+                {
+                    log_info("\t%s supported\n",
+                             GetQueuePropertyName(check_property));
+                }
+            }
+            else
+            {
+                log_info("\t%s not supported, skipped\n",
```
bashbaug, line 81: Can we simplify this logic slightly, by:

1. Getting the `CL_DEVICE_QUEUE_ON_DEVICE_PROPERTIES` in addition to the host queue properties.
2. Checking the `requested_bitfield` against the host queue properties or the device queue properties depending whether this is a host queue or a device queue?

    a. I guess we might need to strip off `CL_QUEUE_ON_DEVICE` and `CL_QUEUE_ON_DEVICE_DEFAULT` if we do this since they don't get included in the query for `CL_DEVICE_QUEUE_ON_DEVICE_PROPERTIES`, but that still seems fairly straightforward.

If we do this then I think we can remove the `all_properties` vector and this entire loop.

I think I convinced myself this test logic is OK for now, but it's a bit hard to follow, and without checking the device queue properties I think it could easily become incorrect in the future.

test_conformance/api/test_queue_array_properties.cpp
```diff
@@ -0,0 +1,273 @@
...
+               "clGetDeviceInfo for CL_DEVICE_QUEUE_ON_HOST_PROPERTIES failed");
+
+    error = clGetDeviceInfo(deviceID, CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES,
+                            sizeof(device_enqueue_caps), &device_enqueue_caps,
+                            NULL);
+    test_error(
+        error,
+        "clGetDeviceInfo for CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES failed");
+    if (!(device_enqueue_caps & CL_DEVICE_QUEUE_SUPPORTED)
+        && requested_size > 0)
```
bashbaug, line 56: Minor: This test has chosen a small enough value for the queue size that it _probably_ will be supported for all devices, but to be sure it could also skip itself if the `requested_size` is greater than `CL_DEVICE_QUEUE_ON_DEVICE_MAX_SIZE`.

test_conformance/api/test_queue_array_properties.cpp
```diff
@@ -0,0 +1,273 @@
...
+                                  set_size, get_properties.data(), NULL);
+    test_error(
+        error,
+        "clGetCommandQueueInfo failed asking for CL_QUEUE_PROPERTIES_ARRAY");
+
+    if (get_properties.size() == 1 && get_properties[0] == 0
+        && test_case.properties.size() == 0)
+    {
+        return TEST_PASS;
+    }
```
bashbaug, line 134: This case can be removed now, since we're requiring that if the incoming properties were `NULL` (for this test, indicated by the size of the properties vector being zero) that the query size return zero.

See: https://github.com/KhronosGroup/OpenCL-Docs/pull/398#issuecomment-672211863, #880.

test_conformance/api/test_queue_array_properties.cpp
gwawiork: I hope this comment is solved in commit https://github.com/KhronosGroup/OpenCL-CTS/pull/869/commits/164f547113d1da99e1672927cf96f44300e11ae2

bashbaug: Yup, looks good, thanks!

bashbaug:
All of my comments are addressed - thanks!  Needs a +1 from another member company.

test_conformance/api/test_queue_array_properties.cpp
```diff
@@ -0,0 +1,294 @@
...
+    int error = CL_SUCCESS;
+    clCommandQueueWrapper queue;
+    std::vector<cl_queue_properties> requested_properties =
+        test_case.properties;
+    log_info("TC description: %s\n", test_case.description.c_str());
+    // first verify if user properties are supported
+    if (requested_properties.size() != 0)
+    {
+        requested_properties.pop_back();
+        cl_command_queue_properties requested_bitfield;
```
EwanC, line 211: should we initialize this to zero in case there's a `CL_QUEUE_SIZE` property but no `CL_QUEUE_PROPERTIES`? could add a test case for that too

test_conformance/api/test_queue_array_properties.cpp
```diff
@@ -0,0 +1,294 @@
...
+                | CL_QUEUE_ON_DEVICE_DEFAULT | CL_QUEUE_PROFILING_ENABLE,
+            0 },
+          "queue, all without CL_QUEUE_SIZE" });
+
+    for (auto test_case : test_cases)
+    {
+        error |= run_test_queue_array_properties(context, deviceID, test_case);
+    }
+    return error;
+}
```
EwanC, line 294: nitpick: newline here

test_conformance/api/test_queue_array_properties.cpp
```diff
@@ -0,0 +1,294 @@
...
+    return TEST_PASS;
+}
+
+int test_queue_array_properties(cl_device_id deviceID, cl_context context,
+                                cl_command_queue queue, int num_elements)
+{
+    int error = CL_SUCCESS;
+    std::vector<test_queue_array_properties_data> test_cases;
+
+    test_cases.push_back({ {}, "queue, NULL properties" });
```
EwanC, line 251: Should we test empty` CL_QUEUE_PROPERTIES`? e.g
```c
 test_cases.push_back( { { CL_QUEUE_PROPERTIES, 0, 0 }, "queue, empty properties" });
```

bashbaug:
I've rebased this PR and addressed the review comments (and fixed a few other bugs I found) in #925.  I'll leave this PR open for reference, but I think we can close it when #925 is merged.

kpet:
#925 now merged, closing.

PR #873, opened 2020-08-06
Specify GCC flag `-frounding-math` on x86

EwanC:
We have been seeing fails in `test_conversions` on x86_64 when converting to the following integer types from `cl_double` with a non-default rounding mode:

```
 char_rtn_double
 char_rtp_double
 int_rtn_double
 int_rtp_double
 long_rtn_double
 long_rtp_double
 long_sat_rtn_double
 long_sat_rtp_double
 short_rtn_double
 short_rtp_double
 uchar_rtp_double
 uint_rtp_double
 ulong_rtp_double
 ushort_rtp_double
```

After investigation it was discovered that `rint()` was incorrectly rounding `cl_double` inputs despite the rounding mode being correctly set using `fesetround()` beforehand. E.g  For 'char_rtn_double' `-3.5` was getting rounding to `-3.0` rather than `-4.0`.

This is a gcc issue https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92164 when using the builtin `rint()` implementation, rather than `std::rint()`, that only presents itself when compiling a Release build with `-O2` rather than Debug.

Adding the compiler flag `-fno-builtin-rint` to the CMake works around this problem, however based on the discussion in the gcc ticket using `-frounding-math` appears to be a more comprehensive fix. As `-frounding-math` tells gcc that the code will be modifying the rounding mode, removing the assumption that the same rounding mode is in effect everywhere.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,8 @@ if(CMAKE_COMPILER_IS_GNUCC OR "${CMAKE_CXX_COMPILER_ID}" MATCHES "(Apple)?Clang"
             STREQUAL "x86")
         set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse -msse2 -mfpmath=sse")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse -msse2 -mfpmath=sse")
+
+        add_cxx_flag_if_supported(-frounding-math)
     endif()
 else()
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /D__SSE__")
--- a/test_common/harness/rounding_mode.h
+++ b/test_common/harness/rounding_mode.h
@@ -16,6 +16,8 @@
 #ifndef __ROUNDING_MODE_H__
 #define __ROUNDING_MODE_H__
 
+#pragma STDC FENV_ACCESS ON
+
 #include "compat.h"
 
 #if (defined(_WIN32) && defined (_MSC_VER))
--- a/test_conformance/conversions/test_conversions.cpp
+++ b/test_conformance/conversions/test_conversions.cpp
@@ -50,8 +50,6 @@
 #include "Sleep.h"
 #include "basic_test_conversions.h"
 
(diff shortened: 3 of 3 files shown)


kbenzie:
Hi @Jeremy-Kemp, @EwanC is on holiday this week and has asked I keep track of this in his place.

> The bug also mentions that `FENV_ACCESS ON` is required in conjunction with `-frounding-math`.
> 
> It appears this is only set in conversions but presumably we should make this more widespread to cover all cases where `fesetround` is called?

Seems the pragma is not supported by GCC (`-frounding-mode` is equivalent) but is an [MSVC](https://docs.microsoft.com/en-us/cpp/preprocessor/fenv-access?view=vs-2019) feature.

Looking around, basic, brute force, contractions, and printf call `set_round()` from the test harness in addition to conversions. I'm wondering if it would be sufficient to set the pragma only in the test harness, or if it should be set in each TU where `set_round()` is called?

The examples all seem to be in the same TU so that's probably the safest bet.

kbenzie:
I've added setting the pragma to the `rounding_mode.h` header, I think this should be sufficient to ensure it's set in all relevant TUs.

gwawiork:
Let me check if these changes do not brake anything. I will run some tests on x86

gwawiork:
All test passed then LGTM

kbenzie:
That's good to hear @gwawiork, thanks. Would you be able to approve this PR so it can eventually be merged?

gwawiork:
I have no access to merge, approved

Jeremy-Kemp:
I would like to conduct some testing with this, as we have RTZ devices which have exposed issues in the past with CTS.

I'll report back once I've had a chance to do so.

kbenzie:
Have you had a chance to run those tests @Jeremy-Kemp?

Jeremy-Kemp:
Not yet, but I'll kick a run today.

EwanC:
> `#pragma STDC FENV_ACCESS ON` could also be removed from `test_conversions.cpp`

Good spot @Jeremy-Kemp, pushed a commit with this change if you could review again to confirm it's what you intended.

PR #874, opened 2020-08-07
skip test cases rather than fail without cl_khr_3d_image_writes

### LINKED ISSUE #872: Lack of cl_khr_3d_image_writes extension causes failure in test_image_streams for CL1.2
In `test_image_set` in test_loops.cpp, there's this code:
```c++
    if ( ( 0 == is_extension_available( device, "cl_khr_3d_image_writes" )) && (imageType == CL_MEM_OBJECT_IMAGE3D) && (formatTestFn == test_write_image_formats) )
    {
        gFailCount++;
        log_error( "-----------------------------------------------------\n" );
        log_error( "FAILED: test writing CL_MEM_OBJECT_IMAGE3D images\n" );
        log_error( "This device does not support the mandated extension cl_khr_3d_image_writes.\n");
        log_error( "-----------------------------------------------------\n\n" );
        return -1;
    }
```

For OpenCL1.2, this extension is optional (https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_EXTENSIONS): 

>Additionally, the following Khronos extension names must be returned by all devices that support OpenCL 2.0, OpenCL 2.1, or OpenCL 2.2. Note that these Khronos extension names are not required for devices supporting OpenCL 3.0:
>
>cl_khr_3d_image_writes
>...

bashbaug:
Fixes #872 

Currently the image streams test will fail if a device does not support `cl_khr_3d_image_writes`, but this extension is only required for OpenCL 2.x devices and is not required for OpenCL 1.x or OpenCL 3.0 devices.  So, rather than failing the test if 3D image writes are unsupported, we can simply skip these test cases instead.

Note that there is already a check that `cl_khr_3d_image_writes` is supported for OpenCL 2.x devices in `test_min_max_device_version`, so we don't need to add any similar checks here.

### DIFF
--- a/test_conformance/images/kernel_read_write/test_loops.cpp
+++ b/test_conformance/images/kernel_read_write/test_loops.cpp
@@ -186,14 +186,15 @@ int test_image_set( cl_device_id device, cl_context context, cl_command_queue qu
     static int printedFormatList = -1;
 
 
-    if ( ( 0 == is_extension_available( device, "cl_khr_3d_image_writes" )) && (imageType == CL_MEM_OBJECT_IMAGE3D) && (formatTestFn == test_write_image_formats) )
+    if ( (imageType == CL_MEM_OBJECT_IMAGE3D) && (formatTestFn == test_write_image_formats) )
     {
-        gFailCount++;
-        log_error( "-----------------------------------------------------\n" );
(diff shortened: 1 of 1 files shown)


kpet:
Two approvals from two companies, very simple change, merging.

PR #875, opened 2020-08-11
API Feature Consistency Test, Part 1

bashbaug:
This is the start of an "API Feature Consistency" test.  It is designed to ensure that when an optional OpenCL 3.0 API feature is not supported that other queries and APIs related to the feature can be called even if they return an error, and that when an optional OpenCL 3.0 API feature is supported, any other required features are also supported.

I've modeled this test very heavily on [Appendix H](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#opencl-3.0-backwards-compatibility) in the OpenCL 3.0 spec.  So far I have added API consistency tests for SVM, the memory consistency model, device-side enqueue, and pipes.

I am mostly looking for high-level feedback on the overall approach before adding additional API feature consistency checks.  I think most of the remaining checks are shorter, so if this looks good, I'll get everything else added in Part 2.

One note: I haven't checked any OpenCL C features in these tests, since I'm assuming this will be covered by a different test (#762?).  I think that's the right thing to do, but if I should add checks for OpenCL C features here also, please do let me know.  Thanks!

### DIFF
--- a/test_conformance/api/test_api_consistency.cpp
+++ b/test_conformance/api/test_api_consistency.cpp
@@ -1,0 +1,92 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "testBase.h"
+#include "harness/testHarness.h"
+
+const char* test_kernel = R"CLC(
+__kernel void test(__global int* dst) {
+    dst[0] = 0;
+}
+)CLC";
+
+int test_consistency_svm(cl_device_id deviceID, cl_context context,
+                         cl_command_queue queue, int num_elements)
+{
+    // clGetDeviceInfo, passing CL_​DEVICE_​SVM_​CAPABILITIES:
+    // May return 0, indicating that device does not support Shared Virtual
+    // Memory.
+    int error;
+
+    const size_t allocSize = 16;
+    clMemWrapper mem;
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+
+    cl_device_svm_capabilities svmCaps = 0;
+    error = clGetDeviceInfo(deviceID, CL_DEVICE_SVM_CAPABILITIES,
+                            sizeof(svmCaps), &svmCaps, NULL);
+    test_error(error, "Unable to query CL_DEVICE_SVM_CAPABILITIES");
+
+    if (svmCaps == 0)
+    {
+        mem =
+            clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, NULL, &error);
+        test_error(error, "Unable to create test buffer");
+
+        error = create_single_kernel_helper(context, &program, &kernel, 1,
+                                            &test_kernel, "test");
+        test_error(error, "Unable to create test kernel");
+
+        // clGetMemObjectInfo, passing CL_MEM_USES_SVM_POINTER
+        // Returns CL_​FALSE if no devices in the context associated with
+        // memobj support Shared Virtual Memory.
+        cl_bool usesSVMPointer;
+        error =
+            clGetMemObjectInfo(mem, CL_MEM_USES_SVM_POINTER,
+                               sizeof(usesSVMPointer), &usesSVMPointer, NULL);
+        test_error(error,
+                   "Unable to query CL_​MEM_​USES_​SVM_​POINTER");
+
+        // Check that the SVM APIs can be called.
+        // It's OK if they return an error.
+        void* ptr0;
+        ptr0 = clSVMAlloc(context, CL_MEM_READ_WRITE, allocSize, 0);
+
+        void* ptr1;
+        ptr1 = clSVMAlloc(context, CL_MEM_READ_WRITE, allocSize, 0);
+
+        cl_uint pattern = 0xAAAAAAAA;
+        clEnqueueSVMMemFill(queue, ptr0, &pattern, sizeof(pattern), allocSize,
+                            0, NULL, NULL);
+        clEnqueueSVMMemcpy(queue, CL_TRUE, ptr1, ptr0, allocSize, 0, NULL,
+                           NULL);
+
+        clEnqueueSVMMap(queue, CL_TRUE, CL_MAP_READ, ptr1, allocSize, 0, NULL,
+                        NULL);
+        clEnqueueSVMUnmap(queue, ptr1, 0, NULL, NULL);
+
+        error = clFinish(queue);
+        test_error(error, "Error calling clFinish after SVM operations");
+
+        clSVMFree(context, ptr0);
+        clEnqueueSVMFree(queue, 1, &ptr1, NULL, NULL, 0, NULL, NULL);
+
+        error = clFinish(queue);
+        test_error(error, "Error calling clFinish after SVM free");
+    }
+
+    return TEST_PASS;
+}
--- a/test_conformance/api/test_api_consistency.cpp
+++ b/test_conformance/api/test_api_consistency.cpp
@@ -121,7 +121,8 @@ int test_consistency_svm(cl_device_id deviceID, cl_context context,
         // supported.
         cl_uint bogus = 0xDEADBEEF;
         clSVMFree(context, &bogus);
-        error = clEnqueueSVMFree(queue, 1, &bogus, NULL, NULL, 0, NULL, NULL);
+        void* pBogus = &bogus;
+        error = clEnqueueSVMFree(queue, 1, &pBogus, NULL, NULL, 0, NULL, NULL);
         test_failure_error(
             error, CL_INVALID_OPERATION,
             "CL_DEVICE_SVM_CAPABILITIES returned 0 but "
--- a/test_conformance/api/test_api_consistency.cpp
+++ b/test_conformance/api/test_api_consistency.cpp
@@ -42,6 +42,8 @@ int test_consistency_svm(cl_device_id deviceID, cl_context context,
 
     if (svmCaps == 0)
     {
+        // Test setup:
+
         mem =
             clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, NULL, &error);
         test_error(error, "Unable to create test buffer");
@@ -110,8 +112,9 @@ int test_consistency_svm(cl_device_id deviceID, cl_context context,
             "CL_DEVICE_SVM_CAPABILITIES returned 0 but "
             "clEnqueueSVMUnmap did not return CL_INVALID_OPERATION");
 
-        error = clFinish(queue);
-        test_error(error, "Error calling clFinish after SVM operations");
+        // If the enqueue calls above did not return errors, a clFinish would be
+        // needed here to ensure the SVM operations are complete before freeing
+        // the SVM pointers.
 
         clSVMFree(context, ptr0);
         error = clEnqueueSVMFree(queue, 1, &ptr1, NULL, NULL, 0, NULL, NULL);
@@ -120,8 +123,8 @@ int test_consistency_svm(cl_device_id deviceID, cl_context context,
             "CL_DEVICE_SVM_CAPABILITIES returned 0 but "
             "clEnqueueSVMFree did not return CL_INVALID_OPERATION");
 
-        error = clFinish(queue);
-        test_error(error, "Error calling clFinish after SVM free");
+        // If the enqueue calls above did not return errors, a clFinish should
+        // be included here to ensure the enqueued SVM free is complete.
 
         // clSetKernelArgSVMPointer, clSetKernelExecInfo
         // Returns CL_INVALID_OPERATION if no devices in the context associated
--- a/test_conformance/api/test_api_consistency.cpp
+++ b/test_conformance/api/test_api_consistency.cpp
@@ -58,6 +58,12 @@ int test_consistency_svm(cl_device_id deviceID, cl_context context,
             clGetMemObjectInfo(mem, CL_MEM_USES_SVM_POINTER,
                                sizeof(usesSVMPointer), &usesSVMPointer, NULL);
         test_error(error, "Unable to query CL_MEM_USES_SVM_POINTER");
+        if (usesSVMPointer != CL_FALSE)
+        {
+            log_error("CL_DEVICE_SVM_CAPABILITIES returned 0 but "
+                      "CL_MEM_USES_SVM_POINTER did not return CL_FALSE\n");
+            return TEST_FAIL;
+        }
 
         // Check that the SVM APIs can be called.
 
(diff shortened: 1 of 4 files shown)


jrprice:
Discussion from August 11th telecon:
We would like to see negative testing here to ensure that APIs and queries that relate to an unsupported feature actually do return the error/value that the spec describes, instead of just allowing them to.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,501 @@
...
+        cl_uint pattern = 0xAAAAAAAA;
+        error = clEnqueueSVMMemFill(queue, ptr0, &pattern, sizeof(pattern),
+                                    allocSize, 0, NULL, NULL);
+        if (error != CL_INVALID_OPERATION)
+        {
+            log_error(
+                "CL_DEVICE_SVM_CAPABILITIES returned 0 but "
+                "clEnqueueSVMMemFill did not return CL_INVALID_OPERATION\n");
+            return TEST_FAIL;
+        }
```
jrprice, line 88: Can make these checks a little more concise by using `test_failure_error`, e.g.:

```
test_failure_error(error, CL_INVALID_OPERATION,
                   "CL_DEVICE_SVM_CAPABILITIES returned 0 but "
                   "clEnqueueSVMMemFill did not return CL_INVALID_OPERATION\n");
```

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,501 @@
...
+        test_error(error, "Unable to create test kernel");
+
+        // clGetMemObjectInfo, passing CL_MEM_USES_SVM_POINTER
+        // Returns CL_FALSE if no devices in the context associated with
+        // memobj support Shared Virtual Memory.
+        cl_bool usesSVMPointer;
+        error =
+            clGetMemObjectInfo(mem, CL_MEM_USES_SVM_POINTER,
+                               sizeof(usesSVMPointer), &usesSVMPointer, NULL);
+        test_error(error, "Unable to query CL_MEM_USES_SVM_POINTER");
```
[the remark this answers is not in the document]
bashbaug: Yes definitely, I missed this one.  Added now.

test_conformance/api/test_api_consistency.cpp
bashbaug: Cool, I was looking for something like this but didn't find it.  I've switched all of these checks to use `test_failure_error`.

It would be nice to have a similar function when checking for a specific value (but not a specific error code).  I looked for this one but I didn't find it.  Does it exist?  I'm thinking of something like `test_value` that could be used to simplify e.g.:

```c
if (u != 0)
{
    log_error(
        "CL_DEVICE_PIPE_SUPPORT returned CL_FALSE, but "
        "CL_DEVICE_PIPE_MAX_PACKET_SIZE returned a non-zero value\n");
    return TEST_FAIL;
}
```

to:

```c
test_value(u, 0,
           "CL_DEVICE_PIPE_SUPPORT returned CL_FALSE, but "
           "CL_DEVICE_PIPE_MAX_PACKET_SIZE returned a non-zero value");
```

bashbaug:
I think I've added all of the negative testing we discussed in the August 11th teleconference.  Thanks!

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,478 @@
...
+        log_error("Atomic capabilities %llx is missing support for at least "
+                  "one required capability %llx!\n",
+                  atomicCaps, requiredCaps);
+        return TEST_FAIL;
+    }
+
+    if ((atomicCaps & CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES) != 0
+        && (atomicCaps & CL_DEVICE_ATOMIC_SCOPE_DEVICE) == 0)
+    {
+        log_info("Check: ATOMIC_SCOPE_ALL_DEVICES is supported, but "
```
gwawiork, line 161: The intention was only print info, no error ?

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,478 @@
...
+                 "ATOMIC_SCOPE_DEVICE is not?\n");
+    }
+
+    if ((atomicCaps & CL_DEVICE_ATOMIC_SCOPE_DEVICE) != 0
+        && (atomicCaps & CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP) == 0)
+    {
+        log_info("Check: ATOMIC_SCOPE_DEVICE is supported, but "
```
gwawiork, line 168: The intention was only print info, no error ?

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,478 @@
...
+        {
+            log_error(
+                "CL_DEVICE_PIPE_SUPPORT returned CL_FALSE, but "
+                "CL_DEVICE_PIPE_MAX_PACKET_SIZE returned a non-zero value\n");
+            return TEST_FAIL;
+        }
+
+        // clCreatePipe
+        // Returns CL_INVALID_OPERATION if no devices in context support Pipes.
+        clMemWrapper mem = clCreatePipe(context, 0, 0, 0, NULL, &error);
```
gwawiork, line 444: pie_packet_size/pipe_max_packets cannot be 0 while creating pipe to get CL_INVALID_OPERATION

test_conformance/api/test_api_consistency.cpp
bashbaug: Yes, this was intentional - it's not clear to me that this is strictly forbidden by the spec, though it is a little strange so I thought it was worth adding to the log.

test_conformance/api/test_api_consistency.cpp
bashbaug: Same as above:

Yes, this was intentional - it's not clear to me that this is strictly forbidden by the spec, though it is a little strange so I thought it was worth adding to the log.

test_conformance/api/test_api_consistency.cpp
bashbaug: Good point - I don't think we've ever specified which error gets returned if two error conditions occur.

I've switched this call to use pipe parameters that would be valid if pipes were supported:

```c
clCreatePipe(context, CL_MEM_HOST_NO_ACCESS, 4, 4, NULL, &error);
```

Note that two error conditions still occur if pipes are not supported, and specifically this call could theoretically return `CL_INVALID_OPERATION` (since no devices in context support Pipes) or `CL_INVALID_PIPE_SIZE` (since the pipe_packet_size exceeds `CL_​DEVICE_​PIPE_​MAX_​PACKET_​SIZE`, which must be zero), but I still think we should keep this check as-is to encourage more uniform behavior.

test_conformance/api/test_api_consistency.cpp
jrprice: I agree, I think it would be useful to add a function that can test arbitrary expressions like this, and I don't believe it exists. I'd suggest something that takes a boolean expression as the first argument to make it usable in more cases, e.g.:
```
// Or test_assert() ?
test_value(ptr0 != NULL || ptr1 != NULL,
           "CL_DEVICE_SVM_CAPABILITIES returned 0 but clSVMAlloc "
           "returned a non-NULL value");
```

It should be a pretty straightforward addition to `errorHelpers.h`. I don't think this change would have to hold up this PR though.

bashbaug: Sounds good, I'll take a look at this for "Part 2".

(I like the idea of switching to a boolean expression.  Maybe `test_condition` instead of `test_value` if we do this though?  I'd be fine with `test_assert` also.)

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+    if (svmCaps == 0)
+    {
+        mem =
+            clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, NULL, &error);
+        test_error(error, "Unable to create test buffer");
+
+        error = create_single_kernel_helper(context, &program, &kernel, 1,
+                                            &test_kernel, "test");
+        test_error(error, "Unable to create test kernel");
```
kpet, line 51: Suggest moving the kernel creation code lower down, right before the kernel is used.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,479 @@
...
+            log_error("CL_DEVICE_SVM_CAPABILITIES returned 0 but "
+                      "CL_MEM_USES_SVM_POINTER did not return CL_FALSE\n");
+            return TEST_FAIL;
+        }
+
+        // Check that the SVM APIs can be called.
+
+        // Returns NULL if no devices in context support Shared Virtual Memory.
+        void* ptr0 = clSVMAlloc(context, CL_MEM_READ_WRITE, allocSize, 0);
+        void* ptr1 = clSVMAlloc(context, CL_MEM_READ_WRITE, allocSize, 0);
```
kpet, line 72: Why two calls? Because two is better than one?

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,479 @@
...
+            "clEnqueueSVMMap did not return CL_INVALID_OPERATION");
+
+        error = clEnqueueSVMUnmap(queue, ptr1, 0, NULL, NULL);
+        test_failure_error(
+            error, CL_INVALID_OPERATION,
+            "CL_DEVICE_SVM_CAPABILITIES returned 0 but "
+            "clEnqueueSVMUnmap did not return CL_INVALID_OPERATION");
+
+        error = clFinish(queue);
+        test_error(error, "Error calling clFinish after SVM operations");
```
kpet, line 114: Why call `clFinish` here? The test will have bailed out if any enqueue command succeeded.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,479 @@
...
+
+        clSVMFree(context, ptr0);
```
kpet, line 116: This is only testing the case where `ptr0` is `NULL` which is guaranteed to be a no-op even when SVM is supported. Better to assign a value to `ptr0` beforehand.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,479 @@
...
+        error = clEnqueueSVMFree(queue, 1, &ptr1, NULL, NULL, 0, NULL, NULL);
+        test_failure_error(
+            error, CL_INVALID_OPERATION,
+            "CL_DEVICE_SVM_CAPABILITIES returned 0 but "
+            "clEnqueueSVMFree did not return CL_INVALID_OPERATION");
+
+        error = clFinish(queue);
+        test_error(error, "Error calling clFinish after SVM free");
```
kpet, line 124: Same comment as above.

test_conformance/api/test_api_consistency.cpp
kpet: Hmm, we should probably create a spec issue for this.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -0,0 +1,479 @@
...
+        // clGetCommandQueueInfo, passing CL_QUEUE_SIZE or
+        // CL_QUEUE_DEVICE_DEFAULT
+        // Returns 0 or NULL if the device associated with command_queue does
+        // not support On-Device Queues.
+
+        error =
+            clGetCommandQueueInfo(queue, CL_QUEUE_SIZE, sizeof(u), &u, NULL);
+        // TODO: is this a valid query?  See:
+        // https://github.com/KhronosGroup/OpenCL-Docs/issues/402
+        // test_error(error, "Unable to query CL_QUEUE_SIZE");
```
kpet, line 308: I think we probably want this to fail. I've commented on the spec issue.

test_conformance/api/test_api_consistency.cpp
bashbaug: I wanted two pointers to for two reasons: first, to have one SVM pointer to "fill" and a different SVM pointer to "copy" to, and second, to have one SVM pointer to free via `clSVMFree` and another SVM pointer to free via `clEnqueueSVMFree`.

Since we're enforcing errors from these functions I suppose this isn't strictly necessary and we can pass bogus arguments to each of these APIs, but is still feels like good hygiene to have a reasonably valid code sequence.

test_conformance/api/test_api_consistency.cpp
bashbaug: Any specific reason why?  I find the test logic easier to follow when this "setup" code is separated.

test_conformance/api/test_api_consistency.cpp
bashbaug: Yes, I suppose this isn't required since we're checking and enforcing that all of the preceding commands return errors, but it would be required otherwise to ensure the SVM operations are complete before freeing the pointers.  Since it's not hurting anything I'm inclined to keep it, but I don't care strongly either way.

test_conformance/api/test_api_consistency.cpp
bashbaug: Well, we cannot allocate a valid SVM pointer when SVM isn't supported, so I'm not sure which value would be better. :smile:

We also have no way of knowing whether `clSVMFree` was a NOP or even if it executed successfully or returned an error, since it has no error code.  I think the best we can do is pass the value of `ptr0` that we have, which will be `NULL`, and be sure the call does not crash.

test_conformance/api/test_api_consistency.cpp
kpet: That's another way of looking at it though the setup code is not explicitly identified here. My reasoning was along a similar line as "declare as close to usage as possible". I guess it's partly down to personal preference. I won't lose sleep over this :).

test_conformance/api/test_api_consistency.cpp
kpet: Ok, I understand the reasoning now. Fine as is.

test_conformance/api/test_api_consistency.cpp
kpet: I think this is confusing and we should remove it. We don't want the test to accidentally work on a broken implementation because of this.

test_conformance/api/test_api_consistency.cpp
kpet: My point was that it's guaranteed to be a no-op whether SVM is supported or not. This is not testing that clSVMFree always is a no-op (not that we can do better than "it doesn't crash" without a lot of effort). I'd just make up a value here (DEADBEEF or similar to make debugging more obvious if this ever triggers a crash).

test_conformance/api/test_api_consistency.cpp
bashbaug: I've created https://github.com/KhronosGroup/OpenCL-Docs/issues/404 to track the spec question.  Suggest that we leave this as-is for now and I will update this check pending resolution of the spec issue.

kpet: Works for me.

test_conformance/api/test_api_consistency.cpp
bashbaug: Sounds good.  Do you want to wait to merge these changes until the spec issue is resolved?

kpet: Not necessarily but we should create a new issue if we're merging this PR ahead of resolving the spec issue so we don't lose track of this.

test_conformance/api/test_api_consistency.cpp
bashbaug: I'll replace these calls with comments instead.

test_conformance/api/test_api_consistency.cpp
bashbaug: I'm hesitant to make this change since it's not clear to me that passing a bogus pointer to `clSVMFree` is any better than passing a `NULL` pointer.  How strongly do you feel about this?

kpet: Well, passing `NULL` is specified to behave the same whether the feature is supported or not. There are bugs that will not be exposed by passing `NULL` but _may_ be exposed by passing a bogus value. I'd say this is unambiguously better.

bashbaug: I'll just change this (and the call to `clEnqueueSVMFree` below) since it's a trivial change, but I will note that the current description for `clSVMFree` says (**emphasis** mine):

> **svm_pointer must be the value returned by a call to clSVMAlloc**. If a NULL pointer is passed in svm_pointer, no action occurs.

It doesn't say what happens if _svm_pointer_ is NOT a value returned by a call to clSVMAlloc, other than the NULL pointer case, but I suspect that many implementations are defensive and will validate that the input is a known SVM pointer before freeing it.  For these implementations, passing a bogus pointer will behave the same as passing a NULL pointer.

As an aide, it would be really nice to describe the expected behavior for `clSVMFree` when passed a bogus pointer in the case when SVM is supported - is this a NOP, undefined behavior, implementation-defined behavior, or something else?

test_conformance/api/test_api_consistency.cpp
```diff
@@ -116,8 +116,13 @@ int test_consistency_svm(cl_device_id deviceID, cl_context context,
...
         // the SVM pointers.
 
-        clSVMFree(context, ptr0);
-        error = clEnqueueSVMFree(queue, 1, &ptr1, NULL, NULL, 0, NULL, NULL);
+        // These calls to free SVM purposefully passes a bogus pointer to the
+        // free function to better test that it they are a NOP when SVM is not
+        // supported.
+        cl_uint bogus = 0xDEADBEEF;
+        clSVMFree(context, &bogus);
+        void* pBogus = &bogus;
```
kpet, line 124: Note that this will use a random stack address as the bogus pointer and not the pattern.

test_conformance/api/test_api_consistency.cpp
kpet: Yes but it doesn't really matter here since this code will only run when SVM is not supported and in that case the implementation should definitely not do anything with the pointer. That's what we're trying to test.

>As an aide, it would be really nice to describe the expected behavior for clSVMFree when passed a bogus pointer in the case when SVM is supported - is this a NOP, undefined behavior, implementation-defined behavior, or something else?

Agree it's worth specifying. I'd lean towards undefined or implementation-defined. This could be checked by a validation layer and would enable implementations to assume the pointer is correct.

test_conformance/api/test_api_consistency.cpp
bashbaug: Yeah, that was intentional: assigning a known bit pattern generates and MSVC error:

`warning C4312: 'type cast': conversion from 'unsigned int' to 'void *' of greater size`

Passing the stack pointer is a bit too clever though, so I've switched this to assign the known bit pattern instead.  I'll just live with the warning for now - there's no warning on with gcc.

(At some point I'd like to tidy up the MSVC warnings, but not now.)

kpet (approved):
LGTM, all comments addressed and other reviews were dismissed after an almost trivial rebase. Merging.

PR #876, opened 2020-08-11
Dynamically select `-cl-std` Build Option for `get_(global|local)_lin…

FranklandJack:
…ear_id`

* Dynamically select the build option `-cl-std` for the
`get_(global|local)_linear_id()` builtin based on the device version
of the driver. Either `CL2.0` or `CL3.0` depending on the driver version,
there is no need to consider earlier OpenCLC versions since this
builtin didn't exist before OpenCLC 2.0.

### DIFF
--- a/test_conformance/basic/test_get_linear_ids.cpp
+++ b/test_conformance/basic/test_get_linear_ids.cpp
@@ -59,7 +59,8 @@ test_get_linear_ids(cl_device_id device, cl_context context, cl_command_queue qu
 
 
     // Create the kernel
-    error = create_single_kernel_helper_with_build_options(context, &program, &kernel, 1, linear_ids_source, "test_linear_ids", "-cl-std=CL2.0");
+    error = create_single_kernel_helper(context, &program, &kernel, 1,
+                                        linear_ids_source, "test_linear_ids");
     if (error)
         return error;
 
--- a/test_conformance/basic/test_global_linear_id.cpp
+++ b/test_conformance/basic/test_global_linear_id.cpp
@@ -80,9 +80,13 @@ test_global_linear_id(cl_device_id device, cl_context context, cl_command_queue 
     streams = clCreateBuffer(context, (cl_mem_flags)(CL_MEM_READ_WRITE), length*sizeof(int), NULL, &err);
     test_error( err, "clCreateBuffer failed.");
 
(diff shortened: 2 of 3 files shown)


test_conformance/basic/test_get_linear_ids.cpp
```diff
@@ -59,7 +59,11 @@ test_get_linear_ids(cl_device_id device, cl_context context, cl_command_queue qu
 
 
     // Create the kernel
-    error = create_single_kernel_helper_with_build_options(context, &program, &kernel, 1, linear_ids_source, "test_linear_ids", "-cl-std=CL2.0");
+    std::string cl_std = "-cl-std=CL";
+    cl_std += (get_device_cl_version(device) == Version(3, 0)) ? "3.0" : "2.0";
```
alycm, line 63: This could be made less fragile by using `>= Version(3, 0)`, i.e. for whenever some future version of OpenCL arrives.

If this sort of `-cl-std=CL2.0` test going to need fixed in more places or if this is it?   If there will be a lot maybe some kind of helper function that returns the highest supported OpenCL C version would be useful.  E.g. CL2.0 for 2.x devices, and the highest version in `CL_​DEVICE_​OPENCL_​C_​ALL_​VERSIONS` for >= 3.0 devices.

alycm: In fact, the code in https://github.com/KhronosGroup/OpenCL-CTS/pull/879 could be the basis of that helper function?

FranklandJack: @alycm Thanks for the feedback. Yes, this assumption that `-cl-std=CL2.0` will need to be updated in quite a few places so I think having a helper function is a good idea. One thing I considered doing is adding this logic into the kernel helper framework where by it would automatically add `-cl-std` using the latest CL C version supported by the device to the list of compiler options if that option hadn't already been defined in the string, that way the calling code wouldn't have to specify the version each time.

bashbaug:
This PR should be revisited after #879 is merged since I think #879 will simplify it considerably.

FranklandJack:
Since #879 has been merged, this PR has been simplified and is now ready for review.

FranklandJack:
> LGTM!
> 
> I think this is another case where we could use `create_single_kernel_helper` with default NULL build options vs. `create_single_kernel_helper_with_build_options` and explicitly NULL build options, but functionally this code is doing the same thing (albeit with a bit more typing).

Thanks for the feedback @bashbaug is really appreciated :) I have updated the `create_single_kernel_helper_with_build_options` instances to `create_single_kernel_helper`.

PR #877, opened 2020-08-12
Change Behviour of non-uniform-work-group tests for OpenCL-3.0

FranklandJack:
* Pass `-cl-std=CL3.0` as build option for OpenCL-3.0.
* For an OpenCL-3.0 driver that optionally does not support non-uniform
work-groups round up the global size and run the tests with uniform
work-groups. This increases coverage and allows testing of the
`get_enqueued_local_size` builtin for uniform work-groups.

### DIFF
--- a/test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
+++ b/test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
@@ -243,75 +243,134 @@ static const char *KERNEL_FUNCTION =
   NL "}"
   NL ;
 
-TestNonUniformWorkGroup::TestNonUniformWorkGroup (const cl_device_id &device, const cl_context &context,
-  const cl_command_queue &queue, const cl_uint dims, const size_t *globalSize, const size_t *localSize, const size_t *buffersSize,
-  const size_t *globalWorkOffset, const size_t *reqdWorkGroupSize)
-  : _device(device), _context(context), _queue(queue), _dims (dims) {
+TestNonUniformWorkGroup::TestNonUniformWorkGroup(
+    const cl_device_id &device, const cl_context &context,
+    const cl_command_queue &queue, const cl_uint dims, size_t *globalSize,
+    const size_t *localSize, const size_t *buffersSize,
+    const size_t *globalWorkOffset, const size_t *reqdWorkGroupSize)
+    : _device(device), _context(context), _queue(queue), _dims(dims)
+{
 
-  if (globalSize == NULL || dims < 1 || dims > 3) {
-    //throw std::invalid_argument("globalSize is NULL value.");
-    // This is method of informing that parameters are wrong.
-    // It would be checked by prepareDevice() function.
-    // This is used because of lack of exception support.
-    _globalSize[0] = 0;
-    return;
-  }
-
-  cl_uint i;
-  _globalWorkOffset_IsNull = true;
-  _localSize_IsNull = true;
-
-  setGlobalWorkgroupSize(globalSize);
-  setLocalWorkgroupSize(globalSize,localSize);
-  for (i = _dims; i < MAX_DIMS; i++) {
-    _globalSize[i] = 1;
-  }
-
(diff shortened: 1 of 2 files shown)


test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
```diff
@@ -482,7 +539,9 @@ int TestNonUniformWorkGroup::prepareDevice () {
   if(_localSize_IsNull == false)
     calculateExpectedValues();
 
-  std::string buildOptions = BUILD_CL_STD_2_0;
+  std::string buildOptions = (get_device_cl_version(_device) == Version(3, 0))
```
gwawiork, line 542: I am not sure if this is needed. 
@bashbaug do we need to change OpenCL C language version in this and other cases ?

bashbaug: Yes, I believe so.  I didn't do an exhaustive check, but I do see that the kernel string for this test includes the `get_enqueued_local_size()` built-in function, which requires OpenCL C 2.0 or OpenCL C 3.0 and was not in earlier versions of OpenCL C.

Jeremy-Kemp: With #879 , the selection could be completely omitted.

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
```diff
@@ -721,50 +782,60 @@ int TestNonUniformWorkGroup::runKernel () {
...
+    }
 }
 
 int SubTestExecutor::calculateWorkGroupSize(size_t &maxWgSize, int testRange) {
   int err;
 
   clProgramWrapper program;
   clKernelWrapper testKernel;
-  std::string buildOptions = BUILD_CL_STD_2_0;
+  std::string buildOptions = (get_device_cl_version(_device) == Version(3, 0))
```
Jeremy-Kemp, line 836: With #879 , the selection could be completely omitted.

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
```diff
@@ -243,75 +243,134 @@ static const char *KERNEL_FUNCTION =
...
-    _globalSize[i] = 1;
-  }
+    // For OpenCL-3.0 support for non-uniform workgroups is optional, it's still
+    // useful to run these tests since we can verify the behavior of the
+    // get_enqueued_local_size() builtin for uniform workgroups, so we round up
+    // the global size to insure uniform workgroups on those 3.0 devices.
+    // We only need to do this when localSize is non-null, otherwise the driver
+    // will select a value for localSize which will be uniform on devices that
+    // don't support non-uniform work-groups.
+    if (nullptr != localSize && get_device_cl_version(device) == Version(3, 0))
```
Jeremy-Kemp, line 271: I think this should be `>= Version(3, 0)` to ensure that future device versions remain compatible here.

Even if we were to make non-uniform a requirement in future versions of the spec, the `CL_DEVICE_NON_UNIFORM_WORK_GROUP_SUPPORT` query would still need to be supported and would therefore always return `CL_TRUE` in that case.

Jeremy-Kemp (changes requested):

FranklandJack:
Since #879 has been merged, this could do with another review before being merged. The clang format changes required to pass travis are quite drastic and obfuscate the functional changes of this patch, so I'll hold off on applying `clang-format` until this PR has been reviewed.

FranklandJack:
I've clang formatted the patch, this hasn't introduced any functional changes, but the PR might need approving again.

bashbaug:
Whoops, I didn't mean to close this PR.

I'll merge this as soon as the formatting issue is fixed.

FranklandJack:
@bashbaug Looks like the formatting job has passed, needs approving one last time since I pushed new commits, but should be good to merge otherwise :)

PR #879, opened 2020-08-12
Dynamically Select `-cl-std` Option

FranklandJack:
* Dynamically select the `-cl-std` option on the basic `sizeof` and
    `wg_barrier` tests to be one of `CL1.X`, `CL2.0` or `CL3.0`.
    Use the most recent CL C standard supported on the device.

### DIFF
--- a/test_common/harness/kernelHelpers.cpp
+++ b/test_common/harness/kernelHelpers.cpp
@@ -29,6 +29,7 @@
 #include <sstream>
 #include <iomanip>
 #include <mutex>
+#include <algorithm>
 
 #if defined(_WIN32)
 std::string slash = "\\";
@@ -809,6 +810,43 @@ int create_single_kernel_helper(cl_context context,
                                 const char *buildOptions,
                                 const bool openclCXX)
 {
+    // For the logic that automatically adds -cl-std it is much cleaner if the
+    // build options have RAII. This buffer will store the potentially updated
+    // build options, in which case buildOptions will point at the string owned
+    // by this buffer.
+    std::string build_options_internal{ buildOptions ? buildOptions : "" };
+
+    // Check the build options for the -cl-std option.
+    if (!buildOptions || !strstr(buildOptions, "-cl-std"))
+    {
+        // If the build option isn't present add it using the latest OpenCL-C
+        // version supported by the device. This allows calling code to force a
+        // particular CL C version if it is required, but also means that
+        // callers need not specify a version if they want to assume the most
+        // recent CL C.
+
+        auto version = get_max_OpenCL_C_for_context(context);
+
+        std::string cl_std{};
+        if (version >= Version(3, 0))
+        {
+            cl_std = "-cl-std=CL3.0";
+        }
+        else if (version >= Version(2, 0) && version < Version(3, 0))
+        {
+            cl_std = "-cl-std=CL2.0";
+        }
+        else
+        {
+            // If the -cl-std build option is not specified, the highest OpenCL
+            // C 1.x language version supported by each device is used when
+            // compiling the program for each device.
+            cl_std = "";
+        }
+        build_options_internal += ' ';
+        build_options_internal += cl_std;
+        buildOptions = build_options_internal.c_str();
+    }
     int error;
     // Create OpenCL C++ program
     if(openclCXX)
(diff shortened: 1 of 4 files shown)


FranklandJack:
@alycm I've updated this PR with some logic to automatically add the `-cl-std=CLX.Y` option using the most recent supported CL C on the device in the case that it is not explicitly passed as a build option. If you like this approach then we can just update the remaining tests that assume `-cl-std=CL2.0` to work for 3.0 by removing the option from the list of build options.

FranklandJack:
@kpet This PR is the one I mentioned yesterday and has the logic for automatically selecting the latest OpenCL C supported by the device. As you pointed out, this would select the most recent version of OpenCL C even when an older version would have been sufficient, e.g. testing OpenCL C 1.2 languge features would use `-cl-std=CL3.0` on a 3.0 device unless the call to `create_single_kernel_helper_with_build_options` explicitly had `-cl-std=CL1.2` passed in its `buildOptions` parameter. I'm not sure if we want this behavior, or whether we should explicitly pass the latest version of OpenCL C when not testing 1.X language features?

test_common/harness/kernelHelpers.cpp
```diff
@@ -809,6 +811,56 @@ int create_single_kernel_helper(cl_context context,
                                 const char *buildOptions,
                                 const bool openclCXX)
 {
+
+    // Check the build options for the -cl-std option.
+    if (!buildOptions || !strstr(buildOptions, "-cl-std"))
```
Jeremy-Kemp, line 816: Should check that `buildOptions` is not `NULL` before passing to `strstr`.

test_common/harness/kernelHelpers.cpp
```diff
@@ -809,6 +811,56 @@ int create_single_kernel_helper(cl_context context,
...
+                                          / sizeof(cl_device_id));
+        error =
+            clGetContextInfo(context, CL_CONTEXT_DEVICES, devices_size_in_bytes,
+                             devices.data(), nullptr);
+        Version min_version{ (std::numeric_limits<int>::max)(),
+                             (std::numeric_limits<int>::max)() };
+        std::for_each(devices.begin(), devices.end(),
+                      [&min_version](cl_device_id device) {
+                          auto version = get_device_cl_version(device);
+                          min_version = (std::min)(version, min_version);
```
Jeremy-Kemp, line 840: Is the logic not reversed here?

If the intent is to select the latest/newest CL C version, then this loop should effectively be `max_version` instead?

Jeremy-Kemp: Resolving, the lowest common denominator is being chosen.

This raises a different question though which I'll post a new comment for.

Jeremy-Kemp, line 840: If there are two devices, one supporting OpenCL 3.0 and one supporting OpenCL 2.0, then -cl-std=CL2.0 will be chosen.

The 3.0 device might not necessarily support OpenCL C 2.0, meaning that we'd need to drop to OpenCL C 1.2.

FranklandJack: Nice catch, I completely overlooked that. I've updated the PR to use 1.2 in the case that it finds a 2.X device and there is already a 3.0 device, or it finds a 3.0 device and there is already a 2.X device. The logic looks a bit convoluted now but I'm not sure if there is a cleaner way.

FranklandJack (commented):
Thanks for your feedback @Jeremy-Kemp, I've addressed your comments :)

test_common/harness/kernelHelpers.cpp
Jeremy-Kemp: *facepalm*. Yes, of course. Resolving.

test_common/harness/kernelHelpers.cpp
Jeremy-Kemp: Thanks, looks good. I don't think there is much that could be done to make it cleaner, there is not getting around just having to loop through every device and compare against what has already been discovered.

test_common/harness/kernelHelpers.cpp
```diff
@@ -809,6 +810,71 @@ int create_single_kernel_helper(cl_context context,
...
+                          }
+                          else
+                          {
+                              current_version =
+                                  (std::min)(device_version, current_version);
+                          }
+                      });
+
+        std::string cl_std{};
+        if (current_version == Version(3, 0))
```
Jeremy-Kemp, line 858: I'm not sure how we can future-proof this part (or if we even need to at the moment) as we don't know if/when we might get a new CL C version.

The worst that happens in this case is that we fall back to 1.2 so shouldn't introduce any problems.

Jeremy-Kemp (approved):
LGMT, thanks!

test_common/harness/kernelHelpers.cpp
FranklandJack: Yeah, this is a good point. It's a bit tricky since we are using the device OpenCL version to determine the OpenCL C version which isn't a 1-1 mapping and it could change with future versions like you pointed out. We could change line 858 to `current_version >= Version(3, 0)`, that way assuming future OpenCL C versions are compatible with OpenCL 3.0 we would at least get the OpenCL 3.0 language features? But I think this logic would probably need to get updated for future OpenCL C versions anyway.

FranklandJack: I've updated this to use `current_version >= Version(3, 0)`.

test_common/harness/kernelHelpers.cpp
```diff
@@ -809,6 +810,71 @@ int create_single_kernel_helper(cl_context context,
...
+        else
+        {
+            // If the -cl-std build option is not specified, the highest OpenCL
+            // C 1.x language version supported by each device is used when
+            // compiling the program for each device.
+            cl_std = "";
+        }
+        std::string updated_build_options = buildOptions ? buildOptions : "";
+        updated_build_options += cl_std;
+        buildOptions = updated_build_options.c_str();
```
FranklandJack, line 876: This will end up pointing at memory that has been freed.

FranklandJack: I've updated this to use a `std::string` declared at the top of the function, it's not ideal but I guess this is a results of trying to use c-style strings and `std::string` together.

test_common/harness/kernelHelpers.cpp
```diff
@@ -809,6 +810,76 @@ int create_single_kernel_helper(cl_context context,
...
+        size_t devices_size_in_bytes{};
+        auto error = clGetContextInfo(context, CL_CONTEXT_DEVICES, 0, nullptr,
+                                      &devices_size_in_bytes);
+        test_error(error, "clGetDeviceInfo failed for CL_CONTEXT_DEVICES");
+        std::vector<cl_device_id> devices(devices_size_in_bytes
+                                          / sizeof(cl_device_id));
+        error =
+            clGetContextInfo(context, CL_CONTEXT_DEVICES, devices_size_in_bytes,
+                             devices.data(), nullptr);
+        auto current_version = get_device_cl_version(devices[0]);
```
bashbaug, line 839: I think this is fine for now, but at some point the device OpenCL C version could be different from the device OpenCL version.  When this occurs we should consider adding a helper function that gets the latest supported OpenCL C version, either by querying [CL_DEVICE_OPENCL_C_ALL_VERSIONS](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_ALL_VERSIONS) (when supported), or falling back to [CL_DEVICE_OPENCL_C_VERSION](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_VERSION) otherwise.

test_common/harness/kernelHelpers.cpp
```diff
@@ -809,6 +810,76 @@ int create_single_kernel_helper(cl_context context,
...
+    // Check the build options for the -cl-std option.
+    if (!buildOptions || !strstr(buildOptions, "-cl-std"))
+    {
+        // If the build option isn't present add it using the latetest OpenCL-C
+        // version supported by the device. This allows calling code to force a
+        // particular CL C version if it is required, but also means that
+        // callers need not specify a version if they want to assume the most
+        // recent CL C.
+
+        // First get all the devices in the context and find the minumum
```
bashbaug, line 828: Nitpick: this comment isn't _quite_ correct, since it's finding the maximum universally supported OpenCL C version rather than the minimum supported OpenCL C version.

I'm not sure how common this will be, but since this is surprisingly tricky I think it'd be helpful to eventually add a helper or utility function that does this so it could be reused in other places.

test_common/harness/kernelHelpers.cpp
FranklandJack: Thanks for your feedback @bashbaug,  is really appreciated! Good point, I considered doing this but I figured that since `CL_DEVICE_OPENCL_C_ALL_VERSIONS` is missing before OpenCL-3.0 it wasn't an ideal way to go about querying the OpenCL C version. With regards to `CL_DEVICE_OPENCL_C_VERSION` as far as I understand, this returns the highest backwards compatible OpenCL C version, so for a 3.0 driver which optionally doesn't support some of the 2.X features, this would be forced to return 1.2, and I don't know if that is the behavior we want here since we are looking for the latest version supported by any device, not the latest backward compatible version?

test_common/harness/kernelHelpers.cpp
FranklandJack: Good point! I'll update the comment and move this functionality into a standalone function with a declaration in `kernelHelpers.h`, that way it can be reused in other places.

test_common/harness/kernelHelpers.cpp
FranklandJack: Perhaps maybe a good place to start would be to add a function getting the latest OpenCL C version using the same method as now (hard coded based on the device version), then we can more easily update this in the future using the correct device queries?

PR #884, opened 2020-08-13
Use float<->half conversion routines from the OpenCL headers

### LINKED ISSUE #870: Use the float<->half conversion routines from the headers
Conversion routines have been added to the headers by https://github.com/KhronosGroup/OpenCL-Headers/pull/60. We need to use them instead of the code in https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/imageHelpers.cpp#L898

kpet:
Fixes #870


test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -208,211 +210,27 @@ CheckD(cl_uint jid, cl_uint tid, void *userInfo)
...
-    return (u.u >> (24-11)) | sign;
+static cl_ushort float2half_rtn( float f ) {
+    return cl_half_from_float(f, CL_HALF_RTN);
 }
 
+static cl_ushort double2half_rte( double f );
+static cl_ushort double2half_rtz( double f );
+static cl_ushort double2half_rtp( double f );
+static cl_ushort double2half_rtn( double f );
+
```
jrprice, line 233: What's the reason for not using `cl_half_from_double` for these ones?

kpet: I'd forgotten that you'd added them as well :). Will do.

kpet:
Agree we need to address that as well. I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/885.

test_common/harness/imageHelpers.cpp
```diff
@@ -985,90 +922,6 @@ cl_ushort convert_float_to_half( float f )
...
-        return (cl_ushort)((int) x | sign);
-    }
-
-    u.u &= 0xFFFFE000U;
-    u.u -= 0x38000000U;
-
-    return (u.u >> (24-11)) | sign;
-}
-
 class TEST
```
jrprice, line 925: Can this class go now? It seems to just be testing the host functions for doing the conversions (and has it's own float->half conversion code in `__vstore_half_rte`).

kpet: Yeah, now removed.

PR #902, opened 2020-08-14
Remove spirv_new binaries and add Python script to generate them

### LINKED ISSUE #21: Handling of in-tree SPIR-V assembly/binaries
A few issues to discuss:
* Do we really want to commit the binaries? Shouldn't we just generate them from the assembly as part of the build system?
* The assembly doesn't use the same syntax as the SPIR-V disassembler which makes it hard to check consistency between the binaries and assembly or check that PRs modify the binaries and assembly in the same manner.

StuartDBrady:
Avoid storing binaries in version control, as it is generally not well suited to this.  Instead, add a script to regenerate the `.spv` files from their sources.

This also changes the `ext_cl_khr_spirv_no_integer_wrap_decoration` assembly source so that the SPIR-V version is listed as 1.0, to ensure that the output of **spirv-dis** exactly matches the assembly.

Fixes #21.


kpet (dismissed):

test_conformance/spirv_new/assemble_spirv.py
```diff
@@ -0,0 +1,163 @@
...
+            if verbose:
+                print(' Assembling {}'.format(asm_file))
+
+            asm_file_root, asm_file_ext = os.path.splitext(asm_file)
+            bin_file = asm_file_root + asm_file_ext.replace('asm', '')
+            bin_file_path = os.path.join(bin_dir, bin_file)
+
+            command = '"{}" --target-env spv1.0 "{}" -o "{}"'.format(
+                spirv_as, asm_file_path, bin_file_path)
+            if os.system(command) != 0:
```
gwawiork, line 61: Can we change executing command to call subprocess method `return_code = subprocess.call(command, shell=True) ` ?
This system method does not work on my windows 10. I got `The system cannot find the path specified.` error message. Probably some issues with escape characters in the path.
The same in validator part.

StuartDBrady: Sounds reasonable. Will do that.

StuartDBrady: Have done this. Does it look okay to you?

kpet: @gwawiork Could you review the updated change please?

gwawiork: Thanks for that it works. I noticed another issue. Looks like validator cannot find files in subdirectory where binary files have been saved
Example:
`python.exe assemble_spirv.py -s spirv_asm -o spirv_bin -a spirv-as.exe -l spirv-val.exe`

I have spirv-as and spirv-val in one directory. I have 2 subdirectories spirv_asm and spirv_bin. Assembling works ok but validation not. Please fix.
Example:

Validating fmod_float4.spv64
The system cannot find the path specified.
ERROR: Failure validating fmod_float4.spv64: see above output.

StuartDBrady: I've pushed another change. Could you test it, please?

gwawiork (approved):
Thanks! LGTM

kpet (approved):

kpet:
Two approvals (from two SPIR-V implementers), merging. Build system integration will follow in a separate PR.

PR #907, opened 2020-08-17
Memory consistency model optionality

### LINKED ISSUE #905: Optionality of memory consistency model for 3.0 conformance
Check device capabilities in the c11 atomics test to ensure only the correct tests are run for a given device.

Jeremy-Kemp:
This PR makes two changes.

1. The bare minimum capabilities for a device reporting >= 3.0 are tested during initialisation.
2. Tests are skipped for >= 3.0 devices where optional capabilities are tested.

Fixes #905

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -959,8 +959,9 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(cl_device_id dev
     // Set up the kernel code
     programSource = PragmaHeader(deviceID)+ProgramHeader(numDestItems)+FunctionCode()+KernelCode(numDestItems);
     programLine = programSource.c_str();
-    if(create_single_kernel_helper_with_build_options(context, &program, &kernel, 1, &programLine, "test_atomic_kernel",
-      gOldAPI ? "" : "-cl-std=CL2.0"))
+    if (create_single_kernel_helper_with_build_options(
+            context, &program, &kernel, 1, &programLine, "test_atomic_kernel",
+            gOldAPI ? "" : nullptr))
     {
       return -1;
     }
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -319,9 +319,7 @@ public:
           case MEMORY_SCOPE_ALL_SVM_DEVICES: {
               if ((gAtomicMemCap & CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES) == 0)
               {
-                  {
-                      return TEST_SKIPPED_ITSELF;
-                  }
+                  return TEST_SKIPPED_ITSELF;
               }
               break;
           }
@@ -355,9 +353,7 @@ public:
           case MEMORY_ORDER_SEQ_CST: {
               if ((gAtomicMemCap & CL_DEVICE_ATOMIC_ORDER_SEQ_CST) == 0)
               {
-                  {
-                      return TEST_SKIPPED_ITSELF;
-                  }
+                  return TEST_SKIPPED_ITSELF;
               }
               break;
           }
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -65,6 +65,8 @@ extern bool gUseHostPtr; // use malloc/free instead of clSVMAlloc/clSVMFree
 extern bool gDebug; // print OpenCL kernel code
 extern int gInternalIterations; // internal test iterations for atomic operation, sufficient to verify atomicity
 extern int gMaxDeviceThreads; // maximum number of threads executed on OCL device
+extern cl_device_atomic_capabilities gAtomicMemCap,
+    gAtomicFenceCap; // atomic memory and fence capabilities for this device
 
 extern const char *get_memory_order_type_name(TExplicitMemoryOrderType orderType);
 extern const char *get_memory_scope_type_name(TExplicitMemoryScopeType scopeType);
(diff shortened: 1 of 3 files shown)


Jeremy-Kemp:
@FranklandJack good point, I do think we should use this opportunity to use the correct build option.

If we gate this PR behind #879 and change the build option string to `NULL`  if `gOldAPI` is not set, would you be happy with that?

FranklandJack:
> @FranklandJack good point, I do think we should use this opportunity to use the correct build option.
> 
> If we gate this PR behind #879 and change the build option string to `NULL` if `gOldAPI` is not set, would you be happy with that?

Yep, sounds good to me, thanks!

Jeremy-Kemp:
Thanks @FranklandJack . I think I've addressed your comment.

This PR should not be merged until #879 has been merged.

FranklandJack:
> Thanks @FranklandJack . I think I've addressed your comment.
> 
> This PR should not be merged until #879 has been merged.

Thanks @Jeremy-Kemp LGTM.

nikhiljnv:
Reviewing the test, it looks like we are skipping the test altogether for OpenCL versions < 2.0, but for OpenCL 3.0, we will run some tests for bare minimum memory consistency requirements (Relaxed, acq_rel, work_group scope).
Should the extra tests that we will run on OpenCL 3.0 are expected to work on OpenCL 1.2 implementation as well?
If yes, any reason not to run them for OpenCL 1.2 conformance?

FranklandJack:
> Reviewing the test, it looks like we are skipping the test altogether for OpenCL versions < 2.0, but for OpenCL 3.0, we will run some tests for bare minimum memory consistency requirements (Relaxed, acq_rel, work_group scope).
> Should the extra tests that we will run on OpenCL 3.0 are expected to work on OpenCL 1.2 implementation as well?
> If yes, any reason not to run them for OpenCL 1.2 conformance?

@nikhiljnv This is something we also ran into when testing this patch locally. We noticed that there is a `-oldAPI` flag that can be passed to the `test_c11_atomics` executable which replaces the c11 atomic calls with 1.2 atomic builtins. As far as I understand this flag needs to be explicitly set when running these tests with a 1.2 driver and since the c11 atomics in 2.0 can't be rewritten in terms of 1.2 atomics and since the `test_c11_atomics` executable is not listed in the [1.2 conformance csv](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/opencl_conformance_tests_12_full.csv) even with the `-oldAPI` flag, we assumed that these test were not supposed to be run for a 1.2 driver.

Since the 1.2 atomics are equivalent to the c11 atomics with `memory_order_relaxed` consistency, and `memory_scope_work_group` scope, I'm also wondering whether if running with `-oldAPI` for the tests that use these memory orderings and scopes would be valid on a 1.2 driver. I don't think they can be run without this flag since the builtins won't exist for a 1.2 driver.

test_conformance/c11_atomics/common.h
```diff
@@ -281,6 +283,92 @@ class CBasicTest : CTest
...
+          case MEMORY_SCOPE_DEVICE: {
+              if ((gAtomicMemCap & CL_DEVICE_ATOMIC_SCOPE_DEVICE) == 0)
+              {
+                  return TEST_SKIPPED_ITSELF;
+              }
+              break;
+          }
+          case MEMORY_SCOPE_ALL_SVM_DEVICES: {
+              if ((gAtomicMemCap & CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES) == 0)
+              {
```
bashbaug, line 321: Super minor: there appears to be an extra set of `{` braces `}` here.  Also below, for `MEMORY_ORDER_SEQ_CST`.

bashbaug:
Regarding running these C11 atomics tests on an OpenCL 1.2 device, I think it could be done, but given that the OpenCL 1.2 atomics are already tested by [test_atomics](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/atomics/test_atomics.cpp) I don't think it's necessary.

nikhiljnv:
Thanks Jeremy !! Looks good to me in that case.

test_conformance/c11_atomics/common.h
```diff
@@ -855,8 +955,9 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(cl_device_id dev
     // Set up the kernel code
     programSource = PragmaHeader(deviceID)+ProgramHeader(numDestItems)+FunctionCode()+KernelCode(numDestItems);
     programLine = programSource.c_str();
-    if(create_single_kernel_helper_with_build_options(context, &program, &kernel, 1, &programLine, "test_atomic_kernel",
-      gOldAPI ? "" : "-cl-std=CL2.0"))
+    if (create_single_kernel_helper_with_build_options(
+            context, &program, &kernel, 1, &programLine, "test_atomic_kernel",
+            gOldAPI ? "" : nullptr))
```
bashbaug, line 960: Sorry I didn't catch this earlier: I believe the updated helper logic creates programs with the latest OpenCL C version  unless another OpenCL C version is provided explicitly.  So, I think we should either pass `-cl-std=CL1.2" specifically in the `gOldAPI` case, or we can remove the `gOldAPI` case entirely and use the latest OpenCL C version in all cases.

This is low priority and it isn't breaking anything right now (I think) because the empty string `""` and NULL build options are treated the same, so we're using the latest OpenCL C version in all cases, even though it looks like there are two different behaviors.

Jeremy-Kemp: Good spot, thanks.

Unless we want to specifically make this test one of the "special" ones that explicitly specifies `-cl-std=CL1.2` then we can just use `nullptr` for all cases.

bashbaug:
This has several reviews and has been open for almost a week... merging!

PR #912, opened 2020-08-20
API Feature Consistency Test, Part 2

bashbaug:
This is part two of the API feature consistency test.  Part 1 was added in #875.

This adds feature consistency tests for:

* Program Scope Global Variables
* Non-Uniform Work Groups
* Read-Write Images
* Creating 2D Images from Buffers
* Depth Images
* Device and Host Timer Synchronization
* Intermediate Language Programs
* Subgroups
* Program Initialization and Clean-Up Kernels
* 3D Image Writes

Other minor changes:

* Adds string decoding for several missing error codes.
* Adds a `test_assert_error` macro to simplify checking for specific non-error-related conditions.
* Updated the consistency check for `CL_QUEUE_SIZE` to require the error `CL_INVALID_COMMAND_QUEUE` when called with a host command queue.


test_common/harness/errorHelpers.h
```diff
@@ -77,6 +77,21 @@
 #define test_failure_warning_ret(errCode, expectedErrCode, msg, retValue) { if( errCode != expectedErrCode ) { print_failure_warning( errCode, expectedErrCode, msg ); warnings++ ; } }
 #define print_failure_warning(errCode, expectedErrCode, msg) log_error( "WARNING: %s! (Got %s, expected %s from %s:%d)\n", msg, IGetErrorString( errCode ), IGetErrorString( expectedErrCode ), __FILE__, __LINE__ );
 
+// generate an error when an assertion is false (not error code related)
+#define test_assert_error(condition, msg)                                      \
+    test_assert_error_ret(condition, msg, TEST_FAIL)
+#define test_assert_error_ret(condition, msg, retValue)                        \
```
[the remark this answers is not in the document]
bashbaug: I think the main advantage of the previous macros is that they print which error occurred (via `IGetErrorString( errCode )`) and not just that the error was not `CL_SUCCESS`.  I agree they could be unified if we choose to change this behavior.  I'd prefer to do this as a separate PR though :).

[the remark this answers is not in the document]
jrprice: Good point, I somehow missed that. No action needed.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +433,686 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+    return TEST_PASS;
+}
+
+int test_consistency_progvar(cl_device_id deviceID, cl_context context,
+                             cl_command_queue queue, int num_elements)
+{
+    // clGetDeviceInfo, passing CL_DEVICE_MAX_GLOBAL_VARIABLE_SIZE
+    // May return 0, indicating that device does not support Program Scope
+    // Global Variables.
+    int error;
```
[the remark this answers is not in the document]
bashbaug: Fixed, at least in this test.  My quick grep shows 550 other instances of `int error` in the CTS tests though.  Probably not all of them are CL errors, but I bet most of them are.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +433,686 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+            sz == 0,
+            "CL_DEVICE_MAX_GLOBAL_VARIABLE_SIZE returned 0 but "
+            "CL_DEVICE_GLOBAL_VARIABLE_PREFERRED_TOTAL_SIZE returned a "
+            "non-zero value");
+
+        // clGetProgramBuildInfo, passing
+        // CL_PROGRAM_BUILD_GLOBAL_VARIABLE_TOTAL_SIZE
+        // Returns 0 if device does not support Program Scope Global Variables.
+
+        error = clGetProgramBuildInfo(
```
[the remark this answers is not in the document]
bashbaug: Yes, definitely, not sure how I missed this one.  Fixed.

bashbaug:
Please hold off merging this for just a bit - I may need to make one more small change as a result of:

https://github.com/KhronosGroup/OpenCL-Docs/pull/418/files#r478742972

I've converted this PR to a draft until the discussion is resolved.  Thanks!

bashbaug:
I've made one small tweak to this test to address an observation in the related spec changes:

https://github.com/KhronosGroup/OpenCL-Docs/pull/418#discussion_r478742972

I think this is ready to go again.  Hopefully that was the last needed change!

Jeremy-Kemp:
I've not had time this week to review the changes, however I have run the test on our implementation and don't have any comments yet based on that.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,689 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+    test_error(error, "Unable to query CL_DEVICE_IL_VERSION");
+    test_assert_error(sz != 0,
+                      "CL_DEVICE_IL_VERSION should return a non-zero size");
+
+    std::string ilVersion = get_device_il_version_string(deviceID);
+
+    error = clGetDeviceInfo(deviceID, CL_DEVICE_ILS_WITH_VERSION, 0, NULL, &sz);
+    test_error(error, "Unable to query CL_DEVICE_ILS_WITH_VERSION");
+
+    if (ilVersion == "" || sz == 0)
```
Jeremy-Kemp, line 894: `CL_DEVICE_ILS_WITH_VERSION` may return an empty array, not an empty string.

I think this should be updated to follow a similar path to the extended versioning test, making sure that a returned `cl_name_version_khr` is empty.

Jeremy-Kemp (changes requested):

test_conformance/api/test_api_consistency.cpp
Jeremy-Kemp: Actually, I think this highlights a problem in extended versioning, not this consistency test.

I'll go and check and then update here.

bashbaug: Cool, I do think this code is correct.  There are two cases to consider:

1. What does `CL_DEVICE_IL_VERSION` return?  This is the single string query added in OpenCL 2.1.  This string is either empty (`""`, indicating that no ILs are supported) or non-empty (ILs supported), but it should not return size equal to zero.

2. What does `CL_DEVICE_ILS_WITH_VERSION` return?  This is the new OpenCL 3.0 array query.  This query may return size equal to zero (no array elements, indicating that no ILs are supported) or a non-zero size (some array elements, indicating that ILs are supported).

The check for `ilVersion == ""` is checking for the first case, and the check for `sz == 0` is checking for the second case.

test_common/harness/errorHelpers.h
```diff
@@ -77,6 +77,21 @@
...
+    {                                                                          \
+        if (!(condition))                                                      \
+        {                                                                      \
+            print_assertion_error(condition, msg);                             \
+            return retValue;                                                   \
+        }                                                                      \
+    }
+#define print_assertion_error(condition, msg)                                  \
+    log_error("ERROR: %s! (!(%s) from %s:%d)\n", msg, #condition, __FILE__,    \
```
kpet, line 92: nitpick: not sure we should always add the `!` after the message.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -262,74 +254,54 @@ int test_consistency_device_enqueue(cl_device_id deviceID, cl_context context,
...
-        // https://github.com/KhronosGroup/OpenCL-Docs/issues/402
-        // test_error(error, "Unable to query CL_QUEUE_SIZE");
-        if (error == CL_SUCCESS && u != 0)
-        {
-            log_error("CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES returned 0 but "
-                      "CL_QUEUE_SIZE returned a non-zero value\n");
-            return TEST_FAIL;
-        }
+        test_failure_error(
+            error, CL_INVALID_COMMAND_QUEUE,
```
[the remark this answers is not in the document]
kpet: Yeah, agree.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -455,13 +415,17 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
-            "clGetPipeInfo did not return CL_INVALID_MEM_OBJECT");
+        // Returns CL_INVALID_OPERATION if no devices in the context associated
+        // with pipe support Pipes.
+        clMemWrapper not_a_pipe =
+            clCreateBuffer(context, CL_MEM_READ_WRITE, 4, NULL, &error);
+        test_error(error, "Unable to create non-pipe buffer");
+
+        error =
+            clGetPipeInfo(not_a_pipe, CL_PIPE_PACKET_SIZE, sizeof(u), &u, NULL);
+        test_assert_error(error == CL_INVALID_OPERATION,
```
kpet, line 426: I think this should test for `error == CL_INVALID_OPERATION || error == CL_INVALID_MEM_OBJECT`

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,689 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+
+    if (maxGlobalVariableSize == 0)
+    {
+        // Test setup:
+
+        error = create_single_kernel_helper(context, &program, &kernel, 1,
+                                            &test_kernel, "test");
+        test_error(error, "Unable to create test kernel");
+
+        size_t sz = 0;
```
kpet, line 474: Probably best to be defensive and initialise the variable to a non-zero value.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,689 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+    {
+        // Test setup:
+
+        mem =
+            clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, NULL, &error);
+        test_error(error, "Unable to create test buffer");
+
+        error = create_single_kernel_helper_with_build_options(
+            context, &program, &kernel, 1, &test_kernel, "test",
+            "-cl-std=CL3.0");
```
kpet, line 543: Could leave the CL C version selection to the framework. That would be a bit more future proof.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,689 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+
+        // Check that both queries return zero:
+        test_assert_error(imagePitchAlignment == 0,
+                          "CL_DEVICE_IMAGE_PITCH_ALIGNMENT returned a non-zero "
+                          "value but CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT "
+                          "returned 0");
+        test_assert_error(imagePitchAlignment == 0,
+                          "CL_DEVICE_IMAGE_PITCH_ALIGNMENT returned 0 but "
+                          "CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT returned a "
+                          "non-zero value");
```
kpet, line 714: These will never fire.  Given the messages, I'm guessing the conditions are incomplete.

bashbaug:
I created https://github.com/KhronosGroup/OpenCL-CTS/issues/933 to track possible changes to the atomic capability queries.  I'd like to merge this PR without these changes, if possible, and I'll create a follow-on PR to fix this last issue.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,689 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+
+        // clGetDeviceInfo, passing CL_DEVICE_EXTENSIONS
+        // Will not describe support for the cl_khr_image2d_from_buffer
+        // extension if device does not support Creating a 2D Image from a
+        // Buffer.
+        test_assert_error(supports_cl_khr_image2d_from_buffer == false,
+                          "Device does not support Creating a 2D Image from a "
```
kpet, line 721: Hmm, it's the presence of the extension that defines whether the feature is supported. I would change the message to say something along the lines of `cl_khr_image2d_from_buffer` is supported but an invalid value was returned for ..."

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,689 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+                            sizeof(imageBaseAddressAlignment),
+                            &imageBaseAddressAlignment, NULL);
+    test_error(error,
+               "Unable to query "
+               "CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT");
+
+    bool supports_cl_khr_image2d_from_buffer =
+        is_extension_available(deviceID, "cl_khr_image2d_from_buffer");
+
+    if (imagePitchAlignment == 0 || imageBaseAddressAlignment == 0)
```
kpet, line 696: Probably should test for the extension string here instead.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,689 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+        test_failure_error(error, CL_INVALID_OPERATION,
+                           "Device does not support Creating a 2D Image from a "
+                           "Buffer but clCreateImageWithProperties did not "
+                           "return CL_INVALID_OPERATION");
+    }
+    else
+    {
+        test_assert_error(supports_cl_khr_image2d_from_buffer,
+                          "Device supports Creating a 2D Image from a Buffer "
+                          "but does not support cl_khr_image2d_from_buffer");
```
kpet, line 755: See comment above

test_conformance/api/test_api_consistency.cpp
Jeremy-Kemp: Perhaps this is where the confusion is coming from (empty string handling).

What exactly do we say the empty string is? In C its just `\0` which has a length of 0. That would mean that `sz` *can* be 0, but that is asserted against above.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,689 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+        test_assert_error(supports_cl_khr_depth_images == false,
+                          "Device does not support Depth Images but does "
+                          "support cl_khr_depth_images");
+    }
+    else
+    {
+        test_assert_error(supports_cl_khr_depth_images,
+                          "Device supports Depth Images but does not support "
+                          "cl_khr_depth_images");
+    }
```
kpet, line 818: A detail: I'm wondering whether "report" wouldn't be better than "support" in the error messages. We're defining support as reporting the extension _and_ at least one format.

kpet (changes requested):

test_conformance/api/test_api_consistency.cpp
Jeremy-Kemp: (For reference, we currently return a size of 1 and have it simply contain `\0`)

test_common/harness/errorHelpers.h
bashbaug: I don't have a strong opinion about this, but I'd prefer to make any changes in a separate PR.  This macro follows the same conventions as the other macros in this file.

test_conformance/api/test_api_consistency.cpp
bashbaug: Yeah, I used to have both of these cases, but I think this is correct with the latest spec updates.  See: https://github.com/KhronosGroup/OpenCL-Docs/pull/418#discussion_r478742972

test_conformance/api/test_api_consistency.cpp
bashbaug: How strongly do you feel about this?  I'm sure this isn't the only test that initializes integral quantities to zero, and we'll return an error if the query fails so I think this is already fairly defensive.

test_conformance/api/test_api_consistency.cpp
bashbaug: This wasn't an option when I created this PR.  Fixed!

test_conformance/api/test_api_consistency.cpp
bashbaug: No, these may fire, since we'll enter the conditional if _either_ of the conditions are true.  These checks are to ensure that if either condition is true then both conditions are true.

test_conformance/api/test_api_consistency.cpp
bashbaug: Following the logic in the spec appendix, the feature is supported / not supported based on the device queries, and the extension must be reported consistent with these queries.  I think this is six of one, half-dozen of the other though, and it really doesn't matter which one gets checked first, so long as everything is consistent.  Are any consistency checks missing?

test_conformance/api/test_api_consistency.cpp
bashbaug: See above.

test_conformance/api/test_api_consistency.cpp
bashbaug: I don't feel strongly about this.  Do you want this changed?  If so, just tell me what to change it to :).

bashbaug:
I've rebased these changes to pick up the changes to the CTS kernel framework to use the latest supported OpenCL C version.  I've replied to the remaining review comments.  If there are any others that are must-fix please call them out, and please be specific what must be changed.  Thanks!

test_common/harness/errorHelpers.h
kpet: Yeah, let's not spend more time on this.

test_conformance/api/test_api_consistency.cpp
kpet: Interesting, I'd missed the latest developments. Assuming we require the function to always be present in the dispatch table, then this is fine. I don't think we say that explicitly but maybe it's "obvious" since we require the function to be present and have a given behaviour.

test_conformance/api/test_api_consistency.cpp
kpet: It's not the only one but this one is particularly bad since we're asserting that the value is 0 directly after the query. An incorrect implementation that doesn't return anything (doesn't write to memory) for `CL_DEVICE_GLOBAL_VARIABLE_PREFERRED_TOTAL_SIZE` will pass the check with `sz` initialised to 0. It wouldn't if `sz` was initialised to a different value. `SIZE_MAX`?

test_conformance/api/test_api_consistency.cpp
kpet: Yup, I read the code a bit too quickly. The second check's condition should probably be `imageBaseAddressAlignment == 0` though.

test_conformance/api/test_api_consistency.cpp
kpet: Well the spec appendix is not normative and my reading of the normative section says otherwise. Both `CL_DEVICE_IMAGE_PITCH_ALIGNMENT` and `CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT` are specified to return 0 when the feature is not supported but only the text for `CL_DEVICE_EXTENSIONS` defines a logical equivalence between the presence of the extension string and support for the feature.

> Are any consistency checks missing?

Not AFAICT. The test will be correct once the comment on the assert above is fixed.

Since this is mostly a readability issue and not a terrible one, I'll resolve this.

test_conformance/api/test_api_consistency.cpp
kpet: I won't lose sleep over that :).

kpet (changes requested):
The assert in the 2d from buffer test needs fixing, my other comments needn't gate this PR IMO.

test_conformance/api/test_api_consistency.cpp
Jeremy-Kemp: I believe the issue is this:

Any device that does not support `CL_DEVICE_IL_VERSION` will fill out `1` for `param_value_size_ret`. Subsequently, `param_value` will be 1 byte and just contain `\0`.

`get_device_il_version_string` calls `get_device_info_string` which constructs an `std::string` using this data, resulting in a string whose length is not 0. Therefore, the constructed string is not equivalent to an `std::string("")`, so `ilVersion == ""` evaluates to `false`.

If my understanding is correct and I'm not missing something obvious (please correct me if I'm wrong), then I think this can be avoided in one of two ways.

1. `get_device_info_string` can be special cased for empty strings
2. The comparison in this test can be changed from `ilVersion == ""` to `strcmp(ilVersion.c_str(), "") == 0`

My preference would be to go with 1. (possibly as a separate PR), as the empty string case may crop up in other tests in the future and is not specific to this test.

test_conformance/api/test_api_consistency.cpp
bashbaug: I'll change the initialization value to `SIZE_MAX`.

test_conformance/api/test_api_consistency.cpp
bashbaug: Yes it should be `imageBaseAddressAlignment` - fixed!

test_conformance/api/test_api_consistency.cpp
bashbaug: Ah, I see.  Yes, we are constructing our `std::string` with:

```c
    std::vector<char> info(size);

    if ((err = clGetDeviceInfo(device, param_name, size, info.data(), NULL))
        != CL_SUCCESS)
    {
        throw std::runtime_error("clGetDeviceInfo failed\n");
    }

    return std::string(info.begin(), info.end());
```

I think we could change this to either `std::string(info.begin(), --info.end());` or `std::string(info.data(), size - 1)`, and get the behavior we are looking for.  Any preference?  I'm leaning towards the latter.

Jeremy-Kemp: No strong preference, so the latter looks good to me. Thanks.

Would you like me to do that as part of a new PR, or just do it with this PR?

bashbaug: I'll take care of it.  Thanks for the quick reply!

bashbaug:
Great thanks!  I think I've addressed the remaining must-fix issues - please take a look.

test_conformance/api/test_api_consistency.cpp
Jeremy-Kemp: Thanks @bashbaug, looks good.

kpet (approved):

Jeremy-Kemp (approved):
Thanks for your help @bashbaug .

test_conformance/api/test_api_consistency.cpp
```diff
@@ -473,14 +437,688 @@ int test_consistency_pipes(cl_device_id deviceID, cl_context context,
...
+        // clGetDeviceInfo, passing CL_DEVICE_EXTENSIONS
+        // Will not describe support for the cl_khr_subgroups extension if
+        // device does not support Subgroups.
+
+        bool supports_cl_khr_subgroups =
+            is_extension_available(deviceID, "cl_khr_subgroups");
+        test_assert_error(supports_cl_khr_subgroups == false,
+                          "Device does not support Subgroups but does "
+                          "support cl_khr_subgroups");
+
```
gwawiork, line 1007: What do you think about adding condition to check if supports_cl_khr_subgroups == cl_true then ifp == cl_true ?

bashbaug: I thought this check would have existed already, but it doesn't look like it does.

I don't mind adding it but would you be OK adding it in a separate PR?  I'm going to need to make a minor modification to this test for #933 anyhow.

gwawiork: We have such verification in subgroups test https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/subgroups/test_ifp.cpp#L356 . 
Then maybe there is no need to add such verification here again.

gwawiork (approved):
LGTM

PR #915, opened 2020-08-20
Fix bug caught by -Wparentheses

einhov:
This fix is required to build with -Wall -Werror with Clang (tested with
Clang 6.0.0).


kpet (approved):

bashbaug:
I'm fine with this change and the check is certainly more readable now, but I'm not sure why this was an error before and my test case isn't producing a warning.  Doesn't `|` have similar precedence as `||` relative to the array subscripting `[]` and the logical not `!`?

https://godbolt.org/z/5hzeqx

einhov:
Seems like I got this a bit mixed up while looking at some other things caught by clang. For this one it's gcc (>= 5) that seems to have caught it, I'll fix the commit message. I believe the old code behaves correctly, but I would still consider it a bug to mix bitwise and logical operations in this situation.

bashbaug:
Gotcha, thanks!  Yes, I see gcc does produce a warning in this case: https://godbolt.org/z/Yd814b, and I agree the code is much more obviously right after this fix.

PR #916, opened 2020-08-22
Only include sys/sysctl.h on for Apple

pierremoreau:
glibc deprecated that header in version 2.30, and removed it in 2.32.

### DIFF
--- a/test_common/harness/testHarness.cpp
+++ b/test_common/harness/testHarness.cpp
@@ -34,7 +34,7 @@
 #include <unistd.h>
 #endif
 
-#if !defined(_WIN32) && !defined(__ANDROID__)
+#if defined(__APPLE__)
 #include <sys/sysctl.h>
 #endif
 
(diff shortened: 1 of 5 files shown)


kpet:
@pierremoreau This is currently marked as a draft. Are you expecting a review? Do you anticipate further changes?

pierremoreau:
I wanted to take a deeper look at those headers and check that those are even needed for Apple (`test_common/harness/testHarness.cpp` definitely needs it, but it didn't look like the other did) and if other `sys/*.h` headers could be removed as well as no longer needed, but ran out of time.
Maybe this can be reviewed as-is as it is enough to make it compile against recent glibc, and open an issue to keep track of the other bits?

kpet (approved):
LGTM, this fixes the issue for me. Suggest we get this is as is to unblock people using the latest glibc.

pierremoreau:
Sounds good, thanks! I'll open another PR with the remaining changes.

Issue #882, opened 2020-08-13
Test ctz under OpenCL 3.0


bashbaug:
Regarding the work item "Test ctz under OpenCL 3.0":

The existing `test_integer_ops` should pretty much work as-is, it just needs to be updated to pass `-cl-std=CL3.0` for an OpenCL 3.0 device rather than `-cl-std=CL2.0`.  The `integer_ops` tests are already running for OpenCL 1.x devices, so I don't think any other changes will be required.

FranklandJack:
@bashbaug I actually have a local patch for this I was working on yesterday but haven't opened a PR for yet.

bashbaug:
Fixed by #917 - closing.

PR #917, opened 2020-08-25
Ctz cl std

FranklandJack:


FranklandJack:
The PR is based on #879.

FranklandJack:
Since #879 has been merged this PR is ready for review.

test_conformance/integer_ops/test_integers.cpp
```diff
@@ -84,11 +84,11 @@ int test_single_param_integer_kernel(cl_command_queue queue, cl_context context,
...
                 sizeName );
 
-    bool isOpenCL20Function = (strcmp(fnName,"ctz") == 0)? true: false;
-
     /* Create kernels */
     programPtr = kernelSource;
-    if( create_single_kernel_helper_with_build_options( context, &program, &kernel, 1, (const char **)&programPtr, "sample_test", isOpenCL20Function ? "-cl-std=CL2.0": "" ) )
+    if (create_single_kernel_helper_with_build_options(
+            context, &program, &kernel, 1, (const char **)&programPtr,
+            "sample_test", nullptr))
```
bashbaug, line 91: This is a very minor comment but: Is there a reason to prefer `create_single_kernel_helper_with_build_options` and NULL build options explicitly vs. `create_single_kernel_helper`, where build options are NULL by default?

bashbaug:
I think this fixes #882?

test_conformance/integer_ops/test_integers.cpp
FranklandJack: Thanks for your feedback @bashbaug. I don't think the `create_single_kernel_helper_with_build_options` has a default value for it's `buildOptions` parameter, although we could definitely add one, which would simplify the calls. Is this what you meant?

FranklandJack:
> I think this fixes #882?

Yep, that is correct :)

test_conformance/integer_ops/test_integers.cpp
bashbaug: Close.  I'm looking at `create_single_kernel_helper`, which already has a default value for `buildOptions`:

```c
extern int create_single_kernel_helper(cl_context context, 
                                       cl_program *outProgram,
                                       cl_kernel *outKernel,
                                       unsigned int numKernelLines,
                                       const char **kernelProgram,
                                       const char *kernelName,
                                       const char *buildOptions = NULL,
                                       const bool openclCXX = false);
```

Longer-term I think we could get rid of `create_single_kernel_helper_with_build_options` entirely, unless we really want a function variant that absolutely requires build options.  For now though, I think we should use `create_single_kernel_helper` when we don't require build options vs. `create_single_kernel_helper_with_build_options` and passing NULL build options explicitly... unless there's a reason to do this that I'm not currently seeing.  Thanks!

FranklandJack: Ah sorry, I misunderstood, yep what you've suggested sounds good, I have updated the PR to use `create_single_kernel_helper` instead of `create_single_kernel_helper_with_build_options`.

PR #924, opened 2020-08-27
api: Add a Negative test clCloneKernel (#747)

lakshmih:
Add a negative test for clCloneKernel that expects the CL_INVALID_VALUE
return code.

Other improvements:
* Remove unused code or at least cast to void.
* Change NULL to nullptr which includes type information.


kpet (changes requested):
A few general comments:
- The prototype for `test_` functions is dictated by the framework. You can't change it.
- It's great to add negative testing for this but this should IMHO be done in a separate test.

kpet:
Correction: these `test_` functions do not correspond to test entry points and my comment above doesn't apply.

lakshmih:
Thanks for the feedback. We will reconsider how best to address this case.

Issue #683, opened 2020-03-17
Reduce the number of binaries


kpet:
Many binaries only contain a small number of tests and duplicate code. This makes maintenance and general management of tests a bit harder. We should aim to reduce the number of binaries.

kpet:
A few hopefully uncontroversial proposals:
- Merge `vec_step` and `vec_align` into a single `vectors` binary
- Merge `select` into `relationals`. The `select` built-in functions is described in the relationals section of the spec and the `relationals` suite already tests `select`.
- Merge `device_timers` into `api`. It contains only two very short running smoke tests.

PR #926, opened 2020-08-28
Merge vec_step and vec_align binaries into a new vectors binary

kpet:
This change just merges the two suites to reuse common functions
and definitions. There is probably room for further code consolidation
but probably best done as a separate change.

Contributes to #683


Jeremy-Kemp (approved):
Thanks @kpet, the changes that aren't just removed/moved files look good.

Another nice reduction of duplicate code.

gwawiork (approved):
LGTM

kpet:
Thanks for the reviews! Two approvals from two member companies, merging.

PR #934, opened 2020-09-03
[MinGW] Fix if statements with stray semicolons after condition

StuartDBrady:
In two places, a stray semicolon between an if statement and its body would cause the body to be executed unconditionally.  Fix this.

### DIFF
--- a/test_common/harness/ThreadPool.cpp
+++ b/test_common/harness/ThreadPool.cpp
@@ -371,8 +371,7 @@ void *ThreadPool_WorkerFunc( void *p )
             {
 #if (__MINGW32__)
                 EnterCriticalSection(&gAtomicLock);
-                if( jobError == CL_SUCCESS );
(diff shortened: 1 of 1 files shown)


StuartDBrady:
Note: I'm not set up for testing on MinGW, which is one of the reasons I've raised a separate PR just for this.

The change seems to bring the code closer into line to that of the GCC(/Clang)-specific and MSVC-specific sections, though, so I think this is the right fix, conceptually.

It might be worth someone checking whether the MSVC-specific code works with current versions of MinGW.

bashbaug (approved):
LGTM - this is pretty obviously wrong :).

I'll merge this tomorrow morning, unless there's a good reason to wait, or unless someone else gets to it first.

gwawiork (approved):
LGTM

kpet (approved):

PR #938, opened 2020-09-03
Ensure basic/test_basic/kernel_preprocessor_macros checks the OpenCL C version correctly for 3.0 onwards.

dgavedissian:
Currently, it fails if `__OPENCL_C_VERSION__` is greater than the version reported by `CL_DEVICE_OPENCL_C_VERSION`.

However (as of the 3.0 provisional spec), it's possible to have an OpenCL 3.0 device which reports 1.2 as the highest backwards-compatible version (from querying `CL_DEVICE_OPENCL_C_VERSION`), but supports 3.0 according to `CL_DEVICE_OPENCL_C_ALL_VERSIONS`.

In that case, this test should only fail if `__OPENCL_C_VERSION__ > CL_DEVICE_OPENCL_C_VERSION` _and_ `__OPENCL_C_VERSION__` is not in the list returned by `CL_DEVICE_OPENCL_C_ALL_VERSIONS`.

I took the liberty of refactoring some of the test to make use of some helper functions in `kernelHelpers.h`.

### DIFF
--- a/test_common/harness/kernelHelpers.cpp
+++ b/test_common/harness/kernelHelpers.cpp
@@ -1584,42 +1584,7 @@ int printDeviceHeader( cl_device_id device )
 
 Version get_device_cl_c_version(cl_device_id device)
 {
-    // Get the device OpenCL version.
     auto device_cl_version = get_device_cl_version(device);
-
-    // If the device version >= 3.0 it must support the
-    // CL_DEVICE_OPENCL_C_ALL_VERSIONS query from which we can extract the most
-    // recent CL C version supported by the device.
-    if (device_cl_version >= Version{ 3, 0 })
-    {
-        size_t opencl_c_all_versions_size_in_bytes{};
-        auto error =
-            clGetDeviceInfo(device, CL_DEVICE_OPENCL_C_ALL_VERSIONS, 0, nullptr,
-                            &opencl_c_all_versions_size_in_bytes);
-        test_error_ret(
-            error, "clGetDeviceInfo failed for CL_DEVICE_OPENCL_C_ALL_VERSIONS",
-            (Version{ -1, 0 }));
-        std::vector<cl_name_version> name_versions(
-            opencl_c_all_versions_size_in_bytes / sizeof(cl_name_version));
-        error = clGetDeviceInfo(device, CL_DEVICE_OPENCL_C_ALL_VERSIONS,
-                                opencl_c_all_versions_size_in_bytes,
-                                name_versions.data(), nullptr);
-        test_error_ret(
-            error, "clGetDeviceInfo failed for CL_DEVICE_OPENCL_C_ALL_VERSIONS",
-            (Version{ -1, 0 }));
-
-        Version max_supported_cl_c_version{};
-        for (const auto &name_version : name_versions)
-        {
-            Version current_version{ CL_VERSION_MAJOR(name_version.version),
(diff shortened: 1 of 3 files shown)


test_common/harness/kernelHelpers.cpp
```diff
@@ -1694,6 +1701,53 @@ Version get_max_OpenCL_C_for_context(cl_context context)
...
+            if (current_version == version)
+            {
+                version_in_all_versions_query = true;
+                break;
+            }
+        }
+    }
+
+    return version >= get_device_cl_c_version(device)
+        || version_in_all_versions_query;
```
bashbaug, line 1748: It may just be a bit too late in the day, but is the first condition backwards?  Seems like the device would support the requested version if `version <= get_device_cl_c_version(device)` instead?

Also, it's fairly minor, but I'd consider swapping these two boolean conditions also, since there's no need to query `CL_DEVICE_OPENCL_C_VERSION` if the version was already found in `CL_DEVICE_OPENCL_C_ALL_VERSIONS`.

test_conformance/basic/test_preprocessors.cpp
```diff
@@ -250,33 +232,28 @@ int test_kernel_preprocessor_macros(cl_device_id deviceID, cl_context context, c
...
-        ;
-    for( p2 = p1; ( *p2 != 0 ) && ( *p2 != '.' ); p2++ )
-        ;
-    for( p3 = p2; ( *p3 != 0 ) && ( *p3 != ' ' ); p3++ )
-        ;
-
-    if( p2 == p3 )
+    if ((results[3] < 100)
+        || (!device_supports_cl_c_version(deviceID,
+                                          Version{ results[3] / 10 })))
```
bashbaug, line 240: This is fairly minor, but I'd consider decoding the result value completely into major and minor versions and using the existing constructor vs. defining a new constructor that works on a combined value.  This seems safer than relying on the current method of converting between a Version and an integer.

test_common/harness/kernelHelpers.cpp
dgavedissian: Yes, you're totally right. Good catch.

Agreed regarding the second change. Rather than store the result of checking CL_DEVICE_OPENCL_C_ALL_VERSIONS, I have changed it to early exit in the for loop instead.

test_conformance/basic/test_preprocessors.cpp
dgavedissian: Sounds reasonable to me!

bashbaug (approved):
Thanks!  LGTM.

test_common/harness/kernelHelpers.cpp
```diff
@@ -1584,7 +1584,45 @@ int printDeviceHeader( cl_device_id device )
...
+                   (Version{ -1, 0 }));
+
+    // Scrape out the major, minor pair from the string.
+    auto major = opencl_c_version[opencl_c_version.find('.') - 1];
+    auto minor = opencl_c_version[opencl_c_version.find('.') + 1];
+
+    return Version{ major - '0', minor - '0' };
+}
+
+Version get_device_latest_cl_c_version(cl_device_id device)
```
alycm, line 1624: Also a comment here, as above.

alycm (approved):
LGTM.

kpet:
Merging as agreed in the 2020/09/15 teleconference.

Issue #939, opened 2020-09-03
c11_atomics: Atomic fence CL kernel is not re-generated when the workgroup size changes


lakshmih:
This issue was originally raised and fixed on cl20_trunk on gitlab. However the fix did not propagate to subsequent CTS versions or this github.

Original issue:
https://www.khronos.org/members/login/bugzilla/show_bug.cgi?id=15676

Original fix merged to cl20_trunk:
https://gitlab.khronos.org/opencl/conformance-tests/-/commit/b71bb46fe0dd4aee9f5d74571a6b6982de24cf96

I've re-summarized it below.

The generator function for atomic_fence uses the current workgroup size to determine the number of non atomic variables per thread. However, when the workgroup size is modified (after the kernel is generated) we don't re-generate the kernel, leading to mismatches during verification.

Following is the sequence in file 
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/c11_atomics/common.h

1) CurrentGroupSize is set to the maximum workgroup size allowed by the device

  // set up work sizes based on device capabilities and test configuration
  error = clGetDeviceInfo(deviceID, CL_DEVICE_MAX_WORK_GROUP_SIZE, sizeof(groupSize), &groupSize, NULL);
  test_error(error, "Unable to obtain max work group size for device");
  `CurrentGroupSize((cl_uint)groupSize);`

2) The CL kernel is generated. The generator function for the kernel uses the workgroup size in determining NumNonAtomicVariablesPerThread(). See calculation of NumNonAtomicVariablesPerThread in test_atomics.cpp

3) Workgroup size is modified depending on the max permitted for the kernel 

    // tune up work sizes based on kernel info
    error = clGetKernelWorkGroupInfo(kernel, deviceID, CL_KERNEL_WORK_GROUP_SIZE, sizeof(groupSize), &groupSize, NULL);
    test_error(error, "Unable to obtain max work group size for device and kernel combo");
 
If the workgroup size changes in step 3 above, the kernel needs to be re-generated. If not, the verify function uses the modified (kernel_max) workgroup size to calculate the reference values whereas the kernel was generated using the original (device_max) workgroup size.

PR #941, opened 2020-09-04
c11_atomics: Fix to iteratively reduce workgroup size  (#939)

lakshmih:
The generator function for atomic_fence uses the current workgroup
size to determine the number of non atomic variables per thread
in the kernel. The kernel should hence be regenerated when the
workgroup size changes.

However regenerating the kernel can itself change the workgroup
size. This change introduces an iterative loop that reduces the
workgroup sizes by 2 each time re-generating the kernel until
we find one that works (or exit at groupsize == 1)

Change-Id: Ic32fe967e32de6643e01c6775f4bddbcad0a299a

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -948,50 +948,75 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(cl_device_id dev
 
   if(deviceThreadCount > 0)
   {
-    cl_ulong usedLocalMemory;
-    cl_ulong totalLocalMemory;
-    cl_uint maxWorkGroupSize;
+      // This loop iteratively reduces the workgroup size by 2 and then
+      // re-generates the kernel with the reduced
+      // workgroup size until we find a size which is admissible for the kernel
+      // being run or reduce the wg size
+      // to the trivial case of 1 (which was separately verified to be accurate
+      // for the kernel being run)
 
-    // Set up the kernel code
-    programSource = PragmaHeader(deviceID)+ProgramHeader(numDestItems)+FunctionCode()+KernelCode(numDestItems);
-    programLine = programSource.c_str();
-    if (create_single_kernel_helper_with_build_options(
-            context, &program, &kernel, 1, &programLine, "test_atomic_kernel",
-            gOldAPI ? "" : nullptr))
-    {
-      return -1;
-    }
-    if(gDebug)
-    {
-      log_info("Program source:\n");
-      log_info("%s\n", programLine);
-    }
-    // tune up work sizes based on kernel info
-    error = clGetKernelWorkGroupInfo(kernel, deviceID, CL_KERNEL_WORK_GROUP_SIZE, sizeof(groupSize), &groupSize, NULL);
-    test_error(error, "Unable to obtain max work group size for device and kernel combo");
+      while ((CurrentGroupSize() > 1))
+      {
+          // Re-generate the kernel code with the current group size
+          if (kernel) clReleaseKernel(kernel);
+          if (program) clReleaseProgram(program);
+          programSource = PragmaHeader(deviceID) + ProgramHeader(numDestItems)
+              + FunctionCode() + KernelCode(numDestItems);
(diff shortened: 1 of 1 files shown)


lakshmih:
@kpet @bashbaug  Please review. This is a fix that was originally merged to the cl20_trunk branch on gitlab (links to old bugzilla issue and fix on the issue description) and not propagated to the 2.1 and future branches.
Please add anyone else that you think needs to be on the review.

test_conformance/c11_atomics/common.h
```diff
@@ -948,51 +948,76 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(cl_device_id dev
...
+      {
+          // Re-generate the kernel code with the current group size
+          if (kernel) clReleaseKernel(kernel);
+          if (program) clReleaseProgram(program);
+          programSource = PragmaHeader(deviceID) + ProgramHeader(numDestItems)
+              + FunctionCode() + KernelCode(numDestItems);
+          programLine = programSource.c_str();
+          if (create_single_kernel_helper_with_build_options(
+                  context, &program, &kernel, 1, &programLine,
+                  "test_atomic_kernel", gOldAPI ? "" : "-cl-std=CL2.0"))
```
Jeremy-Kemp, line 968: Was it intentional to go back to specifying `-cl-std=CL2.0`?

lakshmih: oops no, that was a typo. Thanks for catching it, I'll update.

lakshmih: @Jeremy-Kemp can you review again? Thanks.

lakshmih:
Copying notes from issue description for reviewers:

Original issue:
https://www.khronos.org/members/login/bugzilla/show_bug.cgi?id=15676

Original fix merged to cl20_trunk:
https://gitlab.khronos.org/opencl/conformance-tests/-/commit/b71bb46fe0dd4aee9f5d74571a6b6982de24cf96

I've re-summarized it below.

The generator function for atomic_fence uses the current workgroup size to determine the number of non atomic variables per thread. However, when the workgroup size is modified (after the kernel is generated) we don't re-generate the kernel, leading to mismatches during verification.

Following is the sequence in file
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/c11_atomics/common.h
1)
CurrentGroupSize is set to the maximum workgroup size allowed by the device
// set up work sizes based on device capabilities and test configuration
error = clGetDeviceInfo(deviceID, CL_DEVICE_MAX_WORK_GROUP_SIZE, sizeof(groupSize), &groupSize, NULL);
test_error(error, "Unable to obtain max work group size for device");
CurrentGroupSize((cl_uint)groupSize);

2)
The CL kernel is generated. The generator function for the kernel uses the workgroup size in determining NumNonAtomicVariablesPerThread(). See calculation of NumNonAtomicVariablesPerThread in test_atomics.cpp

3)
Workgroup size is modified depending on the max permitted for the kernel

// tune up work sizes based on kernel info
error = clGetKernelWorkGroupInfo(kernel, deviceID, CL_KERNEL_WORK_GROUP_SIZE, sizeof(groupSize), &groupSize, NULL);
test_error(error, "Unable to obtain max work group size for device and kernel combo");

If the workgroup size changes in step 3 above, the kernel needs to be re-generated. If not, the verify function uses the modified (kernel_max) workgroup size to calculate the reference values whereas the kernel was generated using the original (device_max) workgroup size.

Jeremy-Kemp (approved):
Thanks for the fix @lakshmih .

LGTM, especially as we had this in GitLab. I've sanity checked this on our implementation as well.

bashbaug (approved):
LGTM, especially if these changes were already on GitLab.

bashbaug:
Adding to the OpenCL 3.0 board for more visibility, though this is not a "must fix for OpenCL 3.0" issue.

lakshmih:
Thanks for verifying @Jeremy-Kemp  
@kpet  can you merge this? I'm still not sure about the protocol for merging stuff.

PR #944, opened 2020-09-08
Change Behaviour of C11 Atomic Tests for OpenCL-3.0

FranklandJack:
* Change setup code in `KernelCode()` to use `_explicit` builtin
variants that are common to both OpenCL-2.X and OpenCL-3.0.

* Only test optional supported builtin variants (`_explicit` signature
 memory_order/scope) for OpenCL-3.0.

* Disable program scope global variable and generic address space tests
for a OpenCL-3.0 driver which does not optionally support these
features.

### DIFF
--- a/test_conformance/c11_atomics/common.cpp
+++ b/test_conformance/c11_atomics/common.cpp
@@ -206,3 +206,77 @@ template<> cl_long AtomicTypeExtendedInfo<cl_long>::MaxValue() {return CL_LONG_M
 template<> cl_ulong AtomicTypeExtendedInfo<cl_ulong>::MaxValue() {return CL_ULONG_MAX;}
 template<> cl_float AtomicTypeExtendedInfo<cl_float>::MaxValue() {return CL_FLT_MAX;}
 template<> cl_double AtomicTypeExtendedInfo<cl_double>::MaxValue() {return CL_DBL_MAX;}
+
+cl_int getSupportedMemoryOrdersAndScopes(
+    cl_device_id device, std::vector<TExplicitMemoryOrderType> &memoryOrders,
+    std::vector<TExplicitMemoryScopeType> &memoryScopes)
+{
+    // The CL_DEVICE_ATOMIC_MEMORY_CAPABILITES is missing before 3.0, but since
+    // all orderings and scopes are required for 2.X devices and this test is
+    // skipped before 2.0 we can safely return all orderings and scopes if the
+    // device is 2.X. Query device for the supported orders.
+    if (get_device_cl_version(device) < Version{ 3, 0 })
+    {
+        memoryOrders.push_back(MEMORY_ORDER_EMPTY);
+        memoryOrders.push_back(MEMORY_ORDER_RELAXED);
+        memoryOrders.push_back(MEMORY_ORDER_ACQUIRE);
+        memoryOrders.push_back(MEMORY_ORDER_RELEASE);
+        memoryOrders.push_back(MEMORY_ORDER_ACQ_REL);
+        memoryOrders.push_back(MEMORY_ORDER_SEQ_CST);
+        memoryScopes.push_back(MEMORY_SCOPE_EMPTY);
+        memoryScopes.push_back(MEMORY_SCOPE_WORK_GROUP);
+        memoryScopes.push_back(MEMORY_SCOPE_DEVICE);
+        memoryScopes.push_back(MEMORY_SCOPE_ALL_SVM_DEVICES);
+        return CL_SUCCESS;
+    }
+
+    // For a 3.0 device we can query the supported orderings and scopes
+    // directly.
+    cl_device_atomic_capabilities atomic_capabilities{};
+    test_error(
+        clGetDeviceInfo(device, CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES,
+                        sizeof(atomic_capabilities), &atomic_capabilities,
+                        nullptr),
+        "clGetDeviceInfo failed for CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES\n");
+
+    // Provided we succeeded, we can start filling the vectors.
+    if (atomic_capabilities & CL_DEVICE_ATOMIC_ORDER_RELAXED)
+    {
+        memoryOrders.push_back(MEMORY_ORDER_RELAXED);
+    }
+
+    if (atomic_capabilities & CL_DEVICE_ATOMIC_ORDER_ACQ_REL)
+    {
+        memoryOrders.push_back(MEMORY_ORDER_ACQUIRE);
+        memoryOrders.push_back(MEMORY_ORDER_RELEASE);
+        memoryOrders.push_back(MEMORY_ORDER_ACQ_REL);
+    }
+
+    if (atomic_capabilities & CL_DEVICE_ATOMIC_ORDER_SEQ_CST)
+    {
+        // The functions not ending in explicit have the same semantics as the
+        // corresponding explicit function with memory_order_seq_cst for the
+        // memory_order argument.
+        memoryOrders.push_back(MEMORY_ORDER_EMPTY);
+        memoryOrders.push_back(MEMORY_ORDER_SEQ_CST);
+    }
+
+    if (atomic_capabilities & CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP)
+    {
+        memoryScopes.push_back(MEMORY_SCOPE_WORK_GROUP);
+    }
+
+    if (atomic_capabilities & CL_DEVICE_ATOMIC_SCOPE_DEVICE)
+    {
+        // The functions that do not have memory_scope argument have the same
+        // semantics as the corresponding functions with the memory_scope
+        // argument set to memory_scope_device.
+        memoryScopes.push_back(MEMORY_SCOPE_EMPTY);
+        memoryScopes.push_back(MEMORY_SCOPE_DEVICE);
+    }
+    if (atomic_capabilities & CL_DEVICE_ATOMIC_SCOPE_ALL_DEVICES)
+    {
+        memoryScopes.push_back(MEMORY_SCOPE_ALL_SVM_DEVICES);
+    }
+    return CL_SUCCESS;
+}
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -71,6 +71,10 @@ extern cl_device_atomic_capabilities gAtomicMemCap,
 extern const char *get_memory_order_type_name(TExplicitMemoryOrderType orderType);
 extern const char *get_memory_scope_type_name(TExplicitMemoryScopeType scopeType);
 
+extern cl_int getSupportedMemoryOrdersAndScopes(
+    cl_device_id device, std::vector<TExplicitMemoryOrderType> &memoryOrders,
+    std::vector<TExplicitMemoryScopeType> &memoryScopes);
+
 class AtomicTypeInfo
 {
 public:
@@ -487,16 +491,11 @@ public:
     std::vector<TExplicitMemoryScopeType> memoryScope;
     int error = 0;
 
-    memoryOrder.push_back(MEMORY_ORDER_EMPTY);
-    memoryOrder.push_back(MEMORY_ORDER_RELAXED);
-    memoryOrder.push_back(MEMORY_ORDER_ACQUIRE);
-    memoryOrder.push_back(MEMORY_ORDER_RELEASE);
-    memoryOrder.push_back(MEMORY_ORDER_ACQ_REL);
-    memoryOrder.push_back(MEMORY_ORDER_SEQ_CST);
-    memoryScope.push_back(MEMORY_SCOPE_EMPTY);
-    memoryScope.push_back(MEMORY_SCOPE_WORK_GROUP);
-    memoryScope.push_back(MEMORY_SCOPE_DEVICE);
-    memoryScope.push_back(MEMORY_SCOPE_ALL_SVM_DEVICES);
+    // For OpenCL-3.0 and later some orderings and scopes are optional, so here
+    // we query for the supported ones.
+    test_error_ret(
+        getSupportedMemoryOrdersAndScopes(deviceID, memoryOrder, memoryScope),
+        "getSupportedMemoryOrdersAndScopes failed\n", TEST_FAIL);
 
     for(unsigned oi = 0; oi < memoryOrder.size(); oi++)
     {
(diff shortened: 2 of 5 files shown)


test_conformance/c11_atomics/common.h
```diff
@@ -873,20 +872,23 @@ std::string CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumD
         "  if(get_local_id(0) == 0) // first thread in workgroup\n";
     else
       // global atomics declared in program scope
-      code +=
-      "  if(atomic_fetch_add(&finishedThreads, 1) == get_global_size(0)-1)\n"
-      "    // last finished thread\n";
+      code += "  if(atomic_fetch_add_explicit(&finishedThreads, 1) == "
```
Jeremy-Kemp, line 875: I think the line got a bit muddled, the ordering and scope arguments to the atomic aren't being passed in.

Jeremy-Kemp (changes requested):
Only had a quick skim, will look more tomorrow.

test_conformance/c11_atomics/common.h
```diff
@@ -873,20 +872,23 @@ std::string CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumD
...
+              "get_global_size(0)-1, memory_order_relaxed, "
+              "memory_scope_work_group)\n"
+              "    // last finished thread\n";
```
alycm, line 878: The `memory_order_relaxed` and `memory_scope_work_group` look wrong here, they're free expressions not in the call?

```diff
-if(atomic_fetch_add_explicit(&finishedThreads, 1) == get_global_size(0)-1, memory_order_relaxed, memory_scope_work_group)
+if(atomic_fetch_add_explicit(&finishedThreads, 1, memory_order_relaxed, memory_scope_work_group) == get_global_size(0)-1)
```

If this didn't cause any `clBuildProgram` type errors is this code even traversed?

test_conformance/c11_atomics/common.h
```diff
@@ -807,16 +803,19 @@ std::string CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumD
...
     else
     {
-      code +=
-        "      atomic_store(destMemory+dstItemIdx, finalDest[dstItemIdx]);\n";
+        code += "      atomic_store_explicit(destMemory+dstItemIdx, "
+                "finalDest[dstItemIdx], memory_order_relaxed, "
+                "memory_scope_work_group);\n";
     }
     code +=
       "    }\n"
```
alycm, line 821: Should the barrier on the next line be `CLK_LOCAL_MEM_FENCE` for the `LocalMemory()` case but be `CLK_GLOBAL_MEM_FENCE` otherwise?

Or could just `barrier(CLK_LOCAL_MEM_FENCE | CLK_GLOBAL_MEM_FENCE);` if that's simpler.

test_conformance/c11_atomics/common.h
alycm: Thinking out loud.

It looks like every thread calls `atomic_fetch_add_explicit` with an operand of `1` and the work-item that returns `get_global_size(0)-1` knows it was last so then performs some other action.  The `memory_scope_work_group` is only constraining the ordering of these operations, and `+1` on `finishedThreads` is order invariant, so the `memory_scope_work_group` operation on a global atomic here is just the same as the previous `memory_order_seq_cst`?

The only required response to this comment may be "Yes."

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -230,9 +230,11 @@ class CBasicTestLoad : public CBasicTestMemOrderScope<HostAtomicType, HostDataTy
   {
     std::string memoryOrderScope = MemoryOrderScopeStr();
     std::string postfix(memoryOrderScope.empty() ? "" : "_explicit");
-    return
-      "  atomic_store(&destMemory[tid], tid);\n"
-      "  oldValues[tid] = atomic_load"+postfix+"(&destMemory[tid]"+memoryOrderScope+");\n";
+    return "  atomic_store" + postfix + "(&destMemory[tid], tid"
+        + memoryOrderScope
```
alycm, line 234: I suspect that this is capable of passing `memory_order_acquire` to `atomic_store_explicit`?  If so that is illegal.

If the test is passing `memory_order_acquire` to a `atomic_load_explicit` on the next line then this line should pass `memory_order_release` to the `atomic_store_explicit`.

This will only affect implementations that report `CL_DEVICE_ATOMIC_ORDER_ACQ_REL` (which is likely why it worked for you but is not correct in general).

alycm (commented):
A few things that I think need changed, and a few more comments that are mostly me trying to convince myself that the changes are safe.

test_conformance/c11_atomics/common.h
FranklandJack: Thanks for such quick feedback! Yes, good catch, I guess we were never actually hitting this case, have fixed.

test_conformance/c11_atomics/common.h
```diff
@@ -487,16 +491,13 @@ class CBasicTestMemOrderScope : public CBasicTest<HostAtomicType, HostDataType>
...
-    memoryOrder.push_back(MEMORY_ORDER_ACQ_REL);
-    memoryOrder.push_back(MEMORY_ORDER_SEQ_CST);
-    memoryScope.push_back(MEMORY_SCOPE_EMPTY);
-    memoryScope.push_back(MEMORY_SCOPE_WORK_GROUP);
-    memoryScope.push_back(MEMORY_SCOPE_DEVICE);
-    memoryScope.push_back(MEMORY_SCOPE_ALL_SVM_DEVICES);
+    // For OpenCL-3.0 and later some orderings and scopes are optional, so here
+    // we query for the supported ones.
+    // For OpenCL-3.0 and later some orderings and scopes are optional, so here
+    // we query for the supported ones.
```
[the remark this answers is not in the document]
FranklandJack: Good catch, have fixed.

test_conformance/c11_atomics/common.h
```diff
@@ -807,16 +803,19 @@ std::string CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumD
       "    {\n";
     if(aTypeName == "atomic_flag")
```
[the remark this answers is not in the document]
FranklandJack: Good point, have added your comment.

test_conformance/c11_atomics/common.cpp
```diff
@@ -206,3 +206,90 @@ template<> cl_long AtomicTypeExtendedInfo<cl_long>::MaxValue() {return CL_LONG_M
...
+        memoryScopes.push_back(MEMORY_SCOPE_DEVICE);
+        memoryScopes.push_back(MEMORY_SCOPE_ALL_SVM_DEVICES);
+        return CL_SUCCESS;
+    }
+
+    // For a 3.0 device we can query the supported orderings and scopes
+    // directly.
+    cl_device_atomic_capabilities atomic_capabilities{};
+    test_error(
+        clGetDeviceInfo(device, CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES,
```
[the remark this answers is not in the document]
FranklandJack: I'm not sure, the calling code wasn't setting any fence capabilities before this PR so I don't know if it's needed.

test_conformance/c11_atomics/common.h
FranklandJack: Thanks, I already addressed the arguments missing the call, but yes I think we were not hitting this case when the kernel string was getting built.

FranklandJack: > Thinking out loud.
> 
> It looks like every thread calls `atomic_fetch_add_explicit` with an operand of `1` and the work-item that returns `get_global_size(0)-1` knows it was last so then performs some other action. The `memory_scope_work_group` is only constraining the ordering of these operations, and `+1` on `finishedThreads` is order invariant, so the `memory_scope_work_group` operation on a global atomic here is just the same as the previous `memory_order_seq_cst`?
> 
> The only required response to this comment may be "Yes."

Yes. This sounds reasonable to me. The changes shouldn't make any difference since the `+1` atomic operation is not dependent on order.

test_conformance/c11_atomics/test_atomics.cpp
FranklandJack: Ah good point, before it was always `memory_order_seq_cst`, I'll add some logic to handle these cases.

test_conformance/c11_atomics/common.h
FranklandJack: I'm not entirely sure what the other cases are testing, do you mean [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/c11_atomics/common.h#L841), [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/c11_atomics/common.h#L854) and [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/c11_atomics/common.h#L872).

test_conformance/c11_atomics/test_atomics.cpp
FranklandJack: I've updated this to use `memory_order_release` if the load uses `atomic_order_acquire`. Let me know if this is what you had in mind. It would be super useful if someone could test this on an implementation that supports these orders and confirm that this behaves as expected, unfortunately I can't do this locally.

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -228,11 +229,19 @@ class CBasicTestLoad : public CBasicTestMemOrderScope<HostAtomicType, HostDataTy
...
-    return
-      "  atomic_store(&destMemory[tid], tid);\n"
-      "  oldValues[tid] = atomic_load"+postfix+"(&destMemory[tid]"+memoryOrderScope+");\n";
+      // In the case this test is run with MEMORY_ORDER_ACQUIRE, the store
+      // should be MEMORY_ORDER_RELEASE
+      std::string memoryOrderScopeLoad = MemoryOrderScopeStr();
+      std::string memoryOrderScopeStore =
+          (MemoryOrder() == MEMORY_ORDER_ACQUIRE)
+          ? memoryOrderScopeLoad
+          : ", memory_order_release" + MemoryScopeStr();
```
alycm, line 238: Are these two conditions the wrong way around, i.e. you write `X ? B : A;` but it should be `X ? A : B;` ?

FranklandJack: :facepalm: Updated.

test_conformance/c11_atomics/test_atomics.cpp
alycm: Yes, this is what I had in mind.

test_conformance/c11_atomics/test_atomics.cpp
alycm: Thanks.

test_conformance/c11_atomics/common.h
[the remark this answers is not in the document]
alycm: Thanks.

test_conformance/c11_atomics/common.h
alycm: Thanks.

test_conformance/c11_atomics/common.h
alycm: I was actually referring to the one [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/c11_atomics/common.h#L823).

But my comment no longer makes sense to me.  I think I got mixed up with the nesting, I'll just resolve this point.

test_conformance/c11_atomics/common.cpp
```diff
@@ -206,3 +206,90 @@ template<> cl_long AtomicTypeExtendedInfo<cl_long>::MaxValue() {return CL_LONG_M
...
+        "clGetDeviceInfo failed for CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES\n");
+
+    // Provided we succeeded, we can start filling the vectors.
+    if (atomic_capabilities & CL_DEVICE_ATOMIC_ORDER_RELAXED)
+    {
+        memoryOrders.push_back(MEMORY_ORDER_RELAXED);
+    }
+    else
+    {
+        log_error(
```
Jeremy-Kemp, line 249: Not really an issue at all, but this is also checked during `InitCL`.

test_conformance/c11_atomics/common.cpp
```diff
@@ -206,3 +206,90 @@ template<> cl_long AtomicTypeExtendedInfo<cl_long>::MaxValue() {return CL_LONG_M
...
+        memoryOrders.push_back(MEMORY_ORDER_SEQ_CST);
+    }
+
+    if (atomic_capabilities & CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP)
+    {
+        memoryScopes.push_back(MEMORY_SCOPE_WORK_GROUP);
+    }
+    else
+    {
+        log_error(
```
Jeremy-Kemp, line 276: Not really an issue at all, but this is also checked during `InitCL`.

Jeremy-Kemp (commented):
Yet to run this on an implementation with full capabilities, but this continues to pass on an implementation with a sub-set of capabilities.

FranklandJack:
> Yet to run this on an implementation with full capabilities, but this continues to pass on an implementation with a sub-set of capabilities.

Great! Thanks for testing, is really appreciated.

test_conformance/c11_atomics/common.cpp
FranklandJack: Okay, I think in that case I will remove this. Since this is a utility function it seems a little weird to have it return `TEST_FAIL`, so if this can be avoided we should avoid it.

FranklandJack: Removed.

test_conformance/c11_atomics/common.cpp
FranklandJack: Removed, see above.

alycm (approved):

test_conformance/c11_atomics/common.h
```diff
@@ -487,16 +491,11 @@ class CBasicTestMemOrderScope : public CBasicTest<HostAtomicType, HostDataType>
...
-    memoryOrder.push_back(MEMORY_ORDER_ACQ_REL);
-    memoryOrder.push_back(MEMORY_ORDER_SEQ_CST);
-    memoryScope.push_back(MEMORY_SCOPE_EMPTY);
-    memoryScope.push_back(MEMORY_SCOPE_WORK_GROUP);
-    memoryScope.push_back(MEMORY_SCOPE_DEVICE);
-    memoryScope.push_back(MEMORY_SCOPE_ALL_SVM_DEVICES);
+    // For OpenCL-3.0 and later some orderings and scopes are optional, so here
+    // we query for the supported ones.
+    test_error_ret(
+        getSupportedMemoryOrdersAndScopes(deviceID, memoryOrder, memoryScope),
```
bashbaug, line 497: I reviewed another PR that used a similar pattern earlier today.  In short, I think this will work, but as-written the test_error_ret macro will call the function twice if the first call failed, which might not be expected.

If we're going to use this pattern more regularly then I think we should modify the macro so it assigns to a temporary error code so this pattern will only call the function once.

This doesn't need to be done as part of this PR.

test_conformance/c11_atomics/common.h
```diff
@@ -800,23 +794,31 @@ std::string CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumD
...
+              "    {\n";
+      if (aTypeName == "atomic_flag")
+      {
+          code += "      if(finalDest[dstItemIdx])\n"
+                  "        "
+                  "atomic_flag_test_and_set_explicit(destMemory+dstItemIdx, "
+                  "memory_order_relaxed, memory_scope_work_group);\n"
+                  "      else\n"
+                  "        atomic_flag_clear_explicit(destMemory+dstItemIdx, "
+                  "memory_order_relaxed, memory_scope_work_group);\n";
```
bashbaug, line 815: I know the formatting was done by clang-format, but I wonder if these strings would be easier to read if they were raw string literals?  No action required, just pondering ways to make these tests easier to understand.

bashbaug (dismissed):
LGTM.  I've successfully ran these changes on our implementation coming from OpenCL 2.0.

test_conformance/c11_atomics/common.h
FranklandJack: Good point! Yes, definitely, I have pushed a patch that replace the strings touched in this PR with raw strings.

test_conformance/c11_atomics/common.h
FranklandJack: Thanks for the review! Ah yes, good catch, it's no problem to update the `test_error_ret` macro as part of this MR, I have pushed a patch assigning the parameter to a temporary value so if it is a function call it will only happen once.

FranklandJack:
> LGTM. I've successfully ran these changes on our implementation coming from OpenCL 2.0.

Great! Thanks for testing this out, is really appreciated!

bashbaug (approved):
Thanks!  The change to the test_error_ret macro should help another in-flight PR also.

bashbaug:
@Jeremy-Kemp, can you please give this PR another quick look to see if your requested changes have been addressed?

Jeremy-Kemp (approved):
Sorry, was away last week and didn't get a chance to review before leaving.

LGTM though, thanks!

Issue #943, opened 2020-09-08
Memory leak of test_basic subcase async_copy_global_to_local etc. in test_async_strided_copy.cpp.


letidealsfly:
Hello,

https://github.com/KhronosGroup/OpenCL-CTS/blob/6b1e61f9dedd852a8a33cd97bb0f9bca98e8995a/test_conformance/basic/test_async_strided_copy.cpp#L218

Before `return -1`, there should free inBuffer/outBuffer otherwise it will cause memory leak while verify failed.

Qinyu,
Thanks,

PR #948, opened 2020-09-09
Fix memory leak of test_basic (#943)

letidealsfly:
Fix issue https://github.com/KhronosGroup/OpenCL-CTS/issues/943.

Free buffer while verify failed.

Signed-off-by: Qinyu.Tang <qinyu.tang@verisilicon.com>

### DIFF
--- a/test_conformance/basic/test_async_strided_copy.cpp
+++ b/test_conformance/basic/test_async_strided_copy.cpp
@@ -215,6 +215,8 @@ int test_strided_copy(cl_device_id deviceID, cl_context context, cl_command_queu
                 sprintf(values + strlen( values), "%2x ", outchar[j]);
             sprintf(values + strlen(values), "]");
             log_error("%s\n", values);
+            free(inBuffer);
+            free(outBuffer);
             return -1;
         }
     }


test_conformance/basic/test_async_strided_copy.cpp
```diff
@@ -215,6 +215,8 @@ int test_strided_copy(cl_device_id deviceID, cl_context context, cl_command_queu
                 sprintf(values + strlen( values), "%2x ", outchar[j]);
             sprintf(values + strlen(values), "]");
             log_error("%s\n", values);
+            free(inBuffer);
```
Jeremy-Kemp, line 218: These two allocations could also leak from any early return of `test_error`. Might be better to replace with a vector instead so that there is no chance of leaking from an early return.

letidealsfly: Well, it seems most of the cases have the same problems.
If using `malloc`, might be better to use `goto` in `test_error` or replace all with `vector`.

kpet: Yeah a lot of CTS code doesn't handle error cases well. What @Jeremy-Kemp is suggesting is better but I'd be fine with accepting this patch as is since it fixes the issue for the case where the test runs successfully.

kpet: Agree we should use vectors but probably better to do it for all tests. I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/1033.

kpet (approved):

kpet:
Merging since it fixes one common case (the test runs to completion but fails) and the need for further fixes is tracked in an issue.

PR #950, opened 2020-09-10
Add async_build test to list of tests to skip during offline compilat…

einhov:
…ion mode

The test attempts to build a program from source containing invalid code, to
provoke a build failure. In offline compilation mode, this causes the offline
compiler to give an error, failing the test.


kpet (approved):

alycm (approved):
LGTM.

PR #952, opened 2020-09-10
Remove uninitialised input buffer from test_zero_sized_enqueue

dgavedissian:
If I understand correctly, `test_conformance/api/test_zero_sized_enqueue.cpp` tries to execute a kernel with an NDRange that contains a 0, and asserts whether the output buffer was changed to anything other than 0.

Originally, the kernel was copying data from an input buffer to "clobber" the output buffer. However, this input buffer was initialised, so this test wouldn't fail reliably. This PR removes the input buffer completely, replacing the copy operation in the kernel with just `dst[i] = 1;`. I believe this has the same effect.


kpet (approved):
LGTM. This also makes the test easier to understand.

PR #953, opened 2020-09-10
Use Version type for cl_khr_spir extension version checks

alycm:
The current method of using `std::find(versions, 1.2f)` is unreliable due to
performing a floating point comparison.  I.e. on some compiles it will
spuriously report that the required version is missing and skip the tests.

### DIFF
--- a/test_conformance/spir/main.cpp
+++ b/test_conformance/spir/main.cpp
@@ -141,7 +141,8 @@ static bool is_dir_exits(const char* path)
     return false;
 }
 
-static void get_spir_version(cl_device_id device, std::vector<Version>& versions)
+static void get_spir_version(cl_device_id device,
+                             std::vector<Version> &versions)
 {
     char version[64] = {0};
     cl_int err;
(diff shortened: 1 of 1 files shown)


gwawiork (approved):
LGTM

test_conformance/spir/main.cpp
```diff
@@ -162,11 +163,11 @@ static void get_spir_version(cl_device_id device, std::vector<float>& versions)
...
               std::istream_iterator<std::string>(),
               std::back_inserter(versionVector));
-    for(std::list<std::string>::const_iterator it = versionVector.begin(),
-                                               e  = versionVector.end(); it != e;
-                                               it++)
+    for (auto &v : versionVector)
     {
-        versions.push_back(atof(it->c_str()));
+        auto major = v[v.find('.') - 1];
+        auto minor = v[v.find('.') + 1];
```
kpet, line 169: I hope that by OpenCL 10 or x.10 we will have killed SPIR :).

kpet (approved):

PR #954, opened 2020-09-11
Skip atomic_flag Test on for OpenCL-3.0 not Supporting Device Scope

FranklandJack:
* The `atomic_flag` test assumes support for the `atomic_scope_device`
scope in the global scope test case. Since `atomic_scope_device` is
optional on an OpenCL-3.0 driver, this test should check for support and
skip otherwise.

### DIFF
--- a/test_conformance/c11_atomics/test_atomics.cpp
+++ b/test_conformance/c11_atomics/test_atomics.cpp
@@ -1615,6 +1615,30 @@ public:
       orderStr = std::string(", ") + get_memory_order_type_name(MemoryOrderForClear());
     return orderStr + MemoryScopeStr();
   }
+
+  virtual int ExecuteSingleTest(cl_device_id deviceID, cl_context context,
+                                cl_command_queue queue)
+  {
+      // This test assumes support for the memory_scope_device scope in the case
+      // that LocalMemory() == false. Therefore we should skip this test in that
+      // configuration on a 3.0 driver since supporting the memory_scope_device
+      // scope is optionaly.
+      if (get_device_cl_version(deviceID) >= Version{ 3, 0 })
+      {
+          if (!LocalMemory()
+              && !(gAtomicFenceCap & CL_DEVICE_ATOMIC_SCOPE_DEVICE))
+          {
+              log_info(
+                  "Skipping atomic_flag test due to use of atomic_scope_device "
+                  "which is optionally not supported on this device\n");
+              return 0; // skip test - not applicable
+          }
+      }
+      return CBasicTestMemOrderScope<HostAtomicType,
+                                     HostDataType>::ExecuteSingleTest(deviceID,
+                                                                      context,
+                                                                      queue);
+  }
   virtual std::string ProgramCore()
   {
     std::string memoryOrderScope = MemoryOrderScopeStr();


test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1615,6 +1615,30 @@ class CBasicTestFlag : public CBasicTestMemOrderScope<HostAtomicType, HostDataTy
...
+      // scope is optionaly.
+      if (get_device_cl_version(deviceID) >= Version{ 3, 0 })
+      {
+          if (!LocalMemory()
+              && !(gAtomicMemCap & CL_DEVICE_ATOMIC_SCOPE_DEVICE))
+          {
+              log_info(
+                  "Skipping atomic_flag test due to use of atomic_scope_device "
+                  "which is optionally not supported on this device\n");
+              return TEST_SKIPPED_ITSELF;
```
alycm, line 1634: All the other `ExecuteSingleTest` have code like this:

```c
    if (CheckCapabilities(MemoryScope(), MemoryOrder()) == TEST_SKIPPED_ITSELF)
        return 0; // skip test - not applicable
```

This has two differences to the code here.  The use of `CheckCapabilities(...)`, and `return 0;`.

Does the dependence on `!LocalMemory()` prevent use of `CheckCapabilities(...)`?

As for the return code, I'm not sure if `TEST_SKIPPED_ITSELF` is valid here but `0` would be more consistent with the rest of the tests.

FranklandJack: Thanks for the review! So I'm not 100% certain, but I think that `MemoryScope()` and `MemoryOrder()` are independent of `LocalMemory()`. For a device that doesn't support `memory_scope_device` `MemoryScope()` shouldn't ever return `memory_scope_device` (after #944 is merged). But I think `LocalMemory()` is getting set separately and this is used to select the scope in the kernel source, so this is what we need to use in this case to optionally skip this test?

FranklandJack: Updated to return `0` to be consistent with other tests.

bashbaug (dismissed):
I think this is fine, but it would be good to get another set of eyes on it.

I initially thought this check was disabling more tests than it needed to, but it looks like the `atomic_flag` test kernels for global memory also include calls to acquire and release `atomic_work_item_fence` with memory scope device.

Does this mean that these tests can run on devices that support device scope for fences (not for memory operations)?

Similarly, do we also need to add checks for acquire and release fence support?  I suppose it would be strange to  to only support relaxed fences, and acquire and release fences are currently required for OpenCL 3.0 devices, so in practice the check for device scope is sufficient.

FranklandJack:
Thanks for your review!

> I think this is fine, but it would be good to get another set of eyes on it.
> 
> I initially thought this check was disabling more tests than it needed to, but it looks like the `atomic_flag` test kernels for global memory also include calls to acquire and release `atomic_work_item_fence` with memory scope device.
> 
> Does this mean that these tests can run on devices that support device scope for fences (not for memory operations)?
Good point, I didn't think about that. I will update the query to check against `gAtomicFenceCap` rather than `gAtomicMemCap`.

> Similarly, do we also need to add checks for acquire and release fence support? I suppose it would be strange to to only support relaxed fences, and acquire and release fences are currently required for OpenCL 3.0 devices, so in practice the check for device scope is sufficient.

I'm not sure, as far as I understand the `memory_order_acquire` and `memory_order_release` are required by OpenCL-3.0 for the `atomic_fence` builtin, but if they became optional in future versions of OpenCL, this test would need to be updated. Is that what you were asking, or did I misunderstand the question?

FranklandJack:
> After this change we are properly checking that the device scope is supported for `atomic_work_item_fence`. We are not currently checking that the acquire and release memory orders are supported for fences, but that's OK, because they are currently required for all OpenCL 2.x and newer devices, including OpenCL 3.0 devices.

Thanks for the example, this really makes the explanation clear. Since the orders are currently required for all >= 2.X devices, are you happy to leave this PR as is, or would you prefer that we somehow assert that these orders are supported for memory fences in this case?

bashbaug:
> are you happy to leave this PR as is?

Yeah, I'm find with this PR as-is.  If we ever relax the required atomic fence orders further then I'm sure we'll need to modify more tests than this one!

bashbaug (approved):
Approving as discussed in the September 22 teleconference.

Issue #22, opened 2018-11-09
Many in-tree SPIR-V files don't pass the validator


kpet:

StuartDBrady:
Most of these have been fixed by PR #272 and PR #472 (and possibly also by fixes to the SPIR-V validator).

I have raised issue 3322 in the SPIRV-Tools repository against the remaining validation failure in decorate_nonwritable.spvasm32, decorate_nonwritable.spvasm64. Provided it is accepted that this is a bug in the SPIR-V validator, and not in the tests, this issue can be closed.

StuartDBrady:
It has been determined that the SPIR-V validator was correct in reporting a validation failure in decorate_nonwritable.spv\*, and that it currently fails to report validation failures in decorate_nonreadable.spv\*, decorate_coherent.spv\* and  decorate_volatile.spv\*. As such, I have raised PR #958 to remove these.

kpet:
@StuartDBrady Isn't this now fixed?

StuartDBrady:
@kpet, yes, this is fixed.

PR #958, opened 2020-09-15
Remove invalid SPIR-V image decoration tests (#22)

StuartDBrady:
The Volatile, Coherent, NonWritable and NonReadable decorations can only be applied to images if they are storage images, which require the Shader capability.

Note that in SPIR-V 1.4 and later, the NonWritable decoration can also be applied to objects in the Private and Function storarge classes (although Private also requires the Shader capability), but this was not covered by the existing test for the NonWritable decoration.


kpet (dismissed):
LGTM, this was discussed on https://github.com/KhronosGroup/SPIRV-Tools/issues/3322.

kpet (approved):

bashbaug (approved):
These tests are invalid so removing them seems like the right thing to do.

Do we know why they were added in the first place?  Should we (eventually, not in this PR) add modified tests so we don't lose coverage?  Or are these tests entirely unnecessary and therefore no action is required beyond removal?

StuartDBrady:
> These tests are invalid so removing them seems like the right thing to do.
> 
> Do we know why they were added in the first place? Should we (eventually, not in this PR) add modified tests so we don't lose coverage? Or are these tests entirely unnecessary and therefore no action is required beyond removal?

The commit that added the tests is 05a11d8e499aca1efa1b91f07a9da5c74f6d1214 "Synchronise with Khronos-private Gitlab branch", so I can't say much here.

The original commit seems to be 26bd31d46ae66a6ec96cdf47da408ffd6605f902. I won't comment further on it, other than to say that it added these four image tests, and made no other changes, and there does not seem to have been any detailed rationale provided for the tests that I can see.

PR #959, opened 2020-09-15
Print the latest conformance version passed reported by the device

### LINKED ISSUE #798: Log version of the CTS reported by OpenCL 3 devices
Make sure the value reported via https://github.com/KhronosGroup/OpenCL-Docs/issues/304 is logged so that it is captured as part of official submission packages.

kpet:
Fixes #798

### DIFF
--- a/test_common/harness/deviceInfo.cpp
+++ b/test_common/harness/deviceInfo.cpp
@@ -24,8 +24,8 @@
 
 /* Helper to return a string containing device information for the specified
  * device info parameter. */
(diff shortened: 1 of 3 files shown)


gwawiork (approved):
LGTM

bashbaug (approved):
LTGM!

At some point we should move the other device header queries to use get_device_info_string(), but we don't need to do that now.

Jeremy-Kemp (approved):
It'd be good to also validate the string once the format has been agreed.

PR #960, opened 2020-09-15
simple vector swizzle test

### LINKED ISSUE #891: Test Coverage for .rgba, .xyzw, and .s0123... swizzles
This issue is to track a new test for vector swizzles, for the new .rgba swizzles in OpenCL C 3.0 and the existing .xyzw and .s0123... swizzles.  This was originally tracked by the item "Add a test for vector swizzles, including .xyzw, .rgbz, and .s012... swizzling" in #773.

One of the things we need to decide is how much test coverage we require for adequate test coverage.  For vectors up to size four, we can probably test all combinations of swizzles for all three swizzle formats(4^4 == 256), but for 8-component and 16-component vectors exhaustive testing is probably impossible.

bashbaug:
Fixes #891.

This is a simple test for vector swizzles.

It doesn't exhaustively test all swizzle combinations, or even most of them, but it does test `.xyzw`, `.s0123`, and `.rgba` swizzles (OpenCL 3.0 only) for all types and vector sizes, on both the left hand side of an assignment ("lvalue swizzles") and the right hand side of an assignment ("rvalue swizzles").


test_conformance/basic/test_vector_swizzle.cpp
```diff
@@ -0,0 +1,677 @@
...
+    int hasDouble = is_extension_available(device, "cl_khr_fp64");
+
+    int result = TEST_PASS;
+    result |= test_type<cl_char>("char", device, context, queue);
+    result |= test_type<cl_uchar>("uchar", device, context, queue);
+    result |= test_type<cl_short>("short", device, context, queue);
+    result |= test_type<cl_ushort>("ushort", device, context, queue);
+    result |= test_type<cl_int>("int", device, context, queue);
+    result |= test_type<cl_uint>("uint", device, context, queue);
+    result |= test_type<cl_long>("long", device, context, queue);
```
gwawiork, line 669: Please add conditionally this data type - gHasLong from testHarness

gwawiork (changes requested):
Looks good only need to add verification long/ulong data types. 
The test could be extended later  to swizzles .lo .hi .even .odd

test_conformance/basic/test_vector_swizzle.cpp
bashbaug: Added - thanks!

gwawiork: but `cl_float` should not be there ...

bashbaug: Oh dear, that's an unfortunate copy-paste error.  Fixed!

gwawiork (dismissed):
LGTM

Dithermaster:
Ben, these look good to me! For more exhaustive tests we'd want to use a code generator.
---Dennis

Dithermaster (dismissed):
Ben, these look good to me! For more exhaustive tests we'd want to use a code generator.
---Dennis

test_conformance/basic/test_vector_swizzle.cpp
```diff
@@ -0,0 +1,680 @@
...
+
+    return result;
+}
+
+template <typename T>
+static int test_type(const char* type_name, cl_device_id device,
+                     cl_context context, cl_command_queue queue)
+{
+    return test_vectype<T, 2>(type_name, device, context, queue)
+        | test_vectype<T, 3>(type_name, device, context, queue)
```
lakshmih, line 651: Qualcomm has concerns with the SetKernelArg usage in this PR, specifically for 3-component vectors. 

As per OpenCL-C spec (page 11) "For 3-component vector data types, the size of the data type is 4 * sizeof(component)". Furthermore, from the OpenCL-API spec for clSetKernelArg error conditions, "CL_INVALID_ARG_SIZE if arg_size does not match the size of the data type for an argument that is not a memory object"

This would imply that for 3-component vectors, the size passed to SetKernelArg should be 4 * sizeof(component).

bashbaug: Ah, good catch.  I'm not sure why this wasn't failing on our implementation.

I just pushed a commit that uses the special `vector_size` instead of the number of components in the vector.  Can you please check that this fixes this issue?

lakshmih: Yes, the vector_swizzle test now passes with your patch. Thanks for fixing.

test_conformance/basic/test_vector_swizzle.cpp
```diff
@@ -0,0 +1,681 @@
...
+
+    // lvalue swizzles
+    dst[index++].x = value.x;
+    dst[index++].y = value.x;
+    dst[index++].z = value.x;
+    dst[index++].xyz = value;
+    dst[index++].zyx = value;
+
+    // rvalue swizzles
+    vstore3(value.x, 0, (BASETYPE*)(dst + index++));
```
alycm, line 104: I think for devices that do not support generic adress space all these `(BASETYPE*)` casts should have a `__global` because `dst` is `__global`?

```suggestion
    vstore3(value.x, 0, (__global BASETYPE*)(dst + index++));
```
I'm seeing errors such as

```
kernel.opencl:13:25: error: casting '__global char3 *' to type '__private char *' changes address space of pointer
    vstore3(value.x, 0, (BASETYPE*)(dst + index++));
                        ^          ~~~~~~~~~~~~~~~
```

alycm (changes requested):
LGTM in general, one issue based on running the code, no changes based on reading the code.

test_conformance/basic/test_vector_swizzle.cpp
bashbaug: Good catch - I've fixed all of these.

alycm (approved):
I've rerun these tests and they pass for me now.

test_conformance/basic/test_vector_swizzle.cpp
alycm: Thanks!

alycm:
I'm going to merge this, at least three people have reviewed the code and at least two companies in addition to @bashbaug have run the tests.

PR #961, opened 2020-09-16
add a test case for subbuffer property queries

### LINKED ISSUE #914: Extend CL_MEM_PROPERTIES tests for sub-buffers
See https://github.com/KhronosGroup/OpenCL-Docs/issues/412 and https://github.com/KhronosGroup/OpenCL-Docs/issues/414

bashbaug:
Fixes #914.

This adds several test cases to the `buffer_properties_queries` and `image_properties_queries` tests, for standard buffers, sub-buffers, and standard images.  For these objects, there should be no queried properties, or in other words the queried property size should be zero.

I've also made a small improvement to compile the kernel once and reuse it for all sub-tests vs. compiling it for each sub-test, which should improve the test execution time slightly.

### DIFF
--- a/test_conformance/api/test_mem_object_properties_queries.cpp
+++ b/test_conformance/api/test_mem_object_properties_queries.cpp
@@ -22,16 +22,18 @@
 typedef enum
 {
     image,
-    buffer
-} mem_obj_type;
+    image_with_properties,
+    buffer,
+    buffer_with_properties,
+    subbuffer,
+} test_type;
 
 struct test_data
 {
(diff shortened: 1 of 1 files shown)


gwawiork (approved):
LGTM

bcalidas:
Qualcomm has tested this PR and it passed for us.

Jeremy-Kemp (approved):
Admittedly haven't performed an in-depth review, but its passing on our implementation and looks good having had a quick look.

kpet (approved):

PR #962, opened 2020-09-16
device_execution: Modify the check in profiling subtest

lakshmih:
As per the spec CL_PROFILING_COMMAND_COMPLETE is required to
be equal to CL_PROFILING_COMMAND_END only on devices that
do not support the on-device queue.

For other cases it can be >= CL_PROFILING_COMMAND_END


gwawiork (approved):
LGTM, please fix only formatting issues

kpet (approved):
LGTM. Since this was agreed in the teleconference, I'll merge it straight away.

PR #964, opened 2020-09-16
compiler: Modify the IL check code to check for an empty string

lakshmih:
The called function (get_device_info_string) subtracts 1 from
length before returning it. Hence, to check for an empty string
(that the spec requires the implementation to return when
CL_DEVICE_IL_VERSION is not supported) the check should be modified
to check for empty and not length == 1


bashbaug (approved):
Looks good to me.  I also did a quick grep through the rest of the CTS source and didn't see any other places with a similar bug.

alycm (approved):
LGTM, this is the second review so I will merge now.

Issue #894, opened 2020-08-14
Make sure support for read/write images isn't assumed on CL3


kpet:
With regards to the read-write image support, the test I mentioned in the call yesterday that we know is failing / being run unconditionally for a 3.0 device that optionally does not support read-write images is: `rw_image_access_qualifier` from `basic`. Since I couldn't see a ticket created for this yet I thought I'd put it here before I forget.

_Originally posted by @FranklandJack in https://github.com/KhronosGroup/OpenCL-CTS/issues/773#issuecomment-673955962_

nikhiljnv:
Thanks @kpet  and @FranklandJack for filing the issue !!

I was looking through CTS today and also tried running read_write image tests.

Currently we are checking for OpenCL version < 2.0 to skip the test.
A couple of other tests that may need fixes are -
images/kernel_read_write/
images/samplerlessReads/

https://github.com/KhronosGroup/OpenCL-CTS/blob/ed50fcad2daf98887146122629a8a8e403ecef5c/test_conformance/images/kernel_read_write/main.cpp#L181

FranklandJack:
I have upstreamed a local patch #922 skipping the `rw_image_access_qualifier` test in the cases that images aren't supported, or if on a 3.0 device images are supported but read-write images are optionally not supported. Hope this hasn't duplicated anyone else's work, but I thought since I had it I may as well upstream it.

nikhiljnv:
@FranklandJack Thanks for the PR.
I started looking at this last week, but didn't have any changes ready.
There are couple of other tests I mentioned earlier. So, I will send out PR for those separately.

PR #965, opened 2020-09-17
Add check for Read-Write images support 

nikhiljnv:
Read-Write images have required OpenCL 2.x.
    Read-Write image tests are already being skipped
    for 1.x devices.
    With OpenCL 3.0, read-write images being optional,
    the tests should be run or skipped
    depending on the implementation support.

    Add a check to decide if Read-Write images are
    supported or required to be supported depending
    on OpenCL version and decide if the tests should
    be run or skipped.

    Fixes issue #894


test_common/harness/kernelHelpers.cpp
```diff
@@ -1467,6 +1467,43 @@ int checkFor3DImageSupport( cl_device_id device )
...
+        return CL_IMAGE_FORMAT_NOT_SUPPORTED;
+    }
+
+    auto device_cl_version = get_device_cl_version(device);
+    if (device_cl_version >= Version(3, 0))
+    {
+        // In OpenCL 3.0, Read-Write images are optional.
+        // Check if they are supported.
+        cl_uint are_rw_images_supported{};
+        test_error(
```
bashbaug, line 1483: Minor: I think this will work OK, but the usual pattern is to assign to a separate error code and then check it vs. embedding the call into the macro.  Because of the way the macro is written, I think this will call the function twice if the first call returns an error, which probably isn't intended.

bashbaug (approved):
Mostly LGTM - thanks!

One small comment I would consider changing, but it's not a must-fix.

test_common/harness/kernelHelpers.cpp
bashbaug: Update: no action required here.  PR #944 will fix the macro so the function only gets called once with this pattern.  Thanks!

kpet (approved):

PR #967, opened 2020-09-21
conversions: Use ARM emulation for aarch64

lakshmih:
The host compiler will not always calculate reference values
the same, depending on optimization level.  It generates
instructions that do not respond to CPU rounding mode in
the same way.  Use QCOM rounding mode emulation to correctly
calculate reference values on aarch64.

Change-Id: I4252786da85f77f8f8fcc79c2b8881a2c889e482

### DIFF
--- a/test_conformance/conversions/CMakeLists.txt
+++ b/test_conformance/conversions/CMakeLists.txt
@@ -4,7 +4,7 @@ set (${MODULE_NAME}_SOURCES
       Sleep.cpp test_conversions.cpp basic_test_conversions.cpp
 )
 
-if("${CLConform_TARGET_ARCH}" STREQUAL "ARM")
+if("${CLConform_TARGET_ARCH}" STREQUAL "ARM" OR "${CLConform_TARGET_ARCH}" STREQUAL "ARM64")
     list(APPEND ${MODULE_NAME}_SOURCES fplib.cpp)
 endif()
 
--- a/test_conformance/conversions/basic_test_conversions.cpp
+++ b/test_conformance/conversions/basic_test_conversions.cpp
@@ -21,11 +21,11 @@
 
 #include "harness/mt19937.h"
 
-#if defined( __arm__ ) && defined( __GNUC__ )
+#if (defined(__arm__) || defined(__aarch64__)) && defined(__GNUC__)
 #include "fplib.h"
 #endif
 
-#if defined( __arm__ ) && defined( __GNUC__ )
+#if (defined(__arm__) || defined(__aarch64__)) && defined(__GNUC__)
 /* Rounding modes and saturation for use with qcom 64 bit to float conversion library */
     bool            qcom_sat;
     roundingMode    qcom_rm;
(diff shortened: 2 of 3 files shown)


kpet:
Hmm, this looks suspicious. Do you have an example of the code sequences you mentioned? What toolchain is showing the problem?

lakshmih:
> Hmm, this looks suspicious. Do you have an example of the code sequences you mentioned? What toolchain is showing the problem?

The tool-chain in use is android ndk 21.3.6528147.

The problem is in the function ulong2float, for which our implementation has remained unchanged. However, the host-side reference computes differently when the code is compiled in release vs debug mode.
The code uses a typecast to convert a cl_ulong to float via “(float)l”.  In debug mode, this is done via the instruction “ucvtf   s0, x8”, which converts the 64 bit unsigned int in x8 to the 32 bit float in s0.  
When built in release mode, ulong2float gets vectorized in a loop, and the conversion is done in a two-step process:
“
ucvtf   v0.2d, v0.2d
ucvtf   v1.2d, v1.2d
fcvtn   v0.2s, v0.2d
fcvtn   v1.2s, v1.2d
“
In this context each ucvtf instruction converts two 64 bit ulongs to two 64 bit doubles.  The fcvtn then converts the 64 bit double to 32 bit floats.  This produces slightly different results from the single step process.

When the input ulong is 18446743523953737727u, a direct conversion to float (like in debug mode) results in 18446742974197923840.0.  Via the two-step vectorized process (in release mode) we see the result 18446744073709551616.0.

•	Our GPU implementation produces 18446742974197923840 which matches the direct conversion case.
•	gcc on x64 Linux also produces 18446742974197923840

EwanC:
Unsure of all the implications of this solution, rather than an approach like disabling the optimization in Release mode. However, I tried this branch with our `aarch64` implementation and it passed.

lakshmih:
We have a strong preference for a code based solution over a Makefile based one since we think the former will be more robust. A Makefile based solution is likely to get overwritten in the future and also vulnerable to divergence on different implementations.

bashbaug:
I don't have much to add here, but I have confirmed that our CPU and GPU implementations both convert ulong 18446743523953737727u to float 18446742974197923840.0.

If the reference value is being computed as 18446744073709551616.0 then that seems wrong.

See also https://godbolt.org/z/ordrjW, which shows that the conversion from ulong -> double -> float is not equivalent to a conversion from ulong -> float directly.

jlewis-austin:
It looks fine to me, prioritizing correctness over performance. Just for reference, here's what's happening with the double rounding case here:
integer value: `0xFFFFFF7F FFFFFFFF`
rounded (up) to double 53-bit significand: `0xFFFFFF80 00000000`
rounded (up to nearest even) to float 24-bit significand (value was exactly halfway):  ` 0x1 00000000 00000000`

Rounding directly to 24-bit float finds a 0 in the 25th bit position, so it rounds down to `0xFFFFFF00 00000000`

lakshmih:
@kpet Could you please merge this since there is agreement on the fix. 
Thanks.

kpet (approved):
Looks reasonable to me.

bashbaug:
This has been open for a while and has one approval and no other comments.  Merging.

jlewis-austin:
@lakshmih I just noticed that flib.cpp doesn't implement emulation for RTE mode, but does a cast like the original codepath. Can you verify that your problem was confined to non-RTE modes? We're seeing incorrect reference values on aarch64 with RTE mode.

PR #969, opened 2020-09-22
A first set of updates to the submission template

kpet:

### DIFF
--- a/test_conformance/submission_details_template.txt
+++ b/test_conformance/submission_details_template.txt
@@ -1,5 +1,5 @@
 ##############################################################################
-# OpenCL 2.0 submission details template
+# OpenCL submission details template
 # $Id $
 # $URL $
 ##############################################################################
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
This generally looks good to me.

Regarding removal of `CL_DEVICE_MAX_CLOCK_FREQUENCY`, I think this is fine, but I'd consider removing many (most? all?) of the other device queries also.  The current list of device queries seems fairly arbitrary, and the comprehensive list of device queries can be found in the `test_computeinfo` output if needed.

kpet:
Yes, as discussed in the 2020/09/23 teleconference, removing only this one was arbitrary and I used it as a discussion starter since it seemed the most out of place. Still as discussed in the teleconference, I want us to remove all of these as the conformance process means that the value will be captured in the test logs anyway.

Jeremy-Kemp (approved):
Echo what @bashbaug said.

I don't think it should block this PR though.

kpet:
Two approvals from two implementers. Merging to allow further changes on top.

PR #949, opened 2020-09-09
fix gles build error

ijkmn:
1. gles 3.x also define GL_ES_VERSION_2_0
2. ICD don't export clGetGLContextInfoKHR, should get extension function address
3. remove ' include gl2extQCOM.h '

reopen PR of #651

PR #976, opened 2020-09-24
gles: Fix compile errors.

lakshmih:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -197,6 +197,10 @@ if(LINK_PTHREAD)
     list(APPEND CLConform_LIBRARIES pthread)
 endif()
 
+if(DEFINED USE_GLES3)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DGLES3")
+endif()
+
 if(APPLE)
     find_library(corefoundation CoreFoundation)
     find_library(iokit IOKit)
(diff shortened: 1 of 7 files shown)


gwawiork:
I am trying to compile this on Windows using MSVC but it fails with error:
`cannot open source files: EGL/egl.h`
Where to get these headers and where they should be placed ?
Any additional pre-setup need to be done ?

gwawiork:
> > I am trying to compile this on Windows using MSVC but it fails with error:
> > `cannot open source files: EGL/egl.h`
> > Where to get these headers and where they should be placed ?
> > Any additional pre-setup need to be done ?
> 
> "EGL/egl.h" and the libraries EGL and GLESv2 (with OS specific prefixes/suffixs in the file names) need to be provided by the vendor doing the testing in the same way that the CL equivalents need to be provided. This can be done by putting the EGL headers and libraries alongside the CL variants referred to by CL_INCLUDE_DIR and CL_LIB_DIR CMake variables.

ok Thanks. I could not find any lib for Windows so I suppose this test suite is not supported on Windows. On Linux I got errors which showed me that the PR #949 have to be merged first.

Jeremy-Kemp (commented):
I've tested this out on an implementation that doesn't support gles and it does not cause any issues.

gwawiork (dismissed):
LGTM

kpet (approved):

kpet:
@lakshmih Could you rebase this so it gets tested by the now-fixed CI please?

kpet (approved):

kpet:
Merging since this was reviewed several times already.

Issue #977, opened 2020-09-24
Issue with unload_platform_compiler test when platform is NULL


fhazubski-Intel:
The test `unload_invalid` uses a call `clUnloadPlatformCompiler(nullptr)` which does not seem to have a unified behavior across the implementations. In the [OCL-dev/ocl-icd](https://github.com/OCL-dev/ocl-icd) project the NULL platform is replaced with a default one, causing the API call to return `CL_SUCCESS` which causes the `unload_invalid` test to fail. This seems to be valid as cl_khr_icd extension [mentions](https://github.com/KhronosGroup/OpenCL-Docs/blob/master/ext/cl_khr_icd.asciidoc#L334-L344) that NULL platform can be replaced with a default one and this is an expected behavior.

However in the spec, this behavior (replacing NULL platform with the valid one) seem to be allowed for `clGetPlatformInfo`:
> _platform_ refers to the platform ID returned by clGetPlatformIDs or can be NULL. If _platform_ is NULL, the behavior is implementation-defined.

But it does not seem to be allowed for `clUnloadPlatformCompiler`:
> _platform_ is the platform to unload.

This is a bit confusing and probably requires a further discussion so for now I would propose updating the problematic test to behave consistently.

Test: https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_unload_platform_compiler.cpp#L395

PR #978, opened 2020-09-24
temporarily disable the unload_invalid test case

bashbaug:
see #977 

Different implementations of the OpenCL ICD loader treat a null platform differently, so this case does not reliably return `CL_INVALID_PLATFORM` in all cases.  Temporarily disable this test while we discuss what the proper behavior should be.


alycm (approved):
Disabling this test while we determine expected behaviour here seems sensible.

kpet (approved):

Issue #956, opened 2020-09-14
spirv_new fails if SPIR-V is supported but compiler is not available.


alycm:
Currently an OpenCL 3.0 device with  `CL_DEVICE_COMPILER_AVAILABLE` set to `CL_FALSE` that reports `"SPIR-V_1.0"` for `CL_DEVICE_IL_VERSION` can not pass `test_spirv_new` as:

1. When running `test_spirv_new` directly it attempts to use `clBuildProgram` (I think all the tests should just SKIP in this case).
2. When passing `--compilation-mode binary` to `test_spirv_new` it does not use the same offline compiler script as the rest of the CTS.
3. Said offline compiler script specification [states that the input is OpenCL C](https://github.com/KhronosGroup/OpenCL-CTS/blob/16bb1d83c5e4c4ee1d73f254e7527345751193c7/test_common/harness/cl_offline_compiler-interface.txt#L12).  Changing this specification may be disruptive to implementations that support SPIR-V (or they may all already support this, I don't know).

A workaround is for such a device to just report `""` for `CL_DEVICE_IL_VERSION` as the utility of supporting SPIR-V without a compiler is low (but doing so is consistent with OpenCL C, where `clCreateProgramWithSource` works even if `CL_DEVICE_COMPILER_AVAILABLE` is `CL_FALSE`).

alycm:
I have NOT put this in the 3.0 project due to the potential disruption is point 3. above, and the availability of a workaround.  If someone thinks otherwise then please add this to the 3.0 project.

PR #980, opened 2020-09-24
Skip compiling SPIR-V if a compiler is not available

alycm:
Relates to #956 but too incomplete to call a fix.


alycm:
Actually, I'm not sure this is helpful.  A full version of #956 is really required, i.e. support `--compilation-mode binary`.

PR #982, opened 2020-09-24
Fail feature macro compare if compiler has more features than runtime

### LINKED ISSUE #979: Test compiler features macro need to be fixed in some areas
Add fixes for the following comments:
https://github.com/KhronosGroup/OpenCL-CTS/pull/762
https://github.com/KhronosGroup/OpenCL-CTS/pull/762#discussion_r494445361
https://github.com/KhronosGroup/OpenCL-CTS/pull/762#discussion_r494445638
https://github.com/KhronosGroup/OpenCL-CTS/pull/762#discussion_r494468614

alycm:
Because a C++11 `std::equal` only iterates over the first container, and
matches with items in the second, if the second container contains more items
the check can still pass even though they're not identical.  Just use `==`
instead.

Fixes #979

### DIFF
--- a/test_conformance/compiler/test_feature_macro.cpp
+++ b/test_conformance/compiler/test_feature_macro.cpp
@@ -22,7 +22,7 @@ const char* macro_supported_source = R"(kernel void enabled(global int * buf) {
         int n = get_global_id(0);
         buf[n] = 0;
         #ifndef %s
-            ERROR;
+            #error Feature macro was not defined
         #endif
 })";
 
@@ -31,7 +31,7 @@ const char* macro_not_supported_source =
         int n = get_global_id(0);
         buf[n] = 0;
         #ifdef %s
-            ERROR;
+            #error Feature macro was defined
         #endif
 })";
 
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -686,9 +686,7 @@ int test_consistency_c_features_list(cl_device_id deviceID,
     sort(vec_to_cmp.begin(), vec_to_cmp.end());
     sort(vec_device_feature_names.begin(), vec_device_feature_names.end());
 
-    cl_bool result =
-        std::equal(vec_device_feature_names.begin(),
-                   vec_device_feature_names.end(), vec_to_cmp.begin());
+    cl_bool result = vec_device_feature_names == vec_to_cmp;
     if (result)
```
bashbaug, line 690: I'm not going to lose sleep over this, but I'd find this check easier to understand without the extra boolean variable (and `operator==` doesn't return a `cl_bool` anyhow).

bashbaug (commented):
LGTM - probably should wait for any comments from @gwawiork before merging though.

test_conformance/compiler/test_feature_macro.cpp
alycm: Good point, I've changed it.

lakshmih:
Qualcomm tested the compiler feature_macros sub_test with this change and it passed.

gwawiork (approved):
LGTM

Jeremy-Kemp (approved):
LGTM!

PR #983, opened 2020-09-24
Skip feature macro consistency tests if no compiler is available

alycm:
The test compiles programs to determine compiler support, which isn't possible
if the compiler is not available, so skip the test.


bashbaug (approved):
LGTM!

Jeremy-Kemp (approved):

kpet (approved):

PR #986, opened 2020-09-28
Skip compiler unload/reload tests if compiler is not available

alycm:
Note that tests that use the compiler helper functions but won't work without a
compiler are in the `subtests_to_skip_with_offline_compiler` list.  I didn't do
that here because they tests directly use `clBuildProgram` etc directly,
because they're specifically testing it's behaviour in edge-cases.

### DIFF
--- a/test_conformance/compiler/test_feature_macro.cpp
+++ b/test_conformance/compiler/test_feature_macro.cpp
@@ -732,7 +732,7 @@ int test_features_macro(cl_device_id deviceID, cl_context context,
     // also "may not".
     check_compiler_available(deviceID);
 
-    cl_int error = CL_SUCCESS;
+    int error = TEST_PASS;
     cl_bool supported = CL_FALSE;
     std::string test_macro_name = "";
     std::vector<std::string> supported_features_vec;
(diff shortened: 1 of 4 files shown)


test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -724,21 +724,15 @@ int test_consistency_c_features_list(cl_device_id deviceID,
...
-    }
 
+    // Note: Not checking that the feature array is empty for the compiler not
+    // available case because the specification says "For devices that do not
+    // support compilation from OpenCL C source, this query may return an empty
+    // array."  It "may" return an empty array implies that an implementation
+    // also "may not".
+    check_compiler_available(deviceID);
+
+    cl_int error = CL_SUCCESS;
```
bashbaug, line 735: Minor: since this value represents a test status and not an OpenCL error code now, it more accurately should be an `int` and should be initialized to `TEST_PASS`.  This will not cause any issues in practice as-is.

bashbaug (approved):
LGTM - just one minor comment that can be fixed later (or never).

test_conformance/compiler/test_feature_macro.cpp
alycm: Makes sense, fixed.

bashbaug (approved):
LGTM - thanks!

gwawiork (approved):
LGTM

kpet (approved):

PR #995, opened 2020-09-30
add a basic test for CL_PIPE_PROPERTIES

### LINKED ISSUE #991: Test passing CL_PIPE_PROPERTIES to clGetPipeInfo
The CTS should test the result of passing `CL_PIPE_PROPERTIES` to `clGetPipeInfo`.

As the core spec doesn't actually defined any properties to pass to `clCreatePipe` about the only thing that the CTS can test is that when passing `NULL` properties to `clCreatePipe` you get an empty list (i.e. size 0) back for `CL_PIPE_PROPERTIES` to `clGetPipeInfo`.

bashbaug:
Fixes #991 

This adds a simple test for CL_PIPE_PROPERTIES.  Since pipes can only be created with `NULL` properties there is only one sub-case to test.

### DIFF
--- a/test_conformance/api/test_pipe_properties_queries.cpp
+++ b/test_conformance/api/test_pipe_properties_queries.cpp
@@ -1,0 +1,100 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


gwawiork (approved):
LGTM

test_conformance/api/test_pipe_properties_queries.cpp
```diff
@@ -0,0 +1,100 @@
...
+              "compare pipe properties.\n");
+    return TEST_FAIL;
+}
+
+int test_pipe_properties_queries(cl_device_id deviceID, cl_context context,
+                                 cl_command_queue queue, int num_elements)
+{
+    cl_int error = CL_SUCCESS;
+
+    cl_bool pipeSupport = CL_FALSE;
```
kpet, line 79: Probably best to initialise to `CL_TRUE` so that we're more likely to catch the case where an implementation doesn't touch the memory.

kpet (approved):
One possible (very) minor improvement but fine as is.

alycm (approved):
LGTM.

Jeremy-Kemp (approved):

PR #996, opened 2020-09-30
added extended versioning tests for OpenCL 3.0

### LINKED ISSUE #992: Test extended versioning as a core feature
Currently the `computeinfo` tests for extended versioning only test `cl_khr_extended_versioning`.  These tests should also test the same features for 3.0 core, likely as a duplication due to manage the differences in `_KHR` suffices etc from the extension.

The extended versioning features are used by other 3.0 tests, so there is coverage, but it's easier to convince ourselves of complete coverage if we have a centralized test.

Caveats:
* `CL_DEVICE_OPENCL_C_NUMERIC_VERSION_KHR` is the extension but there is no `CL_DEVICE_OPENCL_C_NUMERIC_VERSION` in  core, so don't copy that part of testing.
* `CL_DEVICE_OPENCL_C_ALL_VERSIONS` is in core but there is no `CL_DEVICE_OPENCL_C_ALL_VERSIONS_KHR` in the extension, so that will require a brand new test.

bashbaug:
Fixes #992 

Adds testing for "extended versioning" for OpenCL 3.0 implementations in addition to implementations supporting `cl_khr_extended_versioning`.

This test assumes (and verifies) that the core enums are equal to the extension enums, so it only tests the queries once.

The only difference between the core functionality and the extension functionality is `CL_DEVICE_OPENCL_C_NUMERIC_VERSION_KHR`, which exists in the extension but was not promoted to a core feature, so this sub-test is skipped if the OpenCL 3.0 implementation does not also support `cl_khr_extended_versioning`.

Note that `CL_DEVICE_OPENCL_C_ALL_VERSIONS` is NOT tested by this test, but it is tested as part of `test_opencl_c_versions`.

### DIFF
--- a/test_conformance/computeinfo/extended_versioning.cpp
+++ b/test_conformance/computeinfo/extended_versioning.cpp
@@ -243,7 +243,8 @@ static int test_extended_versioning_platform_version(cl_platform_id platform)
 
 /* Check that CL_DEVICE{,_OPENCL_C}_NUMERIC_VERSION_KHR return the same versions
  * as CL_DEVICE{,_OPENCL_C}_VERSION */
-static int test_extended_versioning_device_versions(cl_device_id deviceID)
+static int test_extended_versioning_device_versions(int ext,
+                                                    cl_device_id deviceID)
 {
     log_info("Device versions:\n");
 
@@ -260,6 +261,15 @@ static int test_extended_versioning_device_versions(cl_device_id deviceID)
 
     for (const auto& query : device_version_queries)
     {
+        // CL_DEVICE_OPENCL_C_NUMERIC_VERSION_KHR is only supported by
+        // cl_khr_extended_versioning:
+        if (!ext
+            && query.param_name_numeric
+                == CL_DEVICE_OPENCL_C_NUMERIC_VERSION_KHR)
+        {
+            continue;
+        }
+
         const std::vector<char> version_string(
             get_device_string(deviceID, query.param_name_string));
         if (version_string.empty())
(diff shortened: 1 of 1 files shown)


test_conformance/computeinfo/extended_versioning.cpp
```diff
@@ -243,7 +243,8 @@ static int test_extended_versioning_platform_version(cl_platform_id platform)
 
 /* Check that CL_DEVICE{,_OPENCL_C}_NUMERIC_VERSION_KHR return the same versions
  * as CL_DEVICE{,_OPENCL_C}_VERSION */
-static int test_extended_versioning_device_versions(cl_device_id deviceID)
+static int test_extended_versioning_device_versions(int ext,
```
kpet, line 246: Why not use `bool`?

test_conformance/computeinfo/extended_versioning.cpp
```diff
@@ -688,14 +698,41 @@ static int test_extended_versioning_device_built_in_kernels(cl_device_id device)
...
+static_assert(CL_DEVICE_EXTENSIONS_WITH_VERSION
+                  == CL_DEVICE_EXTENSIONS_WITH_VERSION_KHR,
+              "CL_DEVICE_EXTENSIONS_WITH_VERSION mismatch");
+static_assert(CL_DEVICE_ILS_WITH_VERSION == CL_DEVICE_ILS_WITH_VERSION_KHR,
+              "CL_DEVICE_ILS_WITH_VERSION mismatch");
+static_assert(CL_DEVICE_BUILT_IN_KERNELS_WITH_VERSION
+                  == CL_DEVICE_BUILT_IN_KERNELS_WITH_VERSION_KHR,
+              "CL_DEVICE_BUILT_IN_KERNELS_WITH_VERSION mismatch");
+
+static_assert(sizeof(cl_name_version) == sizeof(cl_name_version_khr),
```
kpet, line 721: Mabye you could use `std::is_same` or something a bit stronger than just comparing sizes.

test_conformance/computeinfo/extended_versioning.cpp
```diff
@@ -688,14 +698,41 @@ static int test_extended_versioning_device_built_in_kernels(cl_device_id device)
...
+
+static_assert(CL_MAKE_VERSION(1, 2, 3) == CL_MAKE_VERSION_KHR(1, 2, 3),
+              "CL_MAKE_VERSION mismatch");
+
 int test_extended_versioning(cl_device_id deviceID, cl_context context,
                              cl_command_queue ignoreQueue, int num_elements)
 {
-    if (!is_extension_available(deviceID, "cl_khr_extended_versioning"))
+    int ext = is_extension_available(deviceID, "cl_khr_extended_versioning");
+    int core = get_device_cl_version(deviceID) >= Version(3, 0);
```
kpet, line 731: bool

kpet (changes requested):
Generally looks good. A few minor comments.

gwawiork (approved):
LGTM + kpet comments

alycm (dismissed):
LGTM.  It didn't occur to me that because the enum names have the same values then they don't need tested independently.  This is much tidier than I was thinking.

Jeremy-Kemp (approved):
LGTM with the comments!

test_conformance/computeinfo/extended_versioning.cpp
bashbaug: Fixed.

test_conformance/computeinfo/extended_versioning.cpp
bashbaug: Fixed.  FWIW, `is_extension_available` returns `int` not `bool`.  Is there any reason for this or should `is_extension_available` return `bool` also?

test_conformance/computeinfo/extended_versioning.cpp
bashbaug: We can't use `std::is_same` since they aren't the same type, though they do have the same structure fields in the same order.

Jeremy-Kemp (approved):
Changes LGTM! Need @kpet to verify though.

alycm (approved):

test_conformance/computeinfo/extended_versioning.cpp
kpet: Yeah, it should return `bool` really.

test_conformance/computeinfo/extended_versioning.cpp
kpet: Ouch, I thought we used typedefs. We probably ought to fix this in the headers.

kpet (approved):

PR #1003, opened 2020-10-04
Don't require support for 3D image writes in tests that don't need it

kpet:
Also fix PASSIVE_REQUIRE_* macros to report tests as skipped instead of passed.


bashbaug (approved):
LGTM!

jrprice (approved):

PR #1006, opened 2020-10-07
build: Various improvements. (#743)

lakshmih:
- Remove the build_<platform> scripts that were simply calling cmake in favor of
  using cmake directly
- Move flag  CRT_SECURE_NO_WARNING into a section specifically for visual studio
- Change vendor file selection to just use the file if present
- Add a variable for determining whether to link against pthread
- Delete all lines in CMakeVendor.txt so each implementation can define their own


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.12.4)
```
alycm, line 1: This is quite a big jump, and for example goes beyond the Ubuntu 18.04 default: https://packages.ubuntu.com/bionic/cmake

Which features require this version?

test_conformance/CMakeLists.txt
```diff
@@ -49,9 +49,11 @@ add_subdirectory( subgroups )
 add_subdirectory( workgroups )
 add_subdirectory( pipes )
 add_subdirectory( device_timer )
-add_subdirectory( clcpp )
-add_subdirectory( spirv_new )
-add_subdirectory( spir )
+if(KHRONOS_OFFLINE_COMPILER)
+    add_subdirectory( clcpp )
+    add_subdirectory( spirv_new )
+    add_subdirectory( spir )
```
alycm, line 55: It's not immediately obvious to me why the `spirv_new` and `spir` tests would need `KHRONOS_OFFLINE_COMPILER`.

build_lnx.sh
```diff
@@ -1,12 +0,0 @@
...
-mkdir -p build_lnx
-cd build_lnx
-cmake -G "Unix Makefiles" ../ \
-      -DKHRONOS_OFFLINE_COMPILER=<TO_SET> \
-      -DCL_LIBCLCXX_DIR=<TO_SET> \
-      -DCL_INCLUDE_DIR=<TO_SET> \
-      -DCL_LIB_DIR=<TO_SET> \
-      -DCMAKE_RUNTIME_OUTPUT_DIRECTORY=. \
-      -DOPENCL_LIBRARIES=OpenCL
-make --jobs 8
```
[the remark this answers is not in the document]
kpet: Maybe we should describe the supported option in the README file instead?

.gitignore
```diff
@@ -1,4 +0,0 @@
-# build directories
-build/
```
gwawiork, line 2: It was for me useful. Why do you want to remove this ?

lakshmih: https://cmake.org/cmake/help/latest/manual/cmake.1.html discourages these kinds of in source builds. If you are using an IDE that does an in-source build anyway, the recommended way would be to use a global git ignore (set in ~/.gitconfig > core.excludesFile) to filter out all IDE specific files across all git projects on the computer once and for all

CMakeLists.txt
lakshmih: Agreed that it makes sense to restrict the CMake version requirement to popular long term Linux distros. Ubuntu 16.04 is the oldest one with available package data and that uses CMake 3.5.1, released in 2016. Will put up a patch that uses 3.5.1 as the minimum. Nothing in the change to relies on newer features than this.

build_lnx.sh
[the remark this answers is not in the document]
lakshmih: Will add a README that covers the common options.

test_conformance/CMakeLists.txt
lakshmih: Agreed. Moving those out.

CMakeLists.txt
lakshmih: Lowered to 3.5.1 in the latest patchset, please review.

build_lnx.sh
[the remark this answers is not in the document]
lakshmih: @kpet  @alycm I'll create a separate issue for README since it needs to address more than just the build options.

.gitignore
gwawiork: ok thx

gwawiork (approved):
LGTM

CMakeLists.txt
alycm: Yes, 3.5.1 is fine with me, thanks!

kpet (approved):
Nice change :). Thanks!

lakshmih:
@alycm Can you approve if you are good with this, and we can merge it.
Thanks!

build_lnx.sh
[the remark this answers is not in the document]
alycm: Separate issue is fine.

alycm (approved):
LGTM.

alycm:
Three approvals so merging, the ticket regarding documents options in the readme still needs to be filed though.

alycm:
Sorry, I was looking for an issue re: README and didn't see one, but I missed that there is already a PR: https://github.com/KhronosGroup/OpenCL-CTS/pull/1021

PR #1007, opened 2020-10-08
Reimplement buffer tests

### LINKED ISSUE #439: Uncomment and re-integrate "not implemented" buffer tests.
The buffer tests `buffer_read_half`, `buffer_write_half` and `mem_alloc_ref_flags` hit the unimplemented case in `testHarness.c` and print something like:
```
mem_alloc_ref_flags test currently not implemented
```
This is due to how the buffer test lists are [defined](https://github.com/KhronosGroup/OpenCL-CTS/blob/0b9329a9d679bfe118ca7f4ee321a2a0a315d930/test_conformance/buffers/main.c#L50) in [both](https://github.com/KhronosGroup/OpenCL-CTS/blob/c1ef993f5e85bb1401834a85b506e4694a32adce/test_conformance/buffers/main.c#L50) branches and the existing test implementations being [commented out](https://github.com/KhronosGroup/OpenCL-CTS/blob/0b9329a9d679bfe118ca7f4ee321a2a0a315d930/test_conformance/buffers/test_buffer_read.c#L1145). It would be nice to re-integrate these tests by tweaking the existing test implementations to check if the tested feature is present before running anything and then re-enabling them in the function list.

ellnor01:
Reintegrated and fixed test code for buffer tests buffer_read_half and
buffer_write_half tests.

Added mem_alloc_ref_flags test code, as was previously non-existent,
to test CL_MEM_ALLOC_HOST_PTR. This flag was otherwise untested and
as similar tests within the suite are used to test other cl_mem_flags
it has been assumed that this was the purpose of the test.

Fixes #439


Jeremy-Kemp (approved):
I've had a quick look back to see if I could find any old tests for `mem_alloc_ref_flags` but couldn't find anything either.

Good to have some more FP16 coverage as well, thanks!

ellnor01:
@Jeremy-Kemp sorry didn't realise you'd already reviewed this! I've just updated commit messages and added a final commit to remove unused code

test_conformance/buffers/test_buffer_write.cpp
```diff
@@ -315,40 +315,51 @@ static const char *float_kernel_name[] = { "test_buffer_write_float", "test_buff
...
 const char *buffer_write_half_kernel_code[] = {
-    "__kernel void test_buffer_write_half(__global half *src, __global float *dst)\n"
+    "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n"
+    "__kernel void test_buffer_write_half(__global half *src, __global half "
+    "*dst)\n"
     "{\n"
     "    int  tid = get_global_id(0);\n"
     "\n"
-    "    dst[tid] = vload_half( tid * 2, src );\n"
+    "    dst[tid] = src[tid];\n"
```
kpet, line 324: Fine for now but these tests could also be changed (using vload/vstore) to always run on 2.x implementations even when cl_khr_fp16 is not supported.

kpet (approved):
LGTM

Jeremy-Kemp (approved):
@ellnor01 No problem, thanks for the latest changes! Still LGTM.

PR #1008, opened 2020-10-08
Making object queries tests exhaustive

### LINKED ISSUE #508: Object queries tests should be exhaustive where possible

ellnor01:
Tests which were only testing fixed values have been made exhaustive
and refactored. They only use properties from the core spec.

Associated Macros have been turned into templated functions.

Comparison of devices with cl_device_id rather than vendor ids.

All object queries tests are now in test_queries.cpp.

Fixes #508


james-morrissey-arm:
It looks as if the val arguments are redundant in sampler_param_test and command_queue_param_test;  you can just create local variables to receive the query results as they don't need to be returned to the caller.

test_conformance/api/CMakeLists.txt
```diff
@@ -7,7 +7,6 @@ set(${MODULE_NAME}_SOURCES
          test_retain.cpp
          test_retain_program.cpp
          test_queries.cpp
-         test_queries_compatibility.cpp
```
jrprice, line 10: Presumably this source file should also be deleted?

test_conformance/api/test_queries.cpp
```diff
@@ -131,86 +131,95 @@ int test_get_platform_info(cl_device_id deviceID, cl_context context, cl_command
...
-    test_error( error, "Unable to create sampler to test with" );
-
-    cl_uint refCount;
-    error = clGetSamplerInfo( sampler, CL_SAMPLER_REFERENCE_COUNT, sizeof( refCount ), &refCount, &size );
-    test_error( error, "Unable to get sampler ref count" );
-    if( size != sizeof( refCount ) )
+    T val;
+    int error = clGetSamplerInfo(sampler, param_name, sizeof(val), &val, &size);
+    test_error(error, "Unable to get sampler info");
+    if (val != expected)
```
jrprice, line 142: This could use the `test_assert_error()` helper macro.

test_conformance/api/test_queries.cpp
```diff
@@ -131,86 +131,95 @@ int test_get_platform_info(cl_device_id deviceID, cl_context context, cl_command
...
-    cl_context otherCtx;
-    error = clGetSamplerInfo( sampler, CL_SAMPLER_CONTEXT, sizeof( otherCtx ), &otherCtx, &size );
-    test_error( error, "Unable to get sampler context" );
-    if( otherCtx != context )
-    {
-        log_error( "ERROR: Sampler context does not validate! (expected %p, got %p)\n", context, otherCtx );
-        return -1;
-    }
-    if( size != sizeof( otherCtx ) )
+    if (size != sizeof(val))
```
jrprice, line 146: As above, recommend using `test_assert_error()`.

test_conformance/api/test_queries.cpp
```diff
@@ -131,86 +131,95 @@ int test_get_platform_info(cl_device_id deviceID, cl_context context, cl_command
...
-    if( norm != CL_TRUE )
-    {
-        log_error( "ERROR: Sampler normalized flag does not validate! (expected %d, got %d)\n", (int)CL_TRUE, (int)norm );
-        return -1;
-    }
-    if( size != sizeof( norm ) )
+    error = clGetSamplerInfo(sampler, CL_SAMPLER_REFERENCE_COUNT,
+                             sizeof(refCount), &refCount, &size);
+    test_error(error, "Unable to get sampler ref count");
+    if (size != sizeof(refCount))
```
jrprice, line 199: `test_assert_error()`

test_conformance/api/test_queries.cpp
```diff
@@ -244,25 +253,107 @@ int test_get_sampler_info(cl_device_id deviceID, cl_context context, cl_command_
...
+int command_queue_param_test(cl_command_queue queue,
+                             cl_command_queue_info param_name, T expected,
+                             const char *name)
+{
+    size_t size;
+    T val;
+    int error =
+        clGetCommandQueueInfo(queue, param_name, sizeof(val), &val, &size);
+    test_error(error, "Unable to get command queue info");
+    if (val != expected)
```
jrprice, line 323: `test_assert_error()`

test_conformance/api/test_queries.cpp
```diff
@@ -244,25 +253,107 @@ int test_get_sampler_info(cl_device_id deviceID, cl_context context, cl_command_
...
+    {
+        test_fail("ERROR: Command queue %s did not validate!\n", name);
+    }
+    if (size != sizeof(val))
```
jrprice, line 327: `test_assert_error()`

test_conformance/api/test_queries.cpp
```diff
@@ -273,50 +364,78 @@ int test_get_command_queue_info(cl_device_id deviceID, cl_context context, cl_co
...
+    if (device_props & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE)
     {
-        log_error( "ERROR: Returned size of command queue reference count does not validate! (expected %d, got %d)\n", (int)sizeof( refCount ), (int)size );
-        return -1;
+        // Test out-of-order queues properties if supported
+        num_test_options = OOO_NUM_COMMAND_QUEUE_PROPERTIES;
+        if (version > Version(2, 0) && !is_compatibility)
+        {
+            // Test queue on device if supported
+            num_test_options = ARRAY_SIZE(property_options);
```
jrprice, line 377: This doesn't look quite right - OpenCL 3.0 devices are not required to support on-device queues IIRC.

test_conformance/api/test_queries.cpp
```diff
@@ -273,50 +364,78 @@ int test_get_command_queue_info(cl_device_id deviceID, cl_context context, cl_co
...
 
-    if (size != sizeof(cl_device_id)) {
-        log_error( " ERROR: Returned size of command queue CL_QUEUE_DEVICE does not validate! (expected %d, got %d)\n", (int)sizeof( otherDevice ), (int)size );
-        return -1;
-    }
+        cl_uint refCount;
+        error = clGetCommandQueueInfo(queue, CL_QUEUE_REFERENCE_COUNT,
+                                      sizeof(refCount), &refCount, &size);
+        test_error(error, "Unable to get command queue reference count");
+        if (size != sizeof(refCount))
```
jrprice, line 403: `test_assert_error()`

test_conformance/api/test_queries.cpp
```diff
@@ -244,25 +253,107 @@ int test_get_sampler_info(cl_device_id deviceID, cl_context context, cl_command_
...
 
-int test_get_command_queue_info(cl_device_id deviceID, cl_context context, cl_command_queue ignoreQueue, int num_elements)
+#define MIN_NUM_COMMAND_QUEUE_PROPERTIES 2
+#define OOO_NUM_COMMAND_QUEUE_PROPERTIES 4
+static cl_command_queue_properties property_options[] = {
+    0,
+    CL_QUEUE_PROFILING_ENABLE,
+    CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
+    CL_QUEUE_PROFILING_ENABLE,
+    CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
```
jrprice, line 343: I'm not sure I quite understand this array. Should this be `CL_QUEUE_PROFILING_ENABLE  | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE`, rather than repeating those two properties individually?

jrprice (changes requested):

test_conformance/api/test_queries.cpp
ellnor01: In this case (and a few others below) this removes some of the failure information - it removes the 'name' of the parameter being tested.

test_conformance/api/test_queries.cpp
ellnor01: see above

test_conformance/api/test_queries.cpp
ellnor01: as above

test_conformance/api/test_queries.cpp
ellnor01: as above

test_conformance/api/test_queries.cpp
```diff
@@ -244,79 +249,203 @@ int test_get_sampler_info(cl_device_id deviceID, cl_context context, cl_command_
...
-    test_error( error, "Unable to get device CL_DEVICE_VENDOR_ID" );
-    error = clGetDeviceInfo(deviceID, CL_DEVICE_VENDOR_ID, sizeof(deviceID_vid), &deviceID_vid, NULL );
-    test_error( error, "Unable to get device CL_DEVICE_VENDOR_ID" );
+    auto version = get_device_cl_version(deviceID);
 
-    if( otherDevice_vid != deviceID_vid )
+    // Are on device queues supported
+    bool on_device_supported =
+        (version >= Version(2, 0)
+         || (version >= Version(3, 0) && sizeof(device_queue_props) != 0));
```
jrprice, line 379: Still not quite right - the left-hand side of the `||` will be `true` for any 3.0 device, so the overall expression will always be `true` for 3.0.  Also, the check for device queues on 3.0 shouldn't be testing `sizeof()`,

So I think this needs to be `(version >= Version(2, 0) && version < Version(3, 0)) || (Version >= Version(3, 0) && device_queue_props != 0)`. Not pleasant, but it will be made nicer when Kevin's feature testing framework lands.

jrprice (changes requested):
Almost there, just one more change for 3.0 compatibility.

jrprice (approved):
LGTM, thanks!

PR #1009, opened 2020-10-09
Add new subgroup extensions to the compiler test

gwawiork:
Changes related with extended subgroups.


kpet (approved):
Fine as a point fix but we really ought to change the logic to make it less treacherous :).

kpet:
Straightforward and uncontroversial fix, merging.

Issue #1011, opened 2020-10-12
api/test_zero_sized_enqueue didn't regard CL_SUBMITTED as a valid waiting status


Nuullll:
[test_conformance/api/test_zero_sized_enqueue.cpp@135](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/api/test_zero_sized_enqueue.cpp#L135)
```cpp
    // Verify zero-sized ND range kernel still satisfy event wait list and correct event object
    // is returned
    clEventWrapper ev = NULL;
    clEventWrapper user_ev = clCreateUserEvent(context, &error);
    test_error( error, "user event creation failed." );
    error = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, ndrange30, NULL, 1, &user_ev, &ev);
    test_error( error, "3D zero sized kernel enqueue failed." );
    if (ev == NULL)
    {
        log_error( "ERROR: failed to create an event object\n" );
        return -1;
    }

    cl_int sta;
    error = clGetEventInfo(ev, CL_EVENT_COMMAND_EXECUTION_STATUS, sizeof(cl_int), &sta, NULL);
    test_error( error, "Failed to get event status.");

    if (sta != CL_QUEUED)
    {
        log_error( "ERROR: incorrect zero sized kernel enqueue event status.\n" );
        return -1;
    }
```

`ev` explicitly waits for `user_ev`, status `CL_QUEUED` or `CL_SUBMITTED` are both valid, indicating that `ev` is still waiting.
i.e. checking on `sta` should expand to `sta != CL_QUEUED && sta != CL_SUBMITTED`, I think.

Nuullll:
Also find a reference code section from [test_conformance/events/test_waitlists.cpp@178](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/events/test_waitlists.cpp#L178)

```cpp
        if( status[ 2 ] != CL_QUEUED && status[ 2 ] != CL_SUBMITTED )
        {
            log_error( "ERROR: Test event did not wait for second event before starting! (status of ref: 1: %s, of test: 2: %s)\n", IGetStatusString( status[ 1 ] ), IGetStatusString( status[ 2 ] ) );
            clFinish( queue );
            return -1;
        }
```

which checks waiting status more reasonably.

kpet:
PR merged, closing.

PR #1012, opened 2020-10-12
[api][test_zero_sized_enqueue] Add CL_SUBMITTED check for waiting status

Nuullll:
This addresses #1011

### DIFF
--- a/test_conformance/api/test_zero_sized_enqueue.cpp
+++ b/test_conformance/api/test_zero_sized_enqueue.cpp
@@ -149,7 +149,7 @@ int test_zero_sized_enqueue_helper(cl_device_id deviceID, cl_context context, cl
     error = clGetEventInfo(ev, CL_EVENT_COMMAND_EXECUTION_STATUS, sizeof(cl_int), &sta, NULL);
     test_error( error, "Failed to get event status.");
 
-    if (sta != CL_QUEUED)
+    if (sta != CL_QUEUED && sta != CL_SUBMITTED)
     {
         log_error( "ERROR: incorrect zero sized kernel enqueue event status.\n" );
         return -1;


gwawiork (approved):
LGTM

EwanC (approved):

kpet:
Not sure this is correct. For the event to have a CL_SUBMITTED status, the command must have been flushed.

> Submitted [...] The transition into this state occurs when the command is flushed from the command-queue

The specification defines a number of points at which ways a command can be flushed but it says nothing about implementations flushing commands

A command may reside in the queue until it is flushed either explicitly (a call to clFlush) or implicitly by some other command.

Because commands in a command-queue are not required to begin execution until the command-queue is flushed,

EwanC:
> A command may reside in the queue until it is flushed either explicitly (a call to clFlush) or implicitly by some other command.
> 
> Because commands in a command-queue are not required to begin execution until the command-queue is flushed,

If I'm following right, would it be valid implementation behavior for the `clEnqueueNDRangeKernel` command [here](https://github.com/KhronosGroup/OpenCL-CTS/pull/1012/files#diff-d20cbf2cacd34ebd1ab5bb997c7e1206e482f8ad208db6da44c63d1876e32993R140)
to implicitly flush? Therefore setting the event status to `CL_SUBMITTED`

kpet:
That's what is unclear to me. The spec provides a detailed list of commands that MUST "perform an implicit" flush but it doesn't clearly state whether doing so from another API call or even independently of any API call is allowed or forbidden.

Nuullll:
> That's what is unclear to me. The spec provides a detailed list of commands that MUST "perform an implicit" flush but it doesn't clearly state whether doing so from another API call or even independently of any API call is allowed or forbidden.

Thanks, I agree that the spec is unclear about this part.

I guess a conformance test should only valid the behaviors explicitly stated in the spec. In such a case, I think it's possibly an implementation-defined behavior that doesn't violate the current spec.
Or should we require an update in spec to clarify this issue? I'm not familiar with the OpenCL spec drafting process :P

kpet:
As documented on https://github.com/KhronosGroup/OpenCL-Docs/issues/479, the WG agrees that flushing should be permitted at any time. Merging this change.

PR #1013, opened 2020-10-12
Implement test for CL_KERNEL_PRIVATE_MEM_SIZE

chemis01:

### DIFF
--- a/test_conformance/api/test_kernel_private_memory_size.cpp
+++ b/test_conformance/api/test_kernel_private_memory_size.cpp
@@ -1,0 +1,42 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "harness/errorHelpers.h"
+#include "harness/typeWrappers.h"
+#include <iostream>
+#include "procs.h"
+
+int test_kernel_private_memory_size(cl_device_id deviceID, cl_context context,
+                                    cl_command_queue queue, int num_elements)
+{
+    const char* TEST_KERNEL =
+        R"(__kernel void private_memory( __global ulong *buffer ){
+         volatile __private ulong x[1];
+         buffer[0] = x[0];
+         })";
+
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+    cl_int err = create_single_kernel_helper(context, &program, &kernel, 1,
+                                             &kernels[i], "private_memory");
+    test_error(err, "create_single_kernel_helper");
+    cl_ulong size = CL_ULONG_MAX;
+    err = clGetKernelWorkGroupInfo(kernel, deviceID, CL_KERNEL_PRIVATE_MEM_SIZE,
+                                   sizeof(cl_ulong), &size, nullptr);
+
+    test_error(err, "clGetKernelWorkGroupInfo");
+
+    return TEST_PASS;
+}


test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,58 @@
...
+        ASSERT_SUCCESS(err, "clCreateProgramWithSource");
+        err = clBuildProgram(program, 1, &deviceID, nullptr, nullptr, nullptr);
+        ASSERT_SUCCESS(err, "clBuildProgram");
+
+        /* Create the Kernel*/
+        clKernelWrapper kernel(clCreateKernel(program, "private_memory", &err));
+        ASSERT_SUCCESS(err, "clCreateKernel");
+
+
+        cl_ulong size = -1; // If clGetKernelWorkGroupInfo passes the value
```
Jeremy-Kemp, line 44: Could we use `CL_ULONG_MAX` here please? Using -1 will trigger some errors/warnings in some compilers.

Jeremy-Kemp (changes requested):

test_conformance/api/test_kernel_private_memory_size.cpp
chemis01: Yes, this has been updated in the latest commit. Is this what you were looking for?

Jeremy-Kemp (dismissed):
LGTM, thanks.

I'm not sure how meaningful the `ZERO_MEMORY_KERNEL` test is as the value for `CL_KERNEL_PRIVATE_MEM_SIZE` is unsigned and therefore should never be able to fail.

However, it doesn't hurt either.

test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,58 @@
+#include "harness/errorHelpers.h"
```
EwanC, line 1: Might need a license comment at the top of this file

test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "harness/typeWrappers.h"
+#include <iostream>
+#include <vector>
+#include "procs.h"
```
EwanC, line 5: is this header needed?

test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,58 @@
...
+    R"(    volatile __private ulong x[1];)"
+    R"(    buffer[0] = x[0]; })";
+
+constexpr char const* TWO_MEMORY_KERNEL =
+    R"(__kernel void private_memory( __global ulong *buffer ){ )"
+    R"(    volatile __private ulong x[1];)"
+    R"(    volatile __private ulong y[1];)"
+    R"(    buffer[0] = x[0] = y[0]; })";
+
+static std::vector<char const*> kernels = { ZERO_MEMORY_KERNEL,
```
EwanC, line 22: Could these vectors be `std::array<..., 3>` if they're fixed size

test_conformance/api/test_kernel_private_memory_size.cpp
chemis01: Added a license.

test_conformance/api/test_kernel_private_memory_size.cpp
chemis01: These headers are all needed.

chemis01 (commented):
Updated as per the comments. Please let me know if you have any further comments.

EwanC (approved):

test_conformance/api/test_kernel_private_memory_size.cpp
EwanC: <vector> needs changed to <array> now I guess

Jeremy-Kemp (dismissed):

kpet:
@chemis01 The CI is failing. You may want to take a look.

chemis01:
Updated <vector> to <array> in the headers and CI now passes

EwanC (dismissed):

test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,72 @@
...
+constexpr char const* ONE_MEMORY_KERNEL =
+    R"(__kernel void private_memory( __global ulong *buffer ){ )"
+    R"(    volatile __private ulong x[1];)"
+    R"(    buffer[0] = x[0]; })";
+
+constexpr char const* TWO_MEMORY_KERNEL =
+    R"(__kernel void private_memory( __global ulong *buffer ){ )"
+    R"(    volatile __private ulong x[1];)"
+    R"(    volatile __private ulong y[1];)"
+    R"(    buffer[0] = x[0] = y[0]; })";
```
jrprice, line 34: One of the benefits of raw string literals is that they preserve newlines, so you can avoid having to concatenate multiple literals together to build a kernel, which would make this more readable IMO. Example here:
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/basic/test_intmath.cpp#L38

jrprice (commented):
Can we really reason about how the minimum amount of private memory used by an OpenCL implementation like this?

I know you've used `volatile` to try to prevent the compiler from optimizing out the `private` arrays, but in OpenCL there's no external entities that could ever modify a private variable, so in theory an OpenCL C compiler could still choose to optimize them out.

kpet:
Agree with @jrprice that the spec doesn't guarantee that this test will always work. The CTS should at least check that the query can be made successfully. Beyond that, implementations have a lot of freedom in what they can report and this is something we should discuss. I'd be interested to know what implementations do pass this test.

gwawiork:
Hi. Intel compiler make some optimizations here and we do not pass this test.

chemis01:
Hi everyone, I have made adjustments to this test and it now checks whether or not the API passes.
Please review

test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,33 @@
...
+         volatile __private ulong x[1];
+         buffer[0] = x[0];
+         })";
+
+    cl_int err = 0;
+    clProgramWrapper program(
+        clCreateProgramWithSource(context, 1, &TEST_KERNEL, nullptr, &err));
+    ASSERT_SUCCESS(err, "clCreateProgramWithSource");
+    err = clBuildProgram(program, 1, &deviceID, nullptr, nullptr, nullptr);
+    ASSERT_SUCCESS(err, "clBuildProgram");
```
bashbaug, line 20: Consider using `create_single_kernel_helper` here instead, which supports offline compilation.

test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+    /* Create the Kernel*/
+    clKernelWrapper kernel(clCreateKernel(program, "private_memory", &err));
+    ASSERT_SUCCESS(err, "clCreateKernel");
```
bashbaug, line 24: Is there a reason this test uses `ASSERT_ERROR` instead of `test_error`?  I think most tests use `test_error`, which will simply return a failure vs. throwing an exception.

bashbaug (commented):
I'm a little late to this review, my apologies, but I have one question and one comment.  Thanks!

test_conformance/api/test_kernel_private_memory_size.cpp
chemis01: Ok! I've updated that in the latest patch.

test_conformance/api/test_kernel_private_memory_size.cpp
chemis01: Not really, I'm still new and learning the about the CTS. Thank you for pointing this out to me. I've updated this in the latest patch.

test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,27 @@
...
+    const char* TEST_KERNEL =
+        R"(__kernel void private_memory( __global ulong *buffer ){
+         volatile __private ulong x[1];
+         buffer[0] = x[0];
+         })";
+
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+    cl_int err = create_single_kernel_helper(context, &program, &kernel, 1,
+                                             &kernels[i], "private_memory");
```
bashbaug, line 18: Missing an error check here, e.g. `test_error(err, "create_single_kernel_helper");`

bashbaug (commented):
Thanks!  Just one missing error check now and I think this will be good.

test_conformance/api/test_kernel_private_memory_size.cpp
chemis01: Hi Ben, thank you so much for the comments.
I have added line you suggested and everything should be ready now.

test_conformance/api/test_kernel_private_memory_size.cpp
```diff
@@ -0,0 +1,27 @@
+#include "harness/errorHelpers.h"
```
kpet, line 1: This file is missing a copyright/license header.

kpet (changes requested):

kpet (approved):
LGTM. @bashbaug I believe your comments have been addressed. Could you take another look please?

Jeremy-Kemp (approved):
Thanks for the changes, this looks more suitable now.

jrprice:
Did the wrong commit get landed here somehow? This PR didn't add the new source file to the `CMakeLists.txt` file or to the test list for the api suite, and the code itself doesn't compile.

Issue #330, opened 2019-06-17
kernel_image_methods uses read_only qualifier on images created with CL_MEM_WRITE_ONLY


kpet:

kpet:
PR merged, closing.

PR #1014, opened 2020-10-13
Fix write-only image tests in kernel_image_methods (#330)

james-morrissey-arm:
Pass the right flags to the tests that create images to ensure valid
usage.

Fixes issue #330

### DIFF
--- a/test_conformance/images/kernel_image_methods/test_2D.cpp
+++ b/test_conformance/images/kernel_image_methods/test_2D.cpp
@@ -32,38 +32,42 @@ struct image_kernel_data
 };
 
 static const char *methodTestKernelPattern =
-"typedef struct {\n"
-"    int width;\n"
-"    int height;\n"
-"    int depth;\n"
-"    int widthDim;\n"
-"    int heightDim;\n"
-"    int depthDim;\n"
-"    int channelType;\n"
-"    int channelOrder;\n"
-"    int expectedChannelType;\n"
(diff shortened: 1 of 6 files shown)


test_conformance/images/kernel_image_methods/test_2D.cpp
```diff
@@ -100,6 +108,10 @@ int test_get_image_info_single( cl_context context, cl_command_queue queue, imag
 
     const char* channelTypeName = GetChannelTypeName( imageInfo->format->image_channel_data_type );
     const char* channelOrderName = GetChannelOrderName( imageInfo->format->image_channel_order );
+    const char *image_access_qualifier =
+        (flags == CL_MEM_READ_ONLY) ? "read_only" : "write_only";
+    const char *cl_khr_3d_image_writes_enabler =
+        "#pragma OPENCL EXTENSION cl_khr_3d_image_writes : enable";
```
jrprice, line 114: Adding this unconditionally may cause a problems for devices that don't support 3D image writes (e.g. 1.2 and 3.0 devices that don't support this extension). This should only be added when testing 3D images with `CL_MEM_WRITE_ONLY`, and that test should be skipped (somewhere earlier) if the device doesn't support this functionality.

jrprice (changes requested):

kpet:
This now needs rebasing.

jrprice (approved):

kpet (approved):

Issue #329, opened 2019-06-17
samplerlessReads assumes that supported formats are the same for read_only and read_write images


kpet:

kpet:
PR merged, closing.

PR #1016, opened 2020-10-13
Fix samplerlessReads read_write (#329)

james-morrissey-arm:
The test was trying to create read_write image with read_only formats.
Make it use common image formats for both read_only and read_write
flags when creating images.

Fixes issue #329

### DIFF
--- a/test_conformance/images/samplerlessReads/test_loops.cpp
+++ b/test_conformance/images/samplerlessReads/test_loops.cpp
@@ -97,7 +97,6 @@ int test_image_set( cl_device_id device, cl_context context, cl_command_queue qu
 
     // Grab the list of supported image formats
     cl_image_format *formatList;
-    bool *filterFlags;
     unsigned int numFormats;
     auto version = get_device_cl_version(device);
     if (version < Version(2, 0)) {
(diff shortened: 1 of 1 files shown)


test_conformance/images/samplerlessReads/test_loops.cpp
```diff
@@ -114,14 +113,57 @@ int test_image_set( cl_device_id device, cl_context context, cl_command_queue qu
...
+    if (gTestReadWrite)
+    {
+        cl_image_format *readWriteFormats;
+        unsigned int numReadWriteFormats;
+
+        if (get_format_list(context, imageType, readWriteFormats,
+                            numReadWriteFormats, CL_MEM_KERNEL_READ_AND_WRITE))
+            return -1;
+
+        numFormats = numReadOnlyFormats;
```
gwawiork, line 132: Maybe this should be numReadOnlyFormats + numReadWriteFormats ?

james-morrissey-arm: I agree that the original code wasn't safe there, though the expectation is that there will be at most numReadOnlyFormats matches in the following nested loops.  The patch has been updated to include a break after finding a match in the inner loop, guaranteeing at most one match per iteration of the outer loop, which should make this both safe and explicit.

kpet (approved):
LGTM

test_conformance/images/samplerlessReads/test_loops.cpp
```diff
@@ -109,14 +108,58 @@ int test_image_set( cl_device_id device, cl_context context, cl_command_queue qu
...
+                        readOnlyFormats[i].image_channel_order;
+                    k++;
+                    break;
+                }
+            }
+        }
+
+        numFormats = k;
+
+        delete[] readOnlyFormats;
```
Jeremy-Kemp, line 153: This could be moved to the end of `test_image_set`.

Jeremy-Kemp (commented):
Looks good, one small nit.

bashbaug (commented):
Please feel free to merge these changes if they are blocking progress, but I'm not sure why they are required and I wonder if they're covering up another problem.

If I understand correctly, in the read-write testing case, this changes uses the intersection of the `CL_MEM_READ_ONLY` and `CL_MEM_KERNEL_READ_AND_WRITE` image formats vs. just the `CL_MEM_KERNEL_READ_AND_WRITE` image formats.  Is this correct?

If so, does this mean that on some devices an image is supported for both reading and writing (in the `CL_MEM_KERNEL_READ_AND_WRITE` set) but not supported for just reading (not in the `CL_MEM_READ_ONLY_SET`)?

kpet:
Yes that's correct and that's because the tests as written always require a read-only image of the format being tested in addition to optionally testing read-write.

>If so, does this mean that on some devices an image is supported for both reading and writing (in the CL_MEM_KERNEL_READ_AND_WRITE set) but not supported for just reading (not in the CL_MEM_READ_ONLY_SET)?

This is AFAICT legal for an implementation to do.

I'll merge this since we need this change in master for an upcoming submission but I agree that the tests could be redesigned to remove the need for the format list intersection. I've raised #1101 to look into improving the tests.

PR #1025, opened 2020-10-23
Skip Compiler Unload Tests in Binary Mode

FranklandJack:
The following tests make calls to `clCompileProgram` or
`clBuildProgram` and therefore require a compiler in the runtime:
  * `unload_repeated`
  * `unload_compile_unload_link`
  * `unload_build_unload_create_kernel`
  * `unload_link_different↩`
  * `unload_build_threaded`
  * `unload_build_info`
  * `unload_program_binaries`

Skip these tests if no compiler is present in the runtime.


kpet (approved):
LGTM, makes sense.

PR #1028, opened 2020-10-29
Fix min version for SPIRV test registration.

JohnKesapidesARM:

### DIFF
--- a/test_conformance/spirv_new/main.cpp
+++ b/test_conformance/spirv_new/main.cpp
@@ -82,7 +82,7 @@ void spirvTestsRegistry::addTestClass(baseTestClass *test, const char *testName,
     test_definition testDef;
     testDef.func = test->getFunction();
     testDef.name = testName;
-    testDef.min_version = Version(1, 2);
+    testDef.min_version = version;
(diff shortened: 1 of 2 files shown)


EwanC:
Does this change raise the minimum version required to run the spirv_new tests to [2.1](https://github.com/KhronosGroup/OpenCL-CTS/blob/41cd9c6d9823c15bdd09d3910f580ea39c8b2867/test_conformance/spirv_new/procs.h#L86)?

Our 1.2 implementation runs these tests with `clCreateProgramWithILKHR`

JohnKesapidesARM:
Yeap you are right, the default version when no version is defined is 2.1. I'll change the default to 1.2. 
`#define TEST_SPIRV_FUNC(name) TEST_SPIRV_FUNC_VERSION(name, Version(2, 1))`. We still need the filtering to work properly for 2.2 tests

JohnKesapidesARM:
@EwanC Should be ok now

EwanC (approved):
Looks good, thanks

gwawiork (approved):
LGTM

kpet (approved):

PR #1031, opened 2020-10-30
c11_atomics - change arg type to unsigned int from signed int

### LINKED ISSUE #1030: c11_atomics - argument type do not match function returned type
See line https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/c11_atomics/common.h#L883
there is a literal `1` number give as argument so by default compiler treats this as `signed int` type. 
Reading the specification https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#atomic-functions this argument type described as M should match to C. In this case C is atomic_uint (unsigned int) because https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/c11_atomics/common.h#L718

gwawiork:
Fixes #1030

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


Jeremy-Kemp (approved):
Out of interest have you looked to see if there are any additional mismatches between `A` and `M`?

gwawiork:
> Out of interest have you looked to see if there are any additional mismatches between `A` and `M`?

Hi Jeremy, sorry I do not understand your question.

Jeremy-Kemp:
Sorry it was badly written!

I was checking that there were no more signed/unsigned mismatches but I've had a look and can't see any.

gwawiork:
Ok I have not noticed any more issues.

PR #1032, opened 2020-10-30
Minor refactoring of CL_PROGRAM_SOURCE tests

StuartDBrady:
This is so that only one statement needs to be updated if the checks need to be changed in future.


kpet (approved):
LGTM

Issue #1026, opened 2020-10-27
feature_macro test to stringent for spirv and binary modes


FranklandJack:
The feature macro test: https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_feature_macro.cpp calls `create_single_kernel_helper_create_program` which in binary and spirv mode will call into offline compiler executables. 

In the case of a compilation error `create_single_kernel_helper_create_program` returns an error to the caller in binary/spirv mode since the compiler will not have produced and output file it can read back into the runtime (this doesn't happen in online mode since the CL C isn't compiled during the call to `create_single_kernel_helper_create_program` in this case, this happens later on in a call to `clBuildProgram`). 

Since this feature macro test expects the compiler to fail (it uses this information later on to detect whether a feature macro is supported by the compiler) the return error code of these calls to `create_single_kernel_helper_create_program` should not cause the test itself to fail in spirv/binary modes like it does here: https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_feature_macro.cpp#L143.

As it stands the test is skipped in binary mode, but is not in spirv mode since there needs to be a compiler in the runtime to consume the spirv. By replacing calls to `create_single_kernel_helper_create_program` with calls to `create_single_kernel_helper` and removing subsequent calls to `clBuildProgram` (like here: https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_feature_macro.cpp#L143) in the test, then using the results of the calls to `create_single_kernel_helper` to determine whether compilation succeeded or failed (this will work for all three compilation modes), the above issue could be resolved and this test could be run in binary mode as well as online and spirv.

StuartDBrady:
I'm not especially familiar with the feature macro test in question, so my apologies if my next question is somewhat naïve.

By doing this, would this mean that we'd simply be testing the offline compiler's support of feature macros, and not testing the online compiler? If so, is there value directly related to conformance in testing of the offline compiler in isolation? I thought the point of offline compilation was to test the _consumption_ of binaries and of SPIR-V, not the generation of them. It certainly seems like useful testing to do, but I am not sure this would be testing conformance per se.

[...]

PR #1034, opened 2020-10-30
Skip `features_macro` Test in Offline Mode

FranklandJack:
- [x] Skip the `features_macro` test in offline mode since it makes
calls to `create_single_kernel_helper_create_program` and fails when the
call doesn't return `CL_SUCCESS` which will happen when the offline
compiler fails to compile the kernel which is deliberately supposed to
fail compilation in this test.


gwawiork (approved):
I've read discussion and agree with conclusion LGTM

FranklandJack:
Thanks for reviewing. For reference the discussion in question is #1026.

kpet (approved):

PR #1052, opened 2020-11-09
Restored the embedded reduction factor to bruteforce.

### LINKED ISSUE #1045: Bruteforce is missing the embedded reduction factor
During the consolidation of CTS from GitLab, it seems that the embedded reduction factor that was present in the bruteforce test has been lost.

Other tests (such as conversions) retained the reduction factor.

This issue tracks re-adding the reduction factor into bruteforce.

Jeremy-Kemp:
This change was present on the GitLab branch but missed out during the transition to GitHub.

This change is intentionally as close as possible to the patch on GitLab.

Fixes #1045

### DIFF
--- a/test_conformance/math_brute_force/Utility.h
+++ b/test_conformance/math_brute_force/Utility.h
@@ -229,6 +229,36 @@ void logFunctionInfo(const char *fname, unsigned int float_size, unsigned int is
 
 float getAllowedUlpError(const Func *f, const bool relaxed);
 
+static inline cl_uint setTestScale(size_t typeSize)
+{
+    if (gWimpyMode)
+    {
+        return (cl_uint)typeSize * 2 * gWimpyReductionFactor;
+    }
+    else if (gIsEmbedded)
+    {
+        return EMBEDDED_REDUCTION_FACTOR;
+    }
+    else
+    {
+        return 1;
+    }
+}
+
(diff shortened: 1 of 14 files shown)


EwanC:
Trying to find `EMBEDDED_REDUCTION_FACTOR ` in [gitlab](https://gitlab.khronos.org/opencl/conformance-tests/-/tree/master) to compare. Was it on a particular branch? Only seeing search hits in the conversions tests

EwanC (dismissed):

kpet (commented):
Could the scaling logic be extracted out as a (small) number of utility functions? It seems 2 different blocks of code are repeated over and over.

Jeremy-Kemp:
@kpet I've created the helper functions and used them in locations that this change requires.

kpet (dismissed):
Thanks, this will make maintenance easier.

I find the `set*` name a bit confusing as the functions don't mutate any state. `get*` or `calc*` or something would have been clearer IMHO.

EwanC (dismissed):
> I find the set* name a bit confusing as the functions don't mutate any state. get* or calc* or something would have been clearer IMHO.

Would also have a preference for `get*` etc, but LGTM otherwise

Jeremy-Kemp:
I've pushed a change that simply renames the `get*` functions to `set*` instead.

EwanC (approved):

kpet (approved):

PR #1053, opened 2020-11-09
Fixes for kernel private memory size test

jrprice:
* Add the test source to `CMakeLists.txt`
* Register the test with the api suite
* Fix build error in host code
* Use `uint` instead of `ulong` (not all devices support 64-bit types)


jrprice:
@chemis01 please take a look.

chemis01:
@jrprice Looks good to me, thank you.

kpet (approved):

Issue #832, opened 2020-06-24
Missing tests for CL_KERNEL_ATTRIBUTES


jenatali:
There's only one test for this, as part of the SPIR test. Probably should tests for this query from source and SPIR-V. See https://github.com/microsoft/OpenCLOn12/issues/4.

kpet:
We probably want exhaustive coverage for all attributes and valid combinations thereof:
- `vec_type_hint`
- `work_group_size_hint`
- `reqd_work_group_size`
- `nosvm`
- Any others?

jenatali:
The first 3 were the only ones I noticed were missing.

rjodinchr:
Fixed by PR #1055 (commit `71e26814`), which added `test_conformance/api/test_kernel_attributes.cpp` to thoroughly validate `clGetKernelInfo` queries for `CL_KERNEL_ATTRIBUTES` (`vec_type_hint`, `work_group_size_hint`, and `reqd_work_group_size`).

PR #1055, opened 2020-11-12
Add Test for CL_KERNEL_ATTRIBUTES (#832)

chemis01:
This test generates dummy kernels which have different
combinations of the following attributes:

    * vec_type_hint
    * work_group_size_hint
    * reqd_work_group_size

It then gets the attributes by using clGetKernelInfo
and validates that the attributes returned are correct
by matching the attributes which were used to generate
the kernel are present in the returned string from
clGetKernelInfo.

Additionally, this PR adds some functionality to the Harness to make it easier and more intuitive to know if a device supports the Half or Double types. 

### DIFF
--- a/test_conformance/api/test_kernel_attributes.cpp
+++ b/test_conformance/api/test_kernel_attributes.cpp
@@ -60,12 +60,10 @@ AttributePermutations reqd_vect_work_tests;
 static KernelAttributes generate_vec_type_hint_data(cl_device_id deviceID)
 {
     KernelAttributes vec_type_hint_data;
-    std::vector<std::string> vector_types = { "char",   "uchar", "short",
-                                              "ushort", "int",   "uint",
+    std::vector<std::string> vector_types = { "uchar", "ushort", "uint",
                                               "float" };
     if (gHasLong)
     {
-        vector_types.push_back("long");
         vector_types.push_back("ulong");
     }
     if (device_supports_half(deviceID))
--- a/test_conformance/api/test_kernel_attributes.cpp
+++ b/test_conformance/api/test_kernel_attributes.cpp
@@ -60,6 +60,7 @@ AttributePermutations reqd_vect_work_tests;
 static KernelAttributes generate_vec_type_hint_data(cl_device_id deviceID)
 {
     KernelAttributes vec_type_hint_data;
+    // TODO Test for signed vectors (char/short/int/etc)
     std::vector<std::string> vector_types = { "uchar", "ushort", "uint",
                                               "float" };
     if (gHasLong)
--- a/test_conformance/api/test_kernel_attributes.cpp
+++ b/test_conformance/api/test_kernel_attributes.cpp
@@ -1,0 +1,340 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include <iostream>
+#include <vector>
+#include <string>
+#include <algorithm>
+#include "procs.h"
+#include "harness/errorHelpers.h"
+#include "harness/typeWrappers.h"
+#include "harness/parseParameters.h"
+
+using KernelAttributes = std::vector<std::string>;
+
+static std::string generate_kernel_source(const KernelAttributes& attributes)
+{
+    std::string kernel;
+    for (auto attribute : attributes)
+    {
+        kernel += "__attribute__((" + attribute + "))\n";
+    }
+    kernel += "__kernel void test_kernel(){}";
+    return kernel;
+}
+
+
+using AttributePermutations = std::vector<KernelAttributes>;
+
+// The following combinations have been chosen as they place each of the
+// attribute types in the different orders that they can occur. While distinct
+// permutations would provide a complete overview of the API the sheer number of
+// combinations increases the runtime of this test by an unreasonable amount
+AttributePermutations vect_tests;
+AttributePermutations work_tests;
+AttributePermutations reqd_tests;
+
+AttributePermutations vect_reqd_tests;
+AttributePermutations work_vect_tests;
+AttributePermutations reqd_work_tests;
+
+AttributePermutations vect_work_reqd_tests;
+AttributePermutations work_reqd_vect_tests;
+AttributePermutations reqd_vect_work_tests;
+
+
+// Generate a vector with vec_type_hint(<data_type>) so that it can be used to
+// generate different kernels
+static KernelAttributes generate_vec_type_hint_data(cl_device_id deviceID)
+{
+    KernelAttributes vec_type_hint_data;
+    std::vector<std::string> vector_types = { "char",   "uchar", "short",
+                                              "ushort", "int",   "uint",
+                                              "float" };
+    if (gHasLong)
+    {
+        vector_types.push_back("long");
+        vector_types.push_back("ulong");
+    }
+    if (device_supports_half(deviceID))
+    {
+        vector_types.push_back("half");
+    }
+    if (device_supports_double(deviceID))
+    {
+        vector_types.push_back("double");
+    }
+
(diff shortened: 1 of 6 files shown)


jrprice:
Not a thorough review, but please note that 64-bit integer types (`long`/`ulong`) are optional for embedded profile, so these should not be tested unconditionally. They should be guarded with checks similar to `double` and `half`, e.g. if the device is embedded profile only run them if `cles_khr_int64` is supported.

test_conformance/api/test_kernel_attributes.cpp
```diff
@@ -0,0 +1,322 @@
...
+        std::string attribute_string(attributes.data());
+        attribute_string.erase(
+            std::remove(attribute_string.begin(), attribute_string.end(), ' '),
+            attribute_string.end());
+        for (auto attribute : attribute_permutation)
+        {
+            if (attribute_string.find(attribute) == std::string::npos)
+            {
+                success = false;
+                log_error("ERR: expected: %s\nFound: %s\n", attribute.c_str(),
```
StuartDBrady, line 300: This output could be made a little clearer.  Would suggest `"ERROR: did not find expected attribute: '%s'\n"`, with the full attribute list printed just once if a failure is found for the permutation, with a message such as `"Attribute list reported is: '%s'\n"`.

I would note that this doesn't raise an error if extra attributes were reported that were not in the permutation. Should it?

StuartDBrady: The most recent changes look good to me. I see now that the `CL_KERNEL_ATTRIBUTES` info is not trivial to parse. I think it would be good to check that there are no extraneous attributes, but this could be added some other time.

Would be good to have independent review of the rest of this PR.

StuartDBrady (commented):
This looks good to me but it would be good to have some independent review.

chemis01:
Hi all,
Are there any issues with this PR? I'd be happy to make some changes so that this can be merged.
Thanks

test_conformance/api/test_kernel_attributes.cpp
```diff
@@ -0,0 +1,343 @@
...
+#include <vector>
+#include <string>
+#include <algorithm>
+#include "procs.h"
+#include "harness/errorHelpers.h"
+#include "harness/typeWrappers.h"
+#include "harness/kernelHelpers.h"
+#include "harness/parseParameters.h"
+
+#define __CL_ENABLE_EXCEPTIONS
```
jrprice, line 26: I think this should be removed.

test_conformance/api/test_kernel_attributes.cpp
```diff
@@ -0,0 +1,343 @@
...
+/* Populate the Global Vectors which store individual Kernel Attributes */
+static void populate_single_attribute_tests(
+    /* Vectors to store the different data that fill the attributes */
+    const KernelAttributes& vec_type_hint_data,
+    const KernelAttributes& work_group_size_hint_data,
+    const KernelAttributes& reqd_work_group_size_data)
+{
+    for (auto vector_test : vec_type_hint_data)
+    {
+        /* Initialise vec_type_hint attribute tests */
```
jrprice, line 142: Very nitty style point, but I think we should try to consistently use C++ comments in most places (`//`), as we're generally moving the codebase to be more canonically C++.

test_conformance/api/test_kernel_attributes.cpp
```diff
@@ -0,0 +1,343 @@
...
+    {
+
+        std::string kernel_source_string =
+            generate_kernel_source(attribute_permutation);
+        const char* kernel_src = kernel_source_string.c_str();
+        clProgramWrapper program;
+        clKernelWrapper kernel;
+        cl_int err = create_single_kernel_helper(context, &program, &kernel, 1,
+                                                 &kernel_src, "test_kernel");
+        ASSERT_SUCCESS(err, "create_single_kernel_helper");
```
jrprice, line 282: I think the checks here should use `test_error` instead of `ASSERT_SUCCESS`. The latter will throw an exception and abort the run, which doesn't play nicely when not running this test in isolation.

jrprice (changes requested):
Looks mostly fine, just a couple of nits and one specific thing I think should be changed.

test_conformance/api/test_kernel_attributes.cpp
chemis01: Ok.
Keeping parseParameters.h as it is needed for determining the compilation mode.

jrprice: I was just talking the unused macro definition (GitHub added the extra context automatically). The `kernelHelpers.h` include could be left in for IWYU reasons since there are functions from it used here, but I'm not going to make you change this again before merging.

jrprice (dismissed):
LGTM, thanks for the updates.

As an aside, I'd find it much easier/quicker to review updates to PRs if they were pushed as new commits rather than with amend+force-push. When you rewrite history like this GitHub doesn't let you see what changed since the last time you reviewed, which means you have to review the whole file again. We always squash the history when merging PRs so the extra commits don't matter.

bashbaug (commented):
We're discussing this test internally.  Could you please hold off merging it for a little longer?  Thanks!

kpet:
@bashbaug Any update?

bashbaug:
Sorry for the very slow reply - somehow I missed the "mention" nag above.

We're currently failing this test, though the reason why is somewhat interesting.  As background, we compile through SPIR-V by default even in the compile-from-source path, so we rely on SPIR-V to carry through all attributes.

At the SPIR-V level, we have:

VecTypeHint
A hint to the compiler, which indicates that most operations used in the entry point are explicitly vectorized using a particular vector type. The 16 high-order bits of the Vector Type operand specify the number of components of the vector. The 16 low-order bits of the Vector Type operand specify the data type of the vector.

These are the legal data type values:
0 represents an 8-bit integer value.
1 represents a 16-bit integer value.
2 represents a 32-bit integer value.
3 represents a 64-bit integer value.
4 represents a 16-bit float value.
5 represents a 32-bit float value.
6 represents a 64-bit float value.

Note that SPIR-V does not distinguish between signed and unsigned integer values, so we're losing the signedness of the `vec_type_hint`, and either failing the test for a signed attribute or an unsigned attribute since we have to pick one arbitrarily.

If we want to make short-term forward progress with this PR I think we have several options:

1. Enable the `work_group_size_hint` and `reqd_work_group_size` tests but keep `vec_type_hint` test disabled.
2. Enable all three tests and either pass signed or unsigned integer types for the `vec_type_hint` or warn on this behavior, but don't fail the test.

Longer-term we need to decide what we want to do for this attribute and/or the query and either fix the OpenCL spec, find a way to carry the signedness through SPIR-V, or we'll need to find some other way to carry the attributes through our implementation.

bashbaug:
Hi, we're still failing the new test for the reason described above:

```
$ ./test_conformance/api/test_api kernel_attributes
Compute Device Name = Intel(R) Graphics [0x5916], Compute Device Vendor = Intel(R) Corporation, Compute Device Version = OpenCL 3.0 NEO , CL C Version = OpenCL C 3.0 
Device latest conformance version passed: v2020-11-23-00
sizeof( void*) = 8  (host)
sizeof( void*) = 8  (device)
kernel_attributes...
ERROR: did not find expected attribute: 'vec_type_hint(char2)'
Attribute string reported as: vec_type_hint(uchar2)
...
```

kpet:
@bashbaug Interesting. We've removed the signed cases to get forward progress on this PR (@chemis01 You may want to leave a TODO/FIXME in the code pointing at a new issue in this project). Let us know if you're now passing the updated tests.

test_conformance/api/test_kernel_attributes.cpp
```diff
@@ -0,0 +1,339 @@
...
+AttributePermutations work_tests;
+AttributePermutations reqd_tests;
+
+AttributePermutations vect_reqd_tests;
+AttributePermutations work_vect_tests;
+AttributePermutations reqd_work_tests;
+
+AttributePermutations vect_work_reqd_tests;
+AttributePermutations work_reqd_vect_tests;
+AttributePermutations reqd_vect_work_tests;
```
bashbaug, line 55: This probably does not matter in the grand scheme of things, but I do observe that these permutations are stored in global variables and never freed or cleared, so all of the permutations will persist beyond the execution of this test.  It would be nicer to store them in local variables so the destructors will run when the test completes, or to clear their contents before the test returns pass or fail, but "leaking" the permutations is probably fine for most systems with a reasonable amount of memory.

bashbaug (approved):
Thanks, with the latest changes these tests pass on our implementation.

I left one comment, but please consider it a nice-to-have and not a must-have, and I'm fine with these changes being merged as-is.

kpet:
Remaining comments to be addressed under https://github.com/KhronosGroup/OpenCL-CTS/issues/1248

Issue #522, opened 2019-12-18
Rewrite / redesign api:get_kernel_arg_info


kpet:
We have two different versions of that test (one from the compatibility api suite). Both roughly 5k lines of copy/pasted code. These are essentially unmaintainable and should be rewritten using loops to create combinations programmatically from data tables.

kpet:
https://github.com/KhronosGroup/OpenCL-CTS/pull/1056 fixed this. Closing.

PR #1056, opened 2020-11-17
Redesign clGetKernelArgInfo (#522)

chemis01:
This PR implements changes to the Harness to make it easier and clearer to determine whether a device supports the half and double data types. This is the same changes as in https://github.com/KhronosGroup/OpenCL-CTS/pull/1055.

The remainder of the PR is a complete redesign of clGetKernelArgInfo. It is split into 2 different commits. The first commit removes the previous tests and the second commit implements the redesign.


jrprice:
Not a thorough review, but please note that 64-bit integer types (long/ulong) are optional for embedded profile, so these should not be tested unconditionally. They should be guarded with checks similar to double and half, e.g. if the device is embedded profile only run them if cles_khr_int64 is supported.

test_conformance/api/test_kernel_arg_info.cpp
```diff
@@ -13,5931 +13,786 @@
...
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "uint8*", "globalconstuint8p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "float8*", "globalconstfloat8p",
-        NULL
-    },
-    {
-        "global_const_vector8_restrict_p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST|CL_KERNEL_ARG_TYPE_RESTRICT), "char8*", "globalconstchar8restrictp",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST|CL_KERNEL_ARG_TYPE_RESTRICT), "uchar8*", "globalconstuchar8restrictp",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST|CL_KERNEL_ARG_TYPE_RESTRICT), "short8*", "globalconstshort8restrictp",
-        (const cha
```
StuartDBrady, line 258: These two lines are duplicates of the line above.

StuartDBrady, line 165: Is it really necessary to include "CL_KERNEL_ARG_ADDRESS_" in the output? I think "global" would be better. It is obvious from context what the corresponding name of the enum is in the API.

StuartDBrady, line 168: Likewise.

StuartDBrady, line 171: Likewise.

StuartDBrady, line 174: Likewise.

StuartDBrady, line 185: Likewise, could this be shortened to "read_only" please?

StuartDBrady, line 188: ... and to "write_only", etc.

StuartDBrady, line 205: ... and again to "const", "restrict", etc.? This is where the verbosity of the output is _really_ a problem as there can be multiple qualifiers set.

StuartDBrady, line 263: This logs the "expected" and "actual" info, but the argument as specified in the kernel source can be different to the expected output. Could this also log the argument as specified in the kernel source, please? It is somewhat unhelpful for this not to do so, given that many arguments have the same expected output.

StuartDBrady (changes requested):
Mostly looks good. Some minor issues regarding the test output and some duplicate logic.

Does this improve coverage of the tests? If so, this should be stated in the description of the pull request.

I notice there are typos in one of the commit messages "incresased maintanability" should be "increased maintainability". I expect these messages will be lost when the commits are squashed anyway, so should more be copied into the description of the pull request?

test_conformance/api/test_kernel_arg_info.cpp
StuartDBrady: GitHub's review system is apparently buggy and is showing the wrong lines.

The lines I am referring to are lines 256, 257 and 258:
```
256         || (actual.type_qualifier != expected.type_qualifier)
257         || (actual.type_qualifier != expected.type_qualifier)
258         || (actual.type_qualifier != expected.type_qualifier)
```

StuartDBrady: Of course, here, I'm referring to lines 165, 168, 171 and 174:
```
165             return "CL_KERNEL_ARG_ADDRESS_GLOBAL";
```

StuartDBrady: ... and here, line 185:

```
185             return "CL_KERNEL_ARG_ACCESS_READ_ONLY";
```

StuartDBrady: ... and here, line 205, etc:
```
205         ret += "CL_KERNEL_ARG_TYPE_CONST ";
```

StuartDBrady: Here, I'm referring to line 263 in `compare_expected_actual()`:

```
263         output_difference(expected, actual);
```

StuartDBrady:
GitHub's review system has apparently made a mess of my review.  If you go to the "Files changed" page rather than viewing the review comments in "Conversation", it should hopefully make more sense.

test_conformance/api/test_kernel_arg_info.cpp
StuartDBrady, line 483: If `is_pointer == is_private` we skip. So couldn't we just set `is_pointer = !is_private` and do away with the `is_pointer` loop, here?

Of course, all function arguments including kernel arguments and including all pointer arguments are in the private address space, in OpenCL C. For a pointer, the _pointee_ might be in the global, local or constant address space but the _pointer_ itself is not. However, the use of address space qualifiers here matches the OpenCL API, so I think it's reasonable to rely on "private" meaning "not a pointer" here. It might be worth a comment, though.

StuartDBrady: ... and yes, as we might have guessed, GitHub has made another mess of things.

I'm referring to lines 477-486, here:
```
477             /* OpenCL kernels cannot take "private" pointers */
478             if (is_pointer && is_private)
479             {
480                 continue;
481             }
482             /* Only "private" variables can take values */
483             if (!is_pointer && !is_private)
484             {
485                 continue;
486             }
```

chemis01: Thanks for the comments. I've just pushed some changes which should address these

test_conformance/api/test_kernel_arg_info.cpp
```diff
@@ -13,5931 +13,782 @@
...
```
StuartDBrady, line 553: The lifetime of arg_string is less than that required for its use in expected_args.  Either the arg_string member of the expected_args should be a string rather than a reference to a string, or it should be removed and all_args passed through to compare_kernel_with_expected().

You could test this locally by updating compare_expected_actual(), replacing its condition so that all results are regarded as differing to the expected result.

StuartDBrady: It would be cleaner, IMO, for created_expected_arg_info() to be a method on KernelArgInfo, so that you don't have to pass each of KernelArgInfo's data members as an individual parameter.

StuartDBrady (changes requested):
Looks better. Just one minor issue with object lifetimes and a question about refactoring.

test_conformance/api/test_kernel_arg_info.cpp
StuartDBrady: > The lifetime of arg_string is less than that required for its use in expected_args. Either the arg_string member of the expected_args should be a string rather than a reference to a string, or it should be removed and all_args passed through to compare_kernel_with_expected().

My apologies, I see that the arg_string field of KernelArgInfo itself is a string, not a string reference, so it looks like it should work. It would be cleaner to refactor create_expected_arg_info() IMO, though.

StuartDBrady (changes requested):
Looks good, although I now notice there are some gaps in test coverage.

C permits some flexibility in the ordering of types and qualifiers, and the OpenCL C spec does not contain any language to constrain this. In addition, it is highly probable that all OpenCL C implementations provide this same flexibility, and it is likely that there is code which relies upon it.

The 'unsigned' and 'signed' qualifiers may be placed freely with respect to the type (provided that they 'unsigned' and 'signed' are not mixed together in a single declaration). For example, the following all refer to the same type:
* int
* signed
* signed int
* int signed
* signed signed int signed

Likewise, the following all refer to the same type:
* short
* short int
* signed short
* signed short int
* int short
* short signed
* short signed int
* short int signed
* signed int short
* int signed short
* int short signed
* short short signed signed short signed

Further more, the following are equivalent:
* const unsigned char a
* unsigned const char a
* char const unsigned a
* char unsigned const a
* const char unsigned a
* unsigned char const a
* const const const unsigned const char const const a

Note that it when a pointer is qualified, this has a different effect to qualification of the pointee. The following four types are each distinct:

* char *a
* char *const a
* const char *a
* const char *const a

There are likely some limits imposed by e.g. maximum source line length, etc. but even if we consider only cases without any duplication within them, it is not practical to list them in full in the source code.

I would therefore recommend either adding a _minimal_ set of extra scalar types (i.e. "signed char", "signed int", "signed short", "signed long", "int signed", "char signed", "short unsigned", "long unsigned", "signed", "unsigned"), or otherwise reworking the code to generate different orderings.

It would also be beneficial to ensure that qualifiers of pointers (rather than their pointees) are not reflected in the output of clGetKernelArgInfo().

Realistically, it is likely that any OpenCL C frontend tested by the CTS will be derived from an existing C frontend, and so it is probably not worth placing a whole lot of effort into testing equivalence in these cases, but note that the metadata for clGetKernelArgInfo() is specific to OpenCL, and so it seems worth testing this to some minimal extent.

It may also be worth testing that the ordering of various types of qualifier do not affect the output, e.g. image2d_t read_only vs read_only image2d_t, but I am not aware of the specifics of much of this.

chemis01:
Hi all,
Are there any further issues with this PR? I'd be happy to make some changes so that this can be merged.
Thanks

test_conformance/api/test_kernel_arg_info.cpp
StuartDBrady: The duplicate lines have now been removed.

StuartDBrady: These have been shortened to just "GLOBAL", "LOCAL", "PRIVATE" and "CONSTANT".

StuartDBrady: Has been shortened to just "READ_ONLY".

StuartDBrady: These have been shortened to "CONST", "RESTRICT", etc.

StuartDBrady: Thanks. The changes look good.

test_conformance/api/test_kernel_arg_info.cpp
StuartDBrady: create_expected_arg_info() no longer takes each of the data members of KernelArgInfo as separate parameters, which looks better.

StuartDBrady (approved):
The changes since my previous review look good.

It would be good to have a somewhat thorough independent review of this, though.

test_conformance/api/test_kernel_arg_info.cpp
StuartDBrady: This has been shortened to "WRITE_ONLY".

test_conformance/api/test_kernel_arg_info.cpp
```diff
@@ -1,6 +1,6 @@
 //
-// Copyright (c) 2017 The Khronos Group Inc.
-// 
+// Copyright (c) 2020 The Khronos Group Inc.
```
jlewis-austin, line 2: We made it to 2021 :-)  I've seen others doing "2017, 2012" but I'm not sure what the policy is

test_conformance/api/test_kernel_arg_info.cpp
```diff
@@ -13,5931 +13,828 @@
...
-    },
-    {
-        "global_const_vector8_p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "char8*", "globalconstchar8p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "uchar8*", "globalconstuchar8p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "short8*", "globalconstshort8p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "ushort8*", "globalconstushort8p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "int8*", "globalconstint8p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "uint8*", "globalconstuint8p",
-
```
jlewis-austin, line 339: typo

jlewis-austin, line 796: I wonder if the result logging would be better interleaved with the test runs so test progress is more apparent?

jlewis-austin, line 806: This is a little bit misleading if images aren't supported. If it was moved up to where the test are run, we could customize the message to indicate whether they passed or were skipped.

jlewis-austin, line 464: It would be helpful to see the type here for readability

jlewis-austin (commented):
This is a really nice improvement! Just a few nits. I'm not sure how the coverage compares to the original, but it looks pretty complete. Is the run time similar?

test_conformance/api/test_kernel_arg_info.cpp
chemis01: Done

chemis01: Moved it to only print if the test is run

chemis01: Done, it does make more sense to interleave the output.

chemis01: Done

test_conformance/api/test_kernel_arg_info.cpp
chemis01: Done, I started this at the end of last year and forgot to update this :'D

test_conformance/api/test_kernel_arg_info.cpp
```diff
@@ -13,5931 +13,827 @@
...
```
bashbaug, line 285: I think we generally prefer not to call `ASSERT_SUCCESS` since it will thrown an exception and top execution if a failure is encountered.  Could we call `test_error` or similar instead?

bashbaug (changes requested):
This test is failing on at least one of our conformant devices due to use of too many kernel arguments.  We report 2048 for `CL_DEVICE_MAX_PARAMTER_SIZE` on the device I tested, which is double the required minimum ([link](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_PARAMETER_SIZE)).  Could you please verify that the test does not require too many kernel arguments after this rewrite?

The test seems to be throttling the **number** of kernel arguments (to 100, currently), but it needs to throttle the total **size** of kernel arguments as well.

chemis01:
> This is a really nice improvement! Just a few nits. I'm not sure how the coverage compares to the original, but it looks pretty complete. Is the run time similar?

Yes, from what I can tell there is very little difference between the runtimes of this and the original version

test_conformance/api/test_kernel_arg_info.cpp
chemis01: Replaced the ASSERT's with test_error and made a check to ensure that the limit specified by the device isn't exceeded.

bashbaug (changes requested):
Thanks, we're getting closer now and I can run the test to completion, but I'm still seeing failures that might be test issues.  A few others could be issues on our end - we'll need to investigate:

1. I'm still seeing kernels that are failing compilation due to too many parameters, but it's a lot closer now.  I suspect one of these is happening: the size of vec3s is not being tracked correctly, or we're requiring a minimum (4-byte?) alignment for all parameters, or there are some extra paramters we need to pass implicitly to a kernel that's throwing off the calculation.  I'm not sure if this test should leave a little extra room or if some other solution is needed - open to suggestions.
2. I'm seeing kernels generated with `global read_write  image2d_t  argument0`, but Clang is rightfully (I think) generating an error in this case: https://godbolt.org/z/x1f6M7
3. A bunch of pipe cases are failing.  I'm not entirely sure if these are test errors, Clang errors, or a problem with our implementation.  Have these tests been verified on an implementation that supports pipes?

Thanks for continuing to work on this!

test_conformance/api/test_kernel_arg_info.cpp
```diff
@@ -117,8 +117,17 @@ class KernelArgInfo {
...
     std::string ret;
-    std::string address_qualifier =
-        get_address_qualifier(kernel_arg.address_qualifier);
+
+    const bool is_image = strstr(kernel_arg.arg_type, "image")
+        || strstr(kernel_arg.arg_type, "sampler");
+    std::string address_qualifier = "";
+    // Image Objects are always allocated from the global address space so the
+    // qualifier should not be specified
+    if (!is_image)
```
StuartDBrady, line 126: If `is_image` is true, can't `kernel_arg.address_qualifier` also be `CL_KERNEL_ARG_ADDRESS_LOCAL`, `CL_KERNEL_ARG_ADDRESS_CONSTANT`, or `CL_KERNEL_ARG_ADDRESS_PRIVATE`, here?

If so, then `run_scalar_vector_tests()` should have another `continue` statement added to avoid duplication.

chemis01: No, we set the address qualifier once for images to be global at the start on line 744.

StuartDBrady: I'd missed that image tests are handled in `run_image_tests()`, and (obviously) not `run_scalar_vector_tests()`.

chemis01:
@bashbaug. I pushed a some changes to address the following: 
1) Sets the alignment/space requirement for Vector3's to take up the same space as Vector4's
2) Ensures that image args don't have an explicit address qualifier specified.

Please could you verify that it works?

Regarding 3) the test has been tested on a platform which supports pipes but I believe there are a number of issues with different implementations regarding this. I believe @StuartDBrady knows a bit more on the issue

bashbaug:
> Please could you verify that it works?

Thanks for continuing to look at this, but I'm still seeing a case that is generating too many parameters.  I actually wonder if this has to do with pipes also.  Does the test currently assume that `sizeof(pipe char) == sizeof(char)`?  If so, I don't think we ever explicitly say this in the spec (and we should!), but it looks like Clang and our implementation treats the size of a pipe as the size of a pointer, so up to 8 bytes: https://godbolt.org/z/8xdbWz

Do we need to open a spec bug to clarify the size of a pipe?

We're currently failing a few other cases too, but at first glance those failures appear to be related to https://github.com/KhronosGroup/OpenCL-Docs/pull/558.

chemis01:
You're right it does assume that `sizeof(pipe char) == sizeof(char)`. 
This is addressed in the next patch.
Yes I think it would be worth explicitly stating what the size of a pipe is.

test_conformance/api/test_kernel_arg_info.cpp
```diff
@@ -1,5943 +1,931 @@
...
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST), "float8*", "globalconstfloat8p",
-        NULL
-    },
-    {
-        "global_const_vector8_restrict_p",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST|CL_KERNEL_ARG_TYPE_RESTRICT), "char8*", "globalconstchar8restrictp",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST|CL_KERNEL_ARG_TYPE_RESTRICT), "uchar8*", "globalconstuchar8restrictp",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST|CL_KERNEL_ARG_TYPE_RESTRICT), "short8*", "globalconstshort8restrictp",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_ACCESS_NONE, (const char *)(CL_KERNEL_ARG_TYPE_CONST|CL_KERNEL_ARG_TYPE_RESTRICT), "ushort8*", "globalconstushort8restrictp",
-        (const char *)CL_KERNEL_ARG_ADDRESS_GLOBAL, (const char *)CL_KERNEL_ARG_AC
```
bashbaug, line 546: This code sequence doesn't handle the cases where the `arg_type` is just `signed` or `unsigned` with no other type.  This causes a kernel to be generated with too many parameters.

It might be good to add an assertion that the size is non-zero at the end of this function to catch any errors like these in the future - it took me a while to spot this!

bashbaug (changes requested):
I finally found out why one kernel was still being generated that used too many parameters.

There are still other cases that are failing, at least on our implementation, but this should fix the last of the "too many parameter" kernels.  Thanks!

bashbaug:
Looks like the remaining failures on our implementation fall into three different categories:

1. `char` vs. `signed char`.  See: https://github.com/KhronosGroup/OpenCL-Docs/pull/558

```
Arg Type: Expected: char*	 Actual: signed char*
Argument in Kernel Source Reported as:
global   signed char*  argument14
```

2. Pipes with scalar types and signed or unsigned type modifiers.

```
Arg Type: Expected: uchar	 Actual: unsigned char
Argument in Kernel Source Reported as:
private read_only pipe  uchar  argument92
```

3. Pipes with vector types.

```
Arg Type: Expected: char2	 Actual: char __attribute__((ext_vector_type(2)))
Argument in Kernel Source Reported as:
private read_only pipe  char2  argument9
```

(3) and probably also (2) appear to be due to Clang's handling of pipes.  See: https://godbolt.org/z/cMPP96

```
!5 = !{!"char __attribute__((ext_vector_type(2)))", !"uint*"}
```

I'd be curious to know if any implementations that support pipes and use a largely unmodified Clang are able to pass this test.

chemis01:
@bashbaug thank you for your comments, regarding:
1) Given the confusion and lack of clarity in the OpenCL spec for signed char vs char, and that OpenCL C specifies that a Char is a Signed Char by default, is it acceptable to ignore the "signed char" arguments for the conformance test?

2 & 3) I believe both of these issues are related to Clangs handling of Pipes and a solution is being/has been developed https://reviews.llvm.org/D96151

test_conformance/api/test_kernel_arg_info.cpp
StuartDBrady: This may perhaps be obvious already, but it would probably be best if possible to use `get_expected_arg_type()` here rather than duplicating logic, although it's generally best to avoid doing string handling _at all_ when referring to data types, if you can.

bashbaug:
> 1. is it acceptable to ignore the "signed char" arguments for the conformance test?

Yes, I think so, but please monitor the resolution of https://github.com/KhronosGroup/OpenCL-Docs/pull/558.



> 2 & 3) I believe both of these issues are related to Clangs handling of Pipes and a solution is being/has been developed

Does it make sense to disable the pipe parts of this test by default until this issue is fixed?  We won't be able to pass this test as-is until the Clang solution is complete (and we've had time to integrate it).

I have a few additional pipe-specific comments I'll leave inline in the code as well.

test_conformance/api/test_kernel_arg_info.cpp
bashbaug, line 304: This isn't safe - it's possible that an OpenCL 3.0 device will support pipes but not OpenCL C 2.0, and hence not `-cl-std=CL2.0`.

bashbaug, line 418: I don't see that this test is currently mindful of `CL_DEVICE_MAX_PIPE_ARGS`.  I wonder if it makes more sense to test pipes separately, similar to images, where the number of pipe args can be more tightly controlled.  What do you think?

bashbaug, line 535: This matches the current Clang implementation, but please see https://github.com/KhronosGroup/OpenCL-Docs/issues/563, which concluded that the size of a pipe is implementation-defined.

bashbaug (changes requested):

chemis01:
> Small update, since it's been a few months:
> 
> We're still failing the pipes part of this test due to the Clang issues mentioned above, but we're passing the test otherwise.
> 
> Have you given any thought to merging these changes with the pipes part disabled or optional? I don't think anyone that supports pipes and uses Clang will be able to pass this test otherwise. If this test gets merged as-is we'd probably just stop supporting pipes (which, honestly, we've been considering anyhow).

@bashbaug I've disabled the tests in the latest patch. Please let me know if you are still having problems.

bashbaug (approved):
The latest change that disables the pipes tests work for me - thanks!

Issue #1057, opened 2020-11-17
Conversions test failure with Clang build


sravikumar3393:
Building with clang 8.0.1 on Ubuntu 18.04 resulted in two test failures from 'conversions' test in 'Release' mode:
`float_rtn_uint
double_rtn_ulong`

The two tests pass when built with gcc 7.5.0 compiler.

Error message:
`623) Testing convert_floatn_rtn( uintn ):
Building convert_float_rtn( uint ) test
Building convert_float2_rtn( uint2 ) test
Building convert_float3_rtn( uint3 ) test
Building convert_float4_rtn( uint4 ) test
Building convert_float8_rtn( uint8 ) test
Building convert_float16_rtn( uint16 ) test
Testing... .
Error for vector size 1 found at 0x00000000: *-0x0p+0 vs 0x0p+0

Error for vector size 2 found at 0x00000000: *-0x0p+0 vs 0x0p+0

Error for vector size 3 found at 0x00000000: *-0x0p+0 vs 0x0p+0

Error for vector size 4 found at 0x00000000: *-0x0p+0 vs 0x0p+0

Error for vector size 8 found at 0x00000000: *-0x0p+0 vs 0x0p+0

Error for vector size 16 found at 0x00000000: *-0x0p+0 vs 0x0p+0
Input value: 0x00000000 (convert_float_rtn( uint ))`

`Building convert_double_rtn( ulong ) test
Building convert_double2_rtn( ulong2 ) test
Building convert_double3_rtn( ulong3 ) test
Building convert_double4_rtn( ulong4 ) test
Building convert_double8_rtn( ulong8 ) test
Building convert_double16_rtn( ulong16 ) test
Testing... .
Error for vector size 1 found at 0x000000bf:  *-0x0p+0 vs 0x0p+0

Error for vector size 2 found at 0x000000bf:  *-0x0p+0 vs 0x0p+0

Error for vector size 3 found at 0x000000bf:  *-0x0p+0 vs 0x0p+0

Error for vector size 4 found at 0x000000bf:  *-0x0p+0 vs 0x0p+0

Error for vector size 8 found at 0x000000bf:  *-0x0p+0 vs 0x0p+0

Error for vector size 16 found at 0x000000bf:  *-0x0p+0 vs 0x0p+0
Input value: 0x0000000000000000  (convert_double_rtn( ulong ))
         *** convert_doublen_rtn( ulongn ) FAILED **`

PR is in progress.

PR #1062, opened 2020-11-20
Fix for test_conversions failure with Clang build on Linux #1057

sravikumar3393:
* Avoid optimization by using volatile qualifier
    * Fix both uint2float and ulong2double


sravikumar3393:
"0" is not getting converted to +0.0 in RTN mode because variable "l" is getting optimized out by Clang compiler in the 'Release' version.

jlewis-austin:
The code that was getting optimized is meant to correct common conversion sign bugs (https://bugs.llvm.org/show_bug.cgi?id=47393).
LGTM

kpet (approved):

PR #1065, opened 2020-11-21
Fix build_with_il on devices older than 2.1

jrprice:
The `get_device_il_version_string()` function throws an exception if the
device does not support the `CL_DEVICE_IL_VERSION` query, so don't call
this unless the version is recent enough.


EwanC (approved):
Makes sense :+1:

PR #1070, opened 2020-11-25
gles: Fix compile warnings.

lakshmih:


lakshmih:
Following fixes:

General:
* Improve variable types.
* Make casts explicit to improve readability.
* Improve printf format specifiers.
* Don't warn about using the portable "strdup" on Visual Studio.
* Limit the mark pragma to Apple environments since it's only useful in
  XCode and causes a warning otherwise.
* Remove "harness/threadTesting.cpp" which defines nothing for the last
  three years.
* Move GLES3 macro from CMakeLists.txt to
  test_conformance/gles/CMakeLists.txt to be more specific.

test_common/gles/helpers.cpp:
* Remove needless realWidth and realHeight variables.

test_common/harness/compat.h:
* Don't redefine EXTERN_C on Windows where it's already in the OS
  headers.

test_common/harness/conversions.cpp:
* Fix warning about storing -1 into unsigned long upper limits.

test_common/harness/errorHelpers.h:
* Remove the custom compile time assert in favor of the standard
  "static_assert".

test_common/harness/imageHelpers.cpp:
* Use the standard "static_assert" as supported by Visual Studio 2015
  instead of the custom "ct_assert".

test_common/harness/rounding_mode.h:
* Remove "#pragma STDC FENV_ACCESS ON" which isn't supported by Visual
  Studio nor Android's Q NDK. It may be desirable to conditionally
  reintroduce it as needed in the future.

bashbaug (commented):
This is a nice improvement!  There are certainly fewer warnings with these changes.

I left one comment that I belive will easily eliminate even more warnings.

Are there any specific tests we should run with these changes?  I verified everything builds but I didn't try running any tests since the harnesses are changing and theoretically any test could be affected.

gwawiork (approved):
LGTM. 
I've run full testing set to assure no regression by these changes. All test passed

gwawiork:
> LGTM.
> I've run full testing set to assure no regression by these changes. All test passed

Tested on Linux and Windows

bashbaug (approved):
Please see the comment above - I think this is pretty low-risk and will eliminate even more warnings.  I'm find making this change as a subsequent PR though, if desired.  Thanks!

kpet:
Two approvals, merging this to get forward progress.

PR #1071, opened 2020-11-26
Correct test_consistency_svm

fhazubski-Intel:
Do not use NULL svm pointers.
Add a check for clEnqueueSVMMigrateMem.

Signed-off-by: Filip Hazubski <filip.hazubski@intel.com>

### DIFF
--- a/test_conformance/api/test_api_consistency.cpp
+++ b/test_conformance/api/test_api_consistency.cpp
@@ -79,44 +79,51 @@ int test_consistency_svm(cl_device_id deviceID, cl_context context,
         // CL_INVALID_OPERATION if the device associated with command_queue does
         // not support Shared Virtual Memory.
 
+        // These calls purposefully pass bogus pointers to the functions to
+        // better test that they are a NOP when SVM is not supported.
+        void* bogus0 = (void*)0xDEADBEEF;
+        void* bogus1 = (void*)0xDEADDEAD;
         cl_uint pattern = 0xAAAAAAAA;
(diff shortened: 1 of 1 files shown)


fhazubski-Intel:
Currently _test_consistency_svm_ use `NULL` svm pointers. For all of the functions: `clEnqueueSVMMemcpy`, `clEnqueueSVMMemFill`, `clEnqueueSVMMap`, `clEnqueueSVMUnmap` the specification states that the call returns `CL_INVALID_VALUE` if the svm pointers are `NULL`. For example here for `clEnqueueSVMMemcpy`: https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueSVMMemcpy
Because in the test the expectation is that `CL_INVALID_OPERATION` is returned, I have updated the code to not use `NULL` svm pointers. This will prevent any implementation of OpenCL from returning the unexpected `CL_INVALID_VALUE` value.

Additionally I have added a check for `clEnqueueSVMMigrateMem` function as it seemed to be missing. Again, I ensured that only `CL_INVALID_OPERATION` is allowed to be returned here by the specification.

gwawiork:
The issue  in test_api consistency_svm test case, visible on platforms which do no support SVM feature. 
LGTM !

bashbaug (commented):
Nice catch, LGTM.  Thanks for adding the missing check for `clEnqueueSVMMigrateMem` also.

Needs a non-Intel review before merging.

jrprice (approved):

PR #1073, opened 2020-11-27
Remove command line parameters for platform and device selection

### LINKED ISSUE #711: Remove command line parameters for platform and device selection
The harness currently handles selecting a platform, device and device type with either environment variables or command line parameters (see https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/testHarness.cpp#L165 and https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/testHarness.cpp#L212). Given the state of command line parsing code in general, I suggest we just retain environment variables as a mechanism for now until we rework the command line parsing at least (we probably would want a different interface from the current one).

ellnor01:
Fixes #711

### DIFF
--- a/test_common/harness/testHarness.cpp
+++ b/test_common/harness/testHarness.cpp
@@ -111,14 +111,12 @@ int runTestHarnessWithCheck(int argc, const char *argv[], int testNum,
 
     int err, ret;
     char *endPtr;
-    int based_on_env_var = 0;
 
 
     /* Check for environment variable to set device type */
     char *env_mode = getenv("CL_DEVICE_TYPE");
     if (env_mode != NULL)
     {
-        based_on_env_var = 1;
         if (strcmp(env_mode, "gpu") == 0
             || strcmp(env_mode, "CL_DEVICE_TYPE_GPU") == 0)
             device_type = CL_DEVICE_TYPE_GPU;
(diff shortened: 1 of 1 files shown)


gwawiork:
Hi. 
This change will impact our automated execution. Is there really need to remove this kind of selection without introducing new solution?

kpet:
There is already another way of selecting devices and platforms using the `CL_DEVICE_INDEX`, `CL_PLATFORM_INDEX`  and `CL_DEVICE_TYPE` environment variables. The command-line syntax is quirky and will get in the way of introducing a cleaner command-line parsing logic.

@ellnor01 The help text [there](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/testHarness.cpp#L183) would need updating as well.

kpet:
@gwawiork Any more feedback on this change?

nikhiljnv:
> 
> 
> There is already another way of selecting devices and platforms using the `CL_DEVICE_INDEX`, `CL_PLATFORM_INDEX` and `CL_DEVICE_TYPE` environment variables. The command-line syntax is quirky and will get in the way of introducing a cleaner command-line parsing logic.
> 
> @ellnor01 The help text [there](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/testHarness.cpp#L183) would need updating as well.

I would prefer command-line arguments over env var as relying on env vars can be fragile (it's easy to miss setting them up etc). 
Also, going bit further, just platform_index and device_index doesn't guarantee portability of these indices across different systems due to dependence on enumeration order. We had to add a check for platform name to ensure we indeed run on the intended platform irrespective or enumeration order.

kpet:
@nikhiljnv Agree that's where we want to end up but currently we have a mix of environment variables and command line parameters for device/platform configuration (with both supported in some cases!) and the command line syntax is non-standard/cryptic/atrocious so we will have to break the existing command line interface. What I'm proposing we do is:
1. Remove the current command line syntax and use environment variables for all things device/platform configuration (already the case for device types). That's this PR. This lowers a bar to re-design command line parsing.
2. Re-introduce a clean command line interface when we get to re-designing command line parsing.
3. Deprecate and remove all the environment variables.

It does mean two breaking changes but we can stage them.

ahesham-arm:

PR #1076, opened 2020-12-02
Implement Negative Test for Platform Layer Functions

chemis01:
This change introduces a negative test for clGetPlatformInfo
as well as changes to the Harness to help with
other negative tests.

### DIFF
--- a/test_conformance/api/negative_platform.cpp
+++ b/test_conformance/api/negative_platform.cpp
@@ -1,0 +1,78 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "testBase.h"
+#include "harness/errorHelpers.h"
+#include "harness/testHarness.h"
+
+int test_negative_get_platform_ids(cl_device_id deviceID, cl_context context,
+                                   cl_command_queue queue, int num_elements)
+{
+
+    cl_platform_id platform;
+    bool failed_tests = negative_test(
+        clGetPlatformIDs, CL_INVALID_VALUE,
+        "clGetPlatformIDs should return CL_INVALID_VALUE when: \"num_entries "
+        "is equal to zero and platforms is not NULL\"",
+        0, &platform, nullptr);
+
+    failed_tests |=
+        negative_test(clGetPlatformIDs, CL_INVALID_VALUE,
+                      "clGetPlatformIDs should return CL_INVALID_VALUE "
+                      "when: \"both num_platforms and"
+                      "platforms are NULL\"",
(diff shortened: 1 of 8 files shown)


test_conformance/api/negative_platform.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+int test_negative_get_platform_ids(cl_device_id deviceID, cl_context context,
+                                   cl_command_queue queue, int num_elements)
+{
+
+    // clGetPlatformIDs returns CL_INVALID_VALUE when: "num_entries is equal to
+    // zero and platforms is not NULL"
+    cl_platform_id platform;
+    bool failed_tests = negative_test(clGetPlatformIDs, CL_INVALID_VALUE, 0,
+                                      &platform, nullptr);
```
jrprice, line 29: I'm not sure about this `negative_test` macro. It slightly obfuscates the actual API call being tested (not by much, but it doesn't parse as nicely as an actual function call IMO). Also, I don't think it would work for cases where the error code is an output parameter instead of the return value (e.g. `clCreate*`), and in those cases there's likely two different things you want to check (errcode == expected && return value == nullptr).

Doing the API call and checking the errcode using `test_failure_error` doesn't seem that bad to me, the printed error message would be more useful that just the line number, and the error message replaces the comment describing what is being tested.

```
err = clGetPlatformIDs(0, &platform, nullptr);
test_failure_error(err, CL_INVALID_VALUE,
                   "clGetPlatformIDs should return CL_INVALID_VALUE when: <...>");
```

Are there other tests coming later where this would be more cumbersome?

jrprice (commented):
The tests themselves look fine to me. Just a comment/question on the generic routines being introduced for this.

test_conformance/api/negative_platform.cpp
chemis01: I agree it's not as nice as the actual function call, but not sure if there's an easier 
Regarding the clCreate* functions, this has been accounted for by using a similar function which performs the behaviour you're describing.
I'll update the error messages accordingly.

gwawiork (dismissed):
Very nice solution to implement 'negative tests'. LGTM

test_conformance/api/negative_platform.cpp
jrprice: I'm still not convinced that this `negative_test` macro is an improvement over just calling the APIs directly followed by direct calls to `test_failure_error`, as in my example above. Aside from the reasons I've already mentioned above, the indirection caused by the macro also obfuscates compiler errors (e.g. if you get API arguments wrong) and inhibits IDE autocomplete features.

I guess this is subjective so I won't hold up this PR if everyone else likes the macro, but it'd be nice if you could show a case where this is clearly an improvement over the (IMHO) simpler form shown in my above comment.

bashbaug: I have a slight preference for the direct call method too, though I'm not going to dig in my heels either if folks prefer the macro.

FWIW, I used the direct call method in the API consistency test and it seemed to work out OK.  Example:

```c
        clProgramWrapper ilProgram =
            clCreateProgramWithIL(context, &bogus, sizeof(bogus), &error);
        test_failure_error(
            error, CL_INVALID_OPERATION,
            "Device does not support IL Programs but clCreateProgramWithIL did "
            "not return CL_INVALID_OPERATION");
```

jrprice (approved):
LGTM - thanks for the updates!

test_conformance/api/negative_platform.cpp
jrprice: Thanks for making the change!

gwawiork (approved):
LGTM as well

PR #1079, opened 2020-12-04
Add more relaxed math conformance tests

alan-baker:
* Add relaxed tests for functions (in an implementation that uses
  derived functions) that define an absolute ULP precision requirement,
  but were not tested in the current conformance suite
  * acos
  * asin
  * atan
  * cospi
  * sinpi
  * log10

I've tested these changes (via clvk) on an AMD, Arm, Nvidia and Qualcomm gpu.


EwanC (approved):
LGTM

kpet (approved):
It would be nice to add a symbolic name for 2^-11 and 2^-21 (and maybe define them as hex floating point literals). Not saying it needs to be done here.

It seems the special cases in the verification code handle domain boundaries correctly but this code is very hard to read and could use some love.

PR #1081, opened 2020-12-08
Use highest OpenCL C version

### LINKED ISSUE #1080: Set proper OpenCL C version in -cl-std
Need to modify set of tests to set highest supported OpenCL C version (not fixed version).
The Tests that use `create_single_kernel_helper_with_build_options` and change to `create_single_kernel_helper`. It allows automatically sets highest OpenCL C version to build option -cl-std=

gwawiork:
Fixes #1080 
To cover case when we support OpenCL C 3.0 but not OpenCL C 2.0

### DIFF
--- a/test_conformance/SVM/main.cpp
+++ b/test_conformance/SVM/main.cpp
@@ -215,12 +215,13 @@ cl_int create_cl_objects(cl_device_id device_from_harness, const char** ppCodeSt
     bool extensions_supported = true;
     for (auto extension : extensions_list)
     {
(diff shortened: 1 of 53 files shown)


bashbaug (commented):
LGTM but needs a non-Intel reviewer.

Jeremy-Kemp (approved):
LGTM, should be no functional change (for < 3.0).

EwanC (approved):
Makes sense to me

PR #1087, opened 2020-12-17
Handle NULL hostptr in conformance image tests

nikhiljnv:
As per the spec, clCreateBuffer and clCreateImage return CL_INVALID_HOST_PTR if host_ptr is not NULL,
but CL_MEM_USE_HOST_PTR or CL_MEM_COPY_HOST_PTR are not set in flags.
CTS currently passes non-NULL host_ptr while testing NO_HOST_PTR tests.
Fix CTS to pass null host-ptr in this case to avoid calls to clCreateBuffer/clCreateImage failing.


kpet (commented):
The patch looks correct after a quick scan but that code is very hard to read. The fact we're only finding this now also means that a lot of implementations don't reject the invalid combination.

bashbaug:
> The fact we're only finding this now also means that a lot of implementations don't reject the invalid combination.

@nikhiljnv could you please include a few test command lines that are affected by this change?

From the PR description it sounds like the tests that are affected use the `NO_HOST_PTR` option which is not required by any of the CSV files.  This could explain why nobody has complained about this combination.

FWIW, this command line is failing on our implementation:  `test_image_streams 2D read float CL_RGBA CL_UNORM_INT8 CL_FILTER_NEAREST CL_ADDRESS_CLAMP_TO_EDGE UNNORMALIZED NO_HOST_PTR`

```
read_image (integer coords, float results) *****************************
[CL_RGBA CL_UNORM_INT8            4] - CL_FILTER_NEAREST - CL_ADDRESS_CLAMP_TO_EDGE - UNNORMALIZED
clCreateImage failed (-37)
ERROR: Unable to create 2D image of size 167 x 177 pitch 668 (CL_INVALID_HOST_PTR)
FAILED: [CL_RGBA CL_UNORM_INT8            4] - CL_FILTER_NEAREST - CL_ADDRESS_CLAMP_TO_EDGE - UNNORMALIZED

read_image (unnormalized float coords, float results) *****************************
[CL_RGBA CL_UNORM_INT8            4] - CL_FILTER_NEAREST - CL_ADDRESS_CLAMP_TO_EDGE - UNNORMALIZED
clCreateImage failed (-37)
ERROR: Unable to create 2D image of size 167 x 177 pitch 668 (CL_INVALID_HOST_PTR)
FAILED: [CL_RGBA CL_UNORM_INT8            4] - CL_FILTER_NEAREST - CL_ADDRESS_CLAMP_TO_EDGE - UNNORMALIZED
```

I'll rebuild with the changes in this PR and see if it passes.

nikhiljnv:
The command lines that failed included 
images/kernel_read_write/test_image_streams CL_FILTER_NEAREST CL_RGBA 2D NO_HOST_PTR
images/kernel_read_write/test_image_streams use_pitches CL_FILTER_NEAREST CL_RGBA 2D NO_HOST_PTR

But, I would expect other variants of NO_HOST_PTR should fail in the same way.

@b-sumner ,
Can you take a look as well?

bashbaug:
Thanks!  I can confirm that these two command lines pass with this PR (the first is a subset of the second):

* `test_image_streams 2D read float CL_RGBA CL_UNORM_INT8 CL_FILTER_NEAREST CL_ADDRESS_CLAMP_TO_EDGE UNNORMALIZED NO_HOST_PTR`
* `test_image_streams CL_FILTER_NEAREST CL_RGBA 2D NO_HOST_PTR`

This is still failing for me though:

* `test_image_streams use_pitches CL_FILTER_NEAREST CL_RGBA 2D NO_HOST_PTR`

I think it's failing for legitimate reasons, specifically due to this call:

```
>>>> clCreateImage: context = 0x563ea49f7980, flags = CL_MEM_READ_ONLY (4), format->channel_order = CL_RGBA, format->channel_data_type = CL_UNORM_INT8, desc->type = CL_MEM_OBJECT_IMAGE2D, desc->width = 167, desc->height = 177, desc->depth = 0, desc->array_size = 0, desc->row_pitch = 864, desc->slice_pitch = 0, desc->num_mip_levels = 0, desc->num_samples = 0, desc->mem_object = (nil), host_ptr = (nil) 
<<<< clCreateImage: returned (nil) -> CL_INVALID_IMAGE_DESCRIPTOR
```

since the spec [says](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#image-descriptor):

> image_row_pitch is the scan-line pitch in bytes. **This must be 0 if host_ptr is NULL** and can be either 0 or ≥ image_width × size of element in bytes if host_ptr is not NULL.

I don't think this should hold up this change, but perhaps it means that `use_pitches` and `NO_HOST_PTR` cannot be tested together, at least in some cases.

test_conformance/images/kernel_read_write/test_write_1D.cpp
```diff
@@ -215,6 +215,13 @@ int test_write_image_1D( cl_device_id device, cl_context context, cl_command_que
         }
         else // Either CL_MEM_ALLOC_HOST_PTR, CL_MEM_COPY_HOST_PTR or none
         {
+            char *host_ptr = NULL;
+
+            if (gMemFlagsToUse & CL_MEM_COPY_HOST_PTR)
+            {
+                host_ptr = imageValues;
+            }
```
bashbaug, line 223: Would it be more robust to leave all of the calls into the image helpers unchanged and to instead move logic similar to this into the image helpers themselves?

nikhiljnv: Agree, but this may apply to other cl_mem_flags as well. Can we do this in a separate PR?

nikhiljnv:
@bashbaug 
Anything left here to address before we can merge this?

@b-sumner 
Any concerns from your perspective for this change?

test_conformance/images/kernel_read_write/test_write_1D.cpp
bashbaug: Would you mind taking a quick look at this vs. pushing off to a separate PR?  I think it is a much more robust solution and it'd make many parts of this PR unnecessary.

bashbaug: Specifically, to get equivalent behavior as this PR I think we'd only need to check for `CL_MEM_USE_HOST_PTR` and `CL_MEM_COPY_HOST_PTR`, to ensure we only pass a non-NULL host pointer when we are allowed to:

> CL_INVALID_HOST_PTR if host_ptr is NULL and CL_MEM_USE_HOST_PTR or CL_MEM_COPY_HOST_PTR are set in flags or if host_ptr is not NULL but CL_MEM_COPY_HOST_PTR or CL_MEM_USE_HOST_PTR are not set in flags.

We wouldn't need to check for other `cl_mem_flags`.

nikhiljnv: Updated the change with checks moved to ImageHelper and earlier changes reverted from test_conformance/image/.
Only a small note - If we decide to add new tests to check for NO_HOST_PTR behavior to ensure if implementations return expected error or not, we may have to revisit this or use a different path to avoid overriding NULL host-ptr.

bashbaug (approved):
Thanks!  Latest changes LGTM.

nikhiljnv:
@kpet , @bashbaug 
Can this be merged?

kpet (approved):

PR #1093, opened 2020-12-29
Fix implicit int->float warning inside kernel

jrprice:
This kernel is used to test various compiler options including `-Werror`. Some compilers produce a warning about the implicit conversion which results in this test failing when `-Werror` is used.

This first commit contains the trivial change. The second is just formatting.


kpet (approved):
That looks reasonable to me.

EwanC (approved):

PR #1082, opened 2020-12-11
Reduce number of compilations in buffer suite

ellnor01:
Extracts program and kernel compilation from mem_flags loop
as they were being recompiled unnecessarily.

Fixes #1020

Signed-off-by: Ellen Norris-Thompson <ellen.norris-thompson@arm.com>

PR #1099, opened 2021-01-08
Remove invalid clReleaseEvent calls

jrprice:
These events are now using clEventWrapper to manage their lifetime. That change was made in #1082.

### DIFF
--- a/test_conformance/buffers/test_buffer_fill.cpp
+++ b/test_conformance/buffers/test_buffer_fill.cpp
@@ -566,7 +566,6 @@ int test_buffer_fill( cl_device_id deviceID, cl_context context, cl_command_queu
     void        *outptr[5];
     clProgramWrapper program[5];
     clKernelWrapper kernel[5];
-    clEventWrapper event[2];
     size_t      ptrSizes[5];
     size_t      global_work_size[3];
     int         err;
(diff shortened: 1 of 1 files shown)


kpet (commented):
I was preparing a fix for that issue as well. I believe this change will cause leaks as the wrappers won't release the previously tracked object when a new one is assigned via the address of operator (which isn't really fixable) or via direct assignment (which is fixable). The temporary fix I was about to propose was to set these to NULL after the release so the wrappers don't attempt to release objects that have already been after the last iteration of the loop.

jrprice:
Good point, I missed the fact that this is happening in a loop.

> The temporary fix I was about to propose was to set these to NULL after the release so the wrappers don't attempt to release objects that have already been after the last iteration of the loop.

Can't we just move the event declarations to the loop body? That way their lifetime reflects their actual usage.

Alternatively we could probably just get rid of the events from this test altogether and use `clFinish().

kpet:
Either of these are options worth exploring. I wasn't satisfied with my fix which is why I hadn't uploaded it yet. I'll leave that one to you then. Happy to review.

jrprice:
I did the scope change since it was simplest. Thanks for the quick responses, sorry for stepping on your toes :-)

kpet (approved):
LGTM

PR #1100, opened 2021-01-10
 Fix malloc-size calculation in test imagedim

nikhiljnv:
Use size_t instead of int for imagedim
The size calculation for image with larger dimensions
is overflowing with int values.
Change image dim variables to use size_t
to avoid overflow.
While at it, fix formatting at various places.
Also, use new and delete in place of malloc and free.


test_conformance/basic/test_imagedim.cpp
```diff
@@ -41,8 +41,13 @@ static const char *image_dim_kernel_code =
 static unsigned char *
 generate_8888_image(int w, int h, MTdata d)
 {
-    unsigned char   *ptr = (unsigned char*)malloc(w * h * 4);
-    int             i;
+    unsigned char *ptr =
+        (unsigned char *)malloc(sizeof(unsigned char) * 4 * w * h);
+    unsigned int i;
+    if (ptr == NULL)
```
kpet, line 47: I think we should NOT try to handle allocation failures by testing for NULL pointers, this will clutter the code for no real benefit. I suggest you change this to use `new[]` or a smart pointer.

kpet (changes requested):
Fine with the core change but the allocation error checking should change IMHO.

test_conformance/basic/test_imagedim.cpp
```diff
@@ -41,8 +41,13 @@ static const char *image_dim_kernel_code =
 static unsigned char *
 generate_8888_image(int w, int h, MTdata d)
 {
-    unsigned char   *ptr = (unsigned char*)malloc(w * h * 4);
-    int             i;
+    unsigned char *ptr =
+        (unsigned char *)malloc(sizeof(unsigned char) * 4 * w * h);
```
bashbaug, line 45: I don't think that adding the multiplication by `sizeof(unsigned char)` will hurt anything, but is it required?  I believe `sizeof(unsigned char)` is required to be 1:

https://en.cppreference.com/w/c/language/sizeof

> sizeof(char), sizeof(signed char), and sizeof(unsigned char) always return 1.

nikhiljnv: Hi Ben,
We seem to be exceeding limit 32-bit limit for many of our higher end GPUs where typical size of image tested is 32k x 32k. This change fixes the issue. This way lying around in our internal CTS version. So, upstreamed as is.
The change description is bit misleading.
I believe what fixes the issue is not the value of sizeof(unsigned char), but the fact that this change makes size computation 64-bit instead of 32-bit which caused the overflow.
A cleaner fix would be to use long int or size_t for inputs to this test, but that may be somewhat bigger change. Will check and update PR with a cleaner fix.

bashbaug: I see, makes sense - the important part is the multiplication by `size_t` which changes the calculation from a 32-bit multiplication to a 64-bit multiplication.

https://godbolt.org/z/PPK7sY

Thanks!

test_conformance/basic/test_imagedim.cpp
```diff
@@ -38,24 +38,25 @@ static const char *image_dim_kernel_code =
 "}\n";
 
 
-static unsigned char *
-generate_8888_image(int w, int h, MTdata d)
+static unsigned char *generate_8888_image(size_t w, size_t h, MTdata d)
 {
-    unsigned char   *ptr = (unsigned char*)malloc(w * h * 4);
-    int             i;
+    unsigned char *ptr = new unsigned char[sizeof(unsigned char) * 4 * w * h];
```
bashbaug, line 43: Minor: don't need `sizeof(unsigned char)` here anymore since the allocation is done with `new[]`.

bashbaug (commented):
One small thing to fix but otherwise LGTM - thanks!

test_conformance/basic/test_imagedim.cpp
nikhiljnv: Addressed.

test_conformance/basic/test_imagedim.cpp
nikhiljnv: Addressed.

bashbaug: Thanks!

bashbaug (approved):

kpet (approved):
LGTM, I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/1119 to simplify object and memory allocation management.

PR #1104, opened 2021-01-13
Apply clang-format on math_brute_force

mantognini:
Relates to #625. This change was generated automatically with clang-format. No manual editing was done.

This should allow future updates of the math_brute_force component to focus on content, not formatting, and hopefully reduce code duplication and so on.


mantognini:
NB: I've pushed a minor update. It turned out I was using clang-format-10 which was causing a slightly different (and puzzling) output compared to clang-format-9.

kpet (approved):
Merging straight away so this doesn't go stale.

PR #1105, opened 2021-01-13
Use std::vector for format lists in images suite

### LINKED ISSUE #723: Mismatched new[]/delete and new/free

jrprice:
Avoids memory deallocation issues and generally simplifies the code.

There's a fair bit of churn due to adding lots of `const` qualifiers to function parameters, but it's largely mechanical. The first commit contains the changes, the second is just formatting; I'd recommend reviewing just the first.

Fixes #723

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -3571,8 +3571,8 @@ cl_float CoordWalker::Get(size_t idx, size_t el)
 }
 
 
-void print_read_header(const cl_image_format *format, image_sampler_data *sampler,
-                       bool err, int t)
+void print_read_header(const cl_image_format *format,
+                       image_sampler_data *sampler, bool err, int t)
 {
     const char *addressMode = NULL;
     const char *normalizedNames[2] = { "UNNORMALIZED", "NORMALIZED" };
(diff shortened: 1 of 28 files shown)


kpet (approved):
Nice change :). Many const pointers to `cl_image_format` could be const references but that's best left as a later change I think.

PR #1106, opened 2021-01-14
Test api fix - sampler properties is 0 in compatibility mode

### LINKED ISSUE #1103: test_api get_sampler_info_compatibility failures
This test expects now that in case sampler is created by function `clCreateSampler` then `CL_SAMPLER_PROPERTIES` size is bigger than 0.
see: https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clGetSamplerInfo
```If sampler was created using clCreateSampler, or if the properties argument specified in clCreateSamplerWithProperties was NULL, the implementation must return param_value_size_ret equal to 0, indicating that there are no properties to be returned.```

gwawiork:
Fixes #1103

### DIFF
--- a/test_conformance/api/test_queries.cpp
+++ b/test_conformance/api/test_queries.cpp
@@ -231,23 +231,39 @@ int test_sampler_params(cl_device_id deviceID, cl_context context,
             error,
             "clGetSamplerInfo failed asking for CL_SAMPLER_PROPERTIES size.");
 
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
LGTM but needs a non-Intel reviewer.

kpet (approved):
Ran into this as well. LGTM!

PR #1107, opened 2021-01-15
Use delete[] to deallocate after new[]

jrprice:
Fixes ASan issues flagged in the mem_host_flags suite.

First commit is the trivial change, second is just formatting.


kpet (approved):
LGTM

EwanC (approved):

PR #1108, opened 2021-01-15
Run spirv-val for SPIR-V offline compilation

StuartDBrady:
The common --disable-spirv-validation option has been added to disable
this functionality.

### DIFF
--- a/test_common/harness/kernelHelpers.cpp
+++ b/test_common/harness/kernelHelpers.cpp
@@ -604,7 +604,7 @@ static int get_offline_compiler_output(
                 sourceFilename, outputFilename, openclCXX);
             if (error != CL_SUCCESS) return error;
 
-            // read output file
+            // open output file for reading
             ifs.open(outputFilename.c_str(), std::ios::binary);
             if (!ifs.good())
             {
@@ -614,6 +614,26 @@ static int get_offline_compiler_output(
             }
         }
     }
+
+    if (compilationMode == kSpir_v && !gDisableSPIRVValidation)
+    {
+        std::string runString = gSPIRVValidator + " " + outputFilename;
+
+        int returnCode = system(runString.c_str());
+        if (returnCode == -1)
+        {
+            log_error("Error: failed to invoke SPIR-V validator\n");
+            return CL_COMPILE_PROGRAM_FAILURE;
+        }
+        else if (returnCode != 0)
+        {
+            log_error(
+                "Failed to validate SPIR-V file %s: system() returned 0x%x\n",
+                outputFilename.c_str(), returnCode);
+            return CL_COMPILE_PROGRAM_FAILURE;
+        }
+    }
+
     return CL_SUCCESS;
 }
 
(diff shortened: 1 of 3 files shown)


test_common/harness/kernelHelpers.cpp
```diff
@@ -614,6 +614,26 @@ static int get_offline_compiler_output(
             }
         }
     }
+
+    if (compilationMode == kSpir_v && !gDisableSPIRVValidation)
+    {
+        std::string runString = gSPIRVValidator + " " + outputFilename;
```
jrprice, line 620: Should we be specifying the SPIR-V version/environment to validate against here (i.e. `--target-env=...`)? By default SPIRV-Tools will validate against SPIR-V 1.5, which may let some things through that are not valid for the version of SPIR-V actually being tested.

bashbaug: I'd be fine filing an issue to improve this and saving this change for a subsequent PR.

I guess if we wanted to pick the target SPIR-V environment the "right way" to do that would be to query for the OpenCL version supported by the device?  Note, we still need to add the OpenCL 3.0 environment to the validator...

test_common/harness/kernelHelpers.cpp
```diff
@@ -614,6 +614,26 @@ static int get_offline_compiler_output(
...
+
+        int returnCode = system(runString.c_str());
+        if (returnCode == -1)
+        {
+            log_error("Error: failed to invoke SPIR-V validator\n");
+            return CL_COMPILE_PROGRAM_FAILURE;
```
bashbaug, line 626: This will fail the CTS test if we can't find the validator or if it returns an error code.  Is this the intended behavior?  Debatably this should more of a "warning" than an "error", since it's more of a test of the SPIR-V compiler than the OpenCL implementation under test.

StuartDBrady: The behavior is as intended. @kpet, would you care to comment?

test_common/harness/kernelHelpers.cpp
StuartDBrady: I'd be happy to raise an issue, although I don't have a very clear picture of what the "right way" of handling this is, here, so I'd be equally happy to defer to someone with a clearer picture of this.

test_common/harness/kernelHelpers.cpp
kpet: I personally agree with the current behaviour. There's no point feeding the SPIR-V to the implementation under test if the external toolchain produced an invalid module (except for negative tests but these wouldn't use the offline compilation framework anyway).

bashbaug: This behavior is fine so long as the SPIR-V LLVM translator generates valid SPIR-V and the SPIR-V validator for OpenCL is up-to-date and doesn't have false-positives, but frankly I don't have that level of confidence in either component yet.

I'm not going to lose a ton of sleep over this because there is an option to disable the validator, but it'd be a shame to add these changes to run the validator automatically only to require disabling the validator to run any meaningful tests to completion.

bashbaug (approved):

test_common/harness/kernelHelpers.cpp
kpet: I agree with your assessment of the current quality level of the tools. Part of the reason for enabling this by default is to drive improvements to the SPIR-V LLVM Translator and SPIR-V validator.

PR #1112, opened 2021-01-16
api: Fix testing of local memory size requirement

pierremoreau:
The current test would always end up assuming a minimum of 16 KB
regardless of the exposed OpenCL version, and the logic for testing the
OpenCL version was hard to follow.

Besides fixing the test for OpenCL 1.1 through 2.1, it also
* adds support for OpenCL 2.2, 3.0, and future OpenCL versions (as long
  as `get_device_cl_version()` supports those);
* adapts the error message to mention the currently exposed OpenCL
  version rather than a hardcoded OpenCL 1.1;
* reports the advertised local memory size as KB and not Kb, since local
  memory size is given in bytes.

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -1670,8 +1670,7 @@ int test_min_max_local_mem_size(cl_device_id deviceID, cl_context context, cl_co
     size_t    threads[1], localThreads[1];
     cl_int *localData, *resultData;
     cl_ulong maxSize, kernelLocalUsage, min_max_local_mem_size;
-    cl_char buffer[ 4098 ];
-    size_t length;
+    Version device_version;
     int i;
     int err = 0;
     MTdata d;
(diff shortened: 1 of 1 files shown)


Jeremy-Kemp (approved):
LGTM.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1680,31 +1679,33 @@ int test_min_max_local_mem_size(cl_device_id deviceID, cl_context context, cl_co
     error = clGetDeviceInfo( deviceID, CL_DEVICE_LOCAL_MEM_SIZE, sizeof( maxSize ), &maxSize, 0 );
     test_error( error, "Unable to get max local buffer size" );
 
-    // Device version should fit the regex "OpenCL [0-9]+\.[0-9]+ *.*"
-    error = clGetDeviceInfo( deviceID, CL_DEVICE_VERSION, sizeof( buffer ), buffer, &length );
-    test_error( error, "Unable to get device version string" );
+    try
+    {
+        device_version = get_device_cl_version(deviceID);
+    } catch (const std::runtime_error &e)
```
kpet, line 1685: Fine for this change but longer-term we probably want to introduce a type for these exceptions and catch them in the framework code. Catching individual exceptions like this defeats the point...

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1680,31 +1679,33 @@ int test_min_max_local_mem_size(cl_device_id deviceID, cl_context context, cl_co
...
+    {
+        min_max_local_mem_size = 1L * 1024L;
     }
 
-    if( maxSize < (gIsEmbedded ? 1L * 1024L : min_max_local_mem_size) )
+    if (maxSize < min_max_local_mem_size)
     {
-        log_error( "ERROR: Reported local mem size less than required by OpenCL 1.1 (reported %dKb)\n", (int)( maxSize / 1024L ) );
+        const std::string version_as_string = device_version.to_string();
+        log_error("ERROR: Reported local mem size less than required by OpenCL "
```
kpet, line 1706: It would be helpful to print the required amount as well.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1680,31 +1679,33 @@ int test_min_max_local_mem_size(cl_device_id deviceID, cl_context context, cl_co
...
+    else
+    {
+        min_max_local_mem_size = 1L * 1024L;
     }
 
-    if( maxSize < (gIsEmbedded ? 1L * 1024L : min_max_local_mem_size) )
+    if (maxSize < min_max_local_mem_size)
     {
-        log_error( "ERROR: Reported local mem size less than required by OpenCL 1.1 (reported %dKb)\n", (int)( maxSize / 1024L ) );
+        const std::string version_as_string = device_version.to_string();
```
kpet, line 1705: Could use `auto`

kpet (approved):
LGTM (with some improvement suggestions)

pierremoreau:
Thank you for the comments; I will open an additional PR to address the last two.

PR #1113, opened 2021-01-19
Fix compilation issues for conforming_version.cpp

ellnor01:
Previously this file used the auto keyword to
declare a string. This can cause compilation
issues when used in a later function.


jrprice (dismissed):

ellnor01:
I believe the 'Presubmit / Build macos-11.0' which appears to have failed just didn't run. There is also a successful run for this build.

jrprice:
Yes. Merging as we already have a successful macOS build and this looks like an infra issue.

jrprice (approved):

PR #1114, opened 2021-01-20
Remove compiler options that are specific to Apple

jrprice:
These are causing test failures for non-Apple implementations of
OpenCL running on macOS.

### DIFF
--- a/test_conformance/compiler/test_build_options.cpp
+++ b/test_conformance/compiler/test_build_options.cpp
@@ -61,10 +61,6 @@ const char *optimization_options[] = {
     "-cl-fast-relaxed-math",
     "-w",
     "-Werror",
-#if defined( __APPLE__ )
-    "-cl-opt-enable",
-    "-cl-auto-vectorize-enable"
-#endif
     };
 
 cl_int get_result_from_program( cl_context context, cl_command_queue queue, cl_program program, cl_int *outValue )


EwanC (approved):

kpet:
This is a rather widespread assumption in the CTS (that building on Apple means that the CTS is running against Apple's implementation). Should we find a solution to that problem here? Are we likely to remove most code specific to Apple's implementation anyway given the deprecation? I think this needs more discussion.

jrprice:
> Should we find a solution to that problem here?

I think the solution is to get rid of all of the Apple-implementation specific code (see below). But if others think we should keep it, we could change all of the `#ifdef __APPLE__` to something like `#ifdef __APPLE_OPENCL__` and make that an off-by-default CMake option.

> Are we likely to remove most code specific to Apple's implementation anyway given the deprecation?

I personally think that this is the right thing to do. Aside from causing issues for other OpenCL implementations running on Apple systems, it's a maintenance burden for other CTS developers, and I think it's clear at this point that Apple have no interest in either using it or helping to maintain it. The code will continue to exist in the git history if it's ever needed. If this was the concern though, we could remove all of the Apple-implementation specific code in a single PR so that it's easier to reference in the future.

kpet:
I wouldn't disagree with that outcome. I actually think it would be desirable: there is a lot of code under `__APPLE__` that we're unlikely to want to keep. I just want us to make a clear policy decision that we could refer to in the future instead of eating away at these without a clear direction.

I don't think we should remove the code wholesale though. Some changes are required to run on macOS, regardless of the implementation, and there are other changes that we should consider taking out of `__APPLE__`.

jrprice:
> I don't think we should remove the code wholesale though. Some changes are required to run on macOS, regardless of the implementation, and there are other changes that we should consider taking out of `__APPLE__`.

Yes, definitely. I was talking specifically about code that relates to Apple's OpenCL implementation (e.g. these compiler flags, or the `queue_priority` test in `basic/`), not about all code guarded by `__APPLE__`.

>  I just want us to make a clear policy decision that we could refer to in the future

Sure. As another data point here, we [made a similar policy decision for OpenCL-Headers](https://github.com/KhronosGroup/OpenCL-Headers/pull/32) a while back. I'm happy to wait for a official decision via a WG call again for CTS though, if that helps.

kpet:
That's a useful data point even if the decision on OpenCL-Headers was a much easier one. Let's discuss it next week to get broader feedback.

kpet (approved):
Agreed on the 2021/04/06 teleconference: we can remove code under `__APPLE__` that is specific to Apple's deprecated OpenCL implementation. Support for macOS will be retained.

PR #1115, opened 2021-01-20
Do not require OpenCL C 2.0 on 3.0 devices

jrprice:
The tests in this file were wrongly requiring that all OpenCL 3.0 devices advertise support for OpenCL C 2.0 and all of its (now optional) features.

I'm not sure how this slipped through, since @FranklandJack commented in #968 that the tests were passing on an OpenCL 3.0 implementation that does not support OpenCL C 2.0. Unless I'm missing something here or in the spec.

### DIFF
--- a/test_conformance/compiler/test_opencl_c_versions.cpp
+++ b/test_conformance/compiler/test_opencl_c_versions.cpp
@@ -193,7 +193,7 @@ static int test_CL_DEVICE_OPENCL_C_VERSION_features(cl_device_id device,
     }
 
     const Version clc_version = get_device_cl_c_version(device);
-    if (clc_version >= Version(2, 0))
+    if (clc_version >= Version(2, 0) && clc_version < Version(3, 0))
     {
         bool has_all_OpenCL_C_20_features =
             features.supports__opencl_c_atomic_order_acq_rel
@@ -248,7 +248,7 @@ static int test_CL_DEVICE_OPENCL_C_VERSION_versions(cl_device_id device,
                                                      / sizeof(cl_name_version));
     error = clGetDeviceInfo(device, CL_DEVICE_OPENCL_C_ALL_VERSIONS, sz,
                             device_clc_versions.data(), NULL);
-    test_error(error, "Unable to query CL_DEVICE_OPENCL_C_FEATURES");
+    test_error(error, "Unable to query CL_DEVICE_OPENCL_ALL_VERSIONS");
 
     for (const auto& test_clc_version : test_clc_versions)
     {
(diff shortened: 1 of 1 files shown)


FranklandJack:
Hey @jrprice. I think there might be some confusion stemming from the values of [`clc_version`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_opencl_c_versions.cpp#L112) and [`device_clc_version`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_opencl_c_versions.cpp#L231) in these tests. These variables get their values from calls to [`get_device_cl_c_version`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/kernelHelpers.cpp#L1712) which actually returns the highest fully backwards compatible OpenCL C version supported by the compiler for the device (it does this using the [`Cl_DEVICE_OPENCL_C_VERSION`](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_VERSION) query for devices after OpenCL 1.0). For a 3.0 device, this means that `get_device_cl_c_version` will only return `3.0` if all optional features are supported, in which case, these tests should be run. I think if your 3.0 device doesn't support all these features, then the `CL_DEVICE_OPENCL_C_VERSION` query should return 1.2 and testing for these features in [`test_CL_DEVICE_OPENCL_C_VERSION_features`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_opencl_c_versions.cpp#L226) will be skipped. Similarly in [`test_CL_DEVICE_OPENCL_C_VERSION_versions`](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compiler/test_opencl_c_versions.cpp#L226), only versions less than most recent supported backwards compatible version of CL C will be checked, for a 3.0 device that doesn't support the optional features this would be `{1.0, 1.1, 1.2}`, for a 3.0 device that does, this would be `{1.0, 1.1, 1.2, 2.0, 3.0}`.

I can confirm that this test still passes with a 3.0 driver optionally not supporting these features, however, I don't know if this was the intended behavior of the test since it was written by @bashbaug.

It's been a minute since I looked at this code, so I could have got this completely wrong and misunderstood the spec and your question! Let me know if you think I've got it wrong :)

jrprice:
> These variables get their values from calls to get_device_cl_c_version which actually returns the highest fully backwards compatible OpenCL C version supported by the compiler for the device [...]. For a 3.0 device, this means that get_device_cl_c_version will only return 3.0 if all optional features are supported

This didn't quite sink for me after reading the description of `CL_DEVICE_OPENCL_C_VERSION` a couple of times, but you're absolutely right. I now remember why we deprecated this query :-)

Thanks for the detailed response to help clarify things for me!

PR #1118, opened 2021-01-20
Format code in test_common/miniz

mantognini:
Fix misleading indentation warning in miniz.c (-Wmisleading-indentation) as reported by GCC 7.5

This format the whole miniz directory, related to #625.

If preferred, I can also update the PR to only format the few lines that relates to this warning, as this could help if these files need to be updated with a newer version in the future. (I suppose, an alternative would be to remove these files and use something like [CMake ExternalProject API](https://cmake.org/cmake/help/latest/module/ExternalProject.html#module:ExternalProject), or any other automated solution.)
Let me know your preferences.

### DIFF
--- a/test_common/miniz/miniz.c
+++ b/test_common/miniz/miniz.c
@@ -1,166 +1,227 @@
-/* miniz.c v1.15 - public domain deflate/inflate, zlib-subset, ZIP reading/writing/appending, PNG writing
-   See "unlicense" statement at the end of this file.
-   Rich Geldreich <richgel99@gmail.com>, last updated Oct. 13, 2013
-   Implements RFC 1950: http://www.ietf.org/rfc/rfc1950.txt and RFC 1951: http://www.ietf.org/rfc/rfc1951.txt
+/* miniz.c v1.15 - public domain deflate/inflate, zlib-subset, ZIP
+   reading/writing/appending, PNG writing See "unlicense" statement at the end
+   of this file. Rich Geldreich <richgel99@gmail.com>, last updated Oct. 13,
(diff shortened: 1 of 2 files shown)


mantognini:
Let me know if you have any preferences. As stated above, I can also modify only a few lines to address the warnings (-Wmisleading-indentation) and leave the rest of the file untouched.

EwanC (approved):
Personally happy with formatting the whole files. `miniz.c` is pretty dense right now, especially
 where there are multiple statements on a single line.

StuartDBrady (changes requested):
This diverges from the upstream miniz code in a way that will make updating problematic, if we were to need to do so.  It would be much better to update only the lines that cause warnings, and to make larger changes to the upstream miniz if it's really that important, rather than making them here first.

CC @kpet, IMO the change as-is makes for a larger delta against upstream miniz which hurts maintainability. Should we have a policy against mass reformatting of third-party code?

EwanC:
> This diverges from the upstream miniz code in a way that will make updating problematic, if we were to need to do so. It would be much better to update only the lines that cause warnings, and to make larger changes to the upstream miniz if it's really that important, rather than making them here first.
> 
> CC @kpet, IMO the change as-is makes for a larger delta against upstream miniz which hurts maintainability. Should we have a policy against mass reformatting of third-party code?

Looking at the git history of this file and the dates in the top comment, the value of clang-formatting the code to make it more readable is probably limited. It's so stable and of auxiliary functionality that it's unlikely to get much developer time. So don't have any objections to reducing this change down to only the lines generating the warning. 

An overall policy of not wholesale formatting third-party code to reduce the diff with upstream sounds sensible to me. As an aside, it'd also be interested to collate a list of all the third party code we currently have in the repo, as i'm not sure how much there is.

PR #1124, opened 2021-01-26
Fix test_vector_swizzle overwrite of padding of 3-element vectors

### LINKED ISSUE #1120: test_vector_swizzle checking padding of 3-element vectors
test_vector_swizzle is checking that the value of the padding is unchanged in all 3-element destination vectors,
whether they are written using vstore or directly.
We (IMG) are wondering whether this is correct for the direct stores, or whether they are allowed to overwrite the padding?

stevec611:
Use vstore3 to ensure padding is not overwritten in destination buffer.

Fixes #1120


bashbaug (approved):
LGTM and I've verified this passed on our implementation.  Thanks!

jrprice (approved):

PR #1127, opened 2021-01-28
 Fix command-line function range for bruteforce 

nikhiljnv:
Runnning "test_bruteforce N M" is expected to skip
first N functions and test M functions after it.
When N is 0, the test currently skips M functions
and run all functions thereafter.
Fix the test to honor semantics of these
command-line options to correctly test
first M functions when N is 0.


jlewis-austin (approved):
LGTM

kpet (approved):
LGTM. We really ought to integrate this better in option parsing.

mantognini:
This change broke some flows: when listing tests to be run, they are now always skipped.

Example:
```
$ ./bins/test_bruteforce mad
[snip]
mad...                                                              
Skipping function #53                                               
mad passed                                                          
PASSED sub-test.                                                    
PASSED test.
```


On line 143, we have
```cpp
if (i < gStartTestNumber || i > gEndTestNumber)
```
which is always true: `i` and `gStartTestNumber` are `size_t`, and `i < size_t{-1}` always holds for the values of `i`.

Maybe it's enough to change this line to
```cpp
if ((gStartTestNumber != -1 && i < gStartTestNumber) || i > gEndTestNumber)
```
but I haven't tested this thouroughtly.

PR #1131, opened 2021-01-28
Fix possible size_t overflow in 32-bit builds.

JohnKesapidesARM:
Use cl_ulong temporary values for row/slice_pitch.

### DIFF
--- a/test_conformance/images/kernel_image_methods/test_2D_array.cpp
+++ b/test_conformance/images/kernel_image_methods/test_2D_array.cpp
@@ -269,8 +269,7 @@ int test_get_image_info_2D_array(cl_device_id device, cl_context context,
                 size_t extraHeight = (int)random_log_in_range( 0, 8, seed );
                 slicePitch = rowPitch * (imageInfo.height + extraHeight);
 
-                size = (cl_ulong)slicePitch * (cl_ulong)imageInfo.arraySize * 4
-                    * 4;
+                size = slicePitch * imageInfo.arraySize * 4 * 4;
             } while(  size > maxAllocSize || ( size * 3 ) > memSize );
 
             imageInfo.slicePitch = slicePitch;
--- a/test_conformance/images/kernel_image_methods/test_2D_array.cpp
+++ b/test_conformance/images/kernel_image_methods/test_2D_array.cpp
@@ -244,6 +244,9 @@ int test_get_image_info_2D_array(cl_device_id device, cl_context context,
         for( int i = 0; i < NUM_IMAGE_ITERATIONS; i++ )
         {
             cl_ulong size;
+            cl_ulong slicePitch;
+            cl_ulong rowPitch;
+
             // Loop until we get a size that a) will fit in the max alloc size and b) that an allocation of that
             // image, the result array, plus offset arrays, will fit in the global ram space
             do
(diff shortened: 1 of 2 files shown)


test_conformance/images/kernel_image_methods/test_3D.cpp
```diff
@@ -113,23 +116,26 @@ int test_get_image_info_3D(cl_device_id device, cl_context context,
...
 
                 size_t extraHeight = (int)random_log_in_range( 0, 8, seed );
-                imageInfo.slicePitch = imageInfo.rowPitch * (imageInfo.height + extraHeight);
+                slicePitch = rowPitch * (imageInfo.height + extraHeight);
 
-                size = (cl_ulong)imageInfo.slicePitch * (cl_ulong)imageInfo.depth * 4 * 4;
+                size = (cl_ulong)slicePitch * (cl_ulong)imageInfo.depth * 4 * 4;
             } while(  size > maxAllocSize || ( size * 3 ) > memSize );
 
+            imageInfo.slicePitch = slicePitch;
```
jlewis-austin, line 136: I'm not sure I understand the fix. Aren't we just moving the overflow possibility to here? It seems like a fundamental limitation of the API that rowPitch and slicePitch have to fit in size_t.

test_conformance/images/kernel_image_methods/test_2D_array.cpp
```diff
@@ -252,23 +255,27 @@ int test_get_image_info_2D_array(cl_device_id device, cl_context context,
...
-                } while ((imageInfo.rowPitch % pixelSize) != 0);
+                    rowPitch += extraWidth;
+                } while ((rowPitch % pixelSize) != 0);
 
                 size_t extraHeight = (int)random_log_in_range( 0, 8, seed );
-                imageInfo.slicePitch = imageInfo.rowPitch * (imageInfo.height + extraHeight);
+                slicePitch = rowPitch * (imageInfo.height + extraHeight);
 
-                size = (cl_ulong)imageInfo.slicePitch * (cl_ulong)imageInfo.arraySize * 4 * 4;
+                size = (cl_ulong)slicePitch * (cl_ulong)imageInfo.arraySize * 4
```
jlewis-austin, line 272: the (cl_ulong) cast is unnecessary here

test_conformance/images/kernel_image_methods/test_3D.cpp
```diff
@@ -113,23 +116,26 @@ int test_get_image_info_3D(cl_device_id device, cl_context context,
...
-                } while ((imageInfo.rowPitch % pixelSize) != 0);
+                    rowPitch += extraWidth;
+                } while ((rowPitch % pixelSize) != 0);
 
                 size_t extraHeight = (int)random_log_in_range( 0, 8, seed );
-                imageInfo.slicePitch = imageInfo.rowPitch * (imageInfo.height + extraHeight);
+                slicePitch = rowPitch * (imageInfo.height + extraHeight);
 
-                size = (cl_ulong)imageInfo.slicePitch * (cl_ulong)imageInfo.depth * 4 * 4;
+                size = (cl_ulong)slicePitch * (cl_ulong)imageInfo.depth * 4 * 4;
```
jlewis-austin, line 133: the (cl_ulong) cast is unnecessary here

jlewis-austin (commented):
This definitely cleans it up a bit, but think I'm missing how the fix works.

test_conformance/images/kernel_image_methods/test_3D.cpp
JohnKesapidesARM: The trick here is that for a 32-bit system, if the pitch is more than 32-bit then this `while(  size > maxAllocSize || ( size * 3 ) > memSize );` will cause it to recalculate as maxAllocSize and memSize  are guaranteed to be <=4GB. So if we get to this point outside the loop the pitches are guaranteed to fit.

test_conformance/images/kernel_image_methods/test_2D_array.cpp
JohnKesapidesARM: yeap redundant, removed.

test_conformance/images/kernel_image_methods/test_3D.cpp
JohnKesapidesARM: yeap redundant, removed.

test_conformance/images/kernel_image_methods/test_3D.cpp
jlewis-austin: Thanks, got it. I was looking at the wrong loop scope.

jlewis-austin (approved):
LGTM

kpet (approved):

PR #1141, opened 2021-02-02
Creating common functions for image/kernel_read_write read tests

### LINKED ISSUE #616: De-duplicate code in kernel_read_write
There is a lot of duplicate code in that suite. Move it to the common file introduced in https://github.com/KhronosGroup/OpenCL-CTS/pull/615.

ellnor01:
The second PR in a series to fix #616 

Previously functions used for the 3D image read tests were moved into the common file. In this PR the code is modified to be generic for any of the image types.

This makes it easier for later PRs to use these functions for all image types.


test_conformance/images/kernel_read_write/test_common.cpp
```diff
@@ -105,52 +130,94 @@ void InitFloatCoordsCommon(image_descriptor *imageInfo,
...
         {
-            size_t width_lod, height_lod, depth_lod;
-
-            width_lod =
-                (imageInfo->width >> lod) ? (imageInfo->width >> lod) : 1;
-            height_lod =
-                (imageInfo->height >> lod) ? (imageInfo->height >> lod) : 1;
-            depth_lod =
-                (imageInfo->depth >> lod) ? (imageInfo->depth >> lod) : 1;
+            width_loop = (width_loop >> lod) ? (width_loop >> lod) : 1;
```
kpet, line 157: `lod` means Level Of Detail in that context. At first glance it seems like a meaninful name that should be kept.

ellnor01: Added a commit that restores the _lod

test_conformance/images/kernel_read_write/test_common.cpp
```diff
@@ -141,17 +175,52 @@ void InitFloatCoordsCommon(image_descriptor *imageInfo,
...
 }
 
+cl_mem create_image_of_type(cl_context context, cl_mem_flags mem_flags,
+                            image_descriptor *imageInfo, size_t row_pitch,
+                            size_t slice_pitch, void *host_ptr, cl_int *error)
+{
+    cl_mem image;
+    switch (imageInfo->type)
+    {
+        case CL_MEM_OBJECT_IMAGE3D:
```
jlewis-austin, line 202: We should add a default case to handle unsupported image types (especially since only 3D is supported now. Would be nice to add a comment noting that the implementation is incomplete.

test_conformance/images/kernel_read_write/test_common.cpp
```diff
@@ -141,17 +175,52 @@ void InitFloatCoordsCommon(image_descriptor *imageInfo,
...
+}
+
+static size_t get_image_num_pixels(image_descriptor *imageInfo, size_t width,
+                                   size_t height, size_t depth,
+                                   size_t array_size)
+{
+    size_t image_size;
+    switch (imageInfo->type)
+    {
+        case CL_MEM_OBJECT_IMAGE3D: image_size = width * height * depth; break;
```
jlewis-austin, line 219: We should add a default case to handle unsupported image types (especially since only 3D is supported now. Would be nice to add a comment noting that the implementation is incomplete.

jlewis-austin (changes requested):
This is a great improvement! Just a few comments on error handling

test_conformance/images/kernel_read_write/test_common.cpp
ellnor01: Done. 
Hopefully this is as expected, but it will be removed in a follow up pull request anyway.

test_conformance/images/kernel_read_write/test_common.cpp
ellnor01: Done.
Hopefully this is as expected, but it will be removed in a follow up pull request anyway.

test_conformance/images/kernel_read_write/test_common.cpp
```diff
@@ -34,20 +34,45 @@ cl_sampler create_sampler(cl_context context, image_sampler_data *sdata, bool te
...
+        case CL_MEM_OBJECT_IMAGE2D: height = imageInfo->height; break;
+        case CL_MEM_OBJECT_IMAGE2D_ARRAY:
+            height = imageInfo->height;
+            depth = imageInfo->arraySize;
+            break;
+        case CL_MEM_OBJECT_IMAGE3D:
+            height = imageInfo->height;
+            depth = imageInfo->depth;
+            break;
+    }
```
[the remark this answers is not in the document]
StuartDBrady: The _errCode_ argument to _test_error_ should only ever be an OpenCL error code.

Using `test_error(-1, "ERROR: <some message>")` will print:

```
ERROR: ERROR: <some message> (CL_DEVICE_NOT_FOUND from filename:lineno)
```

It will also return _CL_DEVICE_NOT_FOUND_.  Could you use _log_error_ and have this return _TEST_FAIL_ instead, please?  This change will need to propagate up a bit, to _InitFloatCoordsCommon_ and to the call to it in _test_read_image_.

test_conformance/images/kernel_read_write/test_common.cpp
```diff
@@ -34,122 +34,208 @@ cl_sampler create_sampler(cl_context context, image_sampler_data *sdata, bool te
...
+        case CL_MEM_OBJECT_IMAGE3D:
+            height = imageInfo->height;
+            depth = imageInfo->depth;
+            break;
+        default: test_error(-1, "ERROR: Test does not support image type");
+    }
+    return error;
+}
+
+cl_int InitFloatCoordsCommon(image_descriptor *imageInfo,
```
StuartDBrady, line 62: Can this be marked as static?

(Or should we use the anonymous namespace?)

StuartDBrady (changes requested):

test_conformance/images/kernel_read_write/test_common.cpp
```diff
@@ -34,122 +34,210 @@ cl_sampler create_sampler(cl_context context, image_sampler_data *sdata, bool te
     return sampler;
 }
 
-void InitFloatCoordsCommon(image_descriptor *imageInfo,
-                           image_sampler_data *imageSampler, float *xOffsets,
-                           float *yOffsets, float *zOffsets, float xfract,
-                           float yfract, float zfract, int normalized_coords,
-                           MTdata d, int lod)
+cl_int get_image_dimensions(image_descriptor *imageInfo, size_t &width,
```
StuartDBrady, line 37: If we return _cl_int_, the return type should be changed to _test_status_ (or you could return a _bool_, or define your own type, or ...).

test_conformance/images/kernel_read_write/test_common.cpp
```diff
@@ -161,6 +249,15 @@ int test_read_image(cl_context context, cl_command_queue queue,
     size_t threads[3];
     static int initHalf = 0;
 
+    size_t image_size =
+        get_image_num_pixels(imageInfo, imageInfo->width, imageInfo->height,
+                             imageInfo->depth, imageInfo->arraySize);
+    test_assert_error(0 != image_size, "Invalid image size");
+    size_t width_size, height_size, depth_size;
+    test_error(
```
StuartDBrady, line 257: If _get_image_dimensions_ returns a _test_status_, then that's not a CL error code, and so _test_error_ isn't the right thing to be using here, either.

test_conformance/images/kernel_read_write/test_common.cpp
```diff
@@ -34,122 +34,210 @@ cl_sampler create_sampler(cl_context context, image_sampler_data *sdata, bool te
...
+                             float *yOffsets, float *zOffsets, float xfract,
+                             float yfract, float zfract, int normalized_coords,
+                             MTdata d, int lod)
 {
     size_t i = 0;
-    if (gDisableOffsets)
+    size_t width_loop, height_loop, depth_loop;
+    cl_int error =
+        get_image_dimensions(imageInfo, width_loop, height_loop, depth_loop);
+    if (!error)
```
StuartDBrady, line 74: If _get_image_dimensions_ returns a _test_status_, we shouldn't assume anything about whether _TEST_PASS_ or _TEST_FAIL_ is zero.

test_conformance/images/kernel_read_write/test_common.h
```diff
@@ -63,8 +59,10 @@ int determine_validation_error_offset(
     bool clampingErr = false, clamped = false, otherClampingBug = false;
     int clampedX, clampedY, clampedZ;
 
-    size_t imageWidth = imageInfo->width, imageHeight = imageInfo->height,
-           imageDepth = imageInfo->depth;
+    size_t imageWidth, imageHeight, imageDepth;
+    cl_int error =
+        get_image_dimensions(imageInfo, imageWidth, imageHeight, imageDepth);
+    test_error(error, "invalid image dimensions");
```
StuartDBrady, line 65: If _get_image_dimensions_ returns a _test_status_, we shouldn't be passing that status to _test_error_.

StuartDBrady (changes requested):

test_conformance/images/kernel_read_write/test_common.cpp
ellnor01: I've gone for bool, as test_status feels like we should return TEST_PASS if successful, which isn't true at this point.

test_conformance/images/kernel_read_write/test_common.cpp
ellnor01: test_read_image always returns an error code, so I have return one here.

test_conformance/images/kernel_read_write/test_common.cpp
ellnor01: Hopefully this is resolved by changing to bool rather than test_status.

test_conformance/images/kernel_read_write/test_common.h
ellnor01: Changed this around a bit, does it looks better?
Quick note: This is the first of many more pull requests to refactor this code, unless there are more issues specific to this patch (get_image_dimensions will probably stay as is) can we revisit this on a later change once the code has settled into its final shape?

test_conformance/images/kernel_read_write/test_common.cpp
StuartDBrady: Looks good to me.

I did notice it returns either TEST_FAIL or 0 which get implicitly converted to bools, which isn't ideal, but is not worth holding the change up for.  I agree about fixing these small details later.

test_conformance/images/kernel_read_write/test_common.cpp
StuartDBrady: Looks good to me.

test_conformance/images/kernel_read_write/test_common.cpp
StuartDBrady: Seems reasonable, at least for now.

test_conformance/images/kernel_read_write/test_common.h
StuartDBrady: Looks better.  I agree about dealing with the remaining small details later.

test_conformance/images/kernel_read_write/test_common.cpp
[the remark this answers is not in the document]
StuartDBrady: The `test_error` issue has been dealt with.  @jlewis-austin, does the change here look good to you too?

test_conformance/images/kernel_read_write/test_common.cpp
StuartDBrady: Now marked as static.

test_conformance/images/kernel_read_write/test_common.cpp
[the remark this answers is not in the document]
jlewis-austin: yes, thanks!

test_conformance/images/kernel_read_write/test_common.cpp
jlewis-austin: OK thanks, I didn't realize they were going away.

jlewis-austin (approved):

jlewis-austin (approved):

StuartDBrady (approved):
Looks good to me.

kpet:
The PR has got two approvals and has been open for over 18 months, merging.

Issue #1142, opened 2021-02-03
test_non_uniform_work_group/non_uniform_3d_barriers has segfault if global buffer size is adjusted


hewj03:
Function TestNonUniformWorkGroup::adjustGlobalBufferSize adjusts global buffer size according to query result of CL_DEVICE_MAX_MEM_ALLOC_SIZE. 
If global buffer size is adjusted, there is segfault during kernel execution, because _globalSize is not adjusted accordingly.

testBarriers function in the kernel accesses global buffer with globalIndex=getGlobalIndex(gid2-get_global_offset(2), gid1-get_global_offset(1), gid0-get_global_offset(0));
globalIndex exceeds global buffer bound for some workitems.

PR #1143, opened 2021-02-03
[non_uniform_work_group] Adjust _globalSize after adjusting global buffer size

hewj03:
This fixes out-of-bound access of global buffer within testBarriers. see #1142


hewj03:
@gwawiork could you please help to review?

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
```diff
@@ -733,7 +756,15 @@ int TestNonUniformWorkGroup::runKernel () {
...
 
-  size_t globalBufferSize = adjustGlobalBufferSize(_numOfGlobalWorkItems*sizeof(cl_uint));
+  size_t globalBufferSize =
+      adjustGlobalBufferSize(_numOfGlobalWorkItems * sizeof(cl_uint));
+  if (globalBufferSize != _numOfGlobalWorkItems * sizeof(cl_uint))
+  {
+      // Adjust _globalSize and _numOfGlobalWorkItems accordingly if global
+      // buffer size is adjusted.
+      size_t maxNumOfGlobalWorkItems = globalBufferSize / sizeof(cl_uint);
+      adjustGlobalSize(maxNumOfGlobalWorkItems);
```
gwawiork, line 766: This function can modify _numOfGlobalWorkItems but globalBufferSize will stay the same. In the result we have again
`_numOfGlobalWorkItems *  sizeof(cl_uint) != globalBufferSize `
Am I right ?

hewj03: Yes, you're right. It won't cause new issue in testBarriers because kernel only accesses the global buffer in range [0, _numOfGlobalWorkItems)

hewj03: Done. Update globalBufferSize again to _numOfGlobalWorkItems * sizeof(cl_uint)

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
```diff
@@ -706,12 +706,38 @@ size_t TestNonUniformWorkGroup::adjustGlobalBufferSize(size_t globalBufferSize)
...
 }
 
+void TestNonUniformWorkGroup::adjustGlobalSize(size_t maxNumOfGlobalWorkItems)
+{
+    size_t oldNumOfGlobalWorkItems = _numOfGlobalWorkItems;
+    size_t oldGlobalSize[MAX_DIMS];
+    memcpy(oldGlobalSize, _globalSize, sizeof(_globalSize));
+    // Adjusting _globalSize[1] and _globalSize[2] is enough.
+    // _globalSize[0] is unchanged because it controls non-uniform workgroup.
+    for (int i = (int)_dims - 1; i >= 1; --i)
```
gwawiork, line 721: in such case when _dims == 1 no chance to execute

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
```diff
@@ -706,12 +706,38 @@ size_t TestNonUniformWorkGroup::adjustGlobalBufferSize(size_t globalBufferSize)
   size_t adjustedGlobalBufferSize = globalBufferSize;
   if (deviceMaxAllocObjSize < globalBufferSize) {
     adjustedGlobalBufferSize = deviceMaxAllocObjSize;
-    log_info("globalBufferSize was adjusted from %lu to %lu\n", globalBufferSize, adjustedGlobalBufferSize);
   }
 
   return adjustedGlobalBufferSize;
 }
 
+void TestNonUniformWorkGroup::adjustGlobalSize(size_t maxNumOfGlobalWorkItems)
```
gwawiork, line 714: I am thinking if better solution to have one common function with `adjustGlobalBufferSize`

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
```diff
@@ -706,12 +706,38 @@ size_t TestNonUniformWorkGroup::adjustGlobalBufferSize(size_t globalBufferSize)
...
+    {
+        while (_numOfGlobalWorkItems > maxNumOfGlobalWorkItems
+               && _globalSize[i] >= (2 * _localSize[i]))
+        {
+            _globalSize[i] /= 2;
+            _numOfGlobalWorkItems =
+                _globalSize[0] * _globalSize[1] * _globalSize[2];
+        }
+    }
+    assert(oldNumOfGlobalWorkItems != _numOfGlobalWorkItems
```
gwawiork, line 731: Not sure if such assert is acceptable. Maybe better solution is to return error code and modify variables by reference. Notice that `adjustGlobalBufferSize` function can return error code as `globalBufferSize`. It could be fixed as well when you decide to make one common function

gwawiork (changes requested):

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
hewj03: changed to adjust _globalSize[0] too

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
hewj03: merged into adjustGlobalBufferSize

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
hewj03: assert is removed

test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
```diff
@@ -703,12 +703,46 @@ size_t TestNonUniformWorkGroup::adjustGlobalBufferSize(size_t globalBufferSize)
...
+  size_t adjustedGlobalBufferSize = _numOfGlobalWorkItems * sizeof(cl_uint);
+  log_info("globalSize was adjusted from [%zu, %zu, %zu] to [%zu, %zu, %zu]\n",
+           oldGlobalSize[0], oldGlobalSize[1], oldGlobalSize[2], _globalSize[0],
+           _globalSize[1], _globalSize[2]);
+  log_info("numOfGlobalWorkItems was adjusted from %zu to %zu\n",
+           oldNumOfGlobalWorkItems, _numOfGlobalWorkItems);
+  log_info("globalBufferSize was adjusted from %zu to %zu\n", globalBufferSize,
+           adjustedGlobalBufferSize);
+
   return adjustedGlobalBufferSize;
```
gwawiork, line 746: I have test it with really strange value `deviceMaxAllocObjSize == 164096` and subcase `non_uniform_other_basic`. I noticed that before `return` we have to run `calculateExpectedValues();`. 
Additionally there could be an issue with `_globalWorkOffset` which is not updated after modification `_globalSize`

hewj03: thank you. I've added calculateExpectedValues and adjusted _globalWorkOffset.

hewj03:
@gwawiork could you please take a look again, thanks.

gwawiork (approved):
LGTM

hewj03:
Hi @kpet, could you please review this patch?
This patch fixes an issue that test_non_uniform_work_group/non_uniform_3d_barriers always segfault if global buffer size is adjusted by adjustGlobalBufferSize.

Issue #787, opened 2020-05-20
Narrowing conversion from unsigned long to signed long 


jrprice:
Enabling `-Werror` produces this:
```
../test_common/harness/conversions.cpp:211:14: error: narrowing conversion of ‘18446744071562067968’ from ‘long long unsigned int’ to ‘Long’ {aka ‘long int’} [-Wnarrowing]
  211 |         0, 0 };    // Last two values aren't stored here
```

This will be disabled in the root `CMakeLists.txt` file until fixed.

ellnor01:
This error has been fixed under #901
However reenabling the error has uncovered many other narrowing conversion failures. 
I will create a couple of patches to move the `-Wno-narrowing` flag down to specific suites with loads of issues, fixing some individual errors on the way. Then create new issues to remove the flag from the remaining suites.

ellnor01:
Under #1144 -Wno-narrowing flag is being moved down to individual directories which contain lots of occurrences. I have raised new issues to solve each of these: #1156 #1157 #1158 #1159.  Closing #1144 should now be enough to close this issue.

PR #1144, opened 2021-02-03
Re-enabling narrowing errors

ellnor01:
Fixes narrowing conversion build errors in test_common

Removing disable of narrowing errors in main CMakeLists.txt
and moving it down to specific test_conformance suite's 
CMakeLists.txt where there are many more build errors revealed
from this fix. 

Fixes a few simple issues under test_conformance in the process.

Contributes #787


test_common/harness/testHarness.h
```diff
@@ -26,7 +26,7 @@
 class Version {
 public:
     Version(): m_major(0), m_minor(0) {}
-    Version(int major, int minor): m_major(major), m_minor(minor) {}
+    Version(uint major, uint minor): m_major(major), m_minor(minor) {}
     bool operator>(const Version &rhs) const { return to_int() > rhs.to_int(); }
```
jlewis-austin, line 30: Should we change these to to_uint() for consistency?

ellnor01:
Any comments on here?

kpet:
@ahesham-arm This is failing the CI. It may require a rebase and/or fixes.

ahesham-arm:
@kpet Rebased and all CI checks are now passing.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -274,7 +274,7 @@ struct CommandBufferPrintfTest : public BasicCommandBufferTest
                                      &pattern[0], 0, nullptr, nullptr);
         test_error(error, "clEnqueueWriteBuffer failed");
 
-        cl_int offset[] = { 0, pattern.size() - 1 };
+        size_t offset[] = { 0, pattern.size() - 1 };
```
bashbaug, line 277: Can you please check that this change is correct?  I'm nervous that going from a 32-bit type (`cl_int`) to a possibly 64-bit type (`size_t`) may require some additional changes.

test_conformance/atomics/test_atomics.cpp
```diff
@@ -1292,8 +1292,8 @@ cl_long test_atomic_and_result_long(size_t size, cl_long *startRefValues,
 int test_atomic_and(cl_device_id deviceID, cl_context context,
                     cl_command_queue queue, int num_elements)
 {
-    TestFns set = { 0xffffffff,
-                    0xffffffffffffffffLL,
+    TestFns set = { (int)0xffffffff,
+                    (long long)0xffffffffffffffffLL,
```
bashbaug, line 1296: It might be better to cast these to the types used in the structure, which are guaranteed to have a fixed size, rather than `int` and `long long` that could be different sizes for different compilers.

```suggestion
    TestFns set = { (cl_int)0xffffffff,
                    (cl_long)0xffffffffffffffffLL,
```

ahesham-arm: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
ahesham-arm: `offset` is used in 3 locations in this function, two are trivial loop comparisons against another `size_t`, so those are fine. The other, main one, is a call to `clEnqueueWriteBuffer` which expects a `void*` and the size in bytes. The call is already using `sizeof` so I don't see why this would fail regardless of the data type and/or architecture.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -274,7 +274,7 @@ struct CommandBufferPrintfTest : public BasicCommandBufferTest
...
         error = clEnqueueWriteBuffer(queue, off_mem, CL_TRUE, 0, sizeof(offset),
                                      offset, 0, nullptr, nullptr);
         test_error(error, "clEnqueueWriteBuffer failed");
```
lakshmih, line 280: off_mem is created for sizeof(cl_int) size so I think the enqueuewritebuffer to it using sizeof(size_t) can cause issues.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp
```diff
@@ -115,7 +115,8 @@ struct CreateCommandBufferRepeatedProperties : public BasicCommandBufferTest
             TEST_FAIL);
 
         cl_command_buffer_properties_khr invalid_properties[3] = {
-            CL_COMMAND_BUFFER_FLAGS_KHR, CL_INVALID_PROPERTY, 0
+            CL_COMMAND_BUFFER_FLAGS_KHR,
+            (cl_command_buffer_properties_khr)CL_INVALID_PROPERTY, 0
         };
```
lakshmih, line 120: Not sure CL_INVALID_PROPERTY can be used in a property list like this. The spec only specifies its usage as a return value.

lakshmih (changes requested):
Left some comments

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp
ahesham-arm: I have used `-1` instead.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
ahesham-arm: This is now using `cl_uint` for the type of the array and an assert is added to ensure the casting is safe.

bashbaug (approved):

lakshmih (approved):

bashbaug:

PR #1145, opened 2021-02-04
Fix command line interface for math_brute_force

mantognini:
See https://github.com/KhronosGroup/OpenCL-CTS/pull/1127#issuecomment-770778474 and commit message for further details.

### DIFF
--- a/test_conformance/math_brute_force/main.cpp
+++ b/test_conformance/math_brute_force/main.cpp
@@ -140,7 +140,8 @@ int doTest(const char *name)
         const Func *const temp_func = functionList + i;
         if (strcmp(temp_func->name, name) == 0)
         {
-            if (i < gStartTestNumber || i > gEndTestNumber)
+            if ((gStartTestNumber != -1 && i < gStartTestNumber)
+                || i > gEndTestNumber)
             {
                 vlog("Skipping function #%d\n", i);
                 return 0;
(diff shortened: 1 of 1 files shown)


test_conformance/math_brute_force/main.cpp
```diff
@@ -140,7 +140,8 @@ int doTest(const char *name)
         const Func *const temp_func = functionList + i;
         if (strcmp(temp_func->name, name) == 0)
         {
-            if (i < gStartTestNumber || i > gEndTestNumber)
+            if ((gStartTestNumber != -1 && i < gStartTestNumber)
```
nikhiljnv, line 143: Would it make more sense to track updates to gStartTestNumber and gEndTestNumber via separate flags instead of relying on the initialized values of these variables?

nikhiljnv: LGTM apart from a small comment which may make parsing a little cleaner.

mantognini: This is really just a quick fixup. I do however agree the whole thing could do with a proper cleanup - I mean, those globals and so on are making the code hard to reason about. A proper framework to parse command line options would definitely help too. And the command line interface itself is also non-intuitive: While this is consistent with the documented usage (`-h`), `./math_brute_force 0 0` runs the first test when one might expect it runs no test... This would require a wider discussion, I suppose.

nikhiljnv (approved):
LGTM.

PR #1148, opened 2021-02-05
Implement Negative Tests for clDevice Functions

chemis01:


test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,536 @@
...
+{
+    cl_platform_id platform = getPlatformFromDevice(deviceID);
+
+    cl_device_id devices = nullptr;
+    cl_int err =
+        clGetDeviceIDs(nullptr, CL_DEVICE_TYPE_DEFAULT, 1, &devices, nullptr);
+    test_failure_error_ret(
+        err, CL_INVALID_PLATFORM,
+        "clGetDeviceIDs should return CL_INVALID_PLATFORM when: \"platform is "
+        "not a valid platform\" using a nullptr",
```
kpet, line 75: Passing `nullptr` for the platform is allowed and the behaviour is implementation-defined:

>platform refers to the platform ID returned by clGetPlatformIDs or can be NULL. If platform is
NULL, the behavior is implementation-defined.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,536 @@
...
+    Counts = 1 << 1,
+    Affinity = 1 << 2,
+    All_Schemes = Affinity | Counts | Equally,
+};
+
+static int get_supported_properties(cl_device_id deviceID)
+{
+    size_t number_of_properties = 0;
+    int err = clGetDeviceInfo(deviceID, CL_DEVICE_PARTITION_PROPERTIES, 0,
+                              nullptr, &number_of_properties);
```
kpet, line 293: It's the total size of the data that's returned, you need to divide by `sizeof(cl_device_partition_property)` to get the number of properties.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,536 @@
...
+    test_error(err, "clGetDeviceInfo");
+    std::vector<cl_device_partition_property> supported_properties(
+        number_of_properties);
+    err = clGetDeviceInfo(deviceID, CL_DEVICE_PARTITION_PROPERTIES,
+                          supported_properties.size(),
```
kpet, line 298: `supported_properties.size()` gives the number of elements, not the total size of the data. You'll need to multiply by `sizeof(cl_device_partition_property)`

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,536 @@
...
+}
+
+int test_negative_create_sub_devices(cl_device_id deviceID, cl_context context,
+                                     cl_command_queue queue, int num_elements)
+{
+    int supported_properties = get_supported_properties(deviceID);
+    if (supported_properties == SupportedPartitionSchemes::None)
+    {
+        printf("Device does not support creating subdevices... Skipping\n");
+        return TEST_SKIP;
```
kpet, line 360:
```suggestion
        return TEST_SKIPPED_ITSELF;
```

kpet (changes requested):

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,529 @@
...
+    cl_int err = clGetDeviceInfo(nullptr, CL_DEVICE_TYPE, sizeof(device_type),
+                                 &device_type, nullptr);
+    test_failure_error_ret(
+        err, CL_INVALID_DEVICE,
+        "clGetDeviceInfo should return CL_INVALID_DEVICE when: \"device is not "
+        "a valid device\" with a nullptr",
+        TEST_FAIL);
+
+    err =
+        clGetDeviceInfo(reinterpret_cast<cl_device_id>(context), CL_DEVICE_TYPE,
```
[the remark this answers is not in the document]
bashbaug: This spec issue is related, I think: https://github.com/KhronosGroup/OpenCL-Docs/issues/482

Device IDs are slightly different since you "get" them rather than "create" them and they aren't associated with a context, but there are a lot of similarities to the objects discussed on the issue as well.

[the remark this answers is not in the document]
jlewis-austin: Thanks, that's what I was thinking of. I guess our own discussion has broadened to the more generic question of whether error-checking should be required to distinguish between any two resource objects (specifically those implementing the icd _cl_<object> layout), so it's a question for negative testing on almost every API.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,529 @@
...
+    cl_ulong timer_resolution = 1;
+    if (get_device_cl_version(deviceID) >= Version(3, 0))
+    {
+        cl_int err = clGetPlatformInfo(
+            platform, CL_PLATFORM_HOST_TIMER_RESOLUTION,
+            sizeof(timer_resolution), &timer_resolution, nullptr);
+        test_error(err, "clGetPlatformInfo");
+    }
+
+    if (timer_resolution != 0)
```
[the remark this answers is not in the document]
chemis01: > clGetDeviceAndHostTimer is missing before version 2.1
https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clGetDeviceAndHostTimer

test_common/harness/parseParameters.cpp
```diff
@@ -198,6 +205,32 @@ int parseCustomParam(int argc, const char *argv[], const char *ignore)
                 return -1;
             }
         }
+        else if (!strncmp(argv[i],
+                          "--invalid-object-scenarios=", i_object_length))
+        {
+            if (strlen(argv[i]) > i_object_length)
+            {
+                delArg++;
+                gInvalidObject = 0;
```
jlewis-austin, line 214: It looks like the only way to specify no validity checking at all is by specifying anything at all, like `--invalid-object-scenarios=foo`. If we want to provide this capability, we should add an option like "none" and enforce that it's specified correctly.
Otherwise, since "nullptr" checks are always required, it seems like the whole scheme could be simplified greatly by providing a single boolean argument like `--check-object-types` or such.

test_conformance/api/negative_device.cpp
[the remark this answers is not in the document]
jlewis-austin: Thanks, I missed that

test_common/harness/parseParameters.cpp
kpet: It would be easy to add a "none" option but given that handling the "nullptr" case is mandatory and very simple, I'm not sure I see the point. Do you have a use in mind?

It was designed to take a list so the mechanism is extensible. As part of working group discussions (see https://github.com/KhronosGroup/OpenCL-Docs/issues/482), we have identified several other scenarios for which we may want to add support to the CTS in the future.

jlewis-austin: I may have misinterpreted the intent.  What I was reacting to is the fact that nullptr checks are mandatory, but it still needs to be provided explicitly as an argument. That led me to believe that we wanted to provide the ability to turn off these checks for whatever reason. That can be already be done by omitting the argument altogether, so disregard that part of my comment. 
The question I still have is why we need this argument capability (vs just testing whatever the spec requires for each API) and this level of granularity (vs a binary check/no-check option). Is the idea to provide optional stricter tests for vendors that have implemented those checks?

kpet: You got it. It's a lot less work to make this generic in the CTS than it is to write separate stricter tests and different vendors will handle different cases.

kpet: The plan is also to use those stricter options to test future validation layers.

test_conformance/api/negative_device.cpp
gwawiork: > Passing `nullptr` for the platform is allowed and the behaviour is implementation-defined:
> 
> > platform refers to the platform ID returned by clGetPlatformIDs or can be NULL. If platform is
> > NULL, the behavior is implementation-defined.

Then we have merged incorrect test case [here ](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/api/negative_platform.cpp#L45)? Related with question #1222

kpet: Yes, that code is incorrect. I've replied on the issue and asked @chemis01 to prepare a fix.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,509 @@
...
+        get_uint_device_info(deviceID, CL_DEVICE_PARTITION_MAX_SUB_DEVICES);
+    if (supported_properties & SupportedPartitionSchemes::Equally)
+    {
+        properties[0] = CL_DEVICE_PARTITION_EQUALLY;
+        properties[1] = max_compute_units;
+    }
+    else if (supported_properties & SupportedPartitionSchemes::Counts)
+    {
+        properties[0] = CL_INVALID_DEVICE_PARTITION_COUNT;
+        properties[1] = max_sub_devices;
```
bashbaug, line 359: I'm not quite following what this case is trying to test, but please conform this is intended: Should this be `CL_DEVICE_PARTITION_BY_COUNT` and a list of counts instead?  Do we need a `CL_DEVICE_PARTITION_BY_COUNTS_LIST_END`?

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,509 @@
...
+                                 out_devices, nullptr);
+        test_failure_error_ret(
+            err, CL_INVALID_VALUE,
+            "clCreateSubDevices should return CL_INVALID_VALUE when: \"values "
+            "specified in properties are valid but not supported by the "
+            "device\"",
+            TEST_FAIL);
+    }
+
+    err = clCreateSubDevices(deviceID, properties, properties[1], out_devices,
```
bashbaug, line 412: Please check that this is correct also - it's passing `properties[1]` as `num_devices`?

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,509 @@
...
+static std::vector<cl_device_partition_property>
+get_invalid_properties(int unsupported_properties)
+{
+    if (unsupported_properties & SupportedPartitionSchemes::Equally)
+    {
+        return { CL_DEVICE_PARTITION_EQUALLY, 1, 0 };
+    }
+    else if (unsupported_properties & SupportedPartitionSchemes::Counts)
+    {
+        return { CL_DEVICE_PARTITION_BY_COUNTS, 1, 0 };
```
bashbaug, line 313: I think this is missing a `CL_DEVICE_PARTITION_BY_COUNTS_LIST_END`.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,509 @@
...
+    }
+    return ret;
+}
+
+static std::vector<cl_device_partition_property>
+get_invalid_properties(int unsupported_properties)
+{
+    if (unsupported_properties & SupportedPartitionSchemes::Equally)
+    {
+        return { CL_DEVICE_PARTITION_EQUALLY, 1, 0 };
```
bashbaug, line 309: Not sure this is worth worrying about, but there is a chance partitioning with these properties could exceed `CL_DEVICE_PARTITION_MAX_SUB_DEVICES`.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,509 @@
...
+
+    cl_device_partition_property properties[3] = {};
+    cl_uint max_compute_units =
+        get_uint_device_info(deviceID, CL_DEVICE_MAX_COMPUTE_UNITS);
+    cl_uint max_sub_devices =
+        get_uint_device_info(deviceID, CL_DEVICE_PARTITION_MAX_SUB_DEVICES);
+    if (supported_properties & SupportedPartitionSchemes::Equally)
+    {
+        properties[0] = CL_DEVICE_PARTITION_EQUALLY;
+        properties[1] = max_compute_units;
```
bashbaug, line 354: I don't think this is a valid partitioning since there will only be one output device.

bashbaug (changes requested):
Mostly LGTM, but can you please check the device partitioning cases?

test_conformance/api/negative_device.cpp
chemis01: This function is used to create a set of properties which the implementation does not support. The values in here should be "valid" but for a given implementation should never succeed.

test_conformance/api/negative_device.cpp
chemis01: Yes it is. Updated

test_conformance/api/negative_device.cpp
chemis01: This is a difficult one as the spec isn't clear on how to ensure that CL_DEVICE_PARTITION_FAILED is returned.
Would appreciate some suggestions on this.

test_conformance/api/negative_device.cpp
chemis01: Good catch, this is a mistake and has been updated.

test_conformance/api/negative_device.cpp
bashbaug: OK, I see, these properties should only be used when an implementation does not support `CL_DEVICE_PARTITION_EQUALLY`, to test that `CL_INVALID_VALUE` is returned.

Theoretically I think an implementation could return `CL_DEVICE_INVALID_PARTITION_COUNT` here, if the total number of sub-devices when partitioning equally exceeds the maximum number of sub-devices, but in practices I bet most implementations would check that this partitioning type is supported and hence would return `CL_INVALID_VALUE` first.

I don't think this is worth worrying about.

test_conformance/api/negative_device.cpp
bashbaug: This is still an issue that needs to be solved.

Our CPU device is currently "failing" this test because this partitioning is invalid - it only creates one output device.  Since we check for this condition relatively early we return `CL_DEVICE_PARTITION_FAILED` in the case where `num_devices` is zero rather than `CL_INVALID_VALUE:

```
>>>> clCreateSubDevices: in_device = Intel(R) Core(TM) i5-7300U CPU @ 2.60GHz (CL_DEVICE_TYPE_CPU) (0x558e839082b8), properties = [ CL_DEVICE_PARTITION_EQUALLY = 4 ], num_devices = 0
<<<< clCreateSubDevices -> CL_DEVICE_PARTITION_FAILED
ERROR: clCreateSubDevices should return CL_INVALID_VALUE when: "out_devices is not NULL and num_devices is less than the number of sub-devices created by the partition scheme"! (Got CL_DEVICE_PARTITION_FAILED, expected CL_INVALID_VALUE from /home/bashbaug/git/OpenCL-CTS/test_conformance/api/negative_device.cpp:399)
```

test_conformance/api/negative_device.cpp
bashbaug: Yes, I agree this is a little tricky.

I interpret the spec as saying that `CL_DEVICE_PARTITION_FAILED` will be returned when all parameters are valid but no partitioning will occur.  There are multiple ways to do this, for example: partitioning equally into the same number of compute units as the input device, partitioning by counts into the same number of compute units as the input device, or partitioning by affinity domain when there are no more affinity domains to partition by.

One possible gotcha with this case is that (I think) `properties[1]` (and hence `num_devices`) is non-zero, but `out_devices` is `nullptr`.  I don't see anything prohibiting this in the spec, but it's admittedly a little weird.  Is this valid?  What should happen in this case?

test_conformance/api/negative_device.cpp
kpet: Hmm, the order in which error conditions should be checked and reported is not specified, maybe that's a discussion we should have. In that particular case, and with the spec as it stands, I agree that your implementation should be allowed to return `CL_DEVICE_PARTITION_FAILED`. The test could only check for that particular `CL_INVALID_VALUE` when there exists at least one valid partitioning scheme for the device. That would take care of the ambiguity.

test_conformance/api/negative_platform.cpp
```diff
@@ -42,15 +42,17 @@ int test_negative_get_platform_info(cl_device_id deviceID, cl_context context,
...
-    test_failure_error_ret(
-        err, CL_INVALID_PLATFORM,
-        "clGetPlatformInfo should return CL_INVALID_PLATFORM  when: \"platform "
-        "is not a valid platform\" using a valid object which is NOT a "
-        "platform",
-        TEST_FAIL);
+    cl_int err(CL_SUCCESS);
+    for (auto invalid_platform : get_invalid_objects<cl_platform_id>(deviceID))
+    {
+        err = clGetPlatformInfo(invalid_platform, CL_PLATFORM_VERSION,
```
jlewis-austin, line 48: Behavior when passing a null platform is implementation-defined, so we need to filter out nullptr objects here

kpet: This is handled as a special case in `get_invalid_objects`, see https://github.com/KhronosGroup/OpenCL-CTS/pull/1148/files#diff-877e5e63eeec3f4e2314e73fb4c44f8e997bdfdc9e2fbb738798f2ef91082eebR201.

jlewis-austin: yes indeed, thanks

jlewis-austin:
@chemis01 do you mind if I split off the infrastructure required for the #1318 fix into a separate PR (with attribution of course)? Or you can do it if you'd rather. That would also simplify this one a bit and maybe help get it pushed through faster.

chemis01:
> @chemis01 do you mind if I split off the infrastructure required for the #1318 fix into a separate PR (with attribution of course)? Or you can do it if you'd rather. That would also simplify this one a bit and maybe help get it pushed through faster.

@jlewis-austin I am completely fine with you splitting it off into a separate PR.

jlewis-austin:
> > @chemis01 do you mind if I split off the infrastructure required for the #1318 fix into a separate PR (with attribution of course)? Or you can do it if you'd rather. That would also simplify this one a bit and maybe help get it pushed through faster.
> 
> @jlewis-austin I am completely fine with you splitting it off into a separate PR.

We ended up going with a simpler fix in #1374 to expedite it and keep it separate from the refactor work.

ahesham-arm:
@kpet Rebased and CI checks are now passing.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,520 @@
...
+        test_failure_error_ret(
+            err, CL_INVALID_VALUE,
+            "clCreateSubDevices should return CL_INVALID_VALUE when: \"values "
+            "specified in properties are valid but not supported by the "
+            "device\"",
+            TEST_FAIL);
+    }
+
+    err = clCreateSubDevices(deviceID, properties, max_for_partition + 1,
+                             out_devices.data(), nullptr);
```
bashbaug, line 423: I think there is still a problem here.  On the Intel CPU OpenCL implementation, testing on a CPU with 24 cores, I am seeing this call being generated:

```
>>>> clCreateSubDevices: in_device = 12th Gen Intel(R) Core(TM) i9-12900K (CL_DEVICE_TYPE_CPU) (0x5e7b00958778), properties = [ CL_DEVICE_PARTITION_EQUALLY = 1 ], num_devices = 25
<<<< clCreateSubDevices -> CL_SUCCESS
```

This call is returning `CL_SUCCESS`, and I don't see any reason why it should not return `CL_SUCCESS`: it is only creating 24 sub-devices with this call and the 25th is un-written, but this should be fine.  This certainly is not testing the case where "in_device could not be further partitioned".

Perhaps the properties should be different to test this case?  For example, I believe trying to partition equally into a device with 24 compute units will generate the expected error, since in this case the input device is not being further partitioned:

```
>>>> clCreateSubDevices: in_device = 12th Gen Intel(R) Core(TM) i9-12900K (CL_DEVICE_TYPE_CPU) (0x5f50d1d32778), properties = [ CL_DEVICE_PARTITION_EQUALLY = 24 ], num_devices = 1
ERROR! clCreateSubDevices returned CL_DEVICE_PARTITION_FAILED (-18)
<<<< clCreateSubDevices -> CL_DEVICE_PARTITION_FAILED
```

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,520 @@
...
+        TEST_FAIL);
+
+    if (supported_properties & SupportedPartitionSchemes::Counts)
+    {
+        properties[0] = CL_DEVICE_PARTITION_BY_COUNTS;
+        properties[1] = max_sub_devices + 1;
+        err = clCreateSubDevices(deviceID, properties, max_sub_devices + 1,
+                                 out_devices.data(), nullptr);
+        test_failure_error_ret(
+            err, CL_INVALID_DEVICE_PARTITION_COUNT,
```
bashbaug, line 449: Note: our CPU implementation is returning `CL_DEVICE_PARTITION_FAILED` in this case (and the case below), not `CL_INVALID_DEVICE_PARTITION_COUNT`.  I suppose `CL_INVALID_DEVICE_PARTITION_COUNT` is more descriptive, but `CL_DEVICE_PARTITION_FAILED` seems like it is appropriate too.  Should we accept both error codes, or do we definitely want this case to return `CL_INVALID_DEVICE_PARTITION_COUNT`?

bashbaug (changes requested):

bashbaug:
I'm going to take off "focused review" until the review comments have been addressed - thanks!

PR #1150, opened 2021-02-09
Queue profiling and Device timebase equality test

### LINKED ISSUE #753: Timebase equality between queue profiling and clGetDeviceAndHostTimer untested
The OpenCL spec (§4.2) reads:

>The timestamp returned from clGetEventProfilingInfo for an event on a device and a device
timestamp queried from the same device will always be in the same timebase.

This is not tested by the CTS.

ellnor01:
Adds test to confirm that the timestamp returned from
clGetEventProfilingInfo for an event on a device and a device
timestamp queried from the same device are in the same timebase.

Fixes #753


test_conformance/profiling/profiling_timebase.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+    err = clSetUserEventStatus(uEvent, CL_COMPLETE);
+    test_error(err, "Unable to complete user event");
+
+    err = clWaitForEvents(1, &kEvent);
+    test_error(err, "clWaitForEvents failed");
+
+    cl_ulong deviceTimeAfterCompletion;
+    err =
+        clGetDeviceAndHostTimer(device, &deviceTimeAfterCompletion, &hostTime);
```
gwawiork, line 61: This is optional feature in OpenCL 3.0. You cannot assume this function is always available
https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_device_and_host_timer_synchronization

ellnor01:
Any more comments for this PR?

lakshmih (dismissed):

test_conformance/profiling/profiling_timebase.cpp
```diff
@@ -0,0 +1,104 @@
...
+    err = clGetEventProfilingInfo(kEvent, CL_PROFILING_COMMAND_END,
+                                  sizeof(cl_ulong), &eventEnd, NULL);
+    test_error(err, "clGetEventProfilingInfo failed");
+
+    test_assert_error(deviceTimeBeforeQueue < eventQueue,
+                      "Device timestamp was taken before kernel was queued");
+    test_assert_error(eventQueue < deviceTimeAfterQueue,
+                      "Device timestamp was taken after kernel was queued");
+    test_assert_error(deviceTimeAfterQueue < eventSubmit,
+                      "Device timestamp was taken before kernel was submitted");
```
bashbaug, line 97: I don't think this is a safe check.  The OpenCL execution model says:

> Submitted: The command is flushed from the command-queue and submitted for execution on the device. Once flushed from the command-queue, a command will execute after any prerequisites for execution are met.

I believe this means that the transition to the SUBMITTED state can happen anytime after the QUEUED state, even if the event dependencies defining the "prerequisites for execution" are not yet met.

bashbaug (changes requested):

test_conformance/profiling/profiling_timebase.cpp
ahesham-arm: Agree, this is how I interpreted the quoted paragraph from the specification as well. Assert removed.

test_conformance/profiling/profiling_timebase.cpp
```diff
@@ -0,0 +1,102 @@
...
+                                      &kernelCode, "kernel_empty");
+    test_error(err, "Failed to create kernel");
+
+    cl_ulong deviceTimeBeforeQueue;
+    err = clGetDeviceAndHostTimer(device, &deviceTimeBeforeQueue, &hostTime);
+    test_error(err, "Unable to get starting device time");
+
+    clEventWrapper userEvents = uEvent;
+    err = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, NULL, NULL, 1,
+                                 &userEvents, &kEvent);
```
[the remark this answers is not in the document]
ahesham-arm: Done, thanks.

bashbaug (approved):

bashbaug:

PR #1155, opened 2021-02-12
Reduce difference between files

mantognini:
Another fairly trivial change for math_brute_force that will allow us to reduce code duplication in the long run.

---

Moves `if (!gSkipCorrectnessTesting)` in binary_i.cpp to follow structure of binary.cpp, unary.cpp, macro_unary.cpp and macro_binary.cpp.

Other non-semantic changes include adding/removing new lines and updating comments.


EwanC (approved):

kpet (approved):
LGTM

PR #1160, opened 2021-02-15
Fix Minor memory leaks in test_buffer/compiler.

JohnKesapidesARM:
Fix vairous memory leaks in test_buffer and test_compiler.

Signed-off-by: John Kesapides <john.kesapides@arm.com>

Issue #465, opened 2019-10-02
Fix / redesign / replace object wrapper classes


kpet:
The object wrapper classes defined in `test_common/harness/typeWrappers.h` are currently broken in a number of scenarios which may result in bugs similar to #462/#463. Here are a few obvious issues:

What happens in #462 is that a copy of the wrapper is made but the copy constructor (default implementation) doesn't retain the object which is therefore released twice (once when the copy of the wrapper is destroyed and another time when the original wrapper is destroyed). Do we want to delete the copy constructor to prevent these mistakes? This would make the `Wrapper wrapper = object` idiom impossible as it requires the copy constructor at compilation time (and relies on copy elision for correct execution) but this is easily replaced by a direct construction from the object which would arguably be better anyway.

Another issue is that the copy and move assignment aren't implemented but they aren't deleted either. Which means that the following will lead to `clRelease*` potentially being called on a destroyed object:

```
clEventWrapper myEvent = ...
clEventWrapper otherEvent;
otherEvent = myEvent;
```

Finally these wrappers have been copy/pasted for each type. It would be better to use a template class with specializations for the part that calls the `clRelease*` function. We could also consider deleting them entirely and using the wrapper class that comes with the C++ bindings.

Any thoughts?

rjodinchr:
Fixed by PR #1268 (commit `69f00540`), which overhauled `test_common/harness/typeWrappers.h` with the RAII `Wrapper<T>` template managing proper OpenCL object reference counting (`clRetain*`/`clRelease*`).

PR #1162, opened 2021-02-16
Release clMemObj on reasignment

SebastianLuzynski:
This change fixes a bug introduced in PR #1082.

Signed-off-by: Sebastian Luzynski <sebastian.jozef.luzynski@intel.com>

### DIFF
--- a/test_common/harness/typeWrappers.h
+++ b/test_common/harness/typeWrappers.h
@@ -120,6 +120,7 @@ public:
 
     clMemWrapper &operator=(const cl_mem &rhs)
     {
+        if (mMem != NULL) clReleaseMemObject(mMem);
         mMem = rhs;
         return *this;
     }


SebastianLuzynski:
@kpet, according to OpenCL spec, you cannot use same hostPtr multiple times. You need either to free the buffers or create more hostPtr. In your's pull request #1082 you have to either recreate buffers in inner loop `if ((flag_set[src_flag_id] & CL_MEM_USE_HOST_PTR) || (flag_set[src_flag_id] & CL_MEM_COPY_HOST_PTR))` or apply a check inside `clMemWrapper` so that it calls `clReleaseMemObject` on reassignment.

gwawiork:
I think PR #1082 was added by @ellnor01  . Could you look at this issue ?

jlewis-austin:
I think #1160 should address this. If not, this fix might be incorporated into that one.

kpet:
@SebastianLuzynski As @gwawiork pointed out #1082 was not my change.

I agree the wrapper classes definitly need fixing (see https://github.com/KhronosGroup/OpenCL-CTS/issues/465) but I don't think we should merge a point fix for a single type of object as done in this PR. The first step in tackling these issues IMHO probably is to de-duplicate the existing code with templates.

PR #1163, opened 2021-02-16
Fix compilation error in rtz mode detection

doe300:
I ran into this issue when trying to run the math tests on my custom OpenCL implementation supporting the embedded profile.


mantognini (approved):
LGTM

EwanC (approved):

PR #1166, opened 2021-02-17
Rename files for consistency

mantognini:


EwanC (approved):

gwawiork (approved):
LGTM

gwawiork:
Some recent changes cause failures in my env on function `remquo` can you check on your side if it works properly on master ? Thanks !

mantognini:
It works fine here. I ran `test_bruteforce -w remquo -1` and `test_bruteforce -w remquo -[65536]`.

I don't know the nature of the issue you're seeing but there was a bugfix #1152 that improved error handling in the relevant file. The other changes made recently should not have impacted the behaviour of the test in any way AFAIK. There was also #1153 but it's unlikely to be the issue if only remquo fails.

gwawiork:
> It works fine here. I ran `test_bruteforce -w remquo -1` and `test_bruteforce -w remquo -[65536]`.
> 
> I don't know the nature of the issue you're seeing but there was a bugfix #1152 that improved error handling in the relevant file. The other changes made recently should not have impacted the behaviour of the test in any way AFAIK. There was also #1153 but it's unlikely to be the issue if only remquo fails.

I have kernel build errors. Looks like you have changed third argument of function from reference to value. Could you restore it?
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/math_brute_force/binary_two_results_i.cpp#L141
There was (correct)
                       ` "( in1[i], in2[i], out2 + i );\n" `
Now we have (incorrect)
                        `"( in1[i], in2[i], out2[i] );\n"`

mantognini:
Good catch! I've kept it correct on line 48 but not there. Will fix ASAP.

PR #1167, opened 2021-02-17
Regroup vtbl definitions to one translation unit

mantognini:
Move function declarations to a new header file, with relevant existing documentation.

---

This will allow further cleanups to be easier, such as splitting files to identify undesired differences between float and double versions of the tests.


EwanC (approved):
Nice cleanup 👍

gwawiork (approved):
LGTM

PR #1138, opened 2021-02-01
Reduce difference between files in math_brute_force

mantognini:
The main goal is to reduce code differences in `math_brute_force`. This will allow refactoring and reducing code duplication later, and identify undesired differences between tests (including bugs).

The changes proposed in this PR should not change any existing behaviour. They were identified iteratively using `nvim -d file1 file2 file3 ...`. See also the commit messages for additional details.

PR #1168, opened 2021-02-18
Fix incorrect use of kernel parameter

mantognini:
This issue was introduced in 8ad1088a (Reduce difference between files in math_brute_force (#1138), 2021-02-10).

Cf https://github.com/KhronosGroup/OpenCL-CTS/pull/1166#issuecomment-781143569
CC @gwawiork


gwawiork (approved):
LGTM

EwanC (approved):

PR #1169, opened 2021-02-18
Split math_brute_force files

mantognini:
The first commit split each file into two, one for `float` and one for `double`. The code is not modified, only moved around: every line that is removed is added in a new file. This, of course, introduces some code duplication but that will be improved later more globally along with existing code duplication.

The second patch is a clang-format patch for something that wasn't handled properly by `git-clang-format` on a previous commit but is now detected. The most likely reason behind this seems to be the size of files, which are now individually much smaller.


test_conformance/math_brute_force/binary_i_double.cpp
```diff
@@ -223,112 +129,6 @@ static cl_int BuildKernel_DoubleFn(cl_uint job_id, cl_uint thread_id UNUSED,
...
-    MAKE_HEX_FLOAT(+0x0.000006p-126f, +0x0000006L, -150),
-    MAKE_HEX_FLOAT(+0x0.000004p-126f, +0x0000004L, -150),
-    MAKE_HEX_FLOAT(+0x0.000002p-126f, +0x0000002L, -150),
-    +0.0f
-};
-
-static const size_t specialValuesFloatCount =
-    sizeof(specialValuesFloat) / sizeof(specialValuesFloat[0]);
-
 static const int specialValuesInt[] = {
```
mantognini, line 132: The reason this isn't removed is because there is a typo on line 1065 (previous version) / 295 (new version). This will be dealt with in a separate review to avoid introducing functionality changes in this large patch.

mantognini:
Rebased + forced push to re-trigger tests that were stuck. Content unchanged.

EwanC (approved):
Tried this branch locally and the same number of bruteforce tests run and pass as `master`.

What's the refactoring plan for reducing code duplication going forward? Looking forward to this code being more maintainable.

mantognini:
> What's the refactoring plan for reducing code duplication going forward? Looking forward to this code being more maintainable.

I intend to first spot and reduce trivial differences between duplicated functions, and fold code that is clearly duplicated but this might not be trivial as some functions have different flows. Hopefully, we can make small steps to make the flows uniform in all files.

PR #1172, opened 2021-02-23
Limit buffers sizes to leave some memory for the platform

ouakheli:
Some conformance tests use directly the size returned by the runtime
for max memory size to allocate buffers.
This doesn't leave enough memory for the system to run the tests.

### DIFF
--- a/test_common/harness/deviceInfo.cpp
+++ b/test_common/harness/deviceInfo.cpp
@@ -132,3 +132,42 @@ size_t get_max_param_size(cl_device_id device)
     }
     return ret;
 }
+
+static cl_ulong get_device_info_max_size(cl_device_id device,
+                                         cl_device_info info,
+                                         unsigned int divisor)
+{
+    cl_ulong max_size;
+
+    if (divisor == 0)
+    {
+        throw std::runtime_error("Allocation divisor should not be 0\n");
+    }
+
+    if (clGetDeviceInfo(device, info, sizeof(max_size), &max_size, NULL)
+        != CL_SUCCESS)
+    {
+        throw std::runtime_error("clGetDeviceInfo failed\n");
+    }
+    return max_size / divisor;
+}
+
+cl_ulong get_device_info_max_mem_alloc_size(cl_device_id device,
+                                            unsigned int divisor)
+{
+    return get_device_info_max_size(device, CL_DEVICE_MAX_MEM_ALLOC_SIZE,
+                                    divisor);
+}
+
+cl_ulong get_device_info_global_mem_size(cl_device_id device,
+                                         unsigned int divisor)
+{
+    return get_device_info_max_size(device, CL_DEVICE_GLOBAL_MEM_SIZE, divisor);
+}
+
+cl_ulong get_device_info_max_constant_buffer_size(cl_device_id device,
+                                                  unsigned int divisor)
+{
+    return get_device_info_max_size(device, CL_DEVICE_MAX_CONSTANT_BUFFER_SIZE,
+                                    divisor);
+}
--- a/test_conformance/allocations/main.cpp
+++ b/test_conformance/allocations/main.cpp
@@ -24,8 +24,10 @@
 
 typedef long long unsigned llu;
 
+#define REDUCTION_PERCENTAGE_DEFAULT 50
+
 int g_repetition_count = 1;
-int g_reduction_percentage = 100;
+int g_reduction_percentage = REDUCTION_PERCENTAGE_DEFAULT;
 int g_write_allocations = 1;
 int g_multiple_allocations = 0;
 int g_execute_kernel = 1;
@@ -44,24 +46,9 @@ test_status init_cl(cl_device_id device)
 {
     int error;
 
-    error = clGetDeviceInfo(device, CL_DEVICE_MAX_MEM_ALLOC_SIZE,
-                            sizeof(g_max_individual_allocation_size),
-                            &g_max_individual_allocation_size, NULL);
-    if (error)
-    {
-        print_error(error,
-                    "clGetDeviceInfo failed for CL_DEVICE_MAX_MEM_ALLOC_SIZE");
-        return TEST_FAIL;
-    }
-    error =
-        clGetDeviceInfo(device, CL_DEVICE_GLOBAL_MEM_SIZE,
-                        sizeof(g_global_mem_size), &g_global_mem_size, NULL);
-    if (error)
-    {
-        print_error(error,
-                    "clGetDeviceInfo failed for CL_DEVICE_GLOBAL_MEM_SIZE");
-        return TEST_FAIL;
-    }
+    g_max_individual_allocation_size =
+        get_device_info_max_mem_alloc_size(device);
+    g_global_mem_size = get_device_info_global_mem_size(device);
 
     log_info("Device reports CL_DEVICE_MAX_MEM_ALLOC_SIZE=%llu bytes (%gMB), "
              "CL_DEVICE_GLOBAL_MEM_SIZE=%llu bytes (%gMB).\n",
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -112,6 +112,8 @@ const char *sample_const_max_arg_kernel_pattern =
     "\n"
     "}\n";
 
+#define MAX_REDUCTION_FACTOR 4
+
 int test_min_max_thread_dimensions(cl_device_id deviceID, cl_context context,
                                    cl_command_queue queue, int num_elements)
 {
@@ -564,7 +566,7 @@ int test_min_max_mem_alloc_size(cl_device_id deviceID, cl_context context,
                                 cl_command_queue queue, int num_elements)
 {
     int error;
-    cl_ulong maxAllocSize, memSize, minSizeToTry;
+    cl_ulong maxAllocSize, memSize, minSizeToTry, currentSize;
     clMemWrapper memHdl;
 
     cl_ulong requiredAllocSize;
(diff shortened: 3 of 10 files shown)


test_conformance/api/test_api_min_max.cpp
```diff
@@ -491,12 +491,12 @@ int test_min_max_mem_alloc_size(cl_device_id deviceID, cl_context context, cl_co
...
-    error = clGetDeviceInfo( deviceID, CL_DEVICE_MAX_MEM_ALLOC_SIZE, sizeof( maxAllocSize ), &maxAllocSize, NULL );
-    test_error( error, "Unable to get max mem alloc size from device" );
-
-    error = clGetDeviceInfo( deviceID, CL_DEVICE_GLOBAL_MEM_SIZE, sizeof( memSize ), &memSize, NULL );
-    test_error( error, "Unable to get global memory size from device" );
+    /* Get the max mem alloc size, limit the alloc to half of the available
+     * memory */
+    maxAllocSize = get_device_info_max_mem_alloc_size(
+        deviceID, MAX_DEVICE_MEMORY_SIZE_DIVISOR);
+    memSize = get_device_info_global_mem_size(deviceID,
```
jlewis-austin, line 498: _requiredAllocSize_ already reduces the memSize by a factor of 4 below. If that value is wrong, it seems like the fix should go there rather than introducing a redundant reduction mechansim.

jlewis-austin (commented):
This looks like a pretty fundamental change to code that other implementations are passing without a problem.
I'd suggest starting with a detailed analysis of an individual test to show why the current value is unreasonable. Most tests do reduce the allocation size, so it seems that the question should be more about whether those reductions are correct or not, versus arbitrarily cutting everything in half.

test_conformance/api/test_api_min_max.cpp
ouakheli: A systematic reduction is done for some tests in allocations : test_conformance/allocations/main.cpp
I agree that we don't want a redundant mechanism but it is better if we have it in a single common helper.

I will review the reduction code already done (the division by a factor of 4) and propose a change to that.

ouakheli: Updated the change

ouakheli:
@jlewis-austin It depends really of what the implementation is returning as max size. If the implementation returns the whole available memory in the system, creating a buffer with that size will succeed but the system won't have enough memory left.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -491,47 +493,40 @@ int test_min_max_mem_alloc_size(cl_device_id deviceID, cl_context context, cl_co
...
+    log_info("Reported max allocation size of %lld bytes (%gMB) and global mem "
+             "size of %lld bytes (%gMB).\n",
+             maxAllocSize, maxAllocSize / (1024.0 * 1024.0), memSize,
+             memSize / (1024.0 * 1024.0));
 
     minSizeToTry = maxAllocSize/16;
-    while (maxAllocSize > (maxAllocSize/4)) {
-
+    currentSize = maxAllocSize;
+    while (currentSize >= maxAllocSize / PASSING_FRACTION)
```
jlewis-austin, line 528: It took me a while to figure out the name "PASSING_FRACTION", but I'm having trouble thinking of anything better. Maybe something like MAX_REDUCTION_FACTOR?

test_conformance/api/test_api_min_max.cpp
```diff
@@ -491,47 +493,40 @@ int test_min_max_mem_alloc_size(cl_device_id deviceID, cl_context context, cl_co
...
+        maxAllocSize = memSize;
+    }
 
     if (memSize > (cl_ulong)SIZE_MAX) {
       memSize = (cl_ulong)SIZE_MAX;
     }
 
-    if( maxAllocSize < requiredAllocSize)
+    if ((0 == gIsEmbedded && maxAllocSize < 128 * 1024 * 1024)
+        || maxAllocSize < 1 * 1024 * 1024)
```
jlewis-austin, line 515: These checks can introduce new failures since maxAllocSize has already been reduced. There should be a clear distinction between reported max alloc size and any modifications we do to it. I still have the same objection about reducing allocation size in two different places. Applying this factor up front is reducing coverage for implementations that currently pass the test with their reported size--after this change they would be testing only half the size they're capable of.

test_common/harness/deviceInfo.cpp
```diff
@@ -98,3 +98,42 @@ size_t get_max_param_size(cl_device_id device)
...
+
+    if (divisor == 0)
+    {
+        throw std::runtime_error("Allocation divisor should not be 0\n");
+    }
+
+    if (clGetDeviceInfo(device, info, sizeof(max_size), &max_size, NULL)
+        != CL_SUCCESS)
+    {
+        throw std::runtime_error("clGetDeviceInfo failed\n");
```
jlewis-austin, line 116: Is this caught somewhere? It looks like a change in functionality, since the existing test_error_abort macro just returns an error from the caller and allows further testing to continue.

jlewis-austin (commented):
There's some good fixes in here, but it also introduces functional changes that could impact existing implementations. I think we need to address the fundamental question first--should an implementation report a buffer size that can't be created due to platform memory usage? I tend to think that the allocation tests should be able to pass as-is without adjusting the reported numbers, but welcome other opinions.

ouakheli:
> There's some good fixes in here, but it also introduces functional changes that could impact existing implementations. I think we need to address the fundamental question first--should an implementation report a buffer size that can't be created due to platform memory usage? I tend to think that the allocation tests should be able to pass as-is without adjusting the reported numbers, but welcome other opinions.

Agreed we need more opinions on that before processing with the other comments as this is fundamental for this change.
I don't think any implementation should return something else than the maximum memory available for the buffer creation.
Only the application is aware of the memory needed for the system and the operations that are going to be performed in the system with the buffer.
The implementation can't know what is needed for the platform and this can be different from a platform to another, so if it doesn't report the whole memory it would apply an arbitrary factor and would work for most of the platforms ?

What about passing a parameter to the test ? as we do in test_allocation reduction% ?

jlewis-austin:
I should clarify that most of the reductions look fine, for the tests that aren't specifically testing max alloc size functionality. The main one I'm concerned about is `test_api_min_max`, where there's no "app" to speak of beyond spinning up the platform and doing a single allocation.

That being said, I think you're right about not estimating platform usage, and I got two trains of thought mixed up in the problem statement. The real question should have been whether the CTS lower limits on retries (ie, your PASSING_FRACTION) serve as a "reasonable" minimum value that should succeed, or if the tests should remove all limitations and simply keep re-trying down to size 1. On the max allocation tests, I think it makes sense in either case to always attempt the max size in order to verify that it either fails gracefully or succeeds.

test_common/harness/deviceInfo.cpp
ouakheli: I tried to harmonize this function with get_max_param_size function
Does't really make sense to continue the testing if we clGetDeviceInfo returns an error ?

test_conformance/api/test_api_min_max.cpp
ouakheli: This value is 1M for embedded and 128M otherwise.
I changed the check to use requiredAllocSize for me consistency but I don't see an application reporting 2M or 256M which with the reduction would make it fail ?

test_conformance/api/test_api_min_max.cpp
ouakheli: Change to MAX_REDUCTION_FACTOR

test_conformance/api/test_api_min_max.cpp
```diff
@@ -491,47 +493,40 @@ int test_min_max_mem_alloc_size(cl_device_id deviceID, cl_context context, cl_co
...
 
     minSizeToTry = maxAllocSize/16;
-    while (maxAllocSize > (maxAllocSize/4)) {
-
+    currentSize = maxAllocSize;
+    while (currentSize >= maxAllocSize / PASSING_FRACTION)
+    {
         log_info("Trying to create a buffer of size of %lld bytes (%gMB).\n", maxAllocSize, (double)maxAllocSize/(1024.0*1024.0));
         memHdl = clCreateBuffer( context, CL_MEM_READ_ONLY, (size_t)maxAllocSize, NULL, &error );
         if (error == CL_MEM_OBJECT_ALLOCATION_FAILURE || error == CL_OUT_OF_RESOURCES || error == CL_OUT_OF_HOST_MEMORY) {
```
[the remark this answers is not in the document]
ouakheli: Change to currentSize

test_conformance/allocations/main.cpp
```diff
@@ -24,8 +24,10 @@
 
 typedef long long unsigned llu;
 
+#define REDUCTION_PERCENTAGE_DEFAULT 50
```
bashbaug, line 27: I'm not necessarily opposed to this, but it does change the default behavior of the test pretty significantly, since a "reduction percentage" less than 100% also affects the number of work-items executed.

Should we just change the number of work-items executed unconditionally?  I'd feel more comfortable if this were just changing the max allocation percentage.

bashbaug (commented):
Changes look good to me technically.  I do wonder whether we're losing too much coverage reducing allocation size by 50% in many cases, but we can try this and see.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1575,9 +1542,8 @@ int test_min_max_constant_buffer_size(cl_device_id deviceID, cl_context context,
     MTdata d;
 
     /* Verify our test buffer won't be bigger than allowed */
-    error = clGetDeviceInfo(deviceID, CL_DEVICE_MAX_CONSTANT_BUFFER_SIZE,
-                            sizeof(maxSize), &maxSize, 0);
-    test_error(error, "Unable to get max constant buffer size");
+    maxSize = get_device_info_max_constant_buffer_size(
+        deviceID, MAX_DEVICE_MEMORY_SIZE_DIVISOR);
 
     if ((0 == gIsEmbedded && maxSize < 64L * 1024L) || maxSize < 1L * 1024L)
```
lakshmih, line 1548: Looks like this should be 
+    if ((0 == gIsEmbedded && maxSize*MAX_DEVICE_MEMORY_SIZE_DIVISOR < 64L * 1024L) || maxSize*MAX_DEVICE_MEMORY_SIZE_DIVISOR < 1L * 1024L)

ahesham-arm: Done.

lakshmih (approved):

PR #1173, opened 2021-02-24
fix out of boundary when strcpy

### LINKED ISSUE #1170: out of boundary when strcpy
out of boundary at test_conformance/gles/test_fence_sync.cpp:164

ijkmn:
fix #1170 ,out of boundary when strcpy


kpet (approved):
LGTM. Even better would be to use `std::string` :)

kpet:
The formatting errors are not related to the change which is otherwise trivial, merging.

PR #1174, opened 2021-02-25
Assemble SPIR-V binaries for spirv_new

james-morrissey-arm:
The SPIR-V binaries needed for the spirv_new test are now assembled
automatically using the SPIR-V tools.  If set, the SPIRV_TOOLS_DIR
option is used to locate the tools, otherwise they are expected to
be found in PATH.

Additionally, the source files are listed explicitly to avoid globbing
which is discouraged in the cmake manual.  The long list of spir-v asm
files is stored in a separate file to keep this manageable.

Signed-off-by: Xing Huang <xing.huang@arm.com>


kpet:
Superseded by https://github.com/KhronosGroup/OpenCL-CTS/pull/2200, closing.

PR #1176, opened 2021-03-02
Fold functions for nextafter

mantognini:
Remove unnecessary functions. The first commit adds some asserts to prove that the assumptions used by the second commit are correct.


kpet (approved):
LGTM

EwanC (approved):

PR #1179, opened 2021-03-05
Use C++ headers instead of C headers

mantognini:
Prefer C++ headers and remove some unnecessary includes.


kpet (approved):
LGTM

kpet:
Trivial change, merging with a single review.

PR #1185, opened 2021-03-09
Move code around to reduce differences

mantognini:
Code is moved to reduce the differences between tests for single- and double-precision.

Improve consistency in double-literal.


kpet (approved):
LGTM, trivial code movement, merging.

PR #1191, opened 2021-03-17
Do not dereference null pointer for no matching tests

svenvh:
When invoking for example

    test_c11_atomics test-that-does-not-exist

`parseAndCallCommandLineTests()` would attempt to dereference `resultTestList` which is still a null pointer.


jlewis-austin (approved):
Thanks, just hit this today!

PR #1197, opened 2021-03-24
Limit workgroup size for atomics tests

zsrkmyn:
This avoids extremely large local buffer size and slow run.

### DIFF
--- a/test_conformance/atomics/test_atomics.cpp
+++ b/test_conformance/atomics/test_atomics.cpp
@@ -202,7 +202,7 @@ int test_atomic_function(cl_device_id deviceID, cl_context context, cl_command_q
 
         // Limit workSize to avoid extremely large local buffer size and slow
         // run.
-        if (testFns.NumResultsFn && workSize > 65536) workSize = 65536;
+        if (workSize > 65536) workSize = 65536;
 
         // "workSize" is limited to that of the first dimension as only a 1DRange is executed.
         if( maxSizes[0] < workSize )
--- a/test_conformance/atomics/test_atomics.cpp
+++ b/test_conformance/atomics/test_atomics.cpp
@@ -200,6 +200,10 @@ int test_atomic_function(cl_device_id deviceID, cl_context context, cl_command_q
         error = clGetKernelWorkGroupInfo( kernel, deviceID, CL_KERNEL_WORK_GROUP_SIZE, sizeof( workSize ), &workSize, NULL );
         test_error( error, "Unable to obtain max work group size for device and kernel combo" );
 
+        // Limit workSize to avoid extremely large local buffer size and slow
+        // run.
+        if (testFns.NumResultsFn && workSize > 65536) workSize = 65536;
+
         // "workSize" is limited to that of the first dimension as only a 1DRange is executed.
         if( maxSizes[0] < workSize )
         {


zsrkmyn:
A soft ping.

zsrkmyn:
A soft ping again :-)

test_conformance/atomics/test_atomics.cpp
```diff
@@ -200,6 +200,10 @@ int test_atomic_function(cl_device_id deviceID, cl_context context, cl_command_q
         error = clGetKernelWorkGroupInfo( kernel, deviceID, CL_KERNEL_WORK_GROUP_SIZE, sizeof( workSize ), &workSize, NULL );
         test_error( error, "Unable to obtain max work group size for device and kernel combo" );
 
+        // Limit workSize to avoid extremely large local buffer size and slow
+        // run.
+        if (testFns.NumResultsFn && workSize > 65536) workSize = 65536;
+
         // "workSize" is limited to that of the first dimension as only a 1DRange is executed.
         if( maxSizes[0] < workSize )
```
gwawiork, line 208: What about limitation of `workSize` based on this condition. Watching Intel platforms this values in `maxSizes[0]` are much smaller than 65536. Could you deliver information what values you have in `maxSizes[0]` and `workSize` . I suppose you have broken values in `maxSizes` array.

gwawiork: Additionally please check value `CL_DEVICE_MAX_WORK_GROUP_SIZE`. According to specification  https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_WORK_GROUP_SIZE
`CL_KERNEL_WORK_GROUP_SIZE will be less than or equal to CL_DEVICE_MAX_WORK_GROUP_SIZE for a given kernel object.`

zsrkmyn: @gwawiork, thanks for comments!

With Intel OpenCL CPU runtime, the value 8192, while with Intel FPGA emulator, the max allowed work group size is `64 * 1024 * 1024`, which is quite large. And it can require huge amount of memory and can also be slow to run the test on FPGA emulator (which is actually a CPU device).

gwawiork: I do not understand. Could you write once again what value you have in maxSizes[0] (CL_DEVICE_MAX_WORK_ITEM_SIZES) and in workSize (CL_KERNEL_WORK_GROUP_SIZE) ?

zsrkmyn: `maxSizes[0] = workSize = 64 * 1024 * 1024 = 67108864`

gwawiork: Form me this code is ok. 
Just additional suggestion/idea.  It would be useful to add user command parameter to reduce or force given value just make these test runnable on emulation/silicon environment in different conditions.

gwawiork (dismissed):
LGTM

zsrkmyn:
Thanks @gwawiork for approval!

Could anybody with write access to the repo help take a look?

mkinsner:
@EwanC since you've been active reviewing other CTS changes, would you or a delegate be able to review this small PR?  I've been asked to help push it through the backlog.

Thanks!

test_conformance/atomics/test_atomics.cpp
```diff
@@ -200,6 +200,10 @@ int test_atomic_function(cl_device_id deviceID, cl_context context, cl_command_q
         error = clGetKernelWorkGroupInfo( kernel, deviceID, CL_KERNEL_WORK_GROUP_SIZE, sizeof( workSize ), &workSize, NULL );
         test_error( error, "Unable to obtain max work group size for device and kernel combo" );
 
+        // Limit workSize to avoid extremely large local buffer size and slow
+        // run.
+        if (testFns.NumResultsFn && workSize > 65536) workSize = 65536;
```
kpet, line 205: Why are you also testing `testFns.NumResultsFn` is non-zero?

zsrkmyn: @kpet, thanks for comments. Checking the context of the program, I found that `testFns.NumResultsFn` is used to calculate the local buffer size, and only when it's set, local buffer is required. So if it's not set, we don't need the clamp the `workSize` to limit local buffer usage.

zsrkmyn:
A soft ping.

test_conformance/atomics/test_atomics.cpp
kpet: Would it work to always clamp the size? It is always queried and, assuming I understand correctly, this would help readability I think.

zsrkmyn: Done. Thanks!

kpet (approved):

test_conformance/atomics/test_atomics.cpp
kpet: Looks good, thanks for the changes (and the patience!)

PR #1211, opened 2021-04-06
Remove unsupported code

mantognini:
The GetTime() and associated functions are not fully implemented on Linux. This functionality is assumed to be untested, or unused at best.

Reduce differences between tests by removing this unnecessary feature. It can be (re-)implemented later, if desired, once the math_brute_force component is in better shape.


test_conformance/math_brute_force/main.cpp
```diff
@@ -813,24 +810,6 @@ int main(int argc, const char *argv[])
...
-        if (gSkipCorrectnessTesting)
-            vlog("   \t               ");
-        else
-            vlog("   \t                                        ");
-        if (gWimpyMode) vlog("   ");
-        for (int i = gMinVectorSizeIndex; i < gMaxVectorSizeIndex; i++)
-            vlog("\t  float%s", sizeNames[i]);
-    }
-    else
-    {
```
StuartDBrady, line 830: Although I don't especially agree with the use of "\t" in log output, as it generally doesn't serve the intended purpose of aligning output, was this intended?  Shouldn't the code in the "else" branch be made unconditional, instead?

StuartDBrady: AFAICT, this is to accommodate the difference in string length between "passed" and "Wimp pass". There should have been a comment explaining this, but it's reasonable not to add one right now. However, this difference in string lengths still appears to apply.

StuartDBrady (changes requested):
I think this may need a minor change in printing of the header line in main(), as described.

Otherwise, this is a straightforward removal of unneeded code, and the change looks good.

test_conformance/math_brute_force/main.cpp
mantognini: Right, good catch! I've re-introduced the else branch unconditionally. The output format can be further improved later if desired.

StuartDBrady: Agreed. It seems plausible to me that there was a general accommodation made for the measurement info made in the formatting of the rest of the output, which is no longer needed - but it's best not to change this all in one go.

StuartDBrady (approved):
Looks good to me.

mantognini:
I would welcome review from another company to get this merged. Thank you.

test_conformance/math_brute_force/main.cpp
```diff
@@ -62,8 +62,6 @@ static int32_t gEndTestNumber = -1;
 int gSkipCorrectnessTesting = 0;
 int gStopOnError = 0;
 static bool gSkipRestOfTests;
-int gMeasureTimes = 0;
```
EwanC, line 65: Could we delete `gSimdSize` now too? Can't see anymore uses now the timing code has been removed and it's always hardcoded to 1

mantognini: This is actually part of my next PR. :-)
There are a few other dead globals.

EwanC: nice 😄

EwanC (approved):

PR #1212, opened 2021-04-07
Test api min max - fix printing cl_ulong data type

gwawiork:
There is possibility to display negative numbers by `printf` because of overloading to `int` from `cl_ulong`.

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -1829,7 +1829,7 @@ int test_min_max_constant_args(cl_device_id deviceID, cl_context context,
     log_info(
         "Reported max constant arg count of %llu and max constant buffer "
         "size of %llu. Test will attempt to allocate half of that, or %llu "
-        "buffers of size %z.\n",
+        "buffers of size %zu.\n",
         maxArgs, maxSize, maxArgs, individualBufferSize);
 
     str2 = (char *)malloc(sizeof(char) * 32 * (maxArgs + 2));
@@ -2109,7 +2109,7 @@ int test_min_max_local_mem_size(cl_device_id deviceID, cl_context context,
         * sizeof(cl_int);
     size_t numberOfInts = sizeToAllocate / sizeof(cl_int);
 
-    log_info("Attempting to use %z bytes of local memory.\n", sizeToAllocate);
+    log_info("Attempting to use %zu bytes of local memory.\n", sizeToAllocate);
 
     localData = (cl_int *)malloc(sizeToAllocate);
     d = init_genrand(gRandomSeed);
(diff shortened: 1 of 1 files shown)


EwanC (commented):
Could we use `%zu` rather than `%llu` to print `sizeToAllocate` and `individualBufferSize` since they're `size_t`?

gwawiork:
> Could we use `%zu` rather than `%llu` to print `sizeToAllocate` and `individualBufferSize` since they're `size_t`?

ok fixed. Thanks

EwanC (approved):

pierremoreau:
Shouldn’t it be `%zu` instead of just `%z`? According to the spec, `z` is just a _length modifier_ (which are optional) and not a _conversion format specifier_ which is the mandatory part (in the case of `%zu`, that would be `u`). Maybe `%z` will default to `%zd` which would be for a `signed size_t`?

EwanC (dismissed):

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1465,119 +1791,145 @@ int test_min_max_constant_args(cl_device_id deviceID, cl_context context, cl_com
...
+    test_error(error, "Unable to get max constant buffer size");
     individualBufferSize = (maxSize / 2) / maxArgs;
 
-    log_info("Reported max constant arg count of %d and max constant buffer size of %d. Test will attempt to allocate half of that, or %d buffers of size %d.\n",
-             (int)maxArgs, (int)maxSize, (int)maxArgs, (int)individualBufferSize);
+    log_info(
+        "Reported max constant arg count of %llu and max constant buffer "
+        "size of %llu. Test will attempt to allocate half of that, or %llu "
+        "buffers of size %zu.\n",
+        maxArgs, maxSize, maxArgs, individualBufferSize);
```
bashbaug, line 1833: Minor: maxArgs is a cl_uint, so we'd be OK with just %u instead of %llu for it.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1709,161 +2082,199 @@ int test_min_max_local_mem_size(cl_device_id deviceID, cl_context context, cl_co
...
+    error = clSetKernelArg(kernel, 2, sizeof(streams[1]), &streams[1]);
+    test_error(error, "Unable to set indexed kernel arguments");
 
 
     /* Test running the kernel and verifying it */
     threads[0] = numberOfInts;
     localThreads[0] = 1;
-    log_info("Creating local buffer with %d cl_ints (%d bytes).\n", (int)numberOfInts, (int)sizeToAllocate);
+    log_info("Creating local buffer with %d cl_ints (%zu bytes).\n",
+             (int)numberOfInts, sizeToAllocate);
```
bashbaug, line 2142: numberOfInts is also a size_t, so we may as well print it with %zu while we're at it.

bashbaug (commented):
Mostly LGTM.  Needs a non-Intel reviewer.

EwanC (dismissed):

test_conformance/api/test_api_min_max.cpp
gwawiork: fixed

test_conformance/api/test_api_min_max.cpp
gwawiork: fixed

EwanC (approved):

bashbaug (commented):
Latest version LGTM - thanks!  Already has a non-Intel approval so merging.

PR #1215, opened 2021-04-12
Extended subgroups - use 128bit masks

### LINKED ISSUE #1206: Subgroup tests - modify masks to bitset 128 size to test subgrup sizes max 128
See comment https://github.com/KhronosGroup/OpenCL-CTS/pull/1074#discussion_r603417187

gwawiork:
Partially addresses #1206.


StuartDBrady (changes requested):
The changes generally look good, but there is some duplication that should be removed.

The masks used in `load_masks()` still need to be improved, and the masks should be generated programmatically, but this can be done in a separate PR.

test_conformance/subgroups/test_subgroup_non_uniform_vote.cpp
```diff
@@ -241,21 +262,43 @@ static const char *non_uniform_all_source = R"(
...
+        if(subgroup_local_id < 32) {
+            work_item_mask = work_item_mask_vector.x;
+        } else if(subgroup_local_id < 64) {
+            work_item_mask = work_item_mask_vector.y;
+        } else if(subgroup_local_id < 96) {
+            work_item_mask = work_item_mask_vector.w;
+        } else if(subgroup_local_id < 128) {
+            work_item_mask = work_item_mask_vector.z;
+        }
+        if (elect_work_item & work_item_mask){
```
StuartDBrady, line 299: The same comment regarding the duplication applies here, and in at least 30 other places in this PR. Please address all of them, and not just the ones in this file.

gwawiork: @StuartDBrady I've added changes to avoid code duplication. Please take a look in a free time.

StuartDBrady:
The changes here look okay, but they do not fully fix issue #1206. Please update the description so that it says "Partially fixes issue #1206" and leave the issue open until all of the requested changes have been made.

In particular, I asked: "Please use appropriate bit patterns to ensure reasonable coverage, preferably generated computationally." - this has not yet been acted upon.

The work items mask should be passed to the kernel as a parameter, rather than compiled into the kernel source, and the resulting kernels should then be reused with different masks, reducing the number of compilations performed when running the tests. This can be done in a followup change if needed but should be addressed before #1206 is marked as fixed.

test_conformance/subgroups/test_subgroup_non_uniform_vote.cpp
StuartDBrady: The code is much improved.

Could the description be updated so this change can be merged, please?

StuartDBrady (dismissed):
I've updated the description myself, assuming there is no objection.

It would probably be best to squash this the commits when merging.

kpet:
@gwawiork This should be ready to go after you resolve the merge conflicts.

StuartDBrady (approved):
Suggest applying this as-is, and fixing the remaining aspects of issue #1206 in subsequent PRs.

@kpet, the second commit in this change included some heavy refactoring of code introduced by the first commit - best to squash?

kpet:
This change introduces minor formatting issues but since @gwawiork has been unresponsive for a while, I'll merge it as is to enable further improvements of these tests.

@StuartDBrady Indeed, I typically squash but thanks for the heads up!

PR #1216, opened 2021-04-14
Make variables and functions local to translation unit

mantognini:
Make some global variables local to function, or remove them when actually dead.

@EwanC: As hinted in #1211, this removes `gSimdSize` and others. :-)

### DIFF
--- a/test_conformance/math_brute_force/main.cpp
+++ b/test_conformance/math_brute_force/main.cpp
@@ -99,7 +99,6 @@ cl_mem gOutBuffer[VECTOR_SIZE_COUNT] = { NULL, NULL, NULL, NULL, NULL, NULL };
 cl_mem gOutBuffer2[VECTOR_SIZE_COUNT] = { NULL, NULL, NULL, NULL, NULL, NULL };
 static MTdata gMTdata;
 cl_device_fp_config gFloatCapabilities = 0;
-static cl_device_fp_config gDoubleCapabilities = 0;
 int gWimpyReductionFactor = 32;
 int gWimpyBufferSize = BUFFER_SIZE;
 int gVerboseBruteForce = 0;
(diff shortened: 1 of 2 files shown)


test_conformance/math_brute_force/main.cpp
```diff
@@ -97,28 +97,24 @@ cl_mem gInBuffer2 = NULL;
 cl_mem gInBuffer3 = NULL;
 cl_mem gOutBuffer[VECTOR_SIZE_COUNT] = { NULL, NULL, NULL, NULL, NULL, NULL };
 cl_mem gOutBuffer2[VECTOR_SIZE_COUNT] = { NULL, NULL, NULL, NULL, NULL, NULL };
-uint32_t gComputeDevices = 0;
-uint32_t gSimdSize = 1;
-uint32_t gDeviceFrequency = 0;
 static MTdata gMTdata;
 cl_device_fp_config gFloatCapabilities = 0;
-cl_device_fp_config gDoubleCapabilities = 0;
```
EwanC, line 105: Think we could actually make this function scope where `CL_DEVICE_DOUBLE_FP_CONFIG` is queried

EwanC (commented):
Have a personal preference for using an anonymous namespace over `static` with the globals,
 but don't think that's a convention in any other part of the codebase and would make
 this diff more noisy. So would be a refactor for another day, if we thought it would be an improvement.

mantognini (commented):
> Have a personal preference for using an anonymous namespace over static with the globals,
but don't think that's a convention in any other part of the codebase and would make
this diff more noisy. So would be a refactor for another day, if we thought it would be an improvement.

That would also be my preference as it covers variables, functions and types. If there's a consensus I wouldn't mind using it it future cleanups.

EwanC (approved):

kpet (approved):

PR #1217, opened 2021-04-14
Fold constant global variable

mantognini:
`gWimpyBufferSize` is never modified and is actually not used to modify the number of tests -- `gWimpyReductionFactor` is used for that purpose by some tests, but not all.

This patch removes this unnecessary global variable to simplify the codebase, and reduce differences between tests.


EwanC (approved):
LGTM - nice cleanup

kpet (approved):

PR #1226, opened 2021-04-27
add test for cl_khr_spirv_linkonce_odr

bashbaug:
Adds a basic test for `cl_khr_spirv_linkonce_odr`.

This test compiles two SPIR-V modules and links them together to create the test kernel.

The first SPIR-V module defines two functions, `a` and `b`.  `a` has **LinkOnceODR** linkage, and `b` has **Export** linkage, as usual:

```c
int a(int x) { // LinkOnceODR
    return x + 5;
}

int b(int x) { // Export
    return -x;
}
```

The second SPIR-V module also defines `a` with **LinkOnceODR** linkage, and has an **Import** dependency on `b`.  The `test` kernel calls into both `a` and `b`:

```c
int a(int x) { // LinkOnceODR
    return x + 5;
}

int b(int x); // Import

kernel void test(global int* dst) {
    int id = get_global_id(0);
    dst[id] = a(id) + b(id);
}
```

If the two modules are not linked together properly then the definition of `b` will not be found.

If the **LinkOnceODR** linkage is not recognized properly then there will duplicate definitions of `a`.

If the kernel executes correctly, all elements of the destination buffer will have the value `5`.

NOTE: Assembling these SPIR-V modules requires relatively recent SPIR-V tools with `SPV_KHR_linkonce_odr` support: https://github.com/KhronosGroup/SPIRV-Tools/pull/4161

### DIFF
--- a/test_conformance/spirv_new/spirv_asm/linkage_linkonce_odr_noa_main.spvasm32
+++ b/test_conformance/spirv_new/spirv_asm/linkage_linkonce_odr_noa_main.spvasm32
@@ -12,6 +12,7 @@
                OpDecorate %__spirv_BuiltInGlobalInvocationId BuiltIn GlobalInvocationId
                OpDecorate %__spirv_BuiltInGlobalInvocationId Constant
                OpDecorate %18 FuncParamAttr NoCapture
+               OpDecorate %a LinkageAttributes "a" Import
                OpDecorate %b LinkageAttributes "b" Import
                OpDecorate %__spirv_BuiltInGlobalInvocationId LinkageAttributes "__spirv_BuiltInGlobalInvocationId" Import
        %uint = OpTypeInt 32 0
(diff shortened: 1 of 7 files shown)


kpet (approved):
This looks good but we should also have a test to verify that `LinkOnceODR` actually exports the decorated function. An implementation that just ignores `LinkOnceODR` would pass the test in this PR. Also you might want to rebase this so it runs the latest CI.

bashbaug:
> This looks good but we should also have a test to verify that `LinkOnceODR` actually exports the decorated function. An implementation that just ignores `LinkOnceODR` would pass the test in this PR. Also you might want to rebase this so it runs the latest CI.

Done and done - thanks!

test_conformance/spirv_new/spirv_asm/linkage_linkonce_odr_noa_main.spvasm32
```diff
@@ -0,0 +1,40 @@
...
+               OpCapability Addresses
+               OpCapability Linkage
+               OpCapability Kernel
+               OpExtension "SPV_KHR_linkonce_odr"
+               OpMemoryModel Physical32 OpenCL
+               OpEntryPoint Kernel %17 "test_linkonce_odr" %__spirv_BuiltInGlobalInvocationId
+               OpDecorate %__spirv_BuiltInGlobalInvocationId BuiltIn GlobalInvocationId
+               OpDecorate %__spirv_BuiltInGlobalInvocationId Constant
+               OpDecorate %18 FuncParamAttr NoCapture
+               OpDecorate %b LinkageAttributes "b" Import
```
kpet, line 15: I think you want to use `a` which is exported using `LinkOnceODR` here and not `b` that is exported using `Export`.

bashbaug: Good catch, previously Clang and the SPIR-V LLVM Translator were being too smart and removing the call to `a` completely, which meant the test wasn't actually doing what I was intending for it to be doing.

I've added back the calls to function `a` now in both cases, and added the right `Import` linkage attribute in the `noa` test.  I _think_ this is right now?

bashbaug:
We enabled support for this extension recently.  It would be really great to have some test coverage for it!

kpet (approved):
I have not turned every stone but this now looks to be generally correct after a quick review. It's been open for so long, better to get this in and add any necessary fixes on top.

bashbaug:
Merging as discussed in the February 6th teleconference.

PR #1228, opened 2021-04-27
Reduce scope of variables

mantognini:
Make variables local to loops, with appropriate types. These variables are not read after the loop without being reset first, so this patch doesn't change behaviour.

These variables should now be used for one purpose only, making it easier to reason about the code. This will make future refactoring easier.

### DIFF
--- a/test_conformance/math_brute_force/macro_unary_float.cpp
+++ b/test_conformance/math_brute_force/macro_unary_float.cpp
@@ -150,7 +150,6 @@ int TestMacro_Int_Float(const Func *f, MTdata d, bool relaxedMode)
 {
     TestInfo test_info;
     cl_int error;
-    size_t i, j;
 
     logFunctionInfo(f->name, sizeof(cl_float), relaxedMode);
 
@@ -178,7 +177,7 @@ int TestMacro_Int_Float(const Func *f, MTdata d, bool relaxedMode)
 
     // cl_kernels aren't thread safe, so we make one for each vector size for
     // every thread
-    for (i = gMinVectorSizeIndex; i < gMaxVectorSizeIndex; i++)
+    for (auto i = gMinVectorSizeIndex; i < gMaxVectorSizeIndex; i++)
     {
         size_t array_size = test_info.threadCount * sizeof(cl_kernel);
         test_info.k[i] = (cl_kernel *)malloc(array_size);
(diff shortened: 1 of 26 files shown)


test_conformance/math_brute_force/unary_two_results_float.cpp
```diff
@@ -124,8 +124,6 @@ static cl_int BuildKernelFn(cl_uint job_id, cl_uint thread_id UNUSED, void *p)
 
 int TestFunc_Float2_Float(const Func *f, MTdata d, bool relaxedMode)
 {
-    uint64_t i;
-    uint32_t j, k;
     uint32_t l;
```
EwanC, line 127: Is this `l` variable used anywhere? Can't see anything from a quick scan and wondering if it can be removed

test_conformance/math_brute_force/macro_unary_float.cpp
```diff
@@ -456,7 +454,8 @@ static cl_int Test(cl_uint job_id, cl_uint thread_id, void *data)
             }
 
 
-            for (k = MAX(1, gMinVectorSizeIndex); k < gMaxVectorSizeIndex; k++)
```
EwanC, line 459: Looks like this is variable shadowing the `k` in the outer loop, don't think it makes a difference for correctness but naming the two loop induction variables differently might be more readable

EwanC (approved):
LGTM, some minor comments

test_conformance/math_brute_force/macro_unary_float.cpp
mantognini: Nice observation! I have to admit I had to resort to some automation for this patch and didn't look at this aspect. I'm actually tempted to keep the same name for now because there's a bigger issue with the two loops: there should be only one, most likely. I think other files might be affected too. I'll make a note to investigate this further.

test_conformance/math_brute_force/unary_two_results_float.cpp
mantognini: Indeed, it's not. There are a few others, like `testing_mode` a few lines below. I've actually got pending patches to address unused variables found by Clang's `-Wunused` warning flag.

mantognini (commented):
Thanks for your review.

test_conformance/math_brute_force/macro_unary_float.cpp
EwanC: sounds good, thanks

test_conformance/math_brute_force/unary_two_results_float.cpp
EwanC: cool, should've guessed you'd have a pending patch for this 😄

kpet (approved):

Issue #1222, opened 2021-04-22
platform can be NULL, when querying platform info by clGetPlatformInfo()


lengleng1992:
https://github.com/KhronosGroup/OpenCL-CTS/blob/b8d1ea9962a7bb443e845f5f035b254172fc4ae7/test_conformance/api/negative_platform.cpp#L49

but in spec , The description is as follows:
platform refers to the platform ID returned by clGetPlatformIDs or can be NULL. If platform is
NULL, the behavior is implementation-defined.

If platform is NULL, implementation may call clGetPlatformIDs or internal func to get platform ID.

Therefore, CTS may should not check the return error here.

Similarly, platform in clGetDeviceIDs can also be nullptr.

Thanks!

chemis01:
There is ambiguity in the OpenCL spec which has led to this situation.
While I agree it says that a "NULL Platform is implementation defined", it still specifies that "CL_INVALID_PLATFORM should be returned when platform is not valid"
This then raises the question of what defines a "valid" object which is being discussed https://github.com/KhronosGroup/OpenCL-Docs/issues/482

kpet:
@chemis01 NULL platform is a special case and the CTS is currently wrong: it should not expect an error code when a NULL platform is passed. Could you prepare a patch to fix this please? (see https://github.com/KhronosGroup/OpenCL-CTS/pull/1148#discussion_r573756540 where I reported this as well)

kpet:
Fixed by https://github.com/KhronosGroup/OpenCL-CTS/pull/1229

PR #1229, opened 2021-04-27
Remove NULL Platform Check for Negative clGetPlatformInfo (#1222)

chemis01:
Part of the negative CTS test for clGetPlatformInfo
incorrectly tested with a NULL platform. As this is a
special case it meant that the test was not correct.
This change removes the incorrect part of the test so that
it conforms with the OpenCL specification.


gwawiork (approved):
LGTM

kpet (approved):
Thanks!

PR #1232, opened 2021-05-02
add tests for cl_khr_extended_bit_ops

bashbaug:
Adds tests for `cl_khr_extended_bit_ops`.

The tests are added to the `integer_ops` test.  There are three new tests:

* `extended_bit_ops_extract`: tests signed and unsigned bit extract
* `extended_bit_ops_insert`: tests bit insert
* `extended_bit_ops_reverse`: tests bit reverse


test_conformance/integer_ops/test_extended_bit_ops_extract.cpp
```diff
@@ -0,0 +1,278 @@
...
+        result = 0;
+    }
+    else if (offset + count < sizeof(T) * 8)
+    {
+        result = base << (sizeof(T) * 8 - offset - count);
+        result = result >> (sizeof(T) * 8 - count);
+    }
+    else
+    {
+        result = base >> offset;
```
StuartDBrady, line 81: If we were to rewrite this as:

```
    if (count == 0)
    {
        result = 0;
    }
    else
    {
        result = base << (sizeof(T) * 8 - offset - count);
        result = result >> (sizeof(T) * 8 - count);
    }
```

and then consider the case where `count != 0` and `!(offset + count == sizeof(T) * 8)` (which would hit the `else` case in the original):

* Given `count != 0`, we know that `sizeof(T) * 8 - offset - count < sizeof(T) * 8`.  The `<<` operation therefore cannot overflow.
* Given `offset + count <= sizeof(T) * 8)`, and given `!(offset + count < sizeof(T) * 8)`, we know `offset + count == sizeof(T) * 8`.
* Given `offset + count == sizeof(T) * 8`, we know `sizeof(T) * 8 - offset - count == 0`.
* `result = base << (sizeof(T) * 8 - offset - count)` would therefore simplify to `result = base << 0`, which can be eliminated.
* Given `offset + count == sizeof(T) * 8`, `sizeof(T) * 8 - count == offset`.
* `base >> (sizeof(T) * 8 - count)` would therefore simplify to `base >> offset`.

In short, this is in fact not a special case, but is an unnecessary substitution of the general case.

Please remove the `else` case, and remove the condition of the `else if`.

test_conformance/integer_ops/test_extended_bit_ops_insert.cpp
```diff
@@ -0,0 +1,207 @@
...
+    assert(count <= sizeof(T) * 8);
+    assert(offset + count <= sizeof(T) * 8);
+
+    cl_ulong base = static_cast<cl_ulong>(tbase);
+    cl_ulong insert = static_cast<cl_ulong>(tinsert);
+
+    cl_ulong mask = (count < 64) ? ((1ULL << count) - 1) << offset : ~0ULL;
+    cl_ulong result = ((insert << offset) & mask) | (base & ~mask);
+
+    return static_cast<T>(result);
```
StuartDBrady, line 39: This should be rewritten so as not to rely on implementation-defined behavior for negative values.

StuartDBrady (changes requested):
This generally looks good, aside from some problems with implementation-defined and undefined behavior regarding bitshifts, and a small amount of overcomplication.

Note also that the change needs rebasing.

test_conformance/integer_ops/test_extended_bit_ops_extract.cpp
```diff
@@ -0,0 +1,278 @@
...
+    signed_t result;
+
+    if (count == 0)
+    {
+        result = 0;
+    }
+    else if (offset + count < sizeof(T) * 8)
+    {
+        result = base << (sizeof(T) * 8 - offset - count);
+        result = result >> (sizeof(T) * 8 - count);
```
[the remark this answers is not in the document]
bashbaug: Fixed, though please note that I ended up computing all of the reference results as unsigned types which eliminated the need for casting to a signed type completely.

test_conformance/integer_ops/test_extended_bit_ops_extract.cpp
```diff
@@ -0,0 +1,278 @@
...
+    }
+    else
+    {
+        result = base >> offset;
```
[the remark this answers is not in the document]
bashbaug: Fixed - thanks!

test_conformance/integer_ops/test_extended_bit_ops_extract.cpp
bashbaug: Fixed - thanks!

test_conformance/integer_ops/test_extended_bit_ops_insert.cpp
bashbaug: Fixed by computing the reference results as unsigned types.

bashbaug:
Hi @StuartDBrady , I believe I have addressed all of your review comments.  Sorry for the long delay!  Could you please take a look at the latest code?  Thanks!

lakshmih:

Sreelakshmi Haridas Maruthur
Qualcomm

test_conformance/integer_ops/test_extended_bit_ops_extract.cpp
```diff
@@ -0,0 +1,288 @@
...
+static int test_vectype(cl_device_id device, cl_context context,
+                        cl_command_queue queue)
+{
+    // Because converting from an unsigned type to a signed type is
+    // implementation-defined if the most significant bit is set until C++ 20,
+    // compute all reference results using unsigned types.
+    typedef typename std::make_unsigned<T>::type unsigned_t;
+
+    cl_int error = CL_SUCCESS;
+    int result = TEST_PASS;
```
lakshmih, line 157: This is causing a build failure due to result being unused.

lakshmih (commented):
Currently the reference data is all random.  Suggest adding some non-random test points that are possible corner cases.  Such as 1, -1, 0, and maybe max values.

Sreelakshmi Haridas Maruthur
Qualcomm

test_conformance/integer_ops/test_extended_bit_ops_extract.cpp
bashbaug: Thanks, fixed!  I don't think this warning was enabled when I initially made these changes.

bashbaug:
Ping @StuartDBrady - can you please check if your review comments have been addressed?  Thanks!

bashbaug:
> Currently the reference data is all random. Suggest adding some non-random test points that are possible corner cases. Such as 1, -1, 0, and maybe max values.

Random data is currently used in three places:

1. For the data being extracted by bitfield_extract() ("base").  The offsets and counts are all tested (no randomness).
2. For the data being inserted by bitfield_insert() ("insert").  The data being inserted into is fixed ("base"), and again the offsets and counts are all tested (no randomness).
3. For the data being reversed by bit_reverse().

I think this gives pretty good covereage.  Which data do you suggest making non-random?

test_conformance/integer_ops/test_extended_bit_ops_extract.cpp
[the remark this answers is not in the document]
StuartDBrady: Looks good to me.

StuartDBrady:
The changes look good to me.  Leaving for Qualcomm to finish the review as per their earlier questions about edge cases.

test_conformance/integer_ops/test_extended_bit_ops_insert.cpp
StuartDBrady: Looks good to me.

lakshmih (approved):
Looks good.

Sreelakshmi Haridas Maruthur
Qualcomm

bashbaug:

test_conformance/integer_ops/test_extended_bit_ops_extract.cpp
```diff
@@ -0,0 +1,287 @@
...
+    unsigned_t result = (s ^ x) >> count ^ s;
+
+    return result;
+}
+
+template <typename T>
+static typename std::make_unsigned<T>::type
+cpu_bit_extract_signed(T tbase, cl_uint offset, cl_uint count)
+{
+    typedef typename std::make_signed<T>::type unsigned_t;
```
haonanya, line 47: Hi, @bashbaug , std::make_signed<T>::type does not reflect the literal meaning of unsigned_t, is there just a typo? Thanks very much for your comments.

bashbaug: Let's move discussion to https://github.com/KhronosGroup/OpenCL-CTS/pull/1913.

It's been a while but I think this is a typo and I intended to use unsigned types exclusively.

PR #1237, opened 2021-05-06
Remove unnecessary code from unary_u_*.cpp

mantognini:
Only nan() is tested by unary_u_float.cpp and unary_u_double.cpp. Testing of half_sin, half_tan and half_cos is done in unary_float.cpp and unary_double.cpp.

----

This was spotted because Clang complained `fabsf(s[j])` was a no-op (`-Wabsolute-value`).


EwanC (approved):
Good catch

PR #1238, opened 2021-05-06
Remove dead variables and functions

mantognini:
These were identified using Clang's -Wunused warning flag.

@EwanC: those are the remaining dead variables spotted while working on #1228.


EwanC (approved):
Nice one

kpet (approved):

PR #1245, opened 2021-05-12
Allocations fixes

pierremoreau:
* The buffer_non_blocking test was being wrongly skipped if images were not supported by the device.
* Fix a few typos in the test.


EwanC (approved):

test_conformance/allocations/main.cpp
```diff
@@ -112,6 +112,8 @@ int doTest( cl_device_id device, cl_context context, cl_command_queue queue, All
     int number_of_mems_used;
     cl_ulong max_individual_allocation_size = g_max_individual_allocation_size;
     cl_ulong global_mem_size = g_global_mem_size ;
+    const bool allocate_image =
```
kpet, line 115: Not very intuitive but correct AFAICT.

kpet (approved):

PR #1247, opened 2021-05-13
Fix image pixel reference calculation for CL_{INTENSITY,LUMINANCE} formats

kpet:
As per 6.15.15.7, the first three components have to be set to the luminance value and all components to the intensity value.

### DIFF
--- a/test_common/harness/imageHelpers.h
+++ b/test_common/harness/imageHelpers.h
@@ -489,12 +489,14 @@ void read_image_pixel(void *imageData, image_descriptor *imageInfo, int x,
     }
     else if (format->image_channel_order == CL_INTENSITY)
     {
+        outData[0] = tempData[0];
         outData[1] = tempData[0];
         outData[2] = tempData[0];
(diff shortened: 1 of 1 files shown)


test_common/harness/imageHelpers.h
```diff
@@ -495,6 +495,7 @@ void read_image_pixel(void *imageData, image_descriptor *imageInfo, int x,
     }
     else if (format->image_channel_order == CL_LUMINANCE)
     {
+        outData[0] = tempData[0];
```
EwanC, line 498: Do we need the same change for setting `outData[0]` in the  `CL_INTENSITY` case too? Haven't worked with images much, so might be reading things wrong.

kpet: Yup, well spotted, it was broken too. I've updated the PR to fix that one too.

EwanC (approved):

bashbaug (approved):
That's a little embarassing!  LGTM.

Issue #1234, opened 2021-05-04
Using clCloneKernel on a kernel with zero arguments is not tested


einhov:
This edge case currently has no coverage.
Example kernel: `__kernel void test_kernel() {}`

jainvikas8:
@kpet Please assign this issue to me. Thanks.

rjodinchr:
Fixed by PR #2577 (commit `1bf15290`), which added the `clone_kernel_with_no_args` conformance test in `test_conformance/api/test_clone_kernel.cpp` to verify `clCloneKernel` on kernels with zero arguments.

PR #1251, opened 2021-05-14
Extend test coverage for `clClonekernel`

### LINKED ISSUE #1098: Additional test-coverage for clCloneKernel to ensure updates to cloned kernel args do not affect original kernel args and vice-a-versa
As per the spec, clSetKernelArg on the cloned or original kernel after clCloneKernel shouldn't affect the arguments of other kernel. 
In current CTS, clSetKernelArg before and after clCloneKernel, but that is happening on disjoint sets of arguments and also the original kernel is not being used anywhere after it was cloned.
This issue tracks adding following coverage for test_clone_kernel -

1. clSetKernelArg after clCloneKernel for arguments that were set before calling clCloneKernel to ensure subsequent kernel launches use appropriate arguments for original as well as cloned kernel.

srcKenel = clCreateKernel(program, "xyz);
clSetKernelArg(srcKernel, 0, mem1);
clonedKernel = clCloneKernel(srcKernel);
clSetKernelArg(srcKernel, 0, mem2);
clEnqueueNDRangeKernel(..., clonedKernel, ...)
// Validate results for mem1
clEnqueueNDRangeKernel(..., srcKernel, ...)
// Validate results for mem2
clSetKernelArg(clonedKernel, 0, mem3);
clEnqueueNDRageKernel(..., clonedKernel, ...)
// validate results for mem3

Similar set of test-cases would be needed for other APIs updating arg  clSetKernelArgSVMPointer  and  clSetKernelExecInfo.

jainvikas8:
Add the following test coverage for the `clCloneKernel`

With args:
Use `clSetKernelArg` to set args after kernel is cloned. Enqueue and
read the buffer to validate.

With Execinfo:
Use `clSetKernelExecInfo` after the kernel is cloned and read the buffer
to validate.

With no args:
Clone a kernel with no args and enqueue.

With SVM pointer:
Use `clSetKernelArgSVMPointer` to set args after kernel is cloned.
Enqueue and read the buffer to validate.

This fixes #1098 and #1234 

This PR also replaces `test_error` with `test_assert_error` to report the error during failure.


jainvikas8:
I'm seeing
```
fatal: unable to access 'https://github.com/KhronosGroup/libclcxx.git/': Could not resolve host: github.com
Command exited with code 128
```

How can we restart this?

jainvikas8:
@gwawiork, please review.

mantognini:
> I'm seeing
> 
> ```
> fatal: unable to access 'https://github.com/KhronosGroup/libclcxx.git/': Could not resolve host: github.com
> Command exited with code 128
> ```
> 
> How can we restart this?

I had the same issue on another PR. I ended up re-pushing (on the same branch) to retrigger another run. Hopefully there's a more clever alternative with a simple button to press somewhere, but I couldn't find it.

jainvikas8:
Forced pushed to rebase.

jainvikas8:
Forced pushed again, CI was stuck

gwawiork:
Hi @jainvikas8 
I've check your changes on Windows on Intel GPU  and I am getting 
`ERROR: clFinish failed! (CL_OUT_OF_RESOURCES from C:\NewVPG\Repo\public-github\opensource-qb\OpenCL-CTS\test_conformance\api\test_clone_kernel.cpp:524)
test_exec_enqueue_helper failed for cloneKernel_2 on retry.
clCloneKernel test_cloned_kernel_exec_info failed.
clone_kernel FAILED`

The test process is not able to exit looks like hang. Could you double check on your side if everything is ok ? I will analyze the problem as well.

jainvikas8:
Forced pushed to rebase.

jainvikas8:
> Hi @jainvikas8
> I've check your changes on Windows on Intel GPU and I am getting
> `ERROR: clFinish failed! (CL_OUT_OF_RESOURCES from C:\NewVPG\Repo\public-github\opensource-qb\OpenCL-CTS\test_conformance\api\test_clone_kernel.cpp:524) test_exec_enqueue_helper failed for cloneKernel_2 on retry. clCloneKernel test_cloned_kernel_exec_info failed. clone_kernel FAILED`
> 
> The test process is not able to exit looks like hang. Could you double check on your side if everything is ok ? I will analyze the problem as well.

Have not observed this in the past, normally it should fail for this too.
```  test_exec_enqueue_helper failed for cloneKernel_1 on the retry ```
Will look into reproducing this issue. Thanks.

File: test_conformance/api/test_clone_kernel.cpp [hunk 1]
```diff
@@ -351,51 +466,459 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
+    clKernelWrapper srcKernel;
+
+    cl_int intargs[] = { 1, 2, 3, 4 };
+    cl_device_svm_capabilities svmCaps = 0;
+
+    error = clGetDeviceInfo(deviceID, CL_DEVICE_SVM_CAPABILITIES,
+                            sizeof(svmCaps), &svmCaps, NULL);
+    test_error(error, "Unable to query CL_DEVICE_SVM_CAPABILITIES");
+
+    if (svmCaps != 0)
```
gwawiork, line 596: I think we should return TEST_SKIPPED_ITSELF in case svmCaps == 0 now there is TEST_PASS

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -351,51 +466,459 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
```
jainvikas8, line 809: Will return `TEST_SKIPPED` if svmCaps == 0

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -351,51 +466,459 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
+    }
+
+    return TEST_PASS;
+}
+
+int test_clone_kernel(cl_device_id deviceID, cl_context context,
+                      cl_command_queue queue, int num_elements)
+{
+    if (test_buff_image_multiple_args(deviceID, context, queue, num_elements)
+        != 0)
```
jainvikas8, line 896: Check for `TEST_FAIL` as test can be
```
    TEST_PASS = 0,
    TEST_FAIL = 1,
    TEST_SKIP = 2,
    TEST_SKIPPED_ITSELF = -100,
```

jainvikas8:
Forced pushed to rebase and add 2 new commits to address the review comments.

jainvikas8:
> > Hi @jainvikas8
> > I've check your changes on Windows on Intel GPU and I am getting
> > `ERROR: clFinish failed! (CL_OUT_OF_RESOURCES from C:\NewVPG\Repo\public-github\opensource-qb\OpenCL-CTS\test_conformance\api\test_clone_kernel.cpp:524) test_exec_enqueue_helper failed for cloneKernel_2 on retry. clCloneKernel test_cloned_kernel_exec_info failed. clone_kernel FAILED`
> > The test process is not able to exit looks like hang. Could you double check on your side if everything is ok ? I will analyze the problem as well.
> 
> Have not observed this in the past, normally it should fail for this too.
> ` test_exec_enqueue_helper failed for cloneKernel_1 on the retry`
> Will look into reproducing this issue. Thanks.

@gwawiork I've been unable to reproduce this issue on MALI-GPU in Linux with OpenCL 2.1 and 3.0. So, please let me know your thoughts on this?

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -351,51 +466,468 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
+                                      &srcKernel, &intargs[3])
+            != 0)
+        {
+            test_fail("test_svm_exec_info_helper failed for srcKernel with "
+                      "different values.\n");
+        }
+        clSVMFree(context, svmPtr_srcKernel_1);
+
+        // enqueue - cloneKernel_1 again, to check if the args were not modified
+        if (test_exec_enqueue_helper(context, queue, pBuf, svmPtr_cloneKernel_1,
```
ddabek-i, line 661: From this point on, `pBuf` still contains pointer to `svmPtr_srcKernel_1`.
Since `test_exec_enqueue_helper` does not update the `pBuf`, the `cloneKernel_1` will receive an invalid pointer as argument.

The pointer inside `pBuf` needs to be updated before enqueuing the kernel.

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -351,51 +466,468 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
+                              BufPtr* pBuf, cl_int* svmPtr_Kernel,
+                              cl_kernel* srcKernel, cl_int* value)
+{
+    cl_int error;
+
+    error = clEnqueueSVMMap(queue, CL_TRUE, CL_MAP_READ | CL_MAP_WRITE, pBuf,
+                            sizeof(BufPtr), 0, NULL, NULL);
+    test_error(error, "clEnqueueSVMMap failed");
+    error = clFinish(queue);
+    test_error(error, "clFinish failed");
```
ddabek-i, line 550: This `clFinish` should not be needed, since `clEnqueueSVMMap` is made blocking by passing it `CL_TRUE` as the second argument.

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -351,51 +466,468 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
+    if (svmCaps != 0)
     {
-        test_error( error, "clCloneKernel test failed. Failed to clone structure type argument." );
-        return -1;
+        error = create_single_kernel_helper(context, &program, &srcKernel, 1,
+                                            clone_kernel_test_kernel,
+                                            "set_kernel_exec_info_kernel");
+        test_error(error, "Unable to create srcKernel");
+
+        BufPtr* pBuf =
```
ddabek-i, line 603: It looks like the same structure `pBuf` will be passed to each kernel (as a pointer) and only the `pBuf->store` will be updated before enqueuing each kernel. Not sure if this is intended?

ddabek-i:
@jainvikas8 Please take a look at my review when you have the time.

test_conformance/api/test_clone_kernel.cpp
jainvikas8: I have moved `clSVMFree` to free the buffers in the end.

test_conformance/api/test_clone_kernel.cpp
jainvikas8: done

test_conformance/api/test_clone_kernel.cpp
jainvikas8: It is indeed intended. We use it to test clSetKernelexecinfo with clCloneKernel

lakshmih:
[Qualcomm] We tested this and it passed on our 3.0 implementation. Will need more time for a detailed review.

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -233,16 +236,128 @@ int test_double_arg_clone(cl_device_id deviceID, cl_context context, cl_command_
...
-    if (abs(((cl_double*)pbufRes)[2] - d) > 0.0000001)
+    test_assert_error(abs(((cl_double*)pbufRes)[2] - d) <= 0.0000001,
+                      "clCloneKernel test failed.");
+
+    return TEST_PASS;
+}
+
+int test_args_enqueue_helper(cl_context context, cl_command_queue queue,
+                             cl_kernel* srcKernel, cl_int* value,
+                             cl_mem* bufOut)
```
bashbaug, line 247: Minor: I think we should pass the kernel, the value, and the buffer by value, rather than as a pointer.

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -351,51 +466,467 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
+    error = clFinish(queue);
+    test_error(error, "clFinish failed");
+
+    error = clSetKernelArg(*srcKernel, 0, sizeof(cl_int), value);
+    test_error(error, "clSetKernelArg failed");
+    error = clSetKernelArgSVMPointer(*srcKernel, 1, pBuf);
+    test_error(error, "clSetKernelArgSVMPointer failed");
+    error = clSetKernelExecInfo(*srcKernel, CL_KERNEL_EXEC_INFO_SVM_PTRS,
+                                sizeof(pBuf), pBuf);
+    test_error(error, "clSetKernelExecInfo failed");
```
bashbaug, line 563: This doesn't quite look correct to me.  pBuf is already set as an SVM pointer kernel argument, so it does not also need to be passed via clSetKernelExecInfo, but svmPtr_Kernel does.  Can you please check that we shouldn't be passing svmPtr_Kernel to clSetKernelExecInfo instead of pBuf?

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -351,51 +466,467 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
+
+    delete[] pbuf;
+    delete[] pbufRes;
+
+    return TEST_PASS;
+}
+
+int test_exec_enqueue_helper(cl_context context, cl_command_queue queue,
+                             BufPtr* pBuf, cl_int* svmPtr_Kernel,
+                             cl_kernel* srcKernel, cl_int* value)
```
bashbaug, line 513: See above - I think we can pass most (if not all) of these by value.

Do we need to pass pBuf at all?

Minor: I'd also consider renaming this function to something like test_SVM_enqueue_helper, since this helper could work for other SVM tests and not just tests for clSetKernelExecInfo.

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -351,51 +466,467 @@ int test_clone_kernel(cl_device_id deviceID, cl_context context, cl_command_queu
...
+        test_fail("test_empty_enqueue_helper failed for srcKernel on retry.\n");
     }
 
-    if (bdouble)
+    return TEST_PASS;
+}
+
+int test_svm_enqueue_helper(cl_context context, cl_command_queue queue,
+                            cl_int* svmPtr_Kernel, cl_kernel* srcKernel,
+                            cl_int* value)
```
bashbaug, line 744: See above - do we need another helper here, or can we reuse the previous helper from clSetKernelExecInfo?

bashbaug (changes requested):
This is a nice start but there are a few nice-to-have fixes and please check that the call to clSetKernelExecInfo is correct.

I also see there is a fair amount of code duplication between the SVM and SVM kernel exec info tests - have you considered any ways to reduce this code duplication?

Thanks!

bashbaug:
Removing "focused review" until this PR has an updated owner.

test_conformance/api/test_clone_kernel.cpp
ahesham-arm: The call to `clSetKernelExecInfo` is not needed here, removed.

ahesham-arm:
@bashbaug Addressed outstanding review comments.

PR #1255, opened 2021-05-17
Fix ODR violations in math_brute_force

mantognini:
A program having a type (such as ThreadInfo) defined differently in multiple translation units exhibits undefined behaviour.

This commit fixes such issues in the math_brute_force component by ensuring most types are local to their translation unit with the help of anonymous namespaces. Later refactoring will be able to extract common definitions to a single place.

This patch also removes unnecessary static and typedef keywords. Otherwise, code is only moved around with no change.

---

I appreciate this is rather large patch but it's very mechanical as well.

### DIFF
--- a/test_conformance/math_brute_force/binary_double.cpp
+++ b/test_conformance/math_brute_force/binary_double.cpp
@@ -20,10 +20,12 @@
 
 #include <cstring>
 
+namespace {
+
 const double twoToMinus1022 = MAKE_HEX_DOUBLE(0x1p-1022, 1, -1022);
 
-static int BuildKernel(const char *name, int vectorSize, cl_uint kernel_count,
-                       cl_kernel *k, cl_program *p, bool relaxedMode)
+int BuildKernel(const char *name, int vectorSize, cl_uint kernel_count,
+                cl_kernel *k, cl_program *p, bool relaxedMode)
 {
     const char *c[] = { "#pragma OPENCL EXTENSION cl_khr_fp64 : enable\n",
                         "__kernel void math_kernel",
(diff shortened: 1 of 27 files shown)


test_conformance/math_brute_force/binary_double.cpp
```diff
@@ -20,10 +20,12 @@
 
 #include <cstring>
 
-const double twoToMinus1022 = MAKE_HEX_DOUBLE(0x1p-1022, 1, -1022);
+namespace {
 
-static int BuildKernel(const char *name, int vectorSize, cl_uint kernel_count,
```
EwanC, line 25: In one of the [CI runs](https://ci.appveyor.com/project/Khronoswebmaster/opencl-cts/builds/39199261/job/rs6ku4phah8he58q) on windows seeing

```
 C:\projects\opencl-cts\test_conformance\math_brute_force\binary_double.cpp(25): error C2131: expression did not evaluate to a constant 
```

and for `binary_float.cpp(25)`

mantognini: Yes, indeed. I reverted the const to constexpr change. diff here: https://github.com/KhronosGroup/OpenCL-CTS/compare/b3f457d13e5801371403f1039ddac75d655f3e1b..415c2482cc53bc5973db4de148b5f363c78084e3

EwanC (approved):

test_conformance/math_brute_force/binary_double.cpp
EwanC: Thanks, change looked good otherwise

PR #1258, opened 2021-05-18
Fixes incorrect slice pitch calculation in clCopyImage 1Darray

### LINKED ISSUE #1257: Images clCopyImage 1Darray incorrect slice pitch calculation
For IMAGE1D_ARRAY, the copy region is set up as [imagewidth, arraysize, 1]. This is done in test_copy_generic.cpp here:
160: case CL_MEM_OBJECT_IMAGE1D_ARRAY:
161:     height = imageInfo->arraysize;
162:     depth = 1;

182: size_t region[ 3 ] = { imageInfo->width, height, depth };

However, the slice pitch/padding calculation assumes that the 'height' variable contains the pixel height of the image, which it doesn't for IMAGE1D_ARRAY.

229: size_t mappedSlicePad = mappedSlice - (mappedRow * height);

paulfradgley:
The slice pitch/padding calculation assumed that the 'height' variable contained the pixel height of the image, which it doesn't for IMAGE1D_ARRAY.
Fixes #1257

### DIFF
--- a/test_conformance/images/clCopyImage/test_copy_generic.cpp
+++ b/test_conformance/images/clCopyImage/test_copy_generic.cpp
@@ -228,6 +228,11 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
         }
         size_t mappedSlicePad = mappedSlice - (mappedRow * height);
 
+        // For 1Darray, the height variable actually contains the arraysize,
+        // so it can't be used for calculating the slice padding.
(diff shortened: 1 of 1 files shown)


lakshmih (commented):
Qualcomm tested this change and it passes on our implementation. We are okay with either the change as is or the fix with edits proposed by @jlewis-austin

jlewis-austin (approved):

bashbaug (approved):
Merging as agreed in the May 31 teleconference.

PR #1264, opened 2021-05-25
Add tests for entrypoint cl_khr_suggested_local_work_size

kalchr01:
Tests added within test_conformance/workgroups. The tests cover several
shapes (num dimensions) and sizes of global work size, kernels using
local memory (dynamic and static) and present/non-present global work
offset.

Signed-off-by: Kallia Chronaki <kallia.chronaki@arm.com>

### DIFF
--- a/test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
+++ b/test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
@@ -51,7 +51,7 @@ const char* wg_scan_local_work_group_size = R"(
         return linear_id == 0;
     }
 
-    size_t get_l_size(size_t dim)
+    uint get_l_size(size_t dim)
     {
 #if __OPENCL_VERSION__ < CL_VERSION_2_0
         return get_local_size(dim);
@@ -60,7 +60,7 @@ const char* wg_scan_local_work_group_size = R"(
 #endif
     }
 
-    __kernel void test_wg_scan_local_work_group_size(global size_t *output)
+    __kernel void test_wg_scan_local_work_group_size(global uint *output)
     {
         if(!is_zero_linear_id()) return;
         for (uint i = 0; i < 3; i++)
@@ -69,7 +69,7 @@ const char* wg_scan_local_work_group_size = R"(
         }
     }
     __kernel void test_wg_scan_local_work_group_size_static_local(
-                                            global size_t *output)
+                                            global uint *output)
     {
         __local char c[LOCAL_MEM_SIZE];
     
@@ -80,7 +80,7 @@ const char* wg_scan_local_work_group_size = R"(
         }
     }
     __kernel void test_wg_scan_local_work_group_size_dynlocal(
-                                        global size_t *output,
+                                        global uint *output,
                                         __local char * c)
     {
         if(!is_zero_linear_id()) return;
@@ -111,10 +111,17 @@ bool is_not_odd(size_t a) { return (is_prime(a) || (a % 2 == 0)); }
 /* The numbers we chose in the value_range are to be used for the second and
    third dimension of the global work group size. The numbers below cover many
    different cases: 1024 is a power of 2, 3 is an odd and small prime number, 12
-   is a power of 4 but not a power of 2 1025 is a large odd number, 4217 is the
-   first prime number that is greater than CL_DEVICE_MAX_WORK_GROUP_SIZE and 1
-   is to test the lack of this dimension if the others are present */
-const size_t value_range[] = { 1024, 3, 12, 16, 1025, 4217, 1 };
+   is a multiple of 4 but not a power of 2, 1031 is a large odd and prime number
+   and 1 is to test the lack of this dimension if the others are present */
+const size_t value_range[] = { 1024, 3, 12, 1031, 1 };
+/* The value_range_nD contains numbers to be used for the experiments with 2D
+   and 3D global work sizes. This is because we need smaller numbers so that the
+   resulting number of work items is meaningful and does not become too large.
+   The cases here are: 64 that is a power of 2, 3 is an odd and small prime
+   number, 12 is a multiple of 4 but not a power of 2, 113 is a large prime
+   number
+   and 1 is to test the lack of this dimension if the others are present */
+const size_t value_range_nD[] = { 64, 3, 12, 113, 1 };
 const size_t basic_increment = 16;
 const size_t primes_increment = 1;
 enum num_dims
@@ -131,7 +138,7 @@ int do_test(cl_device_id device, cl_context context, cl_command_queue queue,
     size_t local_work_size[] = { 1, 1, 1 };
     size_t suggested_total_size;
     size_t workgroupinfo_size;
-    size_t kernel_work_size[3] = { 0 };
+    cl_uint kernel_work_size[3] = { 0 };
     clMemWrapper buffer;
     cl_platform_id platform;
 
@@ -146,8 +153,8 @@ int do_test(cl_device_id device, cl_context context, cl_command_queue queue,
 
     if (clGetKernelSuggestedLocalWorkSizeKHR == NULL)
     {
-        log_info(
-            "Extension 'cl_khr_suggested_local_work_size' could not be found.\n");
+        log_info("Extension 'cl_khr_suggested_local_work_size' could not be "
+                 "found.\n");
         return TEST_FAIL;
     }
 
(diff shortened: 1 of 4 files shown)


test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
```diff
@@ -0,0 +1,626 @@
...
+            (clGetKernelSuggestedLocalWorkSizeKHR_fn)
+                clGetExtensionFunctionAddressForPlatform(
+                    platform, "clGetKernelSuggestedLocalWorkSizeKHR");
+
+    if (clGetKernelSuggestedLocalWorkSizeKHR == NULL)
+    {
+        log_info(
+            "Extension 'cl_khr_suggested_local_work_size' could not be found. "
+            "Skipping the test.\n");
+        return TEST_SKIPPED_ITSELF;
```
bashbaug, line 152: I think this case should be an error.  Are there valid circumstances where the extension is supported but `clGetExtensionFunctionAddressForPlatform` returns a NULL function pointer?

test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+    err = clGetKernelSuggestedLocalWorkSizeKHR(queue, scan_kernel, work_dim,
+                                               global_work_offset, test_values,
+                                               local_work_size);
```
bashbaug, line 157: We need to set kernel arguments before calling `clGetKernelSuggestedLocalWorkSize`.  See the error condition:

> CL_INVALID_KERNEL_ARGS if all argument values for kernel have not been set.

(For some devices this may affect what gets returned by `clGetKernelWorkGroupInfo`, also.)

test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
```diff
@@ -0,0 +1,626 @@
...
+    clProgramWrapper scan1kb_program;
+    clKernelWrapper scan1kb_kernel;
+    clProgramWrapper scan4kb_program;
+    clKernelWrapper scan4kb_kernel;
+    clProgramWrapper scan16kb_program;
+    clKernelWrapper scan16kb_kernel;
+    clProgramWrapper scan32kb_program;
+    clKernelWrapper scan32kb_kernel;
+    clProgramWrapper scan_dyn_program;
+    clKernelWrapper scan_dyn_kernel;
```
bashbaug, line 235: This probably isn't the end of the world, but there's a fair amount of copy-paste here.  It it worth moving this into a loop instead, to reduce duplication?

bashbaug (changes requested):
Hi, this is a great start but I found a few issues, thanks!

test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
```diff
@@ -0,0 +1,626 @@
...
+    }
+
+const char* wg_scan_local_work_group_size = R"(
+    bool is_zero_linear_id()
+    {
+        size_t linear_id;
+#if __OPENCL_VERSION__ < CL_VERSION_2_0
+        linear_id = ((get_global_id(2) – get_global_offset(2)) * get_global_size(1) * get_global_size(0)) + 
+                    ((get_global_id(1) – get_global_offset(1)) * get_global_size(0)) + 
+                    (get_global_id(0) – get_global_offset(0));
```
kpet, line 47: The subtractions are not written with the ASCII `-` character (probably some unicode dash). The kernels do not  compile for versions prior to 2.0.

test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
```diff
@@ -0,0 +1,626 @@
...
+        context, &scan1kb_program, &scan1kb_kernel, 1,
+        &wg_scan_local_work_group_size,
+        "test_wg_scan_local_work_group_size_static_local",
+        "-DLOCAL_MEM_SIZE=1024");
+    test_error_ret(err, "create_single_kernel_helper failed", -1);
+    err = create_single_kernel_helper(
+        context, &scan4kb_program, &scan4kb_kernel, 1,
+        &wg_scan_local_work_group_size,
+        "test_wg_scan_local_work_group_size_static_local",
+        "-DLOCAL_MEM_SIZE=4096");
```
Jeremy-Kemp, line 253: Embedded profile devices are only required to support 1KB local memory.

Elsewhere in the test we're already gating test execution based on the value of  `CL_DEVICE_LOCAL_MEM_SIZE`. Please can we also ensure that programs aren't built unless they aren't going to request more than the amount of local memory supported by the device.

kalchr01:
Thank you for the comments. I have addressed some of them in _'test_wg_suggested_local_work_size' fixes_ commit. 
I am working on refactoring the code by adding a loop according to @bashbaug's comment and together I will address comments from @Jeremy-Kemp .
Also one extra comment from @kpet : 
"You're assuming size_t is the same size on host and device and this doesn't hold for all implementations. I suggest using cl_uint to exchange work-group sizes between the host and device. We won't ever need WG sizes greater than 4G"

test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
```diff
@@ -0,0 +1,581 @@
...
+            << local_work_size[1] << ", " << local_work_size[2] << ")\n";
+        return -1;
+    }
+    return err;
+}
+
+int do_test_work_group_suggested_local_size(
+    cl_device_id device, cl_context context, cl_command_queue queue,
+    bool (*skip_cond)(size_t), size_t start, size_t end, size_t incr,
+    cl_long max_local_mem_size, size_t global_work_offset[], num_dims dim)
```
Jeremy-Kemp, line 221: `max_local_mem_size` is not used in this function resulting in compilations and kernel enqueues with potentially too much local memory for the device.

Jeremy-Kemp (changes requested):

test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
kalchr01: Thank you, this has just been addressed in the latest commit

kalchr01:
Updates added to address the mentioned issues. Thank you!

Jeremy-Kemp (changes requested):

test_conformance/workgroups/test_wg_suggested_local_work_size.cpp
```diff
@@ -0,0 +1,583 @@
...
+                                   "test_wg_scan_local_work_group_size_static_local",
+                                   "test_wg_scan_local_work_group_size_dynlocal" };
+    std::string str_local_mem_size[6] = {"-DLOCAL_MEM_SIZE=1",
+                                         "-DLOCAL_MEM_SIZE=1024",
+                                         "-DLOCAL_MEM_SIZE=4096",
+                                         "-DLOCAL_MEM_SIZE=16384",
+                                         "-DLOCAL_MEM_SIZE=32768",
+                                         "-DLOCAL_MEM_SIZE=1" };
+    size_t local_mem_size[6] = {1, 1024, 4096, 16384, 32768, 1};
+    size_t dyn_mem_size[6] = {0, 0, 0, 0, 0, 1024};
```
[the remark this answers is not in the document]
kalchr01: Thank you, I added this check in the next patch.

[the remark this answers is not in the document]
Jeremy-Kemp: Thanks for all your iterations on the local memory changes!

Jeremy-Kemp (commented):
LGTM once the presubmit checks are resolved. Thanks!

kalchr01:
> LGTM once the presubmit checks are resolved. Thanks!

@Jeremy-Kemp: presubmit checks are now passing. Thank you!

Jeremy-Kemp (approved):

kalchr01:
> Hi, this is a great start but I found a few issues, thanks!

@bashbaug: The requested changes are addressed, thank you!

Jeremy-Kemp (commented):
We've just run this test on our prototype, and are finding some of the global dimensions to be problematic in terms of test runtime (vs the functionality being tested).

For example, we see 3D global sizes of 448, 4217, 4217 giving a total number of instances of nearly 8 billion. Even with such trivial kernels to execute, its still too much time to execute in our opinion.

Could we look to reduce this? Do other vendors see this too?

kalchr01:
> We've just run this test on our prototype, and are finding some of the global dimensions to be problematic in terms of test runtime (vs the functionality being tested).
> 
> For example, we see 3D global sizes of 448, 4217, 4217 giving a total number of instances of nearly 8 billion. Even with such trivial kernels to execute, its still too much time to execute in our opinion.
> 
> Could we look to reduce this? Do other vendors see this too?

Hi @Jeremy-Kemp , yes I will have a look into this and decrease a bit the sizes so that they become more meaningful.

kalchr01:
Hello @Jeremy-Kemp, 
I have a patch where I have reduced the dimensions and for a device with CL_DEVICE_MAX_WORK_GROUP_SIZE=512 it produces ND-ranges up to 66387968 instances (for {1013, 256, 256}). 
I am going to push it, but please let me know if you think it is still high... 
Thank you!

kalchr01:
Just uploaded the latest patch whith a reduced number of instances at the 3D tests. In this case I have observed max number of work items to be 7750783 for test values = { 607, 113, 113 } and for CL_DEVICE_MAX_WORK_GROUP_SIZE=512 (model run). 
Tested it on hikey board and it finishes quickly. 
Please let me know if there are any further issues. 
Thank you!

Jeremy-Kemp (approved):
Thank you for latest updates @kalchr01. Really appreciate all the changes and good work for this PR.

PR #1269, opened 2021-06-02
Fix check for image support in test_basic sizeof

BKoscielak:
Currently test for image3d_t can be skipped only in embedded profile. It should also be skipped if images aren't supported by device.

### DIFF
--- a/test_conformance/basic/test_sizeof.cpp
+++ b/test_conformance/basic/test_sizeof.cpp
@@ -292,11 +292,11 @@ int test_sizeof(cl_device_id device, cl_context context, cl_command_queue queue,
             continue;
         }
 
-        if( gIsEmbedded &&
-           0 == strcmp(other_types[i], "image3d_t") &&
-           checkFor3DImageSupport( device ) == CL_IMAGE_FORMAT_NOT_SUPPORTED)
+        if (0 == strcmp(other_types[i], "image3d_t")
+            && checkFor3DImageSupport(device) == CL_IMAGE_FORMAT_NOT_SUPPORTED)
         {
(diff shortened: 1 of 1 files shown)


test_conformance/basic/test_sizeof.cpp
```diff
@@ -292,12 +292,18 @@ int test_sizeof(cl_device_id device, cl_context context, cl_command_queue queue,
...
+        if( 0 == strcmp(other_types[i], "image3d_t" ) )
         {
-            log_info("\n3D images are not supported by this device. Skipping test.\t");
-            continue;
+            if( checkForImageSupport( device ) == CL_IMAGE_FORMAT_NOT_SUPPORTED)
+            {
+                log_info("\nimages are not supported by this device. Skipping test.\t");
+                continue;
+            }
+            if (gIsEmbedded && checkFor3DImageSupport(device) == CL_IMAGE_FORMAT_NOT_SUPPORTED)
```
Jeremy-Kemp, line 302: We could drop the check for `gIsEmbedded` here as `checkFor3DImageSupport` already handles the embedded case.

Jeremy-Kemp (commented):
Looks good once the code formatter is happy!

test_conformance/basic/test_sizeof.cpp
BKoscielak: Then we can also drop `checkForImageSupport`, as `checkFor3DImageSupport` also handles standard image support.

Jeremy-Kemp (approved):

kpet (approved):

PR #1268, opened 2021-06-02
Fix copy and move semantics of wrapper classes

mantognini:
Related to #465.

The first commit removes unnecessary functions. The second one rewrites the wrappers to handle copy and move semantics, while preserving the existing API and removing code duplication. It also adds error handling around clRelase* and clRetain*.

PR #1277, opened 2021-06-18
Fix double-release of memory objects

jrprice:
A recent update to the object wrapper classes (#1268) changed the
behavior of assigning to a wrapper, whereby the wrapped object is now
released upon assignment. A couple of tests were manually calling
`clReleaseMemObject` and then assigning `nullptr` to the wrapper,
resulting in the wrapper calling `clReleaseMemObject` on an object that
had already been destroyed.


mantognini (approved):
Looks appropriate indeed, thanks for the follow up.

EwanC (dismissed):

Jeremy-Kemp (approved):
LGTM!

I'm not sure how to force the `Build macos-latest` action to run to unblock merging?

jrprice:
> I'm not sure how to force the Build macos-latest action to run to unblock merging?

I rebased the PR since it's probably because the macOS config changed on the master branch.

Unfortunately this dismissed all the reviews :-/

EwanC (approved):

Jeremy-Kemp (approved):

Issue #1249, opened 2021-05-14
test_subgroups ifp_ext fails in 2.0


jwhitmoresam:
CTS test "test_subgroups ifp_ext" checks for CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS, which was introduced as a device query in OpenCL 2.1 but is marked as a 2.0 test: https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/subgroups/main.cpp. For 2.0, should not use the global function checkIFPSupport.

rjodinchr:
Fixed by PR #1278 (commit `79f692d8`), which gated `checkIFPSupport` under `if (device_cl_version >= Version(2, 1))` in `test_ifp_ext` because `CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS` was only introduced in OpenCL 2.1.

Issue #1199, opened 2021-03-26
Subgroups tests - data generation for arithmetic function to prevent undefined behavior


gwawiork:
See comments: 
https://github.com/KhronosGroup/OpenCL-CTS/pull/1074#discussion_r572197743
https://github.com/KhronosGroup/OpenCL-CTS/pull/1074#discussion_r597377453
Take into consideration all function especially floating-points types

PR #1278, opened 2021-06-21
subgroups: Fix setting cl_halfs and progress check.

lakshmih:
cl_float testing uses set_value such that a generated cl_ulong of 1 is
stored as 1.0F in a logical sense. However, cl_half values aren't
intrinsic to C++ and generated cl_ulongs less than 1024 in particular
are interpreted bitwise as subnormals. The test fails on compute devices
lacking subnormal support. Perform the logical conversion to cl_half.

Fix independent forward progress check.

### DIFF
--- a/test_conformance/subgroups/main.cpp
+++ b/test_conformance/subgroups/main.cpp
@@ -19,8 +19,10 @@
 #include <string.h>
 #include "procs.h"
 #include "harness/testHarness.h"
+#include "CL/cl_half.h"
 
 MTdata gMTdata;
+cl_half_rounding_mode g_rounding_mode;
 
 test_definition test_list[] = {
     ADD_TEST_VERSION(sub_group_info_ext, Version(2, 0)),
@@ -65,6 +67,22 @@ static test_status InitCL(cl_device_id device)
         {
             ret = TEST_SKIP;
         }
+    }
+    // Determine the rounding mode to be used in float to half conversions in
+    // init and reference code
+    const cl_device_fp_config fpConfig = get_default_rounding_mode(device);
+
+    if (fpConfig == CL_FP_ROUND_TO_NEAREST)
+    {
+        g_rounding_mode = CL_HALF_RTE;
+    }
+    else if (fpConfig == CL_FP_ROUND_TO_ZERO && gIsEmbedded)
+    {
+        g_rounding_mode = CL_HALF_RTZ;
+    }
+    else
+    {
+        assert(false && "Unreachable");
     }
     return ret;
 }
--- a/test_conformance/subgroups/subhelpers.h
+++ b/test_conformance/subgroups/subhelpers.h
@@ -1080,7 +1080,7 @@ template <typename Ty>
 typename std::enable_if<TypeManager<Ty>::is_sb_scalar_type::value>::type
 set_value(Ty &lhs, const cl_ulong &rhs)
 {
-    lhs.data = rhs;
+    lhs.data = cl_half_from_float(static_cast<cl_float>(rhs), CL_HALF_RTE);
 }
 
 // compare for common vectors
--- a/test_conformance/subgroups/subhelpers.h
+++ b/test_conformance/subgroups/subhelpers.h
@@ -1080,7 +1080,7 @@ template <typename Ty>
 typename std::enable_if<TypeManager<Ty>::is_sb_scalar_type::value>::type
 set_value(Ty &lhs, const cl_ulong &rhs)
 {
-    lhs.data = cl_half_from_float(static_cast<cl_float>(rhs), CL_HALF_RTE);
+    lhs.data = convert_float_to_half(static_cast<cl_float>(rhs));
 }
 
 // compare for common vectors
@@ -1292,6 +1292,11 @@ template <typename Ty, typename Fns, size_t TSIZE = 0> struct test
             }
             else if (strstr(TypeManager<Ty>::name(), "half"))
             {
+                if( DetectFloatToHalfRoundingMode(queue) )
+                {
+                    log_error("Unable to detect rounding mode\n");
+                    return TEST_FAIL;
+                }
                 kernel_sstr << "#pragma OPENCL EXTENSION cl_khr_fp16: enable\n";
             }
         }
--- a/test_conformance/subgroups/subhelpers.h
+++ b/test_conformance/subgroups/subhelpers.h
@@ -1292,7 +1292,7 @@ template <typename Ty, typename Fns, size_t TSIZE = 0> struct test
             }
             else if (strstr(TypeManager<Ty>::name(), "half"))
             {
-                if( DetectFloatToHalfRoundingMode(queue) )
+                if (DetectFloatToHalfRoundingMode(queue))
                 {
                     log_error("Unable to detect rounding mode\n");
                     return TEST_FAIL;
--- a/test_conformance/subgroups/subhelpers.h
+++ b/test_conformance/subgroups/subhelpers.h
@@ -28,6 +28,7 @@
 #define NR_OF_ACTIVE_WORK_ITEMS 4
 
 extern MTdata gMTdata;
+extern cl_half_rounding_mode g_rounding_mode;
 
 struct WorkGroupParams
 {
@@ -1080,7 +1081,7 @@ template <typename Ty>
 typename std::enable_if<TypeManager<Ty>::is_sb_scalar_type::value>::type
 set_value(Ty &lhs, const cl_ulong &rhs)
 {
-    lhs.data = convert_float_to_half(static_cast<cl_float>(rhs));
+    lhs.data = cl_half_from_float(static_cast<cl_float>(rhs), g_rounding_mode);
 }
 
 // compare for common vectors
@@ -1292,11 +1293,6 @@ template <typename Ty, typename Fns, size_t TSIZE = 0> struct test
             }
             else if (strstr(TypeManager<Ty>::name(), "half"))
             {
-                if (DetectFloatToHalfRoundingMode(queue))
-                {
-                    log_error("Unable to detect rounding mode\n");
-                    return TEST_FAIL;
-                }
                 kernel_sstr << "#pragma OPENCL EXTENSION cl_khr_fp16: enable\n";
             }
         }
--- a/test_conformance/subgroups/test_ifp.cpp
+++ b/test_conformance/subgroups/test_ifp.cpp
@@ -302,27 +302,16 @@ static test_status checkIFPSupport(cl_device_id device, bool &ifpSupport)
 {
     cl_uint ifp_supported;
     cl_uint error;
-    auto device_cl_version = get_device_cl_version(device);
-    if (device_cl_version < Version(2, 1))
+    error = clGetDeviceInfo(device,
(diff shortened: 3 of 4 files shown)


jlewis-austin:
Fixes #1249

bashbaug (approved):
LGTM.

It'd be nice if there were a better way to catch these "cl_half-as-ushort" types of issues.

kpet:
>It'd be nice if there were a better way to catch these "cl_half-as-ushort" types of issues.

There is one, we could do https://github.com/KhronosGroup/OpenCL-Headers/issues/105.

test_conformance/subgroups/test_ifp.cpp
```diff
@@ -302,16 +302,27 @@ static test_status checkIFPSupport(cl_device_id device, bool &ifpSupport)
...
+    if (device_cl_version < Version(2, 1))
     {
-        print_error(
-            error,
-            "Unable to get CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS "
-            "capability");
-        return TEST_FAIL;
+        // OpenCL 2.0 and below don't provide the CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS property. Instead,
+        // support is determined by the presence of cl_khr_subgroups.
+        ifp_supported = is_extension_available(device, "cl_khr_subgroups");
```
jlewis-austin, line 310: This duplicates the check in test_ifp_ext, which is used to determine if test should skip itself. Can we move that logic in here as well?

test_conformance/subgroups/subhelpers.h
```diff
@@ -1080,7 +1080,7 @@ template <typename Ty>
 typename std::enable_if<TypeManager<Ty>::is_sb_scalar_type::value>::type
 set_value(Ty &lhs, const cl_ulong &rhs)
 {
-    lhs.data = rhs;
+    lhs.data = cl_half_from_float(static_cast<cl_float>(rhs), CL_HALF_RTE);
```
Jeremy-Kemp, line 1083: I think this should also query the rounding mode of the device to select between `CL_HALF_R*` to ensure the correct rounding behaviour.

Jeremy-Kemp (changes requested):

test_conformance/subgroups/subhelpers.h
jlewis-austin, line 1083: This will only give us integer-valued float values and a lot of +Infs; I think we'd want coverage of all the special values (unless that's handled elsewhere).

jlewis-austin (changes requested):

gwawiork (changes requested):

test_conformance/subgroups/subhelpers.h
bashbaug: I thought about this too, but I don't think this is any worse after this change, is it?

jlewis-austin: That's fair. It depends on your definition of worse I guess--we're losing coverage of fractional values, but getting around the blocker of subnormals. I'm OK with letting this in as long as we're tracking it; I'll verify that an issue was generated from the comments on the original PR.

jlewis-austin: #1199 covers this

test_conformance/subgroups/test_ifp.cpp
lakshmih: I went with the approach of moving the check to test_ifp_ext as recommended by @gwawiork because:
- This function gets called from test_ifp_core as well, where it appears we only want to check for CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS (since this is part of core?, anyway that is the logic of the current code)
- When running test_ifp_ext, if version < 2.1 we only check for the string, if >= 2.1 we check for string and CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS.

test_conformance/subgroups/subhelpers.h
lakshmih: Modified to detect and use the device's rounding mode

test_conformance/subgroups/test_ifp.cpp
```diff
@@ -302,16 +302,27 @@ static test_status checkIFPSupport(cl_device_id device, bool &ifpSupport)
 {
     cl_uint ifp_supported;
     cl_uint error;
-    error = clGetDeviceInfo(device,
-                            CL_DEVICE_SUB_GROUP_INDEPENDENT_FORWARD_PROGRESS,
-                            sizeof(ifp_supported), &ifp_supported, NULL);
-    if (error != CL_SUCCESS)
+    auto device_cl_version = get_device_cl_version(device);
+    if (device_cl_version < Version(2, 1))
```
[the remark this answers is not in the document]
lakshmih: Done

test_conformance/subgroups/test_ifp.cpp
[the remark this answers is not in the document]
lakshmih: Done. Also see other comment about why.

jlewis-austin (approved):
Thanks, LGTM!

Jeremy-Kemp (approved):
Thank you for the changes!

lakshmih:
@bashbaug  Ben, can you please merge this?
@jlewis-austin  and @gwawiork  approved the previous patchset and I only made a formatting change in this one.

gwawiork:
@lakshmih I think I have not accepted yet. Let me see at these latest changes

gwawiork:
Changes for `ifp` - LGTM
Changes for `half` - I am getting errors for arithmetic multiplication function. The test expects `0x7bff` (maximal half number) but device calculates and returns `0x7c00` (INF). Any idea how to fix this quickly ? Can we assume that `cl_half_from_float` when return `0x7bff` then it means INF and we can skip comparing returned values ?

jlewis-austin:
We added rounding mode support https://github.com/KhronosGroup/OpenCL-CTS/pull/1278#discussion_r656320825 to the conversion, but not the reference calculation https://github.com/KhronosGroup/OpenCL-CTS/blob/69f0054001438078c11478546b855c06e07e1817/test_conformance/subgroups/subgroup_common_templates.h#L383 @gwawiork could your failure be with a non-RTE rounding mode?

lakshmih:
@gwawiork  Are you using my latest patch, coz I've added the conversion to the reference as well (inside to_half):

https://github.com/KhronosGroup/OpenCL-CTS/pull/1278/files#diff-550bd939e92c1ac5c693e2df7a5414770deb91ba12afcfd0a75a751bfdf49593

jlewis-austin:
> https://github.com/KhronosGroup/OpenCL-CTS/pull/1278/files#diff-550bd939e92c1ac5c693e2df7a5414770deb91ba12afcfd0a75a751bfdf49593

OK nice, I must have gotten into the wrong version when I checked it.

test_conformance/subgroups/subhelpers.h
```diff
@@ -1292,6 +1292,11 @@ template <typename Ty, typename Fns, size_t TSIZE = 0> struct test
             }
             else if (strstr(TypeManager<Ty>::name(), "half"))
             {
+                if (DetectFloatToHalfRoundingMode(queue))
```
jlewis-austin, line 1295: This detector is meant for detecting the rounding mode used on image writes (by running a kernel). That's a special case involving relaxed rounding requirements, and shouldn't be applied to arithmetic operations.

test_conformance/subgroups/subhelpers.h
jlewis-austin: The query used for image tests isn't going to work correctly since it's specific to image writes and unrelated to the default rounding mode (used by the arithmetic operators and conversions). Also, it would execute an image write kernel for every data point in the test (!).

The default mode is always RTE for the full profile, so I think @Jeremy-Kemp  might be referring to the embedded profile, where it can be either RTE or RTZ. You can check this with a CL_DEVICE_SINGLE_FP_CONFIG query, which would be faster than running a kernel, but we'd want to check the impact on run-time before adding it to an arithmetic operator implementation.

The only alternative I can see would be to provide another argument to set_value to pass in the rounding mode. This would be needed anyway if we ever want to use this to support explicit conversion operators with different rounding modes.

test_conformance/subgroups/subhelpers.h
lakshmih: Agreed (took a closer look)
In which case, should I just hard-code to RTE as before? It is required to be supported by the spec right? It is already hardcoded elsewhere in the test so I would argue this patch leaves the test improved in any case and no worse than before.
https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/subgroups/subgroup_common_templates.h#L304

test_conformance/subgroups/subhelpers.h
lakshmih: Agreed and modifying to use  CL_DEVICE_SINGLE_FP_CONFIG query and to allow RTZ for embedded.
This wouldn't have run the kernel for every data point though, coz the detection code is called only once and then used to initialize a global.

test_conformance/subgroups/subhelpers.h
lakshmih: > @lakshmih I think I have not accepted yet. Let me see at these latest changes

@gwawiork  can you review and also check if this fixes your issue?

test_conformance/subgroups/subhelpers.h
jlewis-austin: > This wouldn't have run the kernel for every data point though, coz the detection code is called only once and then used to initialize a global.

Ha, of course, another brain cramp..

test_conformance/subgroups/subhelpers.h
gwawiork: > > @lakshmih I think I have not accepted yet. Let me see at these latest changes
> 
> @gwawiork can you review and also check if this fixes your issue?

Now I got better results. I still see some sporadic issues but it tends to be our implementation issue - 0x7a69 x 0x3c00 is sometimes 0x7a68  while should be 0x7a69

gwawiork (approved):
LGTM

lakshmih:
@bashbaug @Jeremy-Kemp @jlewis-austin  Could you review/merge since @gwawiork  has approved. Thanks.

test_conformance/subgroups/main.cpp
```diff
@@ -66,6 +68,22 @@ static test_status InitCL(cl_device_id device)
...
     }
+    // Determine the rounding mode to be used in float to half conversions in
+    // init and reference code
+    const cl_device_fp_config fpConfig = get_default_rounding_mode(device);
+
+    if (fpConfig == CL_FP_ROUND_TO_NEAREST)
+    {
+        g_rounding_mode = CL_HALF_RTE;
+    }
+    else if (fpConfig == CL_FP_ROUND_TO_ZERO && gIsEmbedded)
```
Jeremy-Kemp, line 79: Sorry, missed the discussion previously on this point.

I don't think that `gIsEmbedded` is required, but works for us so wont let it block this PR.

Jeremy-Kemp (approved):
Thank you again for the changes!

jlewis-austin (approved):

lakshmih:
Ping, is there anything pending from us [Qualcomm] for this to merge?

PR #1283, opened 2021-07-02
Fix memory model issue in `atomic_flag`.

### LINKED ISSUE #134: c11 atomic flag test is not correct in some cases
We're seeing the following failure when running the c11 atomic flag test:
...
       global atomic_flag declared in program, used in generic function, order_relaxed, scope_all_svm_devices...
                (thread count 1024, group size 256)
        local atomic_flag, order_acquire...
                (thread count 256, group size 256)
ERROR: Critical section 233 not visited
ERROR: Reference values did not validate!

The dumped kernel source contains this loop:

 for(cnt = 0; !stop && cnt < threadCount; cnt++) // each thread must find critical section where it is the first visitor
  {
    bool set = atomic_flag_test_and_set_explicit(&destMemory[cnt], memory_order_acquire);
    if (!set)
    {
      uint csIndex = get_enqueued_local_size(0)*get_group_id(0)+cnt;
      // verify that thread is the first visitor
      if(oldValues[csIndex] == 1000000000)
      {
        oldValues[csIndex] = tid; // set the winner id for this critical section
        stop = 1;
      }
      atomic_work_item_fence(CLK_LOCAL_MEM_FENCE, memory_order_release,memory_scope_work_group);
      atomic_flag_clear_explicit(&destMemory[cnt], memory_order_relaxed);
    }
  }

The problem here is that oldValues is in the global address space, but the atomics and fence are on the local address space.  Nothing is ordering the reads and writes to oldValues.  This is a data race.

Jeremy-Kemp:
In `atomic_flag` sub-tests that modify local memory, compilers may re-order memory accesses between the local and global address spaces which can lead to incorrect test failures.

This commit ensures that both local and global memory operations are fenced to prevent this re-ordering from occurring.

Fixes #134.

### DIFF
--- a/test_conformance/c11_atomics/test_atomics.cpp
+++ b/test_conformance/c11_atomics/test_atomics.cpp
@@ -1684,11 +1684,16 @@ public:
       "      }\n";
 
     if (MemoryOrder() == MEMORY_ORDER_ACQUIRE || MemoryOrder() == MEMORY_ORDER_RELAXED)
-      program += "      atomic_work_item_fence(" +
-                 std::string(LocalMemory() ? "CLK_LOCAL_MEM_FENCE | CLK_GLOBAL_MEM_FENCE, " : "CLK_GLOBAL_MEM_FENCE, ") +
-                 "memory_order_release," +
-                 std::string(LocalMemory() ? "memory_scope_work_group" : (UseSVM() ? "memory_scope_all_svm_devices" : "memory_scope_device") ) +
-                 ");\n";
+        program += "      atomic_work_item_fence("
+            + std::string(LocalMemory()
+                              ? "CLK_LOCAL_MEM_FENCE | CLK_GLOBAL_MEM_FENCE, "
+                              : "CLK_GLOBAL_MEM_FENCE, ")
+            + "memory_order_release,"
+            + std::string(LocalMemory()
+                              ? "memory_scope_work_group"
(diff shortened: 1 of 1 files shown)


Jeremy-Kemp:
Thank  you for the comment @jlewis-austin. I agree, there was an omission in the commit whereby the acquire was missed.

I will post an update with the change for review.

jlewis-austin (commented):
Thanks @Jeremy-Kemp, that looks pretty much like what I did for my testing. My only concern with it was that I don't have a good handle on what all the different generated test case kernels look like, so I'd like to dump them and make sure they all make sense. I'll try to get to that ASAP.

jlewis-austin (approved):
I checked out the kernels dumped for a 2.0 implementation and it looks good, aside from some pre-existing coverage gaps (it doesn't exercise the atomic_flag_test_and_set_explicit 2.0 flavors, or use any local variables within the flag region to verify the memory orders.
But for this issue, LGTM. Thanks!

gwawiork (approved):
LGTM

b-sumner:
I think this is a necessary improvement.   I think the test could be strengthened, but this need not wait.  LGTM.

Jeremy-Kemp:
If there are no further objections/comments, please can this be merged in as we're currently at 3 approvals. I'm reluctant to merge my own PR.

kpet:
I'm going to assume that 12 days was enough for people to reply. Merging.

PR #1286, opened 2021-07-15
Fix double free in c11_atomics tests for SVM allocations

gwawiork:
I see errors on Windows in cases svm_atomic_*
`ERROR: clRelease*() failed! (CL_INVALID_MEM_OBJECT from C:\NewVPG\Repo\public-github\opensource-qb\OpenCL-CTS\test_common\harness/typeWrappers.h:70)`
This is related to changes in Wrapper Class introduced in https://github.com/KhronosGroup/OpenCL-CTS/pull/1268

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -1521,16 +1521,13 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(
     if (UseSVM())
     {
         // the buffer object must first be released before the SVM buffer is
-        // freed. Assignment value 0 runs Wrapper Class method reset() defined
-        // in typeWrappers.h which release old object
-        streams[0] = 0;
-        test_error(error, "clReleaseMemObject failed");
+        // freed. The Wrapper Class method reset() will do that
+        streams[0].reset();
         if (gUseHostPtr)
             free(svmAtomicBuffer);
         else
             clSVMFree(context, svmAtomicBuffer);
-        streams[1] = 0;
-        test_error(error, "clReleaseMemObject failed");
+        streams[1].reset();
         if (gUseHostPtr)
             free(svmDataBuffer);
         else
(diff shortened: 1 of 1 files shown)


gwawiork:
@mantognini could you look at this fix ? Thank you!

test_conformance/c11_atomics/common.h
```diff
@@ -1521,15 +1521,14 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(
     if (UseSVM())
     {
         // the buffer object must first be released before the SVM buffer is
-        // freed
-        error = clReleaseMemObject(streams[0]);
+        // freed. Assignment value 0 runs Wrapper Class method reset() defined
+        // in typeWrappers.h which release old object
         streams[0] = 0;
```
bashbaug: It might be clearer to call `reset()` directly here instead.

test_conformance/c11_atomics/common.h
```diff
@@ -1521,15 +1521,14 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(
...
         test_error(error, "clReleaseMemObject failed");
```
bashbaug, line 1527: Minor: After this change there is no error code from `clReleaseMemObject` to check, so this can be removed (see similar case below also).

bashbaug (commented):
Good catch!

Is there any way to check if other tests have similar issues?

gwawiork:
> Good catch!
> 
> Is there any way to check if other tests have similar issues?

I think that results from all cases on Windows would reveal such issues (as in this case). I have some issues with checking the results now so I am not sure if there are some other cases.

gwawiork:
> > Good catch!
> > Is there any way to check if other tests have similar issues?
> 
> I think that results from all cases on Windows would reveal such issues (as in this case). I have some issues with checking the results now so I am not sure if there are some other cases.

There is more:
test_api_context_destructor_callback
test_gl_buffers
test_gl_buffers_getinfo
test_gl_fence_sync

Jeremy-Kemp (approved):

kpet (approved):

PR #1290, opened 2021-07-23
Guard extended subgroup tests with CL_EXPERIMENTAL

StuartDBrady:


kpet:
Closing as these tests are relied upon to test released extensions. Problems with these tests will be reported as specific issues.

PR #1291, opened 2021-07-29
compiler: relax requirement for features on 3.0

kdub:
Section 6.2.1 states that the feature check in opencl_c_versions are
optional on 3.0 C language, but yet this test will fail if a
feature is missing. Make this test check if these are present for 2.0
devices, but not 3.0.

Test: disable __opencl_c_pipes extension on 3.0 driver and pass test.

### DIFF
--- a/test_conformance/compiler/test_opencl_c_versions.cpp
+++ b/test_conformance/compiler/test_opencl_c_versions.cpp
@@ -193,31 +193,27 @@ static int test_CL_DEVICE_OPENCL_C_VERSION_features(cl_device_id device,
     }
 
     const Version clc_version = get_device_cl_c_version(device);
-    if (clc_version >= Version(2, 0))
+    bool has_all_OpenCL_C_20_features =
+        features.supports__opencl_c_atomic_order_acq_rel
+        && features.supports__opencl_c_atomic_order_seq_cst
+        && features.supports__opencl_c_atomic_scope_device
+        && features.supports__opencl_c_atomic_scope_all_devices
+        && features.supports__opencl_c_device_enqueue
+        && features.supports__opencl_c_generic_address_space
+        && features.supports__opencl_c_pipes
+        && features.supports__opencl_c_program_scope_global_variables
+        && features.supports__opencl_c_work_group_collective_functions;
+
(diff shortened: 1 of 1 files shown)


bashbaug:
I think the test is correct as-written, though this is admittedly a confusing part of the spec.

Here are the scenarios to consider:

1. For a device where the **OpenCL** version is less than OpenCL 3.0, this test should not be run.  Rationale: the OpenCL C feature query did not exist before OpenCL 3.0.

2. When the device **OpenCL C** version - specifically [`CL_DEVICE_OPENCL_C_VERSION`](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_VERSION) - is greater than or equal to OpenCL 2.0 we need to check that all of the required OpenCL C 2.0 features are present.  This is because this is the legacy "fully backwards compatible" OpenCL C query.

    * If an OpenCL 3.0 device doesn't support an OpenCL C 2.0 feature (e.g. pipes) then it should not be returning a version greater than OpenCL C 1.2 for this query.
    * It is possible for an OpenCL 3.0 device to return a version greater than OpenCL C 1.2 (e.g. OpenCL C 3.0) for the newer [`CL_DEVICE_OPENCL_C_ALL_VERSIONS`](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_ALL_VERSIONS) query without supporting all OpenCL C 2.0 features.

We have OpenCL 3.0 devices that do not support all OpenCL C 2.0 features (in our case, device-side enqueue), so we are returning 1.2 for `CL_DEVICE_OPENCL_C_VERSION` and 1.0, 1.1, 1.2, and 3.0 (but not 2.0) for `CL_DEVICE_OPENCL_C_ALL_VERSIONS`.

Can you say what is returned for `CL_DEVICE_OPENCL_C_VERSION` when this test is failing on an OpenCL 3.0 device?

bashbaug:
Hi @kdub, this PR has been open for a while.  Does the rationale described above make sense?  Can this PR be closed?  Thanks!

PR #1296, opened 2021-08-06
Fix test_kernel_attributes test issue

fzou1:
Per OpenCL spec on CL_KERNEL_ATTRIBUTES, for kernels not created from OpenCL C
source and the clCreateProgramWithSource API call the string returned from this
query will be empty.
But in test_kernel_attributes test, it read from bc binary and expect to get
kernel attribute, which is not consistent with OpenCL spec.

Fix the test by reading from cl source file and checking kernel
attributes.


bashbaug (changes requested):
I don't think this is the right fix.  Since this case is in the "spir" test suite it would be better to disable the test vs. changing it to compile from OpenCL C source.

Note: If #1297 supersedes this PR can this PR be closed?

fzou1:
Closed. Thank you.

Issue #1284, opened 2021-07-08
test_kernel_attributes test is not correct


fzou1:
OpenCL spec (https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_ATTRIBUTES): For kernels not created from OpenCL C source and the clCreateProgramWithSource API call the string returned from this query will be empty.

But in test_kernel_attributes test, it read from bc binary and expect to get kernel attribute, which is not consistent with OpenCL spec. Related test code:
static bool
test_kernel_attributes(cl_device_id device, cl_uint width, const char *folder)
{
...
        get_bc_file_path(folder, "kernel_attributes", bc_file_path, width);
        clProgramWrapper bcprog = create_program_from_bc(context, bc_file_path);
...
        // Querying the kernel for its attributes.
        kernel = create_kernel_helper(bcprog, "test");
        cl_int err_code = clGetKernelInfo(kernel, CL_KERNEL_ATTRIBUTES, sizeof(attributes), attributes, &res_size);
...
        // Building the expected attributes vector.
        std::vector<std::string> expected;
        expected.push_back(std::string("work_group_size_hint(64,1,1)"));
        expected.push_back(std::string("vec_type_hint(float4)"));

Please have a check. Thanks.

FranklandJack:
Hey, it looks like we've hit this issue also.

As far as I can tell all paths through the `test_kernel_attributes` test get their program from a `clCreateProgramWithBinary` call: https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/spir/run_services.cpp#L193. So, I'm wondering whether based on the spec wording: 

> For kernels not created from OpenCL C source and the clCreateProgramWithSource API call the string returned from this query will be empty.

whether this test should just be removed?

fzou1:
Or should we update test to read from .cl source file?

alycm:
Relevant comments on what `CL_KERNEL_ATTRIBUTES` does for SPIR on current implementations:
* https://github.com/KhronosGroup/OpenCL-CTS/pull/1297#issuecomment-896124934
* https://github.com/KhronosGroup/OpenCL-CTS/pull/1297#issuecomment-896870116

PR #1297, opened 2021-08-10
Temporarily disable the test_kernel_attributes test case

fzou1:
Per OpenCL spec on CL_KERNEL_ATTRIBUTES, for kernels not created from OpenCL C
source and the clCreateProgramWithSource API call the string returned from this
query will be empty.
But in test_kernel_attributes test, it read from bc binary and expect to get
kernel attribute, which is not consistent with OpenCL spec.


fzou1:
#1284

bashbaug (commented):
Makes sense to me, the spec is pretty clear that `CL_KERNEL_ATTRIBUTES` is not supported when a program is not created from source.

There is an interesting question how other companies are passing this test since it has been around for a while, but I think it's OK to disable the test while we figure out what to do.

Note that this is an older SPIR test and not a SPIR-V test.

Needs a non-Intel reviewer - @kpet @alycm @Jeremy-Kemp ?

alycm:
> There is an interesting question how other companies are passing this test since it has been around for a while, but I think it's OK to disable the test while we figure out what to do.

For SPIR we did return attributes for `CL_KERNEL_ATTRIBUTES`.  The 1.2 specification did not have the "the string returned from this query will be empty" line.  That was added in OpenCL 2.1, presumably as part of supporting IL programs.  The `cl_khr_spir` specification does not say anything about this query but I guess we implemented it to pass the test.

What changed is that https://github.com/KhronosGroup/OpenCL-CTS/pull/1055 added a test that enforces the empty string requirement.  Technically that only tests `clCreateProgramWithBinary` for device binaries, not SPIR binaries, but as they go through the same entry point it makes sense that implementations would behave consistently between the two cases.

Perhaps the unified specification accidentally made the OpenCL 1.0 -- 2.0 definition stricter, but the `cl_khr_spir` extension could have been implemented for 2.1+ as well, so the ambiguity isn't new.

alycm (approved):
Given we have two tests expecting contradicting behaviour I'm okay with disabling this test until that is resolved.

Disable this one rather than the other one because this one does not match the unified specification.

fzou1:
Thank you. Could someone help merge this PR? I didn't have write access.

Jeremy-Kemp:
> For SPIR we did return attributes for CL_KERNEL_ATTRIBUTES.

This is the same for our implementation as well. We will return attributes for SPIR inputs which explains why the test has been passing for us.

Jeremy-Kemp (approved):
OK to disable for now, with a proper fix in the future to close out #1284.

Issue #1298, opened 2021-08-12
test_half calls async_work_group_copy with different src and num_gentypes argument values


wenju-he:
test_half has following kernel:
```
        const char *source_local_store_v3[] = {
            "__kernel void test( __global float *p, __global half *f,\n"
            "                   uint extra_last_thread )\n"
            "{\n"
            "   __local ushort data[3*(", local_buf_size, "+1)];\n"
            "   size_t i = get_global_id(0);\n"
            "   size_t lid = get_local_id(0);\n"
            "   size_t last_i = get_global_size(0)-1;\n"
            "   size_t adjust = 0;\n"
            "   size_t lsize = get_local_size(0);\n"
            "   event_t async_event;\n"
            "   if(last_i == i && extra_last_thread != 0) {\n"
            "     adjust = 3-extra_last_thread;\n"
            "   } "
            "   vstore_half3",roundName,"( vload3(i,p-adjust), lid, (__local half *)(&data[0]) );\n"
            "   barrier( CLK_LOCAL_MEM_FENCE ); \n"
            "   async_event = async_work_group_copy((__global ushort *)(f+3*(i-lid)), (__local ushort *)(&data[adjust]), lsize*3-adjust, 0);\n" // investigate later
            "   wait_group_events(1, &async_event);\n"
            "}\n"
        };
```
Since **adjust** depends on the last work-item, all work-items in the last workgroup may have different num_gentypes argument in async_work_group_copy. This results in undefined results according to spec:

"The async copy is performed by all work-items in a work-group and this built-in function must therefore be encountered by all work-items in a work-group executing the kernel with the same argument values; otherwise the results are undefined."

rjodinchr:
Fixed by PR #1299 (commit `67ac6c8d`), which ensured uniform `src` and `num_gentypes` arguments across work-items calling `async_work_group_copy` in `Test_vStoreHalf.cpp`, conforming to OpenCL C section 6.13.11 requirements.

PR #1299, opened 2021-08-12
Fix test_half async_work_group_copy arguments (#1298)

hewj03:
(#1298) Workitems in the last workgroup calls async_work_group_copy with
different argument values depending on 'adjust'. According to spec,
this results in undefined values.


bashbaug (approved):

ayalz:
Fix LGTM.

Could alternatively check `if (last_i - lsize < i)` to complement the `if (last_i == i)`, assuming uniform work groups, but asking explicitly if the group is last is clearer.

Note that the `vstore_half3` looks fine as distinct work-items write to distinct elements of `data` in local memory.

bashbaug:
Merging as agreed in the June 14th teleconference.

PR #1301, opened 2021-08-17
Report unsupported extended subgroup tests as skipped rather than passed

kpet:
Also don't check the presence of extensions for each sub-test.

### DIFF
--- a/test_conformance/subgroups/subhelpers.h
+++ b/test_conformance/subgroups/subhelpers.h
@@ -33,10 +33,9 @@ extern cl_half_rounding_mode g_rounding_mode;
 struct WorkGroupParams
 {
     WorkGroupParams(size_t gws, size_t lws,
-                    const std::vector<std::string> &req_ext = {},
                     const std::vector<uint32_t> &all_wim = {})
         : global_workgroup_size(gws), local_workgroup_size(lws),
-          required_extensions(req_ext), all_work_item_masks(all_wim)
+          all_work_item_masks(all_wim)
     {
         subgroup_size = 0;
         work_items_mask = 0;
@@ -49,7 +48,6 @@ struct WorkGroupParams
     uint32_t work_items_mask;
     int dynsc;
     bool use_core_subgroups;
-    std::vector<std::string> required_extensions;
     std::vector<uint32_t> all_work_item_masks;
 };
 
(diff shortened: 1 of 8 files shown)


test_conformance/subgroups/test_subgroup_ballot.cpp
```diff
@@ -926,11 +926,15 @@ template <typename T> int run_non_uniform_broadcast_for_type(RunTestForType rft)
 int test_subgroup_functions_ballot(cl_device_id device, cl_context context,
                                    cl_command_queue queue, int num_elements)
 {
-    std::vector<std::string> required_extensions = { "cl_khr_subgroup_ballot" };
+    if (!is_extension_available(device, "cl_khr_subgroup_extended_types"))
```
gwawiork, line 929: Did you mean "cl_khr_subgroup_ballot" ?

test_conformance/subgroups/test_subgroup_ballot.cpp
```diff
@@ -926,11 +926,15 @@ template <typename T> int run_non_uniform_broadcast_for_type(RunTestForType rft)
 int test_subgroup_functions_ballot(cl_device_id device, cl_context context,
                                    cl_command_queue queue, int num_elements)
 {
-    std::vector<std::string> required_extensions = { "cl_khr_subgroup_ballot" };
```
gwawiork, line 929: This variable is used in subhelpers.h removing makes dead code in that file.

gwawiork (changes requested):
Have you considered make the only change in subherlpers.h file by setting TEST_SKIPPED_ITSELF ?

test_conformance/subgroups/test_subgroup_ballot.cpp
kpet: Good catch, now fixed.

test_conformance/subgroups/test_subgroup_ballot.cpp
kpet: Yes, might as well remove it all. Done.

kpet:
Thanks for the review!

>Have you considered make the only change in subherlpers.h file by setting TEST_SKIPPED_ITSELF ?

Yes, that was my first change but it doesn't help with reducing the number of times the extension's presence is tested and reported.

test_conformance/subgroups/subhelpers.h
```diff
@@ -1297,19 +1295,6 @@ template <typename Ty, typename Fns, size_t TSIZE = 0> struct test
...
-        for (std::string extension : test_params.required_extensions)
-        {
-            if (!is_extension_available(device, extension.c_str()))
-            {
-                log_info("The extension %s not supported on this device. SKIP "
-                         "testing - kernel %s data type %s\n",
-                         extension.c_str(), kname, TypeManager<Ty>::name());
-                return TEST_PASS;
-            }
-            kernel_sstr << "#pragma OPENCL EXTENSION " + extension
```
gwawiork, line 1309: What about pragmas. Do we need use them or not ?

kpet: AFAICT the pragmas are not required by the extended subgroup specification. @bashbaug Can you confirm?

bashbaug: Correct, no pragmas are (currently) required for the extended subgroups extensions.

StuartDBrady (approved):
Looks good to me.

test_conformance/subgroups/subhelpers.h
kpet: Thanks for confirming Ben!

PR #1302, opened 2021-08-18
Remove unused definitions in CMakeLists.txt

kpet:


jlewis-austin (approved):
Nice! I can't find any usage of CLConform_SOURCE_DIR either. Is that needed?

kpet:
It is used in a few places but probably best replaced with `PROJECT_SOURCE_DIR`. We might also be able to remove that `CONFORMANCE_SUFFIX` variable.

jlewis-austin:
> It is used in a few places but probably best replaced with `PROJECT_SOURCE_DIR`. We might also be able to remove that `CONFORMANCE_SUFFIX` variable.

Yes I see now searching without a space on the end is more helpful. I've done that at least twice now...

PR #1304, opened 2021-08-24
Fix build error for linux with clang-8

### LINKED ISSUE #1303: Linux build fails with clang-8
Returns the following error from brute_force reference math:

```
reference_math.cpp:4646:9: error: absolute value function 'fabs' given an argument of type 'long double' but has parameter of type 'double' which
      may cause truncation of value [-Werror,-Wabsolute-value]
    if (fabs(y) > HEX_DBL(+, 1, 0, +, 970))
        ^

reference_math.cpp:4646:9: note: use function 'std::abs' instead
    if (fabs(y) > HEX_DBL(+, 1, 0, +, 970))
        ^~~~
        std::abs`
```

MR is in progress

jlewis-austin:
-Wabsolute-value warning reported as error (long double truncated to
double)

Fixes #1303

### DIFF
--- a/test_conformance/math_brute_force/reference_math.cpp
+++ b/test_conformance/math_brute_force/reference_math.cpp
@@ -4549,8 +4549,8 @@ long double reference_powl(long double x, long double y)
     if (x != x || y != y) return x + y;
 
     // do the work required to sort out edge cases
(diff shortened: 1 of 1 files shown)


jlewis-austin:
Any objections to this fix?

PR #1307, opened 2021-08-26
Add missing include for gRandomSeed

jrprice:
Fixes a build issue on one of our platforms introduced by #1276.

### DIFF
--- a/test_common/harness/integer_ops_test_info.h
+++ b/test_common/harness/integer_ops_test_info.h
@@ -18,6 +18,7 @@
 #define INTEGER_OPS_TEST_INFO_H
 
 #include "conversions.h"
+#include "testHarness.h"
 
 // TODO: expand usage to other tests.
 


bashbaug:
Merging since this is straightforward also.

I still would be interested to know if there is a way to catch this error earlier.  Thanks!

jrprice:
Missed your comment, sorry, but thanks for merging!

The way to catch this sort of thing earlier would be to try to compile all of our headers standalone as a separate step/job in CI. For example, you can reproduce the particular error that this PR fixes by invoking Clang on the header directly:

```
$ clang++ -w -std=c++11 test_common/harness/integer_ops_test_info.h
test_common/harness/integer_ops_test_info.h:87:20: error: unknown type name
      'gRandomSeed'
    MTdataHolder d(gRandomSeed);
                   ^
1 error generated.
```

I'm not sure if it'd always be this straightforward. There may be other flags that need to be passed to the compiler, e.g. macro definitions or include paths.

PR #1313, opened 2021-08-31
do not directly use media sharing extension function names

### LINKED ISSUE #1254: Unmodified Extension API Names
Another change change in the generated headers PR (KhronosGroup/OpenCL-Headers#161) is to add a function declaration for each of the new APIs.  Previously, this was done inconsistently and only for some extensions.  Example:

```c
extern CL_API_ENTRY cl_command_queue CL_API_CALL
clCreateCommandQueueWithPropertiesKHR(
    cl_context context,
    cl_device_id device,
    const cl_queue_properties_khr* properties,
    cl_int* errcode_ret) CL_API_SUFFIX__VERSION_1_2;
```

While it doesn't happen often in the CTS tests, the "media_sharing" test declares function pointers with the same name as the function APIs, which conflicts with the function declaration in the headers.

```c
clGetDeviceIDsFromDX9MediaAdapterKHR_fn clGetDeviceIDsFromDX9MediaAdapterKHR = NULL;
clCreateFromDX9MediaSurfaceKHR_fn clCreateFromDX9MediaSurfaceKHR = NULL;
clEnqueueAcquireDX9MediaSurfacesKHR_fn clEnqueueAcquireDX9MediaSurfacesKHR = NULL;
clEnqueueReleaseDX9MediaSurfacesKHR_fn clEnqueueReleaseDX9MediaSurfacesKHR = NULL;
```

We should think about how to solve this problem before merging the generated headers PR.  The simplest solution is to name the function pointers differently, such as with a prefix or suffix.  Alternatively, we can look at a solution that provides definitions of the extension APIs as declared in the headers, in which case the calls to these functions can remain unchanged.

bashbaug:
fixes #1254 

This is one possible way to fix #1254, which is caused by function pointers declared at file scope with the same name as the extension function API.  In this version I've appended an underscore to the extension function names.  I'm fine choosing a different convention instead, just let me know what to use, thanks!

### DIFF
--- a/test_conformance/extensions/cl_khr_dx9_media_sharing/main.cpp
+++ b/test_conformance/extensions/cl_khr_dx9_media_sharing/main.cpp
@@ -31,12 +31,12 @@ test_definition test_list[] = { ADD_TEST(context_create),
 
 const int test_num = ARRAY_SIZE(test_list);
 
-clGetDeviceIDsFromDX9MediaAdapterKHR_fn clGetDeviceIDsFromDX9MediaAdapterKHR =
+clGetDeviceIDsFromDX9MediaAdapterKHR_fn clGetDeviceIDsFromDX9MediaAdapterKHR_ =
     NULL;
-clCreateFromDX9MediaSurfaceKHR_fn clCreateFromDX9MediaSurfaceKHR = NULL;
-clEnqueueAcquireDX9MediaSurfacesKHR_fn clEnqueueAcquireDX9MediaSurfacesKHR =
+clCreateFromDX9MediaSurfaceKHR_fn clCreateFromDX9MediaSurfaceKHR_ = NULL;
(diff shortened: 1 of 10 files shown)


bashbaug:
Update: I recently added a `CL_NO_PROTOTYPES` control to the generated headers that can be used to omit all extension function prototypes.  Perhaps it makes more sense to define `CL_NO_PROTOTYPES` instead, in which case declaring function pointers at device scope with the same name as the extension function API will be just fine.

I'll create another PR with these changes and we can choose whichever one makes the most sense.

bashbaug:
I've converted this PR to a draft since I think we should consider #1325 instead, which solves the problem in a much simpler way.

bashbaug:
This is an old PR that is no longer needed, since we used a different solution instead.  Closing.

PR #1314, opened 2021-08-31
suppress MSVC strdup warning

bashbaug:
This PR adds the `_CRT_NONSTDC_NO_DEPRECATE` define for all files when building with MSVC, not just for the harness.  This suppresses a warning related to `strdup` that was being generated for nearly all files, since `strdup` was called from the commonly included header `errorHelpers.h`.


kpet (approved):

bashbaug:
Merging since this is trivial (and eliminates at least some of the MSVC warning noise).

PR #1322, opened 2021-09-15
Avoid use of rand in test_rw_image_access_qualifier

StuartDBrady:
There is no particular reason to limit the domain of unsigned 32-bit integers in the input, so use genrand_int32 directly.

### DIFF
--- a/test_conformance/basic/test_rw_image_access_qualifier.cpp
+++ b/test_conformance/basic/test_rw_image_access_qualifier.cpp
@@ -98,10 +98,13 @@ int test_rw_image_access_qualifier(cl_device_id device_id, cl_context context, c
     return -1;
     }
 
+    MTdata mtData = init_genrand(gRandomSeed);
     /* Fill input array with random values */
     for (i = 0; i < size; i++) {
-        input[i] = (unsigned int)(rand()/((double)RAND_MAX + 1)*255);
+        input[i] = genrand_int32(mtData);
     }
+    free_mtdata(mtData);
+    mtData = NULL;
 
     /* Zero out output array */
     for (i = 0; i < size; i++) {


StuartDBrady:
On second thoughts, I'm a bit confused by this code.  It uses the image channel order of `CL_RGBA` and image channel data type of `CL_UNSIGNED_INT32`, and yet the input values are uniformly distributed pseudo-random values in the domain 0x00000000..0x000000fe (or 0x00000000..0x000000ff with my change). Should the upper bits be used?

jlewis-austin:
> On second thoughts, I'm a bit confused by this code. It uses the image channel order of `CL_RGBA` and image channel data type of `CL_UNSIGNED_INT32`, and yet the input values are uniformly distributed pseudo-random values in the domain 0x00000000..0x000000fe (or 0x00000000..0x000000ff with my change). Should the upper bits be used?

Since the kernel adds 3 to each image value, I'm guessing it's just an arbitrary range that precludes overflow. I suppose it wouldn't hurt to use more of the range just to verify there's no problems in accessing all the bits correctly.

StuartDBrady:
> Since the kernel adds 3 to each image value, I'm guessing it's just an arbitrary range that precludes overflow. I suppose it wouldn't hurt to use more of the range just to verify there's no problems in accessing all the bits correctly.

I can't see any _need_ to preclude overflow, though, as arithmetic for unsigned integer types in both C++ and OpenCL C is guaranteed to be modular (with a modulus of the maximum value for the type plus one).  I think it's best to use genrand_int32 here, as anything else falsely implies a particular need to avoid certain test values.

In any case, we shouldn't be using rand – its implementation is platform-specific and AFAICS we would ideally want test reproducibility that is independent of the host platform.

lakshmih:

EwanC (approved):

bashbaug (approved):
LGTM, other than Ewan's comment about MTDataHolder.

lakshmih (approved):
Looks good.

bashbaug:
@StuartDBrady do you want to switch to MTdataHolder in this PR or do you want to merge as-is?

bashbaug:
Merging as discussed in the January 31st teleconference.

PR #1323, opened 2021-09-15
gles: Fix double frees.

lakshmih:
Remove a few explicit frees in the redirect_buffers test which are
already handled by a wrapper.

### DIFF
--- a/test_conformance/gles/test_buffers.cpp
+++ b/test_conformance/gles/test_buffers.cpp
@@ -205,6 +205,12 @@ int test_buffer_kernel(cl_context context, cl_command_queue queue, ExplicitType 
   if (validate_only) {
     int result = (CheckGLObjectInfo(streams[0], CL_GL_OBJECT_BUFFER, (GLuint)inGLBuffer, (GLenum)0, 0) |
                   CheckGLObjectInfo(streams[2], CL_GL_OBJECT_BUFFER, (GLuint)outGLBuffer, (GLenum)0, 0) );
+
+    for (i = 0; i < 3; i++)
+    {
+        streams[i].reset();
+    }
+
     glDeleteBuffers(1, &inGLBuffer);    inGLBuffer = 0;
     glDeleteBuffers(1, &outGLBuffer);    outGLBuffer = 0;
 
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
Was there truly a double free before?  Or did the `NULL` assignment prevent additional frees?

The only reason we might want to consider keeping this code as-is is because using the wrapper does change the order of the deletion slight: before this change the CL object was deleted first and the GL object was deleted second, whereas after this change the GL object is deleted first and then the CL object is deleted second.

The spec says ([link](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_gl_sharing__memobjs-lifetime-of-shared-objects)):

> If the GL object is deleted through the GL API (e.g. glDeleteBuffers, glDeleteTextures, or glDeleteRenderbuffers), subsequent use of the CL buffer or image object will result in undefined behavior, including but not limited to possible CL errors and data corruption, but may not result in program termination.

It's not entirely clear to me whether releasing the CL object would count as "use of" the CL object, but it could be considered a grey area.

lakshmih:
There was a definite double-free - one from the explicit clReleaseMemObject and the other from assignment to NULL (which, in the wrapper calls reset and hence clReleaseMemObject once more. Almost exactly the same as other double free fixes that have been merged already, eg:
https://github.com/KhronosGroup/OpenCL-CTS/commit/4a03bb79cb8fbd6012b02783e59565cce0b1f376#diff-9453eebbcfb207de5f012c7b24338159957a13907bfc0fa1f2d779cc3cdbf7d6L351

However, the right thing in this case might be to call a reset on the wrapper object like in 
https://github.com/KhronosGroup/OpenCL-CTS/commit/12637114ac81d292861daf4bff2397a36581f712#diff-b203bd4581ecbd7f355b29ddc5c8c6d6ffc0626527d5c7dc7721a95e98d7086eL287

This would retain the relative order of release between CL/GL that you brought up.

I'll make that fix.

bashbaug (approved):
Thanks, latest version LGTM!

lakshmih:
@kpet  can you review as well so we have at least two people that looked at it?

EwanC (approved):

PR #1324, opened 2021-09-15
Fix test_api get_command_queue_info

### LINKED ISSUE #1295: test_queries covers unsupported command queue property combinations
`check_get_command_queue_info_params()` tests various property combinations by selecting from a list of 8 flagsets:
```
0,
CL_QUEUE_PROFILING_ENABLE,
CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
CL_QUEUE_PROFILING_ENABLE | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
CL_QUEUE_ON_DEVICE | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
CL_QUEUE_PROFILING_ENABLE | CL_QUEUE_ON_DEVICE
        | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
CL_QUEUE_ON_DEVICE | CL_QUEUE_ON_DEVICE_DEFAULT
        | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
CL_QUEUE_PROFILING_ENABLE | CL_QUEUE_ON_DEVICE | CL_QUEUE_ON_DEVICE_DEFAULT
        | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE
```
It enables the first two at a minimum, the first four if host OUT_OF_ORDER queues are supported, and all 8 if on-device queues are supported (since they require out of order execution).

This requires any implementation that supports (OUT_OF_ORDER) device queues to also support OUT_OF_ORDER host queues, which is not a spec requirement.

jlewis-austin:
Decouple host and device out-of-order test enabling
Fixes #1295

### DIFF
--- a/test_conformance/api/test_queries.cpp
+++ b/test_conformance/api/test_queries.cpp
@@ -345,25 +345,21 @@ int command_queue_param_test(cl_command_queue queue,
     return 0;
 }
 
-#define MIN_NUM_COMMAND_QUEUE_PROPERTIES 2
-#define OOO_NUM_COMMAND_QUEUE_PROPERTIES 4
-static cl_command_queue_properties property_options[] = {
-    0,
+static cl_command_queue_properties host_required_options[] = {
+    0, CL_QUEUE_PROFILING_ENABLE
+};
 
-    CL_QUEUE_PROFILING_ENABLE,
-
+static cl_command_queue_properties host_out_of_order_options[] = {
     CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
+    CL_QUEUE_PROFILING_ENABLE | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE
+};
 
-    CL_QUEUE_PROFILING_ENABLE | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
-
+static cl_command_queue_properties device_options[] = {
     CL_QUEUE_ON_DEVICE | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
-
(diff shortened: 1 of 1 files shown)


kpet (commented):
This look correct to me but I'm tempted to suggest you change how it's done for maintainability. I think constructing a vector of all flag combinations to be tested would make the code easier to read and maintain. You could just append the contents of the various arrays containing possible flag combinations to that vector depending on what features are supported. The loop would become a simple range for.

test_conformance/api/test_queries.cpp
```diff
@@ -345,87 +346,100 @@ int command_queue_param_test(cl_command_queue queue,
...
+    const size_t host_optional_size = ARRAY_SIZE(host_optional);
+    const size_t device_required_size = ARRAY_SIZE(device_required);
+
+    Version version = get_device_cl_version(deviceID);
 
-    cl_queue_properties host_queue_props, device_queue_props;
-    cl_queue_properties queue_props[] = { CL_QUEUE_PROPERTIES, 0, 0 };
+    const cl_device_info host_queue_query = version >= Version(2, 0)
+        ? CL_DEVICE_QUEUE_ON_HOST_PROPERTIES
+        : CL_DEVICE_QUEUE_PROPERTIES;
```
bashbaug, line 376: Very minor: This is correct and debatably "more right" than the alternative, but `CL_DEVICE_QUEUE_ON_HOST_PROPERTIES` and `CL_DEVICE_QUEUE_PROPERTIES` have the same enum value so the OpenCL implementation being tested will see the same query regardless of the OpenCL version.

bashbaug (approved):
LGTM.  I've verified that the same queue properties are tested before and after this change.

kpet (approved):

PR #1325, opened 2021-09-16
define CL_NO_EXTENSION_PROTOTYPES

### LINKED ISSUE #1254: Unmodified Extension API Names

bashbaug:
This is an alternative to PR #1313.  I think this is a much more straightforward fix, so unless there is some other reason to want extension function prototypes in the near future I think we should go with this solution instead.

Fixes #1254.

In the latest version of the generated headers the `CL_NO_PROTOTYPES` define will suppress declaration of extension functions, so we can continue to declare function pointers with the same name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ add_definitions(-DCL_USE_DEPRECATED_OPENCL_2_0_APIS=1)
 add_definitions(-DCL_USE_DEPRECATED_OPENCL_1_2_APIS=1)
 add_definitions(-DCL_USE_DEPRECATED_OPENCL_1_1_APIS=1)
 add_definitions(-DCL_USE_DEPRECATED_OPENCL_1_0_APIS=1)
+add_definitions(-DCL_NO_PROTOTYPES)
 
 option(USE_CL_EXPERIMENTAL "Use Experimental definitions" OFF)
 if(USE_CL_EXPERIMENTAL)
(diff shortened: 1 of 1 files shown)


bashbaug:
I've changed this to define `CL_NO_EXTENSION_PROTOTYPE` instead, since `CL_NO_PROTOTYPES` will suppress declaration of all OpenCL API functions with the latest generated headers.

bashbaug:
Note:

* I've tested this change with the generated headers on Linux with GLES_IS_SUPPORTED=TRUE and GL_IS_SUPPORTED=TRUE.
* I've also tested this change with the generated headers on Windows with D3D10_IS_SUPPORTED=TRUE and D3D11_IS_SUPPORTED=TRUE.

We will need to merge this before merging the generated headers or the build will break.

PR #1326, opened 2021-09-16
Improve testing of kernel arg info in pipe_info test

StuartDBrady:
The test now checks that CL_KERNEL_ARG_INFO_NOT_AVAILABLE is returned when calling clGetKernelArgInfo() with offline compilation modes.

The correct function name is printed if clGetKernelArgInfo() fails when using online compilation (and not "clSetKernelArgInfo()").

When using online compilation, if the actual arg type is not as expected, the actual arg type is now logged, and the return value is now TEST_FAIL (-1) as per other failures (and not 1).

All other test pass/fail values used in the test now use TEST_PASS and TEST_FAIL instead of 0 and -1 literals.

An unnecessary cast of pipe_kernel_code has been removed.

### DIFF
--- a/test_conformance/pipes/test_pipe_info.cpp
+++ b/test_conformance/pipes/test_pipe_info.cpp
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 #include "procs.h"
+#include "harness/parseParameters.h"
 
 const char* pipe_kernel_code = {
     "__kernel void pipe_kernel(__write_only pipe int out_pipe)\n"
@@ -39,8 +40,7 @@ int test_pipe_info( cl_device_id deviceID, cl_context context, cl_command_queue 
 
     if (pipe_width != returnVal)
     {
-        log_error("Error in clGetPipeInfo() check of pipe packet size\n");
-        return -1;
+        test_fail("Error in clGetPipeInfo() check of pipe packet size\n");
     }
     else
     {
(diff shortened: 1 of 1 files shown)


test_conformance/pipes/test_pipe_info.cpp
```diff
@@ -40,7 +41,7 @@ int test_pipe_info( cl_device_id deviceID, cl_context context, cl_command_queue
     if (pipe_width != returnVal)
     {
         log_error("Error in clGetPipeInfo() check of pipe packet size\n");
-        return -1;
+        return TEST_FAIL;
```
jlewis-austin, line 44: test_fail() includes this functionality (here and 2 places below)

jlewis-austin (dismissed):
LGTM! Just one observation on error utilities; whether it improves readability is debatable, so I'll leave that up to you.

test_conformance/pipes/test_pipe_info.cpp
StuartDBrady: I'm not sure that the macros are named in the best way, but I've rewritten the code to use these macros.  Thanks for pointing this out!

jlewis-austin (dismissed):

test_conformance/pipes/test_pipe_info.cpp
```diff
@@ -52,29 +52,37 @@ int test_pipe_info( cl_device_id deviceID, cl_context context, cl_command_queue
...
     else
     {
         log_info( " CL_PIPE_MAX_PACKETS passed.\n" );
     }
 
-    err = create_single_kernel_helper_with_build_options(context, &program, &kernel, 1, (const char**)&pipe_kernel_code, "pipe_kernel", "-cl-std=CL2.0 -cl-kernel-arg-info");
-    test_error_ret(err, " Error creating program", -1);
+    err = create_single_kernel_helper_with_build_options(
+        context, &program, &kernel, 1, &pipe_kernel_code, "pipe_kernel",
+        "-cl-std=CL2.0 -cl-kernel-arg-info");
```
bashbaug, line 64: This isn't an issue you introduced, but if you feel up to fixing it: hard-coding `-cl-std=CL2.0` here will cause a problem for an OpenCL 3.0 implementation that supports pipes but not some other feature that is required for OpenCL C 2.0.  It'd be better to use a solution like `get_build_options()` in test_kernel_arg_info.cpp.

test_conformance/pipes/test_pipe_info.cpp
```diff
@@ -52,29 +52,37 @@ int test_pipe_info( cl_device_id deviceID, cl_context context, cl_command_queue
...
+    test_error_ret(err, "Error creating program", TEST_FAIL);
```
bashbaug, line 65: Minor/optional: This pattern can  be shortened to `test_err_fail(err, "<error message>")`.

(There's another case a few lines below also.)

test_conformance/pipes/test_pipe_info.cpp
```diff
@@ -52,29 +52,37 @@ int test_pipe_info( cl_device_id deviceID, cl_context context, cl_command_queue
...
+    if (gCompilationMode == kOnline)
     {
-        print_error(err, "ERROR: Bad type qualifier\n");
-        return -1;
+        test_error_ret(err, "clGetKernelArgInfo failed", TEST_FAIL);
+        if (arg_type_qualifier != CL_KERNEL_ARG_TYPE_PIPE)
+        {
+            test_fail("ERROR: Incorrect type qualifier: %i\n",
+                      arg_type_qualifier);
+        }
```
bashbaug, line 78: Minor/optional: This pattern can be shortened using `test_assert_error()`.

bashbaug (dismissed):
LGTM, I left a few comments for possible additional improvements, but this can be merged anytime.

test_conformance/pipes/test_pipe_info.cpp
StuartDBrady: A good suggestion, but I think that would go beyond the intended scope of these changes.

test_conformance/pipes/test_pipe_info.cpp
StuartDBrady: I introduced the `test_error_fail` macro so should have really remembered to use it here! (Except of course this change is quite a bit older than that macro and had sat unreviewed for a long time, and seemed fine as it still rebased cleanly... so my apologies for that.)

test_conformance/pipes/test_pipe_info.cpp
StuartDBrady: I'd prefer not as test_assert_error and similar macros prevent the inclusion of necessary detail in the failure message, instead only allowing a fixed message string. In my opinion, these assert macros are poor.

bashbaug (approved):
LGTM, this has been open a while so no need to wait further IMO.

PR #1327, opened 2021-09-16
api: Enable cl_khr_fp16 when using half types in kernel

lakshmih:


bashbaug:
Possibly related spec issue: https://github.com/KhronosGroup/OpenCL-Docs/issues/21

bashbaug (approved):
This seems reasonable to me - we only include the `half` arguments when the device support the fp16 extension, so it also makes sense to enable the fp16 extension via a pragma.

Debatably we should be able to test a pointer to a scalar `half` without requiring fp16 or an extension pragma, but that's also a bit of a grey area, and we usually include vector `half` types anyhow.

lakshmih:
Thanks @bashbaug  can you merge this?
I'm not sure how to get past the Build macos-latest check.

bashbaug:
It'd be nice to have a review from one more person before merging, just in case.  @kpet or @Jeremy-Kemp perhaps?

I restareted the CI jobs.  Let's see if that gets past the macos-latest check.  If it gets stuck again I think we'd be OK merging regardless, since it looks like it ran correctly, it just didn't report results for some reason.

Jeremy-Kemp (commented):
LGTM. All of the half arguments in the kernels are vectors, so I think its correct to enable the pragma.

lakshmih:
@bashbaug  Can we merge now since @Jeremy-Kemp said it looks good as well. Thanks!

Issue #1250, opened 2021-05-14
Image Helpers: implicit conversion from 'int' to 'float' changes value


mantognini:
Clang 12 (and possibly some older versions as well) report issues such as:

https://github.com/KhronosGroup/OpenCL-CTS/blob/657283799482433cc35d119ddab220dcad1f00aa/test_common/harness/imageHelpers.cpp#L2724-L2727

imageHelpers.cpp:2727:21: error: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Werror,-Wimplicit-const-int-float-conversion]

Another occurence:

https://github.com/KhronosGroup/OpenCL-CTS/blob/657283799482433cc35d119ddab220dcad1f00aa/test_common/harness/imageHelpers.cpp#L2893-L2896

jlewis-austin:
The first one looks like we're relying on the compiler's implementation of (int)2147483648.f to saturate back to 2147483647, which bypasses the purpose of the macro (ie, to provide RNE rounding, and saturation of out of range values). 32-bit integer conversions should promote to doubles for the comparisons, which would avoid the need for all the inexact magic going on in these.

jlewis-austin:
Clang 10.0.4 reports this as `[-Werror,-Wimplicit-int-float-conversion]`, bypassing the current CMakeLists.txt workaround. I'm working on a fix for the root cause

pierremoreau:
This was reported as #855. There was apparently some internal work on fixing it that was reported back in mid 2020, but there has been no update on that work since.

jlewis-austin:
Sorry for the delay, I'm putting up a PR now

Issue #855, opened 2020-07-27
Implicit conversion from int to float changes value of CL_INT_MAX


pierremoreau:
I am getting the following errors when compiling the current master branch using clang 10.0.0:

```
/home/pmoreau/Projects/OpenCL_for_Nouveau/src/OpenCL-CTS/test_common/harness/imageHelpers.cpp:2731:150: error: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Werror,-Wimplicit-int-float-conversion]
                ptr[ i ] = (int)CONVERT_INT( srcVector[ i ], MAKE_HEX_FLOAT( -0x1.0p31f, -1, 31), MAKE_HEX_FLOAT( 0x1.fffffep30f, 0x1fffffe, 30-23), CL_INT_MAX  );
                         ~                                                                                                                           ^~~~~~~~~~
/home/pmoreau/Projects/OpenCL_for_Nouveau/src/OpenCL-Headers/CL/cl_platform.h:250:29: note: expanded from macro 'CL_INT_MAX'
#define CL_INT_MAX          2147483647
                            ^~~~~~~~~~
/home/pmoreau/Projects/OpenCL_for_Nouveau/src/OpenCL-CTS/test_common/harness/imageHelpers.h:68:75: note: expanded from macro 'CONVERT_INT'
#define CONVERT_INT( v, min, max, max_val)  ( v < min ? min : ( v > max ? max_val : round_to_even( v ) ) )
                                                                          ^~~~~~~
/home/pmoreau/Projects/OpenCL_for_Nouveau/src/OpenCL-CTS/test_common/harness/imageHelpers.cpp:2890:188: error: implicit conversion from 'int' to 'float' changes value from 2147483647 to 2147483648 [-Werror,-Wimplicit-int-float-conversion]
                errors[i] = (cl_float)((cl_long) ptr[ i ] - (cl_long) CONVERT_INT( srcVector[ i ], MAKE_HEX_FLOAT( -0x1.0p31f, -1, 31), MAKE_HEX_FLOAT( 0x1.fffffep30f, 0x1fffffe, 30-23), CL_INT_MAX  ));
                                                          ~                                                                                                                                ^~~~~~~~~~
/home/pmoreau/Projects/OpenCL_for_Nouveau/src/OpenCL-Headers/CL/cl_platform.h:250:29: note: expanded from macro 'CL_INT_MAX'
#define CL_INT_MAX          2147483647
[...]

PR #1328, opened 2021-09-16
Fix compilation issue for cases where casts are required

lakshmih:
Fix compilation issues in:
- image_helpers
- subgroups


lakshmih:
@bashbaug  @kpet  can this be merged

lakshmih:
We don't need this anymore, it was seen on the visual studio toolchain at some point.

pierremoreau:
The compilation error was also triggered with clang 10.0.0+ (see #855 and #1250). As of #1252, it is no longer a compilation error (due to disabling -Werror for that particular type of warnings) but the warning in imageHelpers.cpp remains.

PR #1333, opened 2021-09-28
Experiment with macos CI build

alycm:
GitHub CI is in the process of migrating macos-latest to be v11 (rather than
v10.15).  Given our macos CI builds seem to not be reporting completion, but
nothing in the logs indicate why, experiment with an explicit version.

https://github.com/actions/virtual-environments/issues/4060

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         mainmatrix: [true]
-        os: [ubuntu-20.04, macos-latest]
+        os: [ubuntu-20.04, macos-10.15]
         include:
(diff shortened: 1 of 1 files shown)


alycm:
So this PR does not use macos-latest (I'm not proposing that as an actual change, just an experiment).

The GitHub action passes, but we still have the status 

image

I don't know where that `macos-latest` is coming from at all, as this branch doesn't contain that string.  The GitHub CI matrix doesn't contain that string (for this branch).

alycm:
Okay, it was a branch protection rule.  It was checking for `Build macos-latest ` when it should have been checking for `Build macos-latest`.   Note the trailing whitespace in the old rule.

PR #3, opened 2018-03-10
OpenCL 2.2 graphics fixes

pierremoreau:
* Fixes compilation when setting GL_IS_SUPPORTED:
    * due to CMakeLists.txt referencing non-existent files and missing an include folder;
    * due to gl/main.cpp referencing an undefined variable.
* Add OpenGL tests that are present but were not referenced by CMakeLists.txt;
* Expose the graphics interop test toggles via CMake, allowing the user to enable/disable them without modifying CMakeLists.txt;
* Add such a toggle for OpenGL ES tests.

PR #1337, opened 2021-10-04
Fix build, glext should not be used with GLEW

gwawiork:
This issue has been noticed by @pierremoreau some time ago
https://github.com/KhronosGroup/OpenCL-CTS/pull/3#issuecomment-486544336

This issue is described here as well: https://stackoverflow.com/questions/65009898/incompatible-function-prototypes-between-glew-h-and-glext-h

Because glew delivers more definitions than glext we should stay with glew.

### DIFF
--- a/test_common/gl/setup_x11.cpp
+++ b/test_common/gl/setup_x11.cpp
@@ -18,9 +18,6 @@
 #include "testBase.h"
 #include "harness/errorHelpers.h"
 
-#include <GL/gl.h>
-#include <GL/glut.h>
-#include <GL/freeglut.h>
 #include <GL/glx.h>
 #include <CL/cl_ext.h>
 
--- a/test_common/gl/setup_x11.cpp
+++ b/test_common/gl/setup_x11.cpp
@@ -13,7 +13,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#define GL_GLEXT_PROTOTYPES
 
 #include "setup.h"
 #include "testBase.h"
@@ -21,7 +20,6 @@
 
 #include <GL/gl.h>
 #include <GL/glut.h>
-#include <GL/glext.h>
 #include <GL/freeglut.h>
 #include <GL/glx.h>
 #include <CL/cl_ext.h>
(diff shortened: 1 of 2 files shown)


test_common/gl/setup_x11.cpp
```diff
@@ -13,15 +13,13 @@
...
-#define GL_GLEXT_PROTOTYPES
 
 #include "setup.h"
 #include "testBase.h"
 #include "harness/errorHelpers.h"
 
 #include <GL/gl.h>
 #include <GL/glut.h>
-#include <GL/glext.h>
 #include <GL/freeglut.h>
```
bashbaug, line 23: Could/should we clean up a couple of these other headers while we're at it?

Currently `setup.h` includes `gl_headers.h` which looks like it is intended to include all of the necessary OpenGL headers including those for glut or freeglut.  Do we need to include anything more than glx.h here?

test_common/gl/setup_x11.cpp
```diff
@@ -13,15 +13,13 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#define GL_GLEXT_PROTOTYPES
```
bashbaug, line 16: This change is good.

It looks setup_win32.cpp defines GL_GLEXT_PROTOTYPES also, even though win32 isn't using glext.h anymore either.  Could we clean it up too?

bashbaug (dismissed):
Mostly LGTM, just a few possible areas for additional cleanup.

Needs a non-Intel reviewer.

test_common/gl/setup_x11.cpp
gwawiork: I am ok. Removed in commit https://github.com/KhronosGroup/OpenCL-CTS/pull/1337/commits/87127a0046aa57381dbe51e729769ef572c4ab76

test_common/gl/setup_x11.cpp
gwawiork: removed in https://github.com/KhronosGroup/OpenCL-CTS/pull/1337/commits/4db06f50ee26d58454edceaebb6c076b25b718b0

bashbaug (commented):
Thanks, latest version LGTM.  Needs a non-Intel approval.

kpet (approved):

PR #1341, opened 2021-10-09
CTS: Fix compile warnings.

lakshmih:

General:
* Improve variable types.
* Make casts explicit to improve readability.
* Improve printf format specifiers, favoe hex values.
* Don't warn about using the portable "strdup" on Visual Studio.
* Limit the mark pragma to Apple environments since it's only useful in
  XCode and causes a warning otherwise.
* Move GLES3 macro from CMakeLists.txt to
  test_conformance/gles/CMakeLists.txt to be more specific.
* Remove unused variables.
* Fix type narrowing caused by assigning size_ts to ints. This also better
  matches the types of clEnqueueNDRangeKernel's local size, global size,
  and global offset.

test_common/gles/helpers.cpp:
* Remove needless realWidth and realHeight variables.

test_common/harness/compat.h:
* Don't redefine EXTERN_C on Windows where it's already in the OS
  headers.

test_common/harness/conversions.cpp:
* Fix warning about storing -1 into unsigned long upper limits.

test_common/harness/errorHelpers.h:
* Remove the custom compile time assert in favor of the standard
  "static_assert".

test_common/harness/imageHelpers.cpp:
* Use the standard "static_assert" as supported by Visual Studio 2015
  instead of the custom "ct_assert".

test_conformance/subgroups: Fix most compile warnings.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -991,7 +991,7 @@ float get_max_relative_error(const cl_image_format *format,
     {
 #if !defined(_WIN32)
 //#warning Implementations will likely wish to pick a max allowable sampling
-//error policy here that is better than the spec
+// error policy here that is better than the spec
 #endif
         // The spec allows linear filters to return any result most of the time.
         // That's fine for implementations but a problem for testing. After all


test_common/harness/imageHelpers.cpp
```diff
@@ -798,10 +798,10 @@ void get_max_sizes(
     if (image_type == CL_MEM_OBJECT_IMAGE1D)
     {
 
-        size_t M = maximum_sizes[0];
+        double M = static_cast<double>(maximum_sizes[0]);
```
jlewis-austin, line 801: This conversion looks unnecessary since `M` is never used as a double

test_common/harness/imageHelpers.cpp
```diff
@@ -815,17 +815,17 @@ void get_max_sizes(
         {
 
             // Determine the size of the fixed dimension
-            size_t M = maximum_sizes[fixed_dim];
-            size_t A = max_pixels;
+            double M = static_cast<double>(maximum_sizes[fixed_dim]);
```
jlewis-austin, line 818: This causes a change in behavior of the `A / M` divide below, is that intended? If so, it should probably be called out in its own PR. 
To preserve the integer divide, we could just get rid of the double types altogether by replacing `fmin` with `min`.

test_common/harness/imageHelpers.cpp
```diff
@@ -874,9 +874,9 @@ void get_max_sizes(
             assert(x0 > 0 && M > 0);
 
             // Store the size
-            sizes[(*numberOfSizes)][fixed_dim] = M;
-            sizes[(*numberOfSizes)][x0_dim] = x0;
-            sizes[(*numberOfSizes)][x1_dim] = x1;
+            sizes[(*numberOfSizes)][fixed_dim] = static_cast<size_t>(M);
+            sizes[(*numberOfSizes)][x0_dim] = static_cast<size_t>(x0);
```
jlewis-austin, line 878: Why do we need to cast here? Everything is already `size_t`. 
Slightly off-topic, the `x0` and `x1` variable names could be improved since they are not descriptive and look like hex literals.

test_common/harness/imageHelpers.cpp
```diff
@@ -990,7 +990,8 @@ float get_max_relative_error(const cl_image_format *format,
 #else
     {
 #if !defined(_WIN32)
-#warning Implementations will likely wish to pick a max allowable sampling error policy here that is better than the spec
+//#warning Implementations will likely wish to pick a max allowable sampling
+// error policy here that is better than the spec
```
jlewis-austin, line 994: This is worth some discussion. My assumption had been that this warning is desired to alert implementers about the lack of coverage. It would be better to allow it to be set externally so no code modifications are required, but that's out of the scope here.

test_common/CMakeLists.txt
```diff
@@ -21,3 +21,8 @@ set(HARNESS_SOURCES
 )
 
 add_library(harness STATIC ${HARNESS_SOURCES})
+
+if(MSVC)
+    # Don't warn about using the portable "strdup" function.
+    target_compile_definitions(harness PRIVATE _CRT_NONSTDC_NO_DEPRECATE)
```
bashbaug, line 27: I added this to the top-level cmake file a month or so ago: https://github.com/KhronosGroup/OpenCL-CTS/pull/1314

Do we need it here too?

test_common/harness/imageHelpers.cpp
```diff
@@ -1809,7 +1806,7 @@ static float unnormalize_coordinate(const char *name, float coord, float offset,
     switch (addressing_mode)
     {
         case CL_ADDRESS_REPEAT:
-            ret = RepeatNormalizedAddressFn(coord, static_cast<size_t>(extent));
+            ret = RepeatNormalizedAddressFn(coord, static_cast<size_t>( extent) );
```
polsenferreira: Are these diffs actually required?
I understand you're changing code styling here, but it doesn't seem to be necessary for this change, and it is inconsistent with other invocations of the static_cast<> elsewhere on the patch.

test_common/harness/imageHelpers.cpp
```diff
@@ -815,17 +815,17 @@ void get_max_sizes(
         {
 
             // Determine the size of the fixed dimension
-            size_t M = maximum_sizes[fixed_dim];
-            size_t A = max_pixels;
+            double M =  static_cast<double> (maximum_sizes[fixed_dim]);
+            double A =  static_cast<double> (max_pixels);
```
polsenferreira: There seems to be some extraneous spaces here.

test_common/gles/helpers.cpp
```diff
@@ -299,7 +299,7 @@ void * ReadGLTexture( GLenum glTarget, GLuint glTexture,
         GetGLFormatName(readBackFormat),
         GetGLTypeName(readBackType));
 
-    DumpGLBuffer(readBackType, outWidth, outHeight, (void *)outBuffer);
+    DumpGLBuffer(readBackType, outWidth, outHeight, (void*)outBuffer);
```
polsenferreira: Why not a c++ cast here? You seem to have done so elsewhere.

PR #1343, opened 2021-10-14
Remove excessive logging in subgroup tests

StuartDBrady:
This also adds some missing data type logging to the `subgroup_functions_non_uniform_vote` tests.

### DIFF
--- a/test_conformance/subgroups/test_subgroup_non_uniform_vote.cpp
+++ b/test_conformance/subgroups/test_subgroup_non_uniform_vote.cpp
@@ -34,20 +34,10 @@ template <typename T, NonUniformVoteOp operation> struct VOTE
         int last_subgroup_size = 0;
         ii = 0;
 
-        log_info("  sub_group_%s%s... \n",
+        log_info("  sub_group_%s%s(%s)... \n",
                  (operation == NonUniformVoteOp::elect) ? "" : "non_uniform_",
-                 operation_names(operation));
+                 operation_names(operation), TypeManager<T>::name());
 
-        log_info("  test params: global size = %d local size = %d subgroups "
-                 "size = %d data type (%s)\n",
-                 test_params.global_workgroup_size, nw, ns,
-                 TypeManager<T>::name());
-        log_info("               work items mask: %s\n",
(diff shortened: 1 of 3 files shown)


test_conformance/subgroups/test_subgroup_non_uniform_vote.cpp
```diff
@@ -34,20 +34,10 @@ template <typename T, NonUniformVoteOp operation> struct VOTE
         int last_subgroup_size = 0;
         ii = 0;
 
-        log_info("  sub_group_%s%s... \n",
+        log_info("  sub_group_%s%s(%s)... \n",
                  (operation == NonUniformVoteOp::elect) ? "" : "non_uniform_",
-                 operation_names(operation));
+                 operation_names(operation), TypeManager<T>::name());
```
gwawiork, line 39: it would be nice to have somewhere added info about mask testing. Otherwise it looks like the same cases runs in the loop:
  sub_group_non_uniform_any(int)... 
  sub_group_non_uniform_any(int)... passed
  sub_group_non_uniform_any(int)... 
  sub_group_non_uniform_any(int)... passed
  sub_group_non_uniform_any(int)... 
  sub_group_non_uniform_any(int)... passed

polsenferreira: While I agree with you, I think this might be better done on a separate commit.
The commit message specifies "remove excess logging". Your suggestion (which is worth pursuing) is probably outside the scope of this commit.

StuartDBrady: The mask testing is to be reworked so that it doesn't result in an excessive number of compilations, and preferably so that it doesn't result in an excessive number of invocations.  As such, the logging of masks will have to be rewritten, anyway.

In most cases the logging for subgroup extensions includes the expected output and the actual output, but lacks the input, so the logging of masks in many cases did not allow failures to be understood directly without modifying the CTS to add extra debug output - the mask output being removed here isn't necessarily of much use on its own.

gwawiork (approved):
LGTM

PR #1344, opened 2021-10-14
Use maximum subgroup size in sub_group_ballot tests

StuartDBrady:
`sub_group_ballot_bit_count()` and `sub_group_ballot_find_msb()` mask their input according to a subgroup size, which is assumed to be the maximum subgroup size, and not the actual subgroup size excluding non-existent work-items in the "remainder" subgroup.

Fix this as per the clarification made to the OpenCL C specification in revision 3.0.9 for issue KhronosGroup/OpenCL-Docs#626 by pull request KhronosGroup/OpenCL-Docs#689.

### DIFF
--- a/test_conformance/subgroups/test_subgroup_ballot.cpp
+++ b/test_conformance/subgroups/test_subgroup_ballot.cpp
@@ -496,7 +496,7 @@ template <typename Ty, BallotOp operation> struct BALLOT_COUNT_SCAN_FIND
                         | (bs128(mx[wg_offset + wi_id].s1) << 32)
                         | (bs128(mx[wg_offset + wi_id].s2) << 64)
                         | (bs128(mx[wg_offset + wi_id].s3) << 96);
-                    bs &= getImportantBits(wi_id, current_sbs);
+                    bs &= getImportantBits(wi_id, sbs);
                     device_result = my[wg_offset + wi_id].s0;
                     if (operation == BallotOp::ballot_inclusive_scan
                         || operation == BallotOp::ballot_exclusive_scan
@@ -516,7 +516,7 @@ template <typename Ty, BallotOp operation> struct BALLOT_COUNT_SCAN_FIND
                     }
                     else if (operation == BallotOp::ballot_find_lsb)
                     {
-                        for (int id = 0; id < current_sbs; ++id)
+                        for (int id = 0; id < sbs; ++id)
                         {
                             if (bs.test(id))
                             {
@@ -537,7 +537,7 @@ template <typename Ty, BallotOp operation> struct BALLOT_COUNT_SCAN_FIND
                     }
                     else if (operation == BallotOp::ballot_find_msb)
                     {
-                        for (int id = current_sbs - 1; id >= 0; --id)
+                        for (int id = sbs - 1; id >= 0; --id)
                         {
                             if (bs.test(id))
                             {


gwawiork:
> > > I am not sure if any changes to the test are needed.
> > 
> > 
> > Please read issue [KhronosGroup/OpenCL-Docs#626](https://github.com/KhronosGroup/OpenCL-Docs/issues/626).
> > If you need, you can also refer to PR [KhronosGroup/OpenCL-Docs#689](https://github.com/KhronosGroup/OpenCL-Docs/pull/689).
> > > We use subgroup size (sbs in templates) to calculate references to compare results with device results. I see that we set initially subgroup_size to value returned by CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/subgroups/subhelpers.h#L1502). Then we can say that test use CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE for calculation till the last subgroups which can be non uniform. Then for last subgroup when workgroup is non uniform [we recalculate ](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/subgroups/subhelpers.h#L1566) sizes for rest of workitems that should be taken into consideration while calculation reference values.
> > 
> > 
> > We need to use the maximum subgroup size within the dispatch. My understanding is that for sufficiently large dispatches, in practice this is the same as the maximum subgroup size within the the ndrange. I am not sure if we may need to increase the local work size for the ballot tests. We should perhaps use get_max_sub_group_size() directly, although this is dispatch-specific, so must be done within the _same_ dispatch as the test. It's not clear to me if the maximum value returned for CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE for all possible ndranges is allowed to be greater than the maximum subgroup size that would actually be used by any dispatch in practice. If so, then we _must_ use the get_max_sub_group_size() value instead.
> > In cases where the value returned for get_max_sub_group_size() varies depending on the dispatch, we should improve this testing to cover a range of maximum subgroup sizes.
> > In the mean time, could we please make this change? It at least get us closer to testing for correct behavior in the majority of cases.
> 
> In general I agree with this change but with this change the test is failing - sub_group_ballot_find_msb. Probably the test need additional fixes related with "active workitems". Despite of this LGTM.

looking only on active workitem bits  
`bs &= getImportantBits(wi_id, current_sbs);`
not always all bits 
`bs &= getImportantBits(wi_id, sbs);`
make passes.

StuartDBrady:
>> In general I agree with this change but with this change the test is failing - sub_group_ballot_find_msb. Probably the test need additional fixes related with "active workitems". Despite of this LGTM.

> looking only on active workitem bits `bs &= getImportantBits(wi_id, current_sbs);` not always all bits `bs &= getImportantBits(wi_id, sbs);` make passes.

Thanks - the CL implementation in question will need correcting, in that case - the msb operation is specified to work as per Vulkan, using the maximum subgroup size of the dispatch to determine which bits are "important" - the spec is defined in this way so as not to have an adverse effect on performance due to it not being possible to determine the size of the subgroup excluding initially inactive work-items at compile-time.

StuartDBrady:
Can this be reviewed, please?

gwawiork:
> > > In general I agree with this change but with this change the test is failing - sub_group_ballot_find_msb. Probably the test need additional fixes related with "active workitems". Despite of this LGTM.
> 
> > looking only on active workitem bits `bs &= getImportantBits(wi_id, current_sbs);` not always all bits `bs &= getImportantBits(wi_id, sbs);` make passes.
> 
> Thanks - the CL implementation in question will need correcting, in that case - the msb operation is specified to work as per Vulkan, using the maximum subgroup size of the dispatch to determine which bits are "important" - the spec is defined in this way so as not to have an adverse effect on performance due to it not being possible to determine the size of the subgroup excluding initially inactive work-items at compile-time.

Ok finally I think this fix is correct. Then LGTM

gwawiork (dismissed):
LGTM

kpet (approved):

PR #1346, opened 2021-10-22
test_subgroups - Set safe input values for half type and mul, add operations

gwawiork:
While arithmetic operations for half type (especially mul operation) we notice overflow appear in results. In this proposal I set 4 values different than 1 and for all other workitems 1. I think this will be safe no matter how big is subgroup.

### DIFF
--- a/test_conformance/subgroups/subgroup_common_templates.h
+++ b/test_conformance/subgroups/subgroup_common_templates.h
@@ -418,8 +418,6 @@ void fill_and_shuffle_safe_values(std::vector<cl_ulong> &safe_values,
         set_one = 1;
     }
 
-    std::random_device random_device;
-    std::mt19937 mersenne_twister_engine(random_device());
     int offset_size = sb_size - safe_values.size();
     if (offset_size > 0)
     {
@@ -430,6 +428,7 @@ void fill_and_shuffle_safe_values(std::vector<cl_ulong> &safe_values,
         safe_values.resize(sb_size);
     }
 
+    std::mt19937 mersenne_twister_engine(10000);
     std::shuffle(safe_values.begin(), safe_values.end(),
                  mersenne_twister_engine);
 };
@@ -439,19 +438,18 @@ void genrand(Ty *x, Ty *t, cl_int *m, int ns, int nw, int ng)
 {
     int nj = (nw + ns - 1) / ns;
 
+    std::vector<cl_ulong> safe_values;
+    if (operation == ArithmeticOp::mul_ || operation == ArithmeticOp::add_)
+    {
+        fill_and_shuffle_safe_values<Ty>(safe_values, ns);
+    }
+
     for (int k = 0; k < ng; ++k)
     {
         for (int j = 0; j < nj; ++j)
         {
             int ii = j * ns;
             int n = ii + ns > nw ? nw - ii : ns;
-
-            std::vector<cl_ulong> safe_values;
-            if (operation == ArithmeticOp::mul_
-                || operation == ArithmeticOp::add_)
-            {
-                fill_and_shuffle_safe_values<Ty>(safe_values, ns);
-            }
 
             for (int i = 0; i < n; ++i)
             {
--- a/test_conformance/subgroups/subgroup_common_templates.h
+++ b/test_conformance/subgroups/subgroup_common_templates.h
@@ -393,27 +393,52 @@ template <typename Ty> bool is_floating_point()
         || std::is_same<Ty, subgroups::cl_half>::value;
 }
 
+// limit possible input values to avoid arithmetic rouding/overflow issues.
+// for each subgroup values defined different values
+// for rest of workitems set 1
+// shuffle values
+template <typename Ty>
+void fill_and_shuffle_safe_values(std::vector<cl_ulong> &safe_values,
+                                  const int &sb_size)
+{
+    // below values are 0.5f, 0.25f, 2.0f, 4.0f
+    std::vector<cl_ulong> safe_half_values{ 0x3800, 0x3400, 0x4000, 0x4400 };
+    // max possible value is == 5040
+    std::vector<cl_ulong> safe_other_types_values{ 2, 3, 4, 5, 6, 7 };
+
+    cl_ulong set_one = 0x3c00;
+
+    if (std::is_same<Ty, subgroups::cl_half>::value)
+    {
+        safe_values = safe_half_values;
+    }
+    else
+    {
+        safe_values = safe_other_types_values;
+        set_one = 1;
+    }
+
+    std::random_device random_device;
+    std::mt19937 mersenne_twister_engine(random_device());
+    int offset_size = sb_size - safe_values.size();
+    if (offset_size > 0)
+    {
+        safe_values.insert(safe_values.end(), offset_size, set_one);
+    }
+    else
+    {
+        safe_values.resize(sb_size);
+    }
+
+    std::shuffle(safe_values.begin(), safe_values.end(),
+                 mersenne_twister_engine);
+};
+
 template <typename Ty, ArithmeticOp operation>
 void genrand(Ty *x, Ty *t, cl_int *m, int ns, int nw, int ng)
 {
     int nj = (nw + ns - 1) / ns;
 
-    // limit possible input values to avoid arithmetic rouding/overflow issues.
-    // for each subgroup for 4 workitems set 4 values different than 1
-    // for rest of workitems set 1
-    // shuffle values
-    // below values are 0.5f, 0.25f, 2.0f, 4.0f
-    std::vector<cl_ulong> safe_half_values{ 0x3800, 0x3400, 0x4000, 0x4400 };
-    if (std::is_same<Ty, subgroups::cl_half>::value)
-    {
-        std::random_device random_device;
-        std::mt19937 mersenne_twister_engine(random_device());
-        safe_half_values.insert(safe_half_values.end(),
-                                ns - safe_half_values.size(), 0x3c00);
-
-        std::shuffle(safe_half_values.begin(), safe_half_values.end(),
-                     mersenne_twister_engine);
-    }
     for (int k = 0; k < ng; ++k)
     {
         for (int j = 0; j < nj; ++j)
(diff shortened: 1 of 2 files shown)


test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -394,6 +396,22 @@ void genrand(Ty *x, Ty *t, cl_int *m, int ns, int nw, int ng)
...
+    // limit possible input values to avoid arithmetic rouding/overflow issues.
+    // for each subgroup for 4 workitems set 4 values different than 1
+    // for rest of workitems set 1
+    // shuffle values
+    // below values are 0.5f, 0.25f, 2.0f, 4.0f
+    std::vector<cl_ulong> safe_half_values{ 0x3800, 0x3400, 0x4000, 0x4400 };
+    if (std::is_same<Ty, subgroups::cl_half>::value)
+    {
+        std::random_device random_device;
+        std::mt19937 mersenne_twister_engine(random_device());
```
jlewis-austin, line 408: In my experience we generally avoid randomizing seeds for testing, and never without capturing and reporting the seed value so it can be reproduced. Is there a reason we need it here?

test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -394,6 +396,22 @@ void genrand(Ty *x, Ty *t, cl_int *m, int ns, int nw, int ng)
 {
```
jlewis-austin, line 396: I'd recommend changing the function name to something more specific to test_subgroups, like gen_inputs or such. genrand makes it look like a random number generator, but it's much more specialized.

test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -394,6 +396,22 @@ void genrand(Ty *x, Ty *t, cl_int *m, int ns, int nw, int ng)
...
     int nj = (nw + ns - 1) / ns;
 
+    // limit possible input values to avoid arithmetic rouding/overflow issues.
```
jlewis-austin, line 399:
```suggestion
    // limit possible input values to avoid arithmetic rounding/overflow issues.
```

jlewis-austin (commented):
I made a few comments, but still need to review the rest of the infrastructure to remember how these numbers are being used and how results are validated. It's not clear to me yet what the danger is with overflow or rounding. Do random floats not have the same concerns? How does 0 cause overflow in multiplication?

gwawiork:
> I made a few comments, but still need to review the rest of the infrastructure to remember how these numbers are being used and how results are validated. It's not clear to me yet what the danger is with overflow or rounding. Do random floats not have the same concerns? How does 0 cause overflow in multiplication?

Thanks for your comments.
Regarding overflow for multiplication or summation. The test is ready to handle max = 128 work items in the subgroup (but even for lower values like 32 I've noticed overflow). Then in the worst condition when all work items are active the operation can be executed 128 times. In case of previous values were 1,2,3,4. it was easy to reach max value of cl_half type which is 65504 (largest normal number). I agree that for larger subgroups even results for cl_float can be overflown so I can extend using `safe_half_values` even for this data type. 0 value is removed  because most results for multiplication would be calculated as 0.

test_conformance/subgroups/subgroup_common_templates.h
gwawiork: I've set constant seed - see latest

test_conformance/subgroups/subgroup_common_templates.h
gwawiork: done

test_conformance/subgroups/subgroup_common_templates.h
gwawiork: done

test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -388,11 +390,57 @@ template <typename Ty> bool is_floating_point()
...
+template <typename Ty>
+void fill_and_shuffle_safe_values(std::vector<cl_ulong> &safe_values,
+                                  const int &sb_size)
+{
+    // below values are 0.5f, 0.25f, 2.0f, 4.0f
+    std::vector<cl_ulong> safe_half_values{ 0x3800, 0x3400, 0x4000, 0x4400 };
+    // max possible value is == 5040
+    std::vector<cl_ulong> safe_other_types_values{ 2, 3, 4, 5, 6, 7 };
+
+    cl_ulong set_one = 0x3c00;
```
bashbaug, line 406: Minor: I'd find this easier to follow if the intial value was `1` and was only overridden to `0x3c00` for halfs (though see comment above).

test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -388,11 +390,57 @@ template <typename Ty> bool is_floating_point()
...
+// limit possible input values to avoid arithmetic rounding/overflow issues.
+// for each subgroup values defined different values
+// for rest of workitems set 1
+// shuffle values
+template <typename Ty>
+void fill_and_shuffle_safe_values(std::vector<cl_ulong> &safe_values,
+                                  const int &sb_size)
+{
+    // below values are 0.5f, 0.25f, 2.0f, 4.0f
+    std::vector<cl_ulong> safe_half_values{ 0x3800, 0x3400, 0x4000, 0x4400 };
```
bashbaug, line 402: Does this really do what we want it to do?  Since we take the incoming cl_ulong value then convert to float, then convert to half, it seems like we will generate very large half values instead of 0.5h, 0.25h, 2.0h, and 4.0h as intended:

```c++
// set scalar value to half scalar
template <typename Ty>
typename std::enable_if<TypeManager<Ty>::is_sb_scalar_type::value>::type
set_value(Ty &lhs, const cl_ulong &rhs)
{
    lhs.data = cl_half_from_float(static_cast<cl_float>(rhs), g_rounding_mode);
}
```

test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -390,51 +390,73 @@ template <typename Ty> bool is_floating_point()
...
 
-template <typename Ty, ArithmeticOp operation>
-void genrand(Ty *x, Ty *t, cl_int *m, int ns, int nw, int ng)
+// limit possible input values to avoid arithmetic rouding/overflow issues.
+// for each subgroup values defined different values
+// for rest of workitems set 1
+// shuffle values
+template <typename Ty>
+void fill_and_shuffle_safe_values(std::vector<cl_ulong> &safe_values,
+                                  const int &sb_size)
```
PiotrFusik, line 399: Pass `sb_size` by value, not reference.

test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -388,11 +390,57 @@ template <typename Ty> bool is_floating_point()
...
+    }
+
+    int offset_size = sb_size - safe_values.size();
+    if (offset_size > 0)
+    {
+        safe_values.insert(safe_values.end(), offset_size, set_one);
+    }
+    else
+    {
+        safe_values.resize(sb_size);
```
StuartDBrady, line 425: Does this mean that for a subgroup size less than four, not all of the values in safe_half_values will get used, and for a subgroup size less than six, not all of the values in safe_other_type_values will get used.  Is this really appropriate?

test_conformance/subgroups/subgroup_common_templates.h
gwawiork: I are right. Fixed in latest

test_conformance/subgroups/subgroup_common_templates.h
gwawiork: now all types take part from non_one_values

test_conformance/subgroups/subgroup_common_templates.h
gwawiork: fixed

bashbaug (commented):
Thanks!  Latest version LGTM.  Need a non-Intel approver.

test_conformance/subgroups/subgroup_common_templates.h
StuartDBrady: The same question is now equally valid for non_one_values, though!

@jlewis-austin, do you have any thoughts or opinions about this?  I cannot see a good reason that the numerical coverage for small subgroup sizes should be less than that for large subgroup sizes, but perhaps I've missed something in Grzegorz's code?

gwawiork: I am not sure where is the problem. I need to prepare values for particular subgroup size. There is no ability to set more values that subgroup size that why this limitation.

PiotrFusik: > Does this mean that for a subgroup size less than four, not all of the values in safe_half_values will get used, and for a subgroup size less than six, not all of the values in safe_other_type_values will get used.

Yes (now with `non_one_values`).

> Is this really appropriate?

Yes, why not? As @gwawiork noted, we cannot have more values than the subgroup size. And we cannot have too many big numbers in order to prevent multiplication overflows.

PiotrFusik: Also, the whole point of these "magical values" is that they generate no multiplication (or addition) overflow. Testing with the small ones is just as good as the larger ones.

jlewis-austin: > The same question is now equally valid for non_one_values, though!
> 
> @jlewis-austin, do you have any thoughts or opinions about this? I cannot see a good reason that the numerical coverage for small subgroup sizes should be less than that for large subgroup sizes, but perhaps I've missed something in Grzegorz's code?

@StuartDBrady I still have some questions, but haven't had time to dig deep enough to understand the big picture. It seems like coverage would benefit from having more unique values involved, but I'm OK with this letting this go in while we chew on it.

gwawiork:
Could somebody review and merge these changes if no objects?
Latest force-pushed is after rebase with master. 
Thanks !

StuartDBrady (approved):
Looks good to me.

This test still needs work, though – for instance, it seems that for `float` and `double` (and strictly for `half` too, but perhaps with less concern), the test will still pass even if the inputs, intermediate products and/or the final product are rounded to a lesser precision.

StuartDBrady:
If I've understood correctly, the expected product for the mul operation will always be _exactly_ 720.0 for subgroup sizes greater than 5 elements. I can't even see that this tests negative values, something which could be added trivially. (I'm unsure if there's a requirement for subgroup sizes to be powers of two.)

gwawiork:
> If I've understood correctly, the expected product for the mul operation will always be _exactly_ 720.0 for subgroup sizes greater than 4 elements (i.e. at least 5 elements – I'm unsure if there's a requirement for subgroup sizes to be powers of two.) I can't even see that this tests negative values, something which could be added trivially.

In case all work items are active in the subgroup it will be 720

gwawiork:
I understand there is no more objections then please merge.

PR #1349, opened 2021-11-02
Enable mipmap extension pragmas

### LINKED ISSUE #598: image mipmap tests do not enable extensions
The clang OpenCL headers do not expose the image mipmap functions unless cl_khr_mipmap_image* are defined.  However the CTS generated kernel sources do not actually enable the extensions.  I assume they should be enabling the extensions before using them?

Jeremy-Kemp:
Fixes #598

### DIFF
--- a/test_conformance/images/kernel_read_write/test_iterations.cpp
+++ b/test_conformance/images/kernel_read_write/test_iterations.cpp
@@ -39,26 +39,28 @@ static size_t reduceImageSizeRange(size_t maxDimSize) {
 }
 
 const char *read2DKernelSourcePattern =
-"%s\n"
-"__kernel void sample_kernel( read_only %s input,%s __global float *xOffsets, __global float *yOffsets, __global %s%s *results %s)\n"
-"{\n"
-"%s"
-"   int tidX = get_global_id(0), tidY = get_global_id(1);\n"
-"%s"
-"%s"
-"   results[offset] = read_image%s( input, imageSampler, coords %s);\n"
-"}";
+    "%s\n"
+    "__kernel void sample_kernel( read_only %s input,%s __global float "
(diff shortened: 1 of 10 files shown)


bashbaug:
Is this PR still required?  It looks like recent versions of Clang will compile kernels that use the mipmap built-in functions without an extension pragma.  Example: https://godbolt.org/z/dG65aTsTj

Note that Clang does currently require a kernel that uses the mipmap built-in functions to be compiled for OpenCL 2.0 or newer, so compiling with `-cl-std=CL1.2` (or omitting `-cl-std` entirely) will generate an error.  I'm not sure if this is a bug or intentional, but noting it for completeness.

If the PR is still required we should update the conclusions in https://github.com/KhronosGroup/OpenCL-Docs/issues/82, which currently include the mipmap extensions in the set of "extensions that do not require a pragma".

Jeremy-Kemp:
It looks like the requirement to supply pragma was dropped in Clang 13, however we're still supporting implementations using Clang < 13.

Let me double check with our compiler team whether or not we *need* the pragma for these extensions to be present, or if we can patch Clang to allow this extension without the pragma.

bashbaug:
In case it's helpful, I tested via Compiler Explorer all the way back to Clang 4 and it seems to work without the pragma, so the support has been around for a while, albeit still for OpenCL C 2.0 or newer: https://godbolt.org/z/MaxTYqxvs

bashbaug (approved):
I'm fine merging this as-is if it will help to make forward progress.  As observed, the pragmas shouldn't hurt anything, and this extension is already in the set of "should not produce a diagnostic" extensions, even if the pragma is not strictly required (see [comment](https://github.com/KhronosGroup/OpenCL-Docs/issues/82#issuecomment-786411529)).

I've verified that the only changes with this PR are the extension pragmas.

PR #1352, opened 2021-11-09
Improve error handling in subgroup tests

StuartDBrady:


test_common/harness/errorHelpers.h
```diff
@@ -62,6 +62,7 @@ static int vlog_win32(const char *format, ...);
         return TEST_FAIL;                                                      \
     }
 #define test_error(errCode, msg) test_error_ret(errCode, msg, errCode)
```
jlewis-austin, line 64: Is there a use for this version, or should it be noted as deprecated?

jlewis-austin (approved):
One question, but LGTM

test_common/harness/errorHelpers.h
StuartDBrady: test_error and test_error_ret are used in plenty of places.

You can see below that test_error checks if errCode is CL_SUCCESS, and if it is not, prints the specified error message (in the format "ERROR: <msg>! (<CL_ERROR_NAME> from <filename>:<lineno>)")

test_error_ret is for use in cases where the return value of the function from which it is being used is not a CL error code.  Returning TEST_FAIL is a common enough case that it makes sense to have a dedicated function for it, IMO.

There are many places in the code that return a CL error to the harness which I would argue should return a test_status value instead. These don't currently cause a problem because we don't have a CL error equal to -100 (TEST_SKIPPED_ITSELF) yet, but IMO it would be better to use the enum consistently rather than relying on TEST_PASS happening to be equal to CL_SUCCESS.

gwawiork (approved):
LGTM

jlewis-austin (approved):

PR #1355, opened 2021-11-10
Fix conversion data loss in test api min max constant args

BKoscielak:
This change doesn't touch printing, as it should be fixed by https://github.com/KhronosGroup/OpenCL-CTS/pull/1212


jlewis-austin (approved):
LGTM!

gwawiork (approved):
LGTM

PR #1358, opened 2021-11-18
Fix stack-use-after-scope crash in conversions

jrprice:
The way that program sources were being constructed involved capturing
pointers to strings that were allocated on the stack, and then trying
to use them outside of that scope. This change uses a stringstream
defined in the outer scope to build the program instead.


dneto0 (approved):
LGTM.

Thanks!

kpet (approved):

PR #1359, opened 2021-11-24
Fix conversion data loss in test api min max constant buffer size

gwawiork:
It fix issue when bigger sizes above UINT_MAX are reported. Then test works and display results incorrectly.

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -1735,11 +1735,12 @@ int test_min_max_constant_buffer_size(cl_device_id deviceID, cl_context context,
         test_error(error, "clEnqueueReadBuffer failed");
 
         for (i = 0; i < numberOfInts; i++)
-            if (constantData[i] != resultData[i]) {
+            if (constantData[i] != resultData[i])
+            {
                 log_error("Data failed to verify: constantData[%zu]=%d != "
                           "resultData[%zu]=%d\n",
-                          "resultData[%d]=%d\n",
-                          i, constantData[i], i, resultData[i]);
+                          "resultData[%d]=%d\n", i, constantData[i], i,
+                          resultData[i]);
                 free(constantData);
                 free(resultData);
                 free_mtdata(d);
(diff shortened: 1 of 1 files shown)


BKoscielak (approved):

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1297,7 +1298,7 @@ int test_min_max_constant_buffer_size(cl_device_id deviceID, cl_context context,
     size_t    threads[1], localThreads[1];
     cl_int *constantData, *resultData;
     cl_ulong maxSize, stepSize, currentSize, maxGlobalSize, maxAllocSize;
-    int i;
+    cl_ulong i;
```
bashbaug, line 1301: Somewhat pedantic: should this be a `size_t` instead of a `cl_ulong`?  It seems to be used as a loop counter over `numberOfInts` elements, and `numberOfInts` is a `size_t`.

If we do switch to a `size_t` please check the printf formats specifiers also, to be sure we're properly using `%zu` vs. `%llu`.

bashbaug (commented):
Mostly LGTM but have one suggestion for improvement.

Also needs a rebase and a non-Intel reviewer.  Thanks!

test_conformance/api/test_api_min_max.cpp
gwawiork: fixed

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1667,8 +1668,8 @@ int test_min_max_constant_buffer_size(cl_device_id deviceID, cl_context context,
         /* Test running the kernel and verifying it */
         threads[0] = numberOfInts;
         localThreads[0] = 1;
-        log_info("Filling constant buffer with %d cl_ints (%d bytes).\n",
-                 (int)threads[0], (int)(threads[0] * sizeof(cl_int)));
+        log_info("Filling constant buffer with %lld cl_ints (%lld bytes).\n",
+                 threads[0], threads[0] * sizeof(cl_int));
```
bashbaug, line 1672:
```suggestion
        log_info("Filling constant buffer with %zu cl_ints (%zu bytes).\n",
                 threads[0], threads[0] * sizeof(cl_int));
```

Pedantic: `threads` is a `size_t`.

bashbaug (commented):
Latest version LGTM.  Needs a non-Intel reviewer.

bashbaug:
I checked with @gwawiork and even though this PR is old these changes are still relevant.  The PR has been rebased.  Adding "focused review".

PR #1363, opened 2021-11-26
Tidy up code to determine bit mask for ballot scans

StuartDBrady:
It seems more intuitive to set only the bits that are required, rather than to set one more bit than is required, only to clear it again.


gwawiork (approved):
LGTM

PR #1365, opened 2021-12-07
Add cl_khr_command_buffer to list of extensions

EwanC:
cl_khr_command_buffer is now public as a provisional khr extension, which implementations may report. This extension name needs to be added to the hardcoded list of extensions in the compiler tests


kpet (approved):

kpet:
Merging with only one review since this is trivial.

PR #1368, opened 2021-12-17
Initial command-buffer extension tests

EwanC:
Introduce some basic testing of the [cl_khr_command_buffer](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_command_buffer) provisional extension. This code is intended as a starting point from which we can iteratively build up tests for the extension.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
@@ -1,0 +1,532 @@
+//
+// Copyright (c) 2021 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "command_buffer_test_base.h"
+#include "procs.h"
+#include "harness/typeWrappers.h"
+
+#include <algorithm>
+#include <cstring>
+#include <vector>
+
+#define CHECK_VERIFICATION_ERROR(reference, result, index)                     \
+    {                                                                          \
+        if (reference != result)                                               \
+        {                                                                      \
+            log_error("Expected %d was %d at index %u\n", reference, result,   \
+                      index);                                                  \
+            return TEST_FAIL;                                                  \
+        }                                                                      \
+    }
+
+namespace {
+
+// Helper test fixture for constructing OpenCL objects used in testing
+// a variety of simple command-buffer enqueue scenarios.
+struct BasicCommandBufferTest : CommandBufferTestBase
+{
+
+    BasicCommandBufferTest(cl_device_id device, cl_context context,
+                           cl_command_queue queue)
+        : CommandBufferTestBase(device), context(context), queue(queue),
+          command_buffer(nullptr), simultaneous_use(false), num_elements(0)
+    {}
+
+    cl_int TearDown()
+    {
+        if (nullptr != command_buffer)
+        {
+            cl_int error = clReleaseCommandBufferKHR(command_buffer);
+            test_error(error, "clReleaseCommandBufferKHR failed");
+        }
+        return CL_SUCCESS;
+    }
+
+    bool Skip()
+    {
+        cl_command_queue_properties queue_properties;
+        cl_int error = clGetDeviceInfo(
+            device, CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR,
+            sizeof(queue_properties), &queue_properties, NULL);
+        test_error(error,
+                   "Unable to query "
+                   "CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR");
+
+        // Skip if queue properties required, tests don't account for this
+        return (error || queue_properties);
+    }
+
+    cl_int SetUp(int elements)
+    {
+        cl_int error = init_extension_functions();
+        if (error != CL_SUCCESS)
+        {
+            return error;
+        }
+
+        // Query if device supports simultaneous use
+        cl_device_command_buffer_capabilities_khr capabilities;
+        error =
+            clGetDeviceInfo(device, CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR,
+                            sizeof(capabilities), &capabilities, NULL);
+        test_error(error, "Unable to query CL_COMMAND_BUFFER_PROPERTIES_KHR");
+        simultaneous_use =
+            capabilities & CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR;
+
+        if (elements <= 0)
+        {
+            return CL_INVALID_VALUE;
+        }
+        num_elements = static_cast<size_t>(elements);
+
+
+        // Kernel performs a parallel copy from an input buffer to output buffer
+        // is created.
+        const char *kernel_str =
+            R"(
+        __kernel void copy(__global int* in, __global int* out) {
+            size_t id = get_global_id(0);
+            out[id] = in[id];
+        })";
+        const size_t lengths[1] = { std::strlen(kernel_str) };
+
+        program =
+            clCreateProgramWithSource(context, 1, &kernel_str, lengths, &error);
+        test_error(error, "Failed to create program with source");
+
+        error = clBuildProgram(program, 1, &device, nullptr, nullptr, nullptr);
+        test_error(error, "Failed to build program");
+
+        in_mem = clCreateBuffer(context, CL_MEM_READ_ONLY,
+                                sizeof(cl_int) * num_elements, nullptr, &error);
+        test_error(error, "clCreateBuffer failed");
+
+        out_mem =
+            clCreateBuffer(context, CL_MEM_WRITE_ONLY,
+                           sizeof(cl_int) * num_elements, nullptr, &error);
+        test_error(error, "clCreateBuffer failed");
+
+        kernel = clCreateKernel(program, "copy", &error);
+        test_error(error, "Failed to create copy kernel");
+
+        error = clSetKernelArg(kernel, 0, sizeof(in_mem), &in_mem);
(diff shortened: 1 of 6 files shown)


test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -0,0 +1,532 @@
...
+
+    BasicCommandBufferTest(cl_device_id device, cl_context context,
+                           cl_command_queue queue)
+        : CommandBufferTestBase(device), context(context), queue(queue),
+          command_buffer(nullptr), simultaneous_use(false), num_elements(0)
+    {}
+
+    cl_int TearDown()
+    {
+        if (nullptr != command_buffer)
```
kpet, line 49: This works fine but I think we'll probably want to extend the object wrappers at https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/typeWrappers.h

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -0,0 +1,532 @@
...
+        return CL_SUCCESS;
+    }
+
+    /*
+     * Subtests returning an OpenCL error code
+     */
+
+    // Test enqueuing a command-buffer containing a single NDRange command
+    // once
+    cl_int BasicEnqueue();
```
kpet, line 155: Maybe you could have a `Run` function and derive objects for individual test cases?

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -0,0 +1,532 @@
...
+    if (test_fixture.Skip())
+    {
+        return TEST_SKIPPED_ITSELF;
+    }
+
+    error = test_fixture.BasicEnqueue();
+    test_error_ret(error, "Test Failed", TEST_FAIL);
+
+    error = test_fixture.TearDown();
+    test_error_ret(error, "Error in test destruction", TEST_FAIL);
```
kpet, line 437: You could probably introduce some kind of factory utility to perform all of the above for a given test case.

kpet (commented):
Many thanks for uploading this. The code looks good after a first scan. I've added a few suggestions, mostly around how the tests are structured rather than the tests themselves. It might be worth addressing these to keep code duplication under control as we add more tests. I'll do a more detailed review once I get around to running these.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
EwanC: Had a quick try at that but the release/retain template parameters got tripped up by the member variables I'd defined in `CommandBufferTestBase`.  This would make things much nice going forward as our testing expands, so will revisit this

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -0,0 +1,532 @@
...
+        {
+            return error;
+        }
+
+        // Query if device supports simultaneous use
+        cl_device_command_buffer_capabilities_khr capabilities;
+        error =
+            clGetDeviceInfo(device, CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR,
+                            sizeof(capabilities), &capabilities, NULL);
+        test_error(error, "Unable to query CL_COMMAND_BUFFER_PROPERTIES_KHR");
```
kpet, line 84: The error message should say `CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR`.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -0,0 +1,532 @@
...
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+    clMemWrapper in_mem, out_mem;
+    size_t num_elements;
+};
+
+cl_int BasicCommandBufferTest::BasicEnqueue()
+{
+    cl_int error = clCommandNDRangeKernelKHR(
+        command_buffer, queue, nullptr, kernel, 1, nullptr, &num_elements,
```
einhov, line 189: command_queue must be null according to the spec. Also applies to the rest of the clCommand* invocations.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -0,0 +1,532 @@
...
+                           cl_command_queue queue)
+        : CommandBufferTestBase(device), context(context), queue(queue),
+          command_buffer(nullptr), simultaneous_use(false), num_elements(0)
+    {}
+
+    cl_int TearDown()
+    {
+        if (nullptr != command_buffer)
+        {
+            cl_int error = clReleaseCommandBufferKHR(command_buffer);
```
einhov, line 51: We never reach this when a test is skipped, leaking the command buffer.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
EwanC: Good idea, updated to do that

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
EwanC: So because the release/retain function for command-buffers are determined at runtime by querying the platform we can't use those wrapper types with template parameters for release/retain that need a compiler time resolution. I moved this command-buffer release to a destructor for now, but this isn't ideal, will try come up with a better solution.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
EwanC: Used a free function templated on the derived test class

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
EwanC: good spot, fixed

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
EwanC: Another good spot, I've moved this release to the test destructor to fix this leak, but ideally we'd want a wrapper for the command-buffer. Discussion - See https://github.com/KhronosGroup/OpenCL-CTS/pull/1368#discussion_r771424219

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
EwanC: Came up with https://github.com/KhronosGroup/OpenCL-CTS/pull/1368/commits/e9aae637a55e313c8492c7a12939c009bead31c5 , which gives the wrapper a way to access the release/retain functions

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -0,0 +1,483 @@
...
+        const char *kernel_str =
+            R"(
+        __kernel void copy(__global int* in, __global int* out) {
+            size_t id = get_global_id(0);
+            out[id] = in[id];
+        })";
+        const size_t lengths[1] = { std::strlen(kernel_str) };
+
+        program =
+            clCreateProgramWithSource(context, 1, &kernel_str, lengths, &error);
```
aharon-abramson, line 96: You also need to support offline compilation, so please obtain the program by calling create_single_kernel_helper_create_program.

aharon-abramson:
Another thing: you need to query CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR, and if your queue's properties don't satisfy these required properties, you should skip the test (for instance, we return CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR | CL_QUEUE_NO_SYNC_OPERATIONS_INTEL).

EwanC:
> Another thing: you need to query CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR, and if your queue's properties don't satisfy these required properties, you should skip the test (for instance, we return CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR | CL_QUEUE_NO_SYNC_OPERATIONS_INTEL).

I have naive check for this right now, which skips if there are any properties, but i'll change that to checking against the queue.  https://github.com/KhronosGroup/OpenCL-CTS/blob/4cee6a1799472a4aa1b2d56d8333c794426e0881/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp#L47-L58

What i'm trying to figure out is that it's the main() entry point into the test harness passing the queue properties. So need to work out how to initialize the properties of the queue here in a way that gives good test coverage across vendors. https://github.com/KhronosGroup/OpenCL-CTS/blob/4cee6a1799472a4aa1b2d56d8333c794426e0881/test_conformance/extensions/cl_khr_command_buffer/main.cpp#L26-L33

EwanC:
> > Another thing: you need to query CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR, and if your queue's properties don't satisfy these required properties, you should skip the test (for instance, we return CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR | CL_QUEUE_NO_SYNC_OPERATIONS_INTEL).
> 
> I have naive check for this right now, which skips if there are any properties, but i'll change that to checking against the queue.
> 
> What i'm trying to figure out is that it's the main() entry point into the test harness passing the queue properties. So need to work out how to initialize the properties of the queue here in a way that gives good test coverage across vendors.

Added a test now for out-of-order execution with sync-points, where the derived class creates a new command-queue with the out-of-order property set. Rather than replying on the queue properties passed in main. 

https://github.com/KhronosGroup/OpenCL-CTS/blob/e635e8cac3d8d9c002507b5a537e4e313d2fdbcb/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp#L493-L513



Also improved the check for required queue properties https://github.com/KhronosGroup/OpenCL-CTS/blob/e635e8cac3d8d9c002507b5a537e4e313d2fdbcb/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp#L48-L68

aharon-abramson:
Now our implementation passed the test!
Thanks for the updates.

aharon-abramson (approved):

bashbaug (approved):
Merging as discussed on the August 30th teleconference.

PR #1371, opened 2021-12-22
 Fix math tests to allow ftz in relaxed mode.

nikhiljnv:
In recent spec clarification, it is agreed that ftz is
a valid optimization in case of cl-fast-math-relaxed
and doesn't require cl-denorms-are-zero to be passed
explicitly to enforce ftz behavior for implementations
that already support this.
GitHub Spec Issue OpenCL-Docs#579
GitHub Spec Issue OpenCL-Docs#597
GitHub CTS Issue OpenCL-CTS#1267


test_conformance/math_brute_force/unary_float.cpp
```diff
@@ -435,7 +435,7 @@ cl_int Test(cl_uint job_id, cl_uint thread_id, void *data)
 
                 if (fail)
                 {
-                    if (ftz)
+                    if (ftz || relaxedMode)
```
bashbaug, line 438: There are a bunch of other places with `if (ftz)` checks.  Should they be updated too?

nikhiljnv: Yes. Just wanted to get some eyes before I can proceed further. 
If the proposed fix looks good, I can update the other files too. 
This is likely to increase runtime in some cases as this possibly is going to require checking against two reference computations (Check against the default first and then against the relaxed reference computation.

nikhiljnv: Have updated at other places too. Let me know if you have any further suggestions.

nikhiljnv:
@bashbaug
Can you take a look and let me know if there is anything left here to address?
If not, can we merge this one?

bashbaug (dismissed):
These changes look good to me - thanks!

I found a couple of other checks that may possibly need updating:

```sh
$ grep -r "if (ftz" *
...
binary_two_results_i_double.cpp:                if (ftz && fail)
binary_two_results_i_float.cpp:                if (ftz && fail)
i_unary_double.cpp:                    if (ftz && IsDoubleSubnormal(s[j]))
i_unary_float.cpp:                    if (ftz && IsFloatSubnormal(s[j]))
...
```

The "i_unary" cases are all for `ilogb` and hence probably are not affected by denorms, but the "binary_two_results_i" cases are for `remquo`.  Should this be updated for fast relaxed math also?

bashbaug:
For easier future reference, the correct links to the spec issue and PR are:

* https://github.com/KhronosGroup/OpenCL-Docs/issues/579
* https://github.com/KhronosGroup/OpenCL-Docs/pull/597

nikhiljnv:
@alycm @kpet 
Can one of you take a look at this one and help approve the same?
Ben had already approved, but had one last suggestion to extend the fix to few other places which is addressed in the latest patch.

EwanC:
> > These changes look good to me - thanks!
> > I found a couple of other checks that may possibly need updating:
> > ```shell
> > $ grep -r "if (ftz" *
> > ...
> > binary_two_results_i_double.cpp:                if (ftz && fail)
> > binary_two_results_i_float.cpp:                if (ftz && fail)
> > i_unary_double.cpp:                    if (ftz && IsDoubleSubnormal(s[j]))
> > i_unary_float.cpp:                    if (ftz && IsFloatSubnormal(s[j]))
> > ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The "i_unary" cases are all for `ilogb` and hence probably are not affected by denorms, but the "binary_two_results_i" cases are for `remquo`. Should this be updated for fast relaxed math also?
> 
> Have fixed for remaining places mentioned here. Let me know if we are good to merge this one.

Did a grep for a slightly different variant of `ftz` conditional code in the bruteforce tests. 

```sh
$ grep "&& ftz"
binary_double.cpp:                if (fail && ftz)
binary_float.cpp:    if (isFDim && ftz) RestoreFPState(&oldMode);
binary_float.cpp:                    if (fail && ftz)
binary_i_double.cpp:                if (fail && ftz)
binary_i_float.cpp:                if (fail && ftz)
binary_operator_double.cpp:                if (fail && ftz)
binary_operator_float.cpp:                if (fail && ftz)
ternary_double.cpp:                    if (fail && ftz)
ternary_float.cpp:                    if (fail && ftz)
unary_two_results_float.cpp:        if (isFract && ftz) RestoreFPState(&oldMode);
```

Do any of these need updated too?

nikhiljnv:
I just did a grep for ftz and went through all files to avoid missing out places to edit. I have updated few more places that I found.
Hopefully, this covers all, but let me know if anything is still missed out.

EwanC (approved):

nikhiljnv:
Merged as agreed in OpenCL WG call on 07/05.

PR #1372, opened 2021-12-23
Limit size of kernel args no more than 255

haonanya:
Minimum limit for OpFunctionCall actual arguments is 255 according to
https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#_universal_validation_rules

Signed-off-by: Haonan Yang <haonan.yang@intel.com>

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -19,6 +19,10 @@
 #include <ctype.h>
 #include <string.h>
 
+// Minimum limit for OpFunctionCall actual arguments is 255 according to
+// https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#_universal_validation_rules
+const unsigned int SPIRVMinLimitForFuncArgSize = 255;
+
 const char *sample_single_param_kernel[] = {
     "__kernel void sample_test(__global int *src)\n"
     "{\n"
@@ -348,6 +352,7 @@ int test_min_max_read_image_args(cl_device_id deviceID, cl_context context,
                  (int)(maxParameterSize / deviceAddressSize));
         maxReadImages = (unsigned int)(maxParameterSize / deviceAddressSize);
     }
+    maxReadImages = std::min(maxReadImages, SPIRVMinLimitForFuncArgSize);
 
     /* Create a program with that many read args */
     programSrc = (char *)malloc(strlen(sample_read_image_kernel_pattern[0])
@@ -357,13 +362,13 @@ int test_min_max_read_image_args(cl_device_id deviceID, cl_context context,
 
     strcpy(programSrc, sample_read_image_kernel_pattern[0]);
     strcat(programSrc, "read_only image2d_t srcimg0");
-    for (i = 0; i < maxReadImages - 1; i++)
+    for (i = 0; i < maxReadImages - 2; i++)
     {
         sprintf(readArgLine, readArgPattern, i + 1);
         strcat(programSrc, readArgLine);
     }
     strcat(programSrc, sample_read_image_kernel_pattern[1]);
-    for (i = 0; i < maxReadImages; i++)
+    for (i = 0; i < maxReadImages - 1; i++)
     {
         sprintf(
             readArgLine,
@@ -385,7 +390,7 @@ int test_min_max_read_image_args(cl_device_id deviceID, cl_context context,
 
     /* Create some I/O streams */
     streams = new clMemWrapper[maxReadImages + 1];
-    for (i = 0; i < maxReadImages; i++)
+    for (i = 0; i < maxReadImages - 1; i++)
     {
         image_data[0] = i;
         image_result += image_data[0];
@@ -399,7 +404,7 @@ int test_min_max_read_image_args(cl_device_id deviceID, cl_context context,
     test_error(error, "Unable to set kernel arguments");
 
     /* Set the arguments */
-    for (i = 1; i < maxReadImages + 1; i++)
+    for (i = 1; i < maxReadImages; i++)
     {
         error =
             clSetKernelArg(kernel, i, sizeof(streams[i - 1]), &streams[i - 1]);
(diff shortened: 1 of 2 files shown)


test_conformance/api/test_api_min_max.cpp
```diff
@@ -19,6 +19,12 @@
 #include <ctype.h>
 #include <string.h>
 
+#define MIN(X, Y) (((X) < (Y)) ? (X) : (Y))
```
bashbaug, line 22: Could we use `std::min` here vs. defining our own `MIN`?

test_conformance/api/test_api_min_max.cpp
```diff
@@ -357,13 +364,13 @@ int test_min_max_read_image_args(cl_device_id deviceID, cl_context context,
 
     strcpy(programSrc, sample_read_image_kernel_pattern[0]);
     strcat(programSrc, "read_only image2d_t srcimg0");
-    for (i = 0; i < maxReadImages - 1; i++)
+    for (i = 0; i < maxReadImages - 2; i++)
```
bashbaug, line 367: I don't understand why this changed from `maxReadImages - 1` to `maxReadImages - 2`.  Seems that this will change the test behavior even for devices that do not approach the SPIR-V function argument limit.  Was this intended?

Same goes for the other changes to loop limits in this PR.

bashbaug (commented):
I've left a few specific comments on these changes.

More generally though, do we need something in the core API or OpenCL C spec to limit the number of function arguments independent of SPIR-V?

We already have this special-case in the description of `CL_DEVICE_MAX_PARAMETER_SIZE` ([link](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_PARAMETER_SIZE)):

> The minimum value is 1024 for devices that are not of type CL_DEVICE_TYPE_CUSTOM. For this minimum value, only a maximum of 128 arguments can be passed to a kernel

There is no limit for devices that report greater than this minimum value though.  Should there be?

lakshmih:
[Qualcomm] Our concern is that this change would affect test behavior even on implementations that don't support SPIR-V. Further discussion is needed.

bashbaug:
I don't think we can merge this yet.  @haonanya, can you please look at my review comments above?

haonanya:
> I don't think we can merge this yet. @haonanya, can you please look at my review comments above?

Hi, @bashbaug, thanks for your review, I will have a look.

test_conformance/api/test_api_min_max.cpp
haonanya: This doesn't change original behavior. The test is always running successfully without SPIR-V validation, I am keeping size of function parameter equal to 255. The reason why I am changing code from `maxReadImages - 1` to `maxReadImages - 2` is that there are `__global float *result` and `read_only image2d_t srcimg0` as function parameters, available function parameters are 253. Finally we will get:
`_kernel void sample_test( __global float *result, read_only image2d_t srcimg0, read_only image2d_t srcimg1, ... read_only image2d_t srcimg253 )`

test_conformance/api/test_api_min_max.cpp
haonanya: done, thanks very much.

test_conformance/api/test_api_min_max.cpp
haonanya: ping @bashbaug, thanks very much.

bashbaug: Hi, sorry for the delay.

I still think this change needs to be generalized slightly so it will not affect devices that do not approach the SPIR-V function argument limit.

As an experiment, I ran this test on one of our GPU devices with CL_DEVICE_MAX_READ_IMAGE_ARGS = 128 with the OpenCL intercept layer and [DumpProgramSource](https://github.com/intel/opencl-intercept-layer/blob/master/docs/controls.md#dumpprogramsource-bool) enabled, to see how kernels were generated before and after these changes.

Before these changes I see kernels of the form:

```c
__kernel void sample_test( __global float *result, read_only image2d_t srcimg0, ..., read_only image2d_t srcimg127 )
```

So, all 128 images are tested.

After these changes, I see kernels of the form:

```c
__kernel void sample_test( __global float *result, read_only image2d_t srcimg0, ..., read_only image2d_t srcimg126 )
```

So, only 127 images are tested.

Note that this kernel doesn't approach the SPIR-V function argument limit.  Is this intended?  Or, should we be limiting the number of images only when `maxReadImages` is sufficiently large to exceed the SPIR-V function argument limit?

bashbaug:
I've filed the two issues in OpenCL-Docs described above.

test_conformance/api/test_api_min_max.cpp
bashbaug: Hi @haonanya, want to be sure you saw the comment above - thanks!

haonanya: yes, thanks for your explanation.

nikhiljnv:
Apologize for late feedback here.
If I am reading it correctly, 255 is the minimum limit for actual function arguments of SPIR-V. Does this mean number of kernel arguments should be no more than 255 or implementations supporting SPIR-V need to support minimum of upto 255 kernel arguments and some implementations can choose to support higher limit?

Also, if we decide to cap max number of kernel arguments to 255, is it fair to say all implementations need to limit  CL_DEVICE_MAX_PARAMETER_SIZE to 2048? For implementations already reporting CL_DEVICE_MAX_PARAMETER_SIZE  > 2048, possibly there may be applications using >255 arguments which may start to fail with this change and this change may potentially break backward compatibility.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -19,6 +19,10 @@
 #include <ctype.h>
 #include <string.h>
 
+// Minimum limit for OpFunctionCall actual arguments is 255 according to
+// https://www.khronos.org/registry/SPIR-V/specs/unified1/SPIRV.html#_universal_validation_rules
```
alycm, line 23:
```suggestion
// https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_universal_limits
```

AFAICT it's actually the limits not the validation rules section that says 255.

ahesham-arm:
@bashbaug look what I found :)

PR #1374, opened 2022-01-02
Remove invalid negative_get_platform_info testcase

### LINKED ISSUE #1318: Invalid test case for negative_get_platform_info
The first test case for `test api negative_platform_info` 
https://github.com/KhronosGroup/OpenCL-CTS/blob/1f26e1d8ba372f4c638f9c0cdae7566e349b9b9a/test_conformance/api/negative_platform.cpp#L46 casts a `cl_device_id` to a `cl_platform_id` and expects an error of `CL_INVALID_PLATFORM`.

Per https://github.com/KhronosGroup/OpenCL-Docs/issues/482, implementations are only responsible for null-checking to check validity. This particular case actually dispatches correctly since the dispatch table is present--any valid CL object from the same platform would suffice for the platform argument. 

This test case should be deleted.

jlewis-austin:
Fixes #1318

### DIFF
--- a/test_conformance/api/negative_platform.cpp
+++ b/test_conformance/api/negative_platform.cpp
@@ -42,18 +42,8 @@ int test_negative_get_platform_info(cl_device_id deviceID, cl_context context,
 {
     cl_platform_id platform = getPlatformFromDevice(deviceID);
 
-    cl_int err =
-        clGetPlatformInfo(reinterpret_cast<cl_platform_id>(deviceID),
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
LGTM.

I initially thought that we could modify this test case to pass `NULL` as the "invalid" platform ID, but the spec says that behavior is implementation-defined for a `NULL` platform ID, so I agree removing this case is the only viable option.

kpet (approved):
LGTM. Very straightforward as a standalone fix. Rebasing the other PR on top of this should be easy.

PR #1375, opened 2022-01-03
test_spir sampler_enumeration - Fix incorrect use image channel data type and filtering mode

gwawiork:
Current test implementation sets CL_RGBA + CL_SIGNED_INT32 for read_imagef() function and use CLK_FILTER_LINEAR sampler filtering mode. 
This is incorrect and described in [spec](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_C.html#image-read-and-write-functions) as undefined behavior. 
This PR fixes this by exchanging in kernel image data type description what allow set in test configuration phase proper channel data type (read_imagef will use CL_FLOAT)
The change has no impact on output spir files. Only kernel modifications.
Inside the kernel changes:
from:
`__kernel void sample_kernel( read_only image2d_t input, __global float *xOffsets, __global float *yOffsets, __global float4 *results )`

to:

```
typedef read_only image2d_t read_only_image2d_float;

__kernel void sample_kernel(read_only_image2d_float input, __global float *xOffsets, __global float *yOffsets, __global float4 *results )
```


bashbaug (commented):
LGTM.

I was initially confused how this PR solved the issue but the "spir" test host code is setup to examine the type for each kernel argument - see `typeinfo.h`.  When the kernel argument type is `image2d_t` the host code will create and use an image with channel data type equal to `CL_SIGNED_INT32`, which is incompatible with `read_imagef`.  Changing the type to `read_only_image2d_float` instead will cause the host code to create and use an image with channel data type equal to `CL_FLOAT` instead, which fixes this issue.

Needs a non-Intel reviewer.

kpet:
@alycm Is that something you guys can review? (IIRC you support cl_khr_spir).

Jeremy-Kemp (approved):
LGTM, thanks @gwawiork.

I've also verified this on our implementation (we support `cl_khr_spir`).

EwanC (approved):
Can confirm this MR passes on Codeplay's implementation of `cl_khr_spir`. Can't say I'm too familiar with this area of the CTS, but the solution descriptions in the MR discussion as to why this change fixes the issue make sense.

PR #1377, opened 2022-01-05
Change memory order and scope for atomics that gate final results being stored.

### LINKED ISSUE #1370: C11 Atomics: Relaxed memory ordering in atomic_load tests
CTS can generate test cases that are invalid. We have noticed this in some of the `atomic_load` sub-tests.

In the kernel below, there is nothing preventing a compiler from re-ordering the `atomic_load_explicit`/`atomic_store_explicit` calls around the `atomic_fetch_add_explicit` call. It seems the intent of the atomic add is to gate the final load/store however we don't think that is strong enough to achieve the desired effect.

For this particular case, we don't think the structure of the kernel can be used due to the nature of `memory_order_relaxed`; making the test case (and others like it) invalid.

```C
/*  -cl-std=CL3.0 */
__global volatile atomic_uint finishedThreads = ATOMIC_VAR_INIT(0);
__global volatile atomic_int destMemory[1024] = {
    ATOMIC_VAR_INIT(255),
    ATOMIC_VAR_INIT(255),
    ....
    ATOMIC_VAR_INIT(255),
    ATOMIC_VAR_INIT(255)};
 
void test_atomic_function(uint tid, uint threadCount, uint numDestItems, volatile __global atomic_int *destMemory, __global int *oldValues)
{
    atomic_store_explicit(&destMemory[tid], tid, memory_order_relaxed, memory_scope_device);
    oldValues[tid] = atomic_load_explicit(&destMemory[tid], memory_order_relaxed, memory_scope_device);
}
 
__kernel void test_atomic_kernel(uint threadCount, uint numDestItems, volatile __global int *finalDest, __global int *oldValues)
{
    uint tid = get_global_id(0);
 
    test_atomic_function(tid, threadCount, numDestItems, destMemory, oldValues);
 
    // Copy final values to host reachable buffer
 
    if (atomic_fetch_add_explicit(&finishedThreads, 1u,
                                  memory_order_relaxed,
                                  memory_scope_work_group) == get_global_size(0) - 1) // last finished thread
        for (uint dstItemIdx = 0; dstItemIdx < numDestItems; dstItemIdx++)
 
            finalDest[dstItemIdx] =
                atomic_load_explicit(destMemory + dstItemIdx,
                                     memory_order_relaxed,
                                     memory_scope_work_group);
}
```

Jeremy-Kemp:
memory_order_acq_rel with memory_scope_device is now used to guarantee that the correct memory consistency is observed before final results are stored.

Previously it was possible for kernels to be generated that all used relaxed memory ordering, which could lead to false-positive failures.

Fixes #1370

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -1111,7 +1111,7 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(
     if (!LocalMemory() && DeclaredInProgram())
     {
         if (((gAtomicMemCap & CL_DEVICE_ATOMIC_SCOPE_DEVICE) == 0)
-            || ((gAtomicMemCap & CL_DEVICE_ATOMIC_ORDER_ACQ_REL == 0)))
+            || ((gAtomicMemCap & CL_DEVICE_ATOMIC_ORDER_ACQ_REL) == 0))
         {
             log_info("\t\tTest disabled\n");
             return 0;
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -1108,6 +1108,15 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(
         log_info("\t\tTest disabled\n");
         return 0;
     }
+    if (!LocalMemory() && DeclaredInProgram())
+    {
+        if (((gAtomicMemCap & CL_DEVICE_ATOMIC_SCOPE_DEVICE) == 0)
+            || ((gAtomicMemCap & CL_DEVICE_ATOMIC_ORDER_ACQ_REL == 0)))
+        {
+            log_info("\t\tTest disabled\n");
+            return 0;
+        }
+    }
 
     // set up work sizes based on device capabilities and test configuration
     error = clGetDeviceInfo(deviceID, CL_DEVICE_MAX_WORK_GROUP_SIZE,
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -1041,8 +1041,8 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
             // global atomics declared in program scope
             code += R"(
                 if(atomic_fetch_add_explicit(&finishedThreads, 1u,
-                                           memory_order_relaxed,
-                                           memory_scope_work_group)
+                                           memory_order_acq_rel,
+                                           memory_scope_device)
                    == get_global_size(0)-1) // last finished thread
                    )";
         code += "    for(uint dstItemIdx = 0; dstItemIdx < numDestItems; "


Jeremy-Kemp:
I have manually reviewed several (but not all) kernels from the various suites in c11_atomics to verify that the generated code for multiple cases is correct.

bashbaug (commented):
This seems reasonable to me.  My only questions are:

1. Are there any other places that need to be changed as well?  Not sure what the most efficient way is to check this.
2. Because the stronger atomic memory order and wider atomic scope are optional for OpenCL 3.0 devices, do any checks need to be updated so these test cases do not incorrectly run on devices that do not support these atomic capabilities?

Thanks for fixing this!

Jeremy-Kemp:
> Are there any other places that need to be changed as well? Not sure what the most efficient way is to check this.

I've had a look around to try and find other potential cases but as yet have not surfaced anything. Its quite a tricky test to parse and figure out all the cases as there are so many combinations.
>Because the stronger atomic memory order and wider atomic scope are optional for OpenCL 3.0 devices, do any checks need to be updated so these test cases do not incorrectly run on devices that do not support these atomic capabilities?

Good spot! Yes, I think this could be a problem and this change will potentially reduce the number of test cases that current implementations will be able to target. I think any 3.0 implementation that uplifted from 1.2 without adding 2.x functionality will be at risk here.

I think we could either exclude the affected tests if the implementation does not support device scope and/or acq/rel - or we could look at reworking the kernels in some way which seems like a non-trivial amount of work.

jlewis-austin:
We've run with this fix several times without failure; we do see occasional failures without it.

bashbaug:
> We've run with this fix several times without failure; we do see occasional failures without it.

Nice, this is encouraging.

Were we going to do any other checking to ensure these tests would only run on devices that support the updated atomic memory order and scope?

Jeremy-Kemp:
> We've run with this fix several times without failure; we do see occasional failures without it.

Great, thanks for testing it out!

> Were we going to do any other checking to ensure these tests would only run on devices that support the updated atomic memory order and scope?

Yes, definitely. This is still on my list to look at.

test_conformance/c11_atomics/common.h
```diff
@@ -1041,8 +1041,8 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
             // global atomics declared in program scope
             code += R"(
                 if(atomic_fetch_add_explicit(&finishedThreads, 1u,
-                                           memory_order_relaxed,
-                                           memory_scope_work_group)
+                                           memory_order_acq_rel,
+                                           memory_scope_device)
```
jlewis-austin, line 1045: Note #1398 duplicates (and extends) this change; some questions about it support being optional in 3.0 are raised there.

bashbaug:
I think this is a possible duplicate of #1398, or at least they're both trying to solve similar problems.

@lakshmih can you please check this PR and see if it solves your problem?

lakshmih:
We tested this change and found that we still need a part of our fix which I have raised as a PR into Jeremy's branch 
https://github.com/Jeremy-Kemp/OpenCL-CTS/pull/1

Jeremy-Kemp:
The group discussed the additional changes on the 14/04/2022 teleconference. I will be pulling in the updates from https://github.com/Jeremy-Kemp/OpenCL-CTS/pull/1 shortly.

lakshmih:
> Thanks @lakshmih for testing out this PR, and opening up a further modification. I've taken a closer look, and I believe your changes are an improvement over what we have here, rather than a duplication.
> 
> If I'm understanding [Jeremy-Kemp#1](https://github.com/Jeremy-Kemp/OpenCL-CTS/pull/1) correctly, the `memory_order_acq_rel` is not sufficient as it only applies to `finishedThreads`. Therefore, inconsistency can still be present for `destMemory` as that memory operation may still be re-ordered before the `memory_order_acq_rel` on `finishedThreads`.
> 
> Have I followed the intent correctly?

@Jeremy-Kemp  That part of the kernel was handling two cases - LocalMemory() and DeclaredInProgram(). What our modification does is to change the scope to memory_scope_device for the atomic calls in the DeclaredInProgram() case since the loads and stores are across the entire global size and not on a particular workgroup.

Jeremy-Kemp:
Thanks @lakshmih . Upon further inspection (and the discussion on the call) I came to that conclusion. Thanks for the changes!

lakshmih (approved):
Tested and approved from Qualcomm.

test_conformance/c11_atomics/common.h
```diff
@@ -1053,15 +1073,15 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
...
+                                                      memory_scope_device);)";
         }
         else
         {
             code += R"(
                 finalDest[dstItemIdx] =
                     atomic_load_explicit(destMemory+dstItemIdx,
                                          memory_order_relaxed,
-                                         memory_scope_work_group);)";
+                                         memory_scope_device);)";
```
bashbaug, line 1084: This isn't hurting anything, but I don't think it should be required either: since this is a relaxed load, there are no memory ordering constraints, so the memory scope should be irrelevant.

PR #1378, opened 2022-01-06
Fix memory leaks

lakshmih:
Fixed memory leaks in: buffers, basic, and vectors

### DIFF
--- a/test_conformance/basic/test_vector_swizzle.cpp
+++ b/test_conformance/basic/test_vector_swizzle.cpp
@@ -631,8 +631,8 @@ static int test_vectype(const char* type_name, cl_device_id device,
 
         const char* xyzw_source = TestInfo<N>::kernel_source_xyzw;
         error = create_single_kernel_helper(
-            context, &program, &kernel, 1, &xyzw_source, "test_vector_swizzle_xyzw",
-            buildOptions.c_str());
+            context, &program, &kernel, 1, &xyzw_source,
+            "test_vector_swizzle_xyzw", buildOptions.c_str());
         test_error(error, "Unable to create xyzw test kernel");
 
         result |= test_vectype_case(value, reference, context, kernel, queue);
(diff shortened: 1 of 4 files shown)


test_conformance/basic/test_vector_swizzle.cpp
```diff
@@ -628,35 +625,50 @@ static int test_vectype(const char* type_name, cl_device_id device,
...
     // RGBA swizzles for OpenCL 3.0 and newer:
-    const Version device_version = get_device_cl_version(device);
-    if (device_version >= Version(3, 0))
     {
-        const char* rgba_source = TestInfo<N>::kernel_source_rgba;
-        error = create_single_kernel_helper(
-            context, &program, &kernel, 1, &rgba_source,
-            "test_vector_swizzle_rgba", buildOptions.c_str());
-        test_error(error, "Unable to create rgba test kernel");
+        clProgramWrapper program;
```
EwanC, line 657: minor: Could have these in the `if (device_version >= Version(3, 0))` scope to save introducing another level of indentation

EwanC (approved):

bashbaug (approved):
LGTM, nice catch(es).

Is there anything we could do to make this more robust / less error-prone in the future?  The cases involving the wrapper are especially subtle.

The good news I suppose is that the OpenCL Intercept Layer [LeakChecking](https://github.com/intel/opencl-intercept-layer/blob/master/docs/controls.md#leakchecking-bool) was able to detect all of these leaks (and verify that they are fixed after this PR).  I should get in the habit of running that on all of my PRs, since I was the one who introduced the vector swizzle leak :cry:.

lakshmih:
@kpet can you merge, has two approvals now.

Kerilk:
> The good news I suppose is that the OpenCL Intercept Layer [LeakChecking](https://github.com/intel/opencl-intercept-layer/blob/master/docs/controls.md#leakchecking-bool) was able to detect all of these leaks (and verify that they are fixed after this PR).

I did the same with the [object-lifetime](https://github.com/KhronosGroup/OpenCL-Layers/tree/main/object-lifetime) layer and observed both the leaks and their disappearance with the fix. If we had a way to run the tests in CI, then integrating the layers to the run would make sense I think.

PR #1382, opened 2022-01-19
Improve testing of sub_group_ballot

StuartDBrady:

### DIFF
--- a/test_conformance/subgroups/test_subgroup_ballot.cpp
+++ b/test_conformance/subgroups/test_subgroup_ballot.cpp
@@ -31,45 +31,26 @@ template <typename Ty> struct BALLOT
 
     static void gen(Ty *x, Ty *t, cl_int *m, const WorkGroupParams &test_params)
     {
-        // no work here
-        int gws = test_params.global_workgroup_size;
-        int lws = test_params.local_workgroup_size;
-        int sbs = test_params.subgroup_size;
-        int non_uniform_size = gws % lws;
-    }
-
-    static test_status chk(Ty *x, Ty *y, Ty *mx, Ty *my, cl_int *m,
-                           const WorkGroupParams &test_params)
-    {
-        int wi_id, wg_id, sb_id;
         int gws = test_params.global_workgroup_size;
         int lws = test_params.local_workgroup_size;
         int sbs = test_params.subgroup_size;
         int sb_number = (lws + sbs - 1) / sbs;
-        int current_sbs = 0;
-        cl_uint expected_result, device_result;
         int non_uniform_size = gws % lws;
         int wg_number = gws / lws;
         wg_number = non_uniform_size ? wg_number + 1 : wg_number;
         int last_subgroup_size = 0;
 
-        for (wg_id = 0; wg_id < wg_number; ++wg_id)
+        for (int wg_id = 0; wg_id < wg_number; ++wg_id)
         { // for each work_group
             if (non_uniform_size && wg_id == wg_number - 1)
             {
                 set_last_workgroup_params(non_uniform_size, sb_number, sbs, lws,
                                           last_subgroup_size);
             }
-
-            for (wi_id = 0; wi_id < lws; ++wi_id)
-            { // inside the work_group
-                // read device outputs for work_group
-                my[wi_id] = y[wi_id];
-            }
-
-            for (sb_id = 0; sb_id < sb_number; ++sb_id)
+            for (int sb_id = 0; sb_id < sb_number; ++sb_id)
             { // for each subgroup
                 int wg_offset = sb_id * sbs;
+                int current_sbs;
                 if (last_subgroup_size && sb_id == sb_number - 1)
                 {
                     current_sbs = last_subgroup_size;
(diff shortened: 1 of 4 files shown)


test_conformance/subgroups/test_subgroup_ballot.cpp
```diff
@@ -78,25 +126,54 @@ template <typename Ty> struct BALLOT
                 {
                     current_sbs = wg_offset + sbs > lws ? lws - wg_offset : sbs;
                 }
-                for (wi_id = 0; wi_id < current_sbs; ++wi_id)
+
+                bs128 expected_result_bs = 0;
+
+                std::set<int> active_work_items;
```
polsenferreira, line 132: I would suggest moving this outside of all loops, and use .clear() here.
However, this would be a minor performance tune, and ultimately I don't think it ought to block the commit.

test_conformance/subgroups/test_subgroup_ballot.cpp
```diff
@@ -31,45 +31,93 @@ template <typename Ty> struct BALLOT
...
+                        v = genrand_bool(gMTdata);
+                    }
+                    else if (genrand_bool(gMTdata))
+                    {
+                        v = 1U << ((genrand_int32(gMTdata) % 31) + 1);
+                    }
+                    else
+                    {
+                        v = genrand_int32(gMTdata);
+                    }
```
polsenferreira, line 77: Not sure that the probabilities on this sequence match what's expected, as from what i can tell, the first branch is executed 50% of the time, and the following two branches 25% each.

It might not matter at all, but I felt that perhaps I should mention it.

test_conformance/subgroups/test_subgroup_ballot.cpp
StuartDBrady: How expensive do you think the bs128 use would be in comparison?

test_conformance/subgroups/test_subgroup_ballot.cpp
```diff
@@ -78,25 +126,54 @@ template <typename Ty> struct BALLOT
...
+                    if (test_params.work_items_mask.test(wi_id))
+                    {
+                        bool predicate = (mx[wg_offset + wi_id].s0 != 0);
+                        expected_result_bs |= (bs128(predicate) << wi_id);
+                        active_work_items.insert(wi_id);
+                    }
+                }
+                if (active_work_items.empty())
+                {
+                    continue;
```
polsenferreira, line 144: It might be worth pointing out why this is being skipped.
Given that the following loop is iterating over the contents of the container, imediatelyfollowed the loop restart, perhaps this condition is not necessary.

polsenferreira (commented):
I think the changes look fine to me.

Due to the complexity of the loops, it is difficult to do a functional review, so it is possible I missed something.
An extra pair of eyes will give us some more confidence.

test_conformance/subgroups/test_subgroup_ballot.cpp
polsenferreira: Sorry, github gives more context than I intended. I meant strictly the std::set, and nothing else.
My reasoning was that clear() might recycle previous heap allocations.

StuartDBrady: Right, but might avoiding having bs128s falling in and out of scope, too, also recycle previous heap allocations? I wonder if either might be premature optimization, though.

polsenferreira: I don't know what bs128 is, and it seems unrelated to std::set (from my cursory view).
This is fine as-is. I merely wanted to provide some feedback on this line and am not expecting any changes :)
Ultimately whatever gains this may have might be immesurable when compared to the time consumed by the rest of the function.

test_conformance/subgroups/test_subgroup_ballot.cpp
StuartDBrady: Yeah, that seemed fairly reasonable to me at the time I wrote it. It certainly didn't seem worth doing anything more complicated. The point here is that if we just used random 32-bit integers, they would be true 99.9999999767% of the time. The testing should be improved at some point so that with wide warp widths, there's a reasonable chance of the predicate being the same for all threads... but for now, this is better than what we had.

test_conformance/subgroups/test_subgroup_ballot.cpp
StuartDBrady: The bs128 is a 128-bit [std::bitset](https://en.cppreference.com/w/cpp/utility/bitset). I expect it will involve some heap use.

test_conformance/subgroups/test_subgroup_ballot.cpp
gwawiork: I see now this is not necessary because in next lines we start iterating over `active_work_items` so in case it is empty we skip running checking results.

gwawiork (approved):
LGTM

test_conformance/subgroups/test_subgroup_ballot.cpp
StuartDBrady: Perhaps worth profiling this, but there's a fair chance that this code will be changing quite a bit anyway.

test_conformance/subgroups/test_subgroup_ballot.cpp
StuartDBrady: Assuming there's no objection here.

test_conformance/subgroups/test_subgroup_ballot.cpp
StuartDBrady: It's certainly not necessary but I think the structure of this still makes some sense. If there are no active work items, there is no expected result, and it makes sense not to enter the scope containing a variable for such a result.

PR #1383, opened 2022-01-24
Check for non-uniform work-group support

FranklandJack:
Only run sub-group tests with non-uniform work-groups on OpenCL 3.0 and
later if it is supported by the device.


test_conformance/subgroups/test_workitem.cpp
```diff
@@ -251,8 +252,21 @@ int test_work_item_functions(cl_device_id device, cl_context context,
 
     global = local * 5;
 
-    // Make sure we have a flexible range
-    global += 3 * local / 4;
+    // Non-uniform work-groups are an optional feature from 3.0 onward.
+    cl_bool device_supports_non_uniform_wg = CL_TRUE;
+    if (get_device_latest_cl_c_version(device) >= Version(3, 0))
```
alycm, line 257: I don't know if this is just a standard convention, but checking for support of OpenCL C versions for an API property seems odd.  E.g. a 3.0 device without a device compiler will not support any OpenCL C version, but may support non-uniform work-groups.

FranklandJack: Sorry this should be `get_device_cl_version`, I've updated.

bashbaug (approved):
LGTM!

PR #1386, opened 2022-02-04
Added simple test for CL_DEVICE_PRINTF_BUFFER_SIZE.

Jeremy-Kemp:


test_conformance/printf/test_printf.cpp
```diff
@@ -991,6 +991,8 @@ test_status InitCL( cl_device_id device )
 {
     uint32_t device_frequency = 0;
     uint32_t compute_devices = 0;
+    size_t printf_buff_size = 0;
+    const size_t printf_buff_size_req = !gIsEmbedded ? 1048576 : 1024;
```
kpet, line 995: Maybe `1014*1024UL`?

kpet (commented):
Good addition. Maybe this would be better structured as a test rather than as part of the init code?

Jeremy-Kemp:
> Maybe this would be better structured as a test rather than as part of the init code?

I debated this as it is such a small test, wasn't sure if it warranted it or not. I'll spin it out to its own test case to make it more explicit - especially in the failure case.

kpet (approved):
LGTM, thanks!

test_conformance/printf/test_printf.cpp
```diff
@@ -825,73 +825,75 @@ int test_address_space_4(cl_device_id deviceID, cl_context context, cl_command_q
...
+
+    ADD_TEST(vector_0),        ADD_TEST(vector_1),
+    ADD_TEST(vector_2),        ADD_TEST(vector_3),
+    ADD_TEST(vector_4),
+
+    ADD_TEST(address_space_0), ADD_TEST(address_space_1),
+    ADD_TEST(address_space_2), ADD_TEST(address_space_3),
+    ADD_TEST(address_space_4),
+
+    ADD_TEST(buffer_size),
```
EwanC, line 896: Should this be `ADD_TEST_VERSION(buffer_size, Version(1, 2))`, I think `CL_DEVICE_PRINTF_BUFFER_SIZE` was added in OpenCL 1.2

Jeremy-Kemp: Currently `InitCL` takes care of that for all the tests in the printf suite.

I agree though it'd be a nice change to make in general but also likely applies to a greater number of tests than just this one.

bashbaug (approved):

test_conformance/printf/test_printf.cpp
EwanC: ah missed that, thanks

EwanC (approved):

PR #1390, opened 2022-02-22
Initial CTS for external semaphore and memory extensions

nikhiljnv:
Initial CTS for external semaphore and memory extensions


test_conformance/vulkan/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+        ../../test_common/harness/genericThread.cpp
+        ../../test_common/harness/errorHelpers.cpp
+        ../../test_common/harness/threadTesting.cpp
+        ../../test_common/harness/testHarness.cpp
+        ../../test_common/harness/kernelHelpers.cpp
+        ../../test_common/harness/mt19937.cpp
+        ../../test_common/harness/msvc9.c
+        ../../test_common/harness/parseParameters.cpp
+        ../../test_common/harness/deviceInfo.cpp
+        ../../test_common/harness/crc32.cpp
```
kpet, line 47: The binary should be linked against the harness lib instead of re-building the harness. See how other binaries now do it.

test_conformance/vulkan/main.cpp
```diff
@@ -0,0 +1,275 @@
+/******************************************************************
+//
+//  OpenCL Conformance Tests
+// 
+//  Copyright:	(c) 2008-2011 by Apple Inc. All Rights Reserved.
```
kpet, line 5: Might want to fix the copyrights.

test_conformance/vulkan/main.cpp
```diff
@@ -0,0 +1,275 @@
...
+//
+******************************************************************/
+
+#include <stdio.h>
+#include <stdlib.h>
```
kpet, line 10: Suggest using the C++ headers if necessary.

test_conformance/vulkan/main.cpp
```diff
@@ -0,0 +1,275 @@
...
+        log_info(" Device does not support cl_khr_external_memory or cl_khr_external_semaphore\n");
+        log_info(" TEST SKIPPED\n");
+        return CL_SUCCESS;
+    }
+    init_cl_vk_ext(platform);
+
+ // Execute tests.
+      
+ // Note: don't use the entire harness, because we have a different way of obtaining the device (via the context)
+    errNum = parseAndCallCommandLineTests( argCount, argList, devices[device_no], test_num, test_list,  true, 0, 1024 );
```
kpet, line 273: Maybe we should add support to the harness instead of duplicating the code here?

test_conformance/vulkan/procs.h
```diff
@@ -0,0 +1,23 @@
+/******************************************************************
+//
+//  OpenCL Conformance Tests
+// 
+//  Copyright:	(c) 2008-2011 by Apple Inc. All Rights Reserved.
+//
+******************************************************************/
+
+#include "../../test_common/harness/mt19937.h"
```
kpet, line 9: `harness/mt19937.h` ought to work

test_conformance/vulkan/procs.h
```diff
@@ -0,0 +1,23 @@
...
+
+#pragma mark -
+#pragma Misc tests
```
kpet, line 12: Why the pragma?

test_conformance/vulkan/test_vulkan_api_consistency.cpp
```diff
@@ -0,0 +1,496 @@
...
+
+    cmd_queue = clCreateCommandQueue(context, deviceID, 0, &errNum);
+    test_error(errNum, "Unable to create command queue");
+   
+    uint32_t bufferSize = 32;
+    cl_device_id devList[] = {deviceID, NULL};
+
+    #ifdef _WIN32
+        if (!is_extension_available(devList[0], "cl_khr_external_memory_win32")) {
+            throw std::runtime_error(" Device does not support cl_khr_external_memory_win32 extension \n");
```
kpet, line 48: Don't we want to skip the test instead?

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper_types.hpp
```diff
@@ -0,0 +1,390 @@
...
+};
+
+enum VulkanImageType
+{
+    VULKAN_IMAGE_TYPE_1D = VK_IMAGE_TYPE_1D,
+    VULKAN_IMAGE_TYPE_2D = VK_IMAGE_TYPE_2D,
+    VULKAN_IMAGE_TYPE_3D = VK_IMAGE_TYPE_3D
+};
+
+enum VulkanFormat
```
kpet, line 153: What's the purpose of these duplicate definitions? These will be a maintenance problem.

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
```diff
@@ -0,0 +1,2228 @@
...
+    }
+
+    VkShaderModuleCreateInfo vkShaderModuleCreateInfo = {};
+    vkShaderModuleCreateInfo.sType    = VK_STRUCTURE_TYPE_SHADER_MODULE_CREATE_INFO;
+    vkShaderModuleCreateInfo.pNext    = NULL;
+    vkShaderModuleCreateInfo.flags    = 0;
+    vkShaderModuleCreateInfo.codeSize = paddedCode.size();
+    vkShaderModuleCreateInfo.pCode    = (const uint32_t *)(void *)paddedCode.c_str();
+
+    vkCreateShaderModule(m_device, &vkShaderModuleCreateInfo, NULL, &m_vkShaderModule);
```
kpet, line 868: Creating a shader module from GLSL code is not standard Vulkan. We either need to pre-compile the SPIR-V or integrate glslang.

test_conformance/vulkan/vulkan_interop_common/vulkan_utility.hpp
```diff
@@ -0,0 +1,55 @@
+#ifndef _vulkan_utility_hpp_
+#define _vulkan_utility_hpp_
+
+#include "vulkan_wrapper_types.hpp"
+#include <vector>
+#include <ostream>
+#include <string.h>
+#include <map>
+
+#define ARRAY_SIZE(a)  (sizeof(a) / sizeof(a[0]))
```
kpet, line 10: This is already provided by the harness

kpet (changes requested):
A very quick first pass on this. It's great to see tests starting to appear but we have a lot of work on our hands to integrate this into the CTS. A few overall questions/comments:

- You should use object wrappers and RAII instead of flags, totos, etc. Much easier to reason about and maintain.
- Have you considered using existing Vulkan bindings (e.g. https://github.com/KhronosGroup/Vulkan-Hpp) rather than rolling your own?
- Any reason Vulkan error codes are not checked at all?
- Please avoid the use of malloc/free and prefer `std::string`, `std::vector`, etc
- Consider introducing/using device query wrappers from https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/deviceInfo.h rather than copy/pasting code.

test_conformance/vulkan/main.cpp
```diff
@@ -0,0 +1,275 @@
...
+        errNum = memcmp(uuid, vkDevice.getPhysicalDevice().getUUID(), VK_UUID_SIZE);
+        if (errNum == 0) {
+            break;
+        }
+    }
+    if (device_no >= num_devices) {
+         fprintf(stderr, "OpenCL error:"
+               " No Vulkan-OpenCL Interop capable GPU found.\n");
+    }
+    if (!( is_extension_available(devices[device_no], "cl_khr_external_memory")  && is_extension_available(devices[device_no], "cl_khr_external_semaphore"))) {
```
paulfradgley, line 263: Will it be possible to run a subset of tests if only one of these extensions is supported?

test_conformance/vulkan/vulkan_interop_common/vulkan_utility.cpp
```diff
@@ -0,0 +1,700 @@
...
+{
+    std::vector<VulkanExternalSemaphoreHandleType> externalSemaphoreHandleTypeList;
+
+#if _WIN32 
+    if (IsWindows8OrGreater()) {
+        externalSemaphoreHandleTypeList.push_back(VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_WIN32_NT);
+    }
+    externalSemaphoreHandleTypeList.push_back(VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_WIN32_KMT);
+#else
+    externalSemaphoreHandleTypeList.push_back(VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_FD);
```
paulfradgley, line 184: Can we change this to allow the tests to run on platforms that only support SYNC_FD?

test_conformance/vulkan/test_vulkan_api_consistency.cpp
nikhiljnv: I see two cases here -->
1. cl_khr_external_memory/cl_khr_external_semaphore is not supported. In which case, we should have already skipped the test. 
2. cl_khr_external_memory/cl_khr_external_semaphore is supported. In which case, not supporting any of windows specific handles, should be an error. While it doesn't need to be cl_khr_external_memory_win32 and can be cl_khr_external_memory_dx in future, the test can not be skipped unconditionally and should evaluate minimum required support before deciding on skipping or error out.

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
nikhiljnv: Ack. We are working on updating this to avoid using NV extension.

test_conformance/vulkan/main.cpp
nikhiljnv: Currently, we have not isolated memory sharing and semaphore sharing cleanly to allow running only one of them.

test_conformance/vulkan/vulkan_interop_common/vulkan_utility.cpp
nikhiljnv: We currently don't have SYNC_FD support in this PR. But, we can allow only SYNC_FD when we add support for SYNC_FD.

nikhiljnv:
Thanks Kevin for the comments. More comments inline, but some of these sound like P1 and will probably start looking at addressing these after fixing immediate build and CI issues and addressing more specific comments affecting functionality and gaps wrt spec conformance.

> A very quick first pass on this. It's great to see tests starting to appear but we have a lot of work on our hands to integrate this into the CTS. A few overall questions/comments:
> 
>     * You should use object wrappers and RAII instead of flags, totos, etc. Much easier to reason about and maintain.
> 
[nikhiljnv] Will look into this.
>     * Have you considered using existing Vulkan bindings (e.g. https://github.com/KhronosGroup/Vulkan-Hpp) rather than rolling your own?
> 
[nikhiljnv] Will review and see if the same can be reused.
>     * Any reason Vulkan error codes are not checked at all?
> 
[nikhiljnv] No particular reason. This should be P0 after immediate build issues.
>     * Please avoid the use of malloc/free and prefer `std::string`, `std::vector`, etc
> 
[nikhiljnv] Sure. This is kind of P0 after immediate build issues to avoid rework in future.
>     * Consider introducing/using device query wrappers from https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/deviceInfo.h rather than copy/pasting code.
[nikhiljnv] Agreed. P0 after immediate build issues to avoid rework in future.

test_conformance/vulkan/vulkan_interop_common/vulkan_utility.hpp
nikhiljnv: Seems like a tiny fix. Will fix it along with build and CI issues.

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper_types.hpp
nikhiljnv: Seem like just convention with renaming. Will consider as part of cleanup and reusing standard Vulkan headers.

test_conformance/vulkan/main.cpp
nikhiljnv: Will consider this in a separate PR.

test_conformance/vulkan/procs.h
nikhiljnv: Will consider this as part of the same PR while fixing build and CI issues.

test_conformance/vulkan/CMakeLists.txt
nikhiljnv: Will consider as part of this PR, but likely to get pushed to a separate PR.

test_conformance/vulkan/vulkan_interop_common/vulkan_utility.hpp
nikhiljnv: This is fixed in latest revision.

test_conformance/vulkan/procs.h
nikhiljnv: Fixed.

test_conformance/vulkan/procs.h
nikhiljnv: Was borrowed from some other tests. Removed.

test_conformance/vulkan/main.cpp
nikhiljnv: This is fixed in latest revision.

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
```diff
@@ -0,0 +1,2073 @@
...
+
+#ifdef _WIN32
+#define NOMINMAX
+#include <Windows.h>
+#include <dxgi1_2.h>
+#include <aclapi.h>
+#endif
+#include <vulkan/vulkan.h>
+#include "vulkan_wrapper.hpp"
+#ifdef __linux__ && !__ANDROID__
```
bcalidas, line 25: change to #ifdef defined(__linux__) && !defined(__ANDROID__)

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
```diff
@@ -0,0 +1,2073 @@
...
+#include <gnu/libc-version.h>
+#include <dlfcn.h>
```
bcalidas, line 27: insert 

#elif defined(__ANDROID__)
#include <dlfcn.h>

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
```diff
@@ -0,0 +1,2073 @@
...
+///////////////////////////////////
+
+VulkanInstance::VulkanInstance(const VulkanInstance &instance)
+    : m_vkInstance(instance.m_vkInstance),
+      m_physicalDeviceList(instance.m_physicalDeviceList)
+{}
+
+VulkanInstance::VulkanInstance(): m_vkInstance(VK_NULL_HANDLE)
+{
+#if defined(__linux__)
```
bcalidas, line 60: change to #if defined(__linux__) && !defined(__ANDROID__)

bcalidas:
From our perspective, PR is now good to merge

PR #1391, opened 2022-02-24
Refactor kernel execution in subgroup tests

StuartDBrady:
This will simplify subsequent changes to test::run() in cases where multiple run-and-check sequences are needed for the testing of a single kernel.

### DIFF
--- a/test_conformance/subgroups/subhelpers.h
+++ b/test_conformance/subgroups/subhelpers.h
@@ -1322,73 +1322,129 @@ inline bool compare_ordered(const subgroups::cl_half &lhs, const int &rhs)
     return cl_half_to_float(lhs.data) == rhs;
 }
 
-// Run a test kernel to compute the result of a built-in on an input
-static int run_kernel(cl_context context, cl_command_queue queue,
-                      cl_kernel kernel, size_t global, size_t local,
-                      void *idata, size_t isize, void *mdata, size_t msize,
-                      void *odata, size_t osize, size_t tsize = 0)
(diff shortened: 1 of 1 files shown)


test_conformance/subgroups/subhelpers.h
```diff
@@ -1322,73 +1322,129 @@ inline bool compare_ordered(const subgroups::cl_half &lhs, const int &rhs)
...
+                   size_t l, Ty *id, size_t is, Ty *mid, Ty *mod, cl_int *md,
+                   size_t ms, Ty *od, size_t os, size_t ts = 0)
+        : context(c), queue(q), kernel(k), global(g), local(l), idata(id),
+          isize(is), mapin_data(mid), mapout_data(mod), mdata(md), msize(ms),
+          odata(od), osize(os), tsize(ts)
+    {
+        has_status = false;
+        run_failed = false;
+    }
+    cl_context context;
```
svenvh, line 1337: Most of these members could be `private` I believe (except for `run_failed`).  I am okay to leave that for a future cleanup though, as there are other classes/structs in this file that expose all of their members.

svenvh (approved):

test_conformance/subgroups/subhelpers.h
StuartDBrady: Agreed - I will open an issue for improving the coding style within the subgroup tests once the functionality is in a slightly better shape, and will include data member visibility as a concern.

PR #1398, opened 2022-03-07
Change memory scope for atomic fetch and load calls in kernel

lakshmih:
Change the memory scope from memory_scope_work_group to
memory_scope_device so the ordering applies across all work items

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -1042,7 +1042,7 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
             code += R"(
                 if(atomic_fetch_add_explicit(&finishedThreads, 1u,
                                            memory_order_relaxed,
-                                           memory_scope_work_group)
+                                           memory_scope_device)
                    == get_global_size(0)-1) // last finished thread
                    )";
         code += "    for(uint dstItemIdx = 0; dstItemIdx < numDestItems; "
@@ -1053,7 +1053,7 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
                 finalDest[dstItemIdx] =
                     atomic_flag_test_and_set_explicit(destMemory+dstItemIdx,
                                                       memory_order_relaxed,
-                                                      memory_scope_work_group);)";
+                                                      memory_scope_device);)";
         }
         else
         {
@@ -1061,7 +1061,7 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
                 finalDest[dstItemIdx] =
                     atomic_load_explicit(destMemory+dstItemIdx,
                                          memory_order_relaxed,
-                                         memory_scope_work_group);)";
+                                         memory_scope_device);)";
         }
     }
     code += "}\n"
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -1031,20 +1031,40 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
         }
         code += "\n";
     }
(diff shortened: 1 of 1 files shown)


alycm (commented):
Is this change required because the test is actually incorrect when using `memory_scope_work_group`?  It's quite hard to follow if this test is run with multiple work-groups or not.

If the change is required for correctness then support for `memory_scope_device` is optional in OpenCL 3.0 and it's not clear to me that will be checked by this test. `memory_scope_work_group` is mandatory, so it is safe to use in all contexts.  For wider scopes there is machinery in this file to only use supported scopes but this change may bypass that?

Sorry for the basic questions, I did try to answer them by skimming around this file but these tests are quite complicated.

lakshmih:
> Is this change required because the test is actually incorrect when using `memory_scope_work_group`? It's quite hard to follow if this test is run with multiple work-groups or not.
> 
> If the change is required for correctness then support for `memory_scope_device` is optional in OpenCL 3.0 and it's not clear to me that will be checked by this test. `memory_scope_work_group` is mandatory, so it is safe to use in all contexts. For wider scopes there is machinery in this file to only use supported scopes but this change may bypass that?
> 
> Sorry for the basic questions, I did try to answer them by skimming around this file but these tests are quite complicated.

Our interpretation - depending on the NDRange used and the verification logic - is  that the test intends to check integrity of loads and stores across the entire global size and hence needs the scope to device.

alycm:
Okay, thank you for the clarification.

Will the test be correctly skipped on devices that do not support `memory_scope_device`?  I.e. those that do not have `CL_DEVICE_ATOMIC_SCOPE_DEVICE` set in [`CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES`](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES).

My concern is that this PR changes the code from using a mandatory feature (`memory_scope_work_group`) to a feature that is optional in 3.0 (`memory_scope_device`) without also changing the code that checks whether a given configuration is expected to work.

It is possible that something else in the test means that such configs will already be skipped, but that is not immediately clear to me based on skimming the code.

Jeremy-Kemp:
I'm on vacation so can't review properly, but this looks related/the same as our PR #1377. I will be able to review properly next week.

test_conformance/c11_atomics/common.h
```diff
@@ -1053,15 +1053,15 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
                 finalDest[dstItemIdx] =
                     atomic_flag_test_and_set_explicit(destMemory+dstItemIdx,
                                                       memory_order_relaxed,
-                                                      memory_scope_work_group);)";
+                                                      memory_scope_device);)";
```
alycm, line 1056: It looks like it is possible to reach this line for the `LocalMemory()` path.  That means `destMemory` could be `__local`.  While `memory_scope_device` is allowed for local memory accesses it provides no additional guarantees over `memory_scope_work_group` for that case.

If it avoids need to skip this configuration for the local memory case then this value could be changed based on `LocalMemory()`.

test_conformance/c11_atomics/common.h
```diff
@@ -1053,15 +1053,15 @@ CBasicTest<HostAtomicType, HostDataType>::KernelCode(cl_uint maxNumDestItems)
...
         }
         else
         {
             code += R"(
                 finalDest[dstItemIdx] =
                     atomic_load_explicit(destMemory+dstItemIdx,
                                          memory_order_relaxed,
-                                         memory_scope_work_group);)";
+                                         memory_scope_device);)";
```
alycm, line 1064: Same comment as above re: `LocalMemory()`

lakshmih:
Not sure why the automated checks on this didnt progress so I resolved the comments since I addressed it in the latest patchset.

bashbaug:
Is this still needed now that #1377 is merged?

PR #1399, opened 2022-03-07
Use volatile qualifier to prevent optimizations

lakshmih:
Use volatile to prevent clang optimizations, fix int2float

### DIFF
--- a/test_conformance/conversions/basic_test_conversions.cpp
+++ b/test_conformance/conversions/basic_test_conversions.cpp
@@ -696,7 +696,8 @@ static void int2short( void *out, void *in){ ((cl_short*) out)[0] = ((cl_int*) i
 static void int2uint( void *out, void *in){ ((cl_uint*) out)[0] = ((cl_int*) in)[0]; }
 static void int2float( void *out, void *in)
 {
-    cl_int l = ((cl_int*) in)[0];
+    // Use volatile to prevent optimization by Clang compiler
+    volatile cl_int l = ((cl_int *)in)[0];
     ((float*) out)[0] = (l == 0 ? 0.0f : (float) l);        // Per IEEE-754-2008 5.4.1, 0's always convert to +0.0
 }
 static void int2double( void *out, void *in)


jlewis-austin (approved):
This change looks good, but I'm wondering if we should go ahead and hit all integer to floating-point permutations to avoid this issue coming up over and over? 
Also, if we do that I would love to change the variable "l" to literally anything else 😬

test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -696,7 +696,8 @@ static void int2short( void *out, void *in){ ((cl_short*) out)[0] = ((cl_int*) i
 static void int2uint( void *out, void *in){ ((cl_uint*) out)[0] = ((cl_int*) in)[0]; }
 static void int2float( void *out, void *in)
 {
-    cl_int l = ((cl_int*) in)[0];
+    // Use volatile to prevent optimization by Clang compiler
+    volatile cl_int l = ((cl_int *)in)[0];
```
alycm, line 700: I assume that the root cause of whatever Clang is doing is that type-punning is undefined behaviour?  If so then using `volatile` is really just papering over the problem.

Does something like the below work?  I haven't actually compiled or run the code so apologies for errors.  Note that the compiler will almost certainly optimize away the memcpy's.  It should, however, be a safe and defined way of handling the type punning.

```c++
static void int2float(void *out, void *in)
{
    cl_int l;
    memcpy(&l, in, sizeof(cl_int));
    float f = (l == 0 ? 0.0f : (float) l);  // Per IEEE-754-2008 5.4.1, 0's always convert to +0.0
    memcpy(out, &f, sizeof(float));
}
```

lakshmih: As mentioned earlier, this pattern repeats elsewhere in the test as well. See previous merged PR here:  https://github.com/KhronosGroup/OpenCL-CTS/pull/1062

alycm: Oh, I see.  I completely misunderstood the intent of the change.  If the input and output types are well typed should this function really be taking `void *` parameters?

Regarding a clang issue Jim, you reference [this issue](https://bugs.llvm.org/show_bug.cgi?id=47393) in the earlier MR, but as you say it sounds rtn specific.

It's hard for me to know if this is correct, but it does not have the issue that I thought it did so I will resolve this thread.

lakshmih:
@bashbaug  Can this be merged? Thanks.

lakshmih:
Created https://github.com/KhronosGroup/OpenCL-CTS/issues/1405 for follow-up work

PR #1402, opened 2022-03-14
Improve cl_khr_subgroup_shuffle* test coverage

StuartDBrady:
Test cases where the index/mask/delta is greater than or equal to the
maximum subgroup size.  These are cases that return undefined results
but are not undefined behavior.

The index/mask/delta values now include values less than twice the
subgroup size, and 0xffffffff.

Testing for sub_group_shuffle_xor() already allowed inputs that were
greater or equal to the subgroup size for the last subgroup in a
workgroup, but did not properly account for this in the verification
function, potentially resulting in out of bounds accesses.

### DIFF
--- a/test_conformance/subgroups/subgroup_common_templates.h
+++ b/test_conformance/subgroups/subgroup_common_templates.h
@@ -481,12 +481,12 @@ template <typename Ty, ShuffleOp operation> struct SHF
 
     static void gen(Ty *x, Ty *t, cl_int *m, const WorkGroupParams &test_params)
     {
-        int i, ii, j, k, l, n, delta;
+        int i, ii, j, k, n, delta;
+        cl_uint l;
         int nw = test_params.local_workgroup_size;
         int ns = test_params.subgroup_size;
         int ng = test_params.global_workgroup_size;
         int nj = (nw + ns - 1) / ns;
-        int d = ns > 100 ? 100 : ns;
         ii = 0;
         ng = ng / nw;
         for (k = 0; k < ng; ++k)
@@ -498,33 +498,10 @@ template <typename Ty, ShuffleOp operation> struct SHF
                 for (i = 0; i < n; ++i)
                 {
                     int midx = 4 * ii + 4 * i + 2;
-                    l = (int)(genrand_int32(gMTdata) & 0x7fffffff)
-                        % (d > n ? n : d);
-                    switch (operation)
-                    {
-                        case ShuffleOp::shuffle:
-                        case ShuffleOp::shuffle_xor:
-                            // storing information about shuffle index
-                            m[midx] = (cl_int)l;
-                            break;
-                        case ShuffleOp::shuffle_up:
-                            delta = l; // calculate delta for shuffle up
-                            if (i - delta < 0)
-                            {
-                                delta = i;
-                            }
-                            m[midx] = (cl_int)delta;
-                            break;
-                        case ShuffleOp::shuffle_down:
-                            delta = l; // calculate delta for shuffle down
-                            if (i + delta >= n)
-                            {
-                                delta = n - 1 - i;
-                            }
-                            m[midx] = (cl_int)delta;
-                            break;
-                        default: break;
-                    }
+                    l = (((cl_uint)(genrand_int32(gMTdata) & 0x7fffffff) + 1)
+                         % (ns * 2 + 1))
+                        - 1;
+                    m[midx] = l;
                     cl_ulong number = genrand_int64(gMTdata);
                     set_value(t[ii + i], number);
                 }
@@ -542,7 +519,8 @@ template <typename Ty, ShuffleOp operation> struct SHF
     static test_status chk(Ty *x, Ty *y, Ty *mx, Ty *my, cl_int *m,
                            const WorkGroupParams &test_params)
     {
-        int ii, i, j, k, l, n;
+        int ii, i, j, k, n;
+        cl_uint l;
         int nw = test_params.local_workgroup_size;
         int ns = test_params.subgroup_size;
         int ng = test_params.global_workgroup_size;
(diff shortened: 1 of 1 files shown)


gwawiork:
I made some experiment and modified your proposal adding in [line](https://github.com/KhronosGroup/OpenCL-CTS/blob/9ae9efb0373fc32f65e0b188337fecdbf3c2b9ef/test_conformance/subgroups/subgroup_common_templates.h#L573)  `else` condition in function `chk` to see how many values is undefined.
The second modification was to leave `gen` function without changes and I've run again test to see undefined values. 
I see a lot of undefined values `if (tr_idx < 0 || tr_idx >= n)`  in first case (all changes you propose) and in second case there is such case (0). So in my opinion your changes opposite to the title adds a lot of undefined values which are skipped while verification.

StuartDBrady:
> I see a lot of undefined values `if (tr_idx < 0 || tr_idx >= n)` in first case (all changes you propose) and in second case there is such case (0). So in my opinion your changes opposite to the title adds a lot of undefined values which are skipped while verification.

How do you explain this for `shuffle_xor`, for which there was no code in `gen` to limit the masks used, unlike for `shuffle_up` and `shuffle_down`? Without this, I would expect when XORed with the subgroup local ID, it may be the case (depending on the maximum subgroup size) that the resulting IDs would go beyond the subgroup size, for the last subgroup in the workgroup.

Having asked that, I accept that we now test cases that we did not previously for all of the shuffle ops, and will update the PR description and commit message to make that clearer.

I would still appreciate an answer for my question above regarding `shuffle_xor`, though. (Perhaps the workgroup size is a multiple of your subgroup size, which should be fixed by choosing a more appropriate workgroup size?)

gwawiork:
> I would still appreciate an answer for my question above regarding `shuffle_xor`, though. (Perhaps the workgroup size is a multiple of your subgroup size, which should be fixed by choosing a more appropriate workgroup size?)

Yes I see that  `shuffle_xor` should have kind of guard to not allow undefined values. It could be done by setting 0
` l = l ^ sb_local_id > sbs_size ? 0 : l`

StuartDBrady:
> > I would still appreciate an answer for my question above regarding `shuffle_xor`, though. (Perhaps the workgroup size is a multiple of your subgroup size, which should be fixed by choosing a more appropriate workgroup size?)
> 
> Yes I see that `shuffle_xor` should have kind of guard to not allow undefined values. It could be done by setting 0 ` l = l ^ sb_local_id > sbs_size ? 0 : l`

No, because although using these other inputs has an undefined result, it does not not have undefined behavior, and so they should still be tested.

StuartDBrady:
@gwawiork, I have updated this PR to go further in extending the testing of cases with undefined results for the cl_khr_subgroup_shuffle* extensions, and have updated the title, description and commit message to make it clearer that the main change is to the scope of the testing performed. However, this still affects testing of `sub_group_shuffle_xor()` as we discussed.

gwawiork:
> @gwawiork, I have updated this PR to go further in extending the testing of cases with undefined results for the cl_khr_subgroup_shuffle* extensions, and have updated the title, description and commit message to make it clearer that the main change is to the scope of the testing performed. However, this still affects testing of `sub_group_shuffle_xor()` as we discussed.

Thanks. I think I understand your point.  Let's test all inputs.

gwawiork (approved):
LGTM

test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -498,33 +498,9 @@ template <typename Ty, ShuffleOp operation> struct SHF
...
-                            {
-                                delta = n - 1 - i;
-                            }
-                            m[midx] = (cl_int)delta;
-                            break;
-                        default: break;
-                    }
+                    l = (((cl_uint)(genrand_int32(gMTdata) & 0x7fffffff) + 1)
+                         % (ns * 2 + 1))
+                        - 1;
```
bashbaug, line 503: Is this variable `l` used anywhere?  If not, can it be removed?

bashbaug: Th variable in the line above this `midx` doesn't look like it is used, either.

test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -567,32 +544,42 @@ template <typename Ty, ShuffleOp operation> struct SHF
...
+                        // shuffle down - treat l as delta
+                        case ShuffleOp::shuffle_down:
+                            if (l >= ns) skip = true;
+                            tr_idx = i + l;
                             break;
                         default: break;
                     }
 
-                    if (!compare(rr, tr))
+                    if (!skip && tr_idx < n)
```
bashbaug, line 570: I think this is OK, though perhaps a bit redundant.  In the cases above when `skip` is true, will there ever be a case when the index is valid (`tr_idx < n`)?  If not, debatably the check for `skip` is debatably unnecessary.

bashbaug (commented):
LGTM, this can be merged anytime.

test_conformance/subgroups/subgroup_common_templates.h
StuartDBrady: Will fix - my apologies!

test_conformance/subgroups/subgroup_common_templates.h
StuartDBrady: Yes, there will be such cases. For example, if `i` is 1 and `l` is 0xffffffff, then for _shuffle_up_, `tr_idx` would be 2 and `skip` would be _true_, and for _shuffle_down_, `tr_idx` would be 0 and `skip` would be again be _true_.

The check for `tr_idx < n` will likely be extended in future with a more elaborate test of the "work_item_mask" as used elsewhere, but the use of `skip` would remain unchanged.

Note that where `skip` is set, it is done on the basis of `ns` (the _maximum_ subgroup size), whereas the comparison with `tr_idx` is done against `n` (the _actual_ subgroup size). These relate to separate requirements in the spec, and so in my view, it's cleaner to keep the cases separate here, too.

If we were to move the `tr_idx < n` check into the individual cases, we'd need to merge it into _all four_ cases including _shuffle_ and _shuffle_xor_, with either a somewhat less-than-obvious expression simplification done for _shuffle_up_ and _shuffle_down_, or otherwise with unnecessary duplication of `tr_idx < n` as-is in all four cases.

test_conformance/subgroups/subgroup_common_templates.h
StuartDBrady: I have added the missing line.

bashbaug (approved):
Merging as we discussed in the March 29th teleconference.

Issue #1140, opened 2021-02-02
Kernel_read_write test - should other channel_data_type need max absolute error?


heiX-christing:
Should other channel_data_type need max absolute error? 
For example: CL_UNORM_INT_101010, CL_UNORM_SHORT_565 and so on.

https://github.com/KhronosGroup/OpenCL-CTS/blob/4abea6f761f6320e420b815352b2a7d7bc850c0a/test_common/harness/imageHelpers.cpp#L808

rjodinchr:
Fix: Implemented in PR #1406 (commit 35c21a8e: "imageHelpers: add CL_UNORM_SHORT_{555, 565} in get_max_absolute_error (#1406)").

PR #1406, opened 2022-03-24
imageHelpers: add CL_UNORM_SHORT_{555, 565} in get_max_absolute_error

rjodinchr:
Working on a device supporting CL_UNORM_SHORT_565 image data type, I
noticed that the max absolute error authorized was not the right one
for such image data type.

Also because of normalization, there is always an absolute error
authorized whatever the filtering of the sampler.

Ref #1140

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -912,6 +912,8 @@ void get_max_sizes(
 float get_max_absolute_error(const cl_image_format *format,
                              image_sampler_data *sampler)
 {
+    if (sampler->filter_mode == CL_FILTER_NEAREST) return 0.0f;
+
     switch (format->image_channel_data_type)
     {
         case CL_SNORM_INT8: return 1.0f / 127.0f;
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -912,8 +912,6 @@ void get_max_sizes(
 float get_max_absolute_error(const cl_image_format *format,
                              image_sampler_data *sampler)
 {
-    if (sampler->filter_mode == CL_FILTER_NEAREST) return 0.0f;
-
     switch (format->image_channel_data_type)
     {
         case CL_SNORM_INT8: return 1.0f / 127.0f;
@@ -924,6 +922,8 @@ float get_max_absolute_error(const cl_image_format *format,
 #ifdef CL_SFIXED14_APPLE
         case CL_SFIXED14_APPLE: return 0x1.0p-14f;
 #endif
+        case CL_UNORM_SHORT_555:
+        case CL_UNORM_SHORT_565: return 1.0f / 31.0f;
         default: return 0.0f;
     }
 }


bashbaug:
@outofcontrol can you please check why the CLA bot (and all other CI) seems to be stuck for this PR?  Thank you!

outofcontrol:
Look like a reviewer is required  to sign off first?

bashbaug:
Discussed via the mailing list - the CI jobs appear to be "stuck" for no good reason.

@rjodinchr can you please close and re-open your PR to see if that gets them "unstuck"?  If you cannot for whatever reason, please let me know, and I'll do my best to assist.

I'll add some review comments in the meantime.  Thank you for your patience!

test_common/harness/imageHelpers.cpp
```diff
@@ -912,8 +912,6 @@ void get_max_sizes(
 float get_max_absolute_error(const cl_image_format *format,
                              image_sampler_data *sampler)
 {
-    if (sampler->filter_mode == CL_FILTER_NEAREST) return 0.0f;
```
bashbaug, line 915: I tried to follow what this test is doing but it wasn't immediately apparent to me.

Do we know why the absolute error was forced to zero previously for NEAREST filtering?  Why is it correct to remove this check?  Removing the check means we'll have looser error bounds for formats beyond just CL_UNORM_SHORT_555 and CL_UNORM_SHORT_565.

test_common/harness/imageHelpers.cpp
```diff
@@ -924,6 +922,8 @@ float get_max_absolute_error(const cl_image_format *format,
 #ifdef CL_SFIXED14_APPLE
         case CL_SFIXED14_APPLE: return 0x1.0p-14f;
 #endif
+        case CL_UNORM_SHORT_555:
+        case CL_UNORM_SHORT_565: return 1.0f / 31.0f;
```
bashbaug, line 926: Do we care the the error bounds should be tighter (1.0f / 63.0f?) for the channel with six bits for CL_UNORM_SHORT_565?

bashbaug (commented):
Hi, can you please provde a little background which test was failing without these changes and why these changes allow the test to pass?  Thanks!

test_common/harness/imageHelpers.cpp
rjodinchr: I have not used cl images that much, so I might be wrong, but here is what I think:

The `filter_mode` allows you to have different strategies on which elements of the image to consider when the coordinates used to access are floating point coordinates.

Independently from the final coordinates chosen, an absolute error is created because of normalization:
Let's take `CL_UNORM_SHORT_555` for example
You start with a floating point value in the range `[0.0, 1.0]` , this value will be stored using 5bit (so in the integer range `[0, 31]`). During this first transformation, some data will be lost.

This is the same with all normalized formats.
Even when the `filter_mode` is `CL_FILTER_NEAREST` this error exists.

test_common/harness/imageHelpers.cpp
rjodinchr: I am not sure, but it would mean reworking the CTS quite a lot to have it properly implemented.

outofcontrol:
Appears closing and re-opening has got the builds running! Thank you @rjodinchr

lakshmih:
Qualcomm built and tested this patch. No issues were seen

test_common/harness/imageHelpers.cpp
jlewis-austin: I have the same concern as @bashbaug. The motivation for this whole function is unclear, but my best guess is that it's intended to allow an arbitrary amount of extra error for normalized types when in CL_FILTER_LINEAR mode. Removing the LINEAR conditional drastically loosens the requirements for all normalized types when in NEAREST mode so I don't think we should make that change.

As Ben mentioned, it would be helpful to get background on which test cases are fixed by this change.

test_common/harness/imageHelpers.cpp
jlewis-austin: Note that an (incorrect) attempt at the actual spec requirement for conversions is already in place in `get_max_relative_error` and includes these image types. The calling code uses this `get_max_absolute_error` function to override those limits by zeroing out any error within these (very loose) bounds.

rjodinchr:
This PR comes from a run of the OpenCL-CTS using [clvk](https://github.com/kpet/clvk) on several intel GPUs.
I don't have much time to allocate to this PR right now, so let's reduce it to just adding the cases for CL_UNORM_SHORT_555 and CL_UNORM_SHORT_565.

jlewis-austin (approved):
LGTM. This correctness of this function (and much of the code in general) is still unclear, but the changes are consistent with existing code and only affect CL_FILTER_LINEAR mode.

outofcontrol:
> Is there something missing on my side to have this PR merged?

@rjodinchr Nothing required from you. Waiting on reviewers to finalize and merge.

bashbaug (approved):

PR #1409, opened 2022-03-28
test_api_min_max.cpp: use size_t for get_global_id() value

JasonTTang:
In some rare cases where get_global_id() is larger than 2G, the 32bit int type would convert the value into a negative integer.

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -22,7 +22,7 @@
 const char *sample_single_param_kernel[] = {
     "__kernel void sample_test(__global int *src)\n"
     "{\n"
-    "    int  tid = get_global_id(0);\n"
+    "    size_t  tid = get_global_id(0);\n"
     "\n"
     "}\n"
 };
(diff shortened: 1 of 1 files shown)


test_conformance/api/test_api_min_max.cpp
```diff
@@ -22,16 +22,16 @@
...
     "}\n"
 };
 
 const char *sample_single_param_write_kernel[] = {
     "__kernel void sample_test(__global int *src)\n"
     "{\n"
-    "    int  tid = get_global_id(0);\n"
-    "     src[tid] = tid;\n"
+    "    size_t  tid = get_global_id(0);\n"
+    "    src[tid] = tid;\n"
```
jlewis-austin, line 34: There's still a potential issue assigning `size_t` to `int` here. We should make that cast explicit, and determine if the test is affected by the overflow situation.

jlewis-austin (commented):
Looks good! Just had one question where the tid is used as `int` data

test_conformance/api/test_api_min_max.cpp
JasonTTang: This kernel is unused, so I'm just removing it.

JasonTTang:
I am sorry, 'git push' didn't update this PR, and I somehow created a new one [https://github.com/KhronosGroup/OpenCL-CTS/pull/1410](https://github.com/KhronosGroup/OpenCL-CTS/pull/1410)

bashbaug:
Closing - #1410 is merged.

PR #1410, opened 2022-03-29
test_api_min_max.cpp: use size_t for get_global_id() value

JasonTTang:
In some rare cases where get_global_id() is larger than 2G, the 32bit int type would convert the value into a negative integer.


bashbaug (approved):
LGTM!

I'll merge after approval from @jlewis-austin , since he had comments on #1409.

jlewis-austin (approved):
LGTM!

PR #1411, opened 2022-03-31
Fix sub_group_ballot_find_msb/lsb tests

StuartDBrady:
As per the OpenCL Extension Specification § 38.6 Ballots:

   If no bits representing predicate values from all work items in
   the subgroup are set in the bitfield value then the return value
   is undefined.

The case with no bits set is still worth testing, as it does not result
in undefined behavior, but only an undefined return value.


bashbaug (approved):
LTGM!

bashbaug:
I'm going to go ahead and merge this - @kpet or @alycm or @jlewis-austin if you have any review comments we can address after the merge.

Issue #1096, opened 2021-01-05
Update readme


AnastasiaStulova:
It would be good to add basic information about code structure, formatting guidelines, build instructions, etc in the README file.

EwanC:
I feel like https://github.com/KhronosGroup/OpenCL-CTS/pull/1427 contains enough information to close this issue, and anything else we'd like to have in the README could be a new issue.

PR #1413, opened 2022-04-05
Update Github Actions CI and add Windows

kpet:
- Add one Windows build to Github Actions
- Remove Appveyor config
- Move a few build steps out of the script
- Use Ninja as the generator (makes for more readable logs)
- Add build cache (except on Windows where it seems to break)

Change-Id: Ida90ee1842af98aff86e5144ab7b9766480378c9


EwanC (approved):
An observation that before this change I could be lazy and run `presubmit.sh` to clean build the CTS locally. I think `presubmit.sh` should just be for CI so no objections to this change, but given we don't have any instructions in our README on how to build the project, it was nice to have all the steps in one place. This is probably just a thought towards us making progress with issue https://github.com/KhronosGroup/OpenCL-CTS/issues/1096

presubmit.sh
```diff
@@ -38,35 +38,36 @@ fi
...
       -DCMAKE_TOOLCHAIN_FILE=${TOOLCHAIN_FILE} \
       -DCMAKE_RUNTIME_OUTPUT_DIRECTORY=./bin \
-      -DOPENCL_LIBRARIES="-lOpenCL -lpthread" \
+      -DOPENCL_LIBRARIES="${CMAKE_OPENCL_LIBRARIES_OPTION}" \
       -DUSE_CL_EXPERIMENTAL=ON \
-      -DGL_IS_SUPPORTED=${BUILD_OPENGL_TEST} \
-      ..
-make -j2
+      -DGL_IS_SUPPORTED=${BUILD_OPENGL_TEST}
+cmake --build . -j3 --config Release
```
bashbaug, line 72: Mostly for my own education: is there a reason to pick `-j3` here vs. the previous `-j2`, or the generic `-j` or `--parallel`?

presubmit.sh
```diff
@@ -38,35 +38,36 @@ fi
...
-    sudo apt-get -y install libglu1-mesa-dev freeglut3-dev mesa-common-dev libglew-dev
 fi
-# Prepare headers
-git clone https://github.com/KhronosGroup/OpenCL-Headers.git
-cd OpenCL-Headers
-ln -s CL OpenCL # For OSX builds
-cd ..
 
 # Get and build loader
 git clone https://github.com/KhronosGroup/OpenCL-ICD-Loader.git
```
bashbaug, line 44: Is there a reason to fetch the headers as part of the GitHub actions YAML file but the ICD loader as part of this presubmit script?

bashbaug (approved):
LGTM, have a few questions, but more for my own education.  Thanks!

presubmit.sh
kpet: Some of the Github hosted runners have more than 2 cores and I've found having a few more jobs in flight generally speeds up jobs.

presubmit.sh
kpet: The headers could be moved very easily whereas moving the ICD loader fetch and build steps to the YAML file requires porting the whole cross-compilation logic. It should be done but it was more than I had time for when looking at this. Over time, I think we should move everything to the YAML file.

bashbaug: Cool, makes sense, this PR is a step in the right direction then (even if it breaks @EwanC 's workflow!).

If it's helpful I do fetch, build, and use the ICD loader via the YAML file in one of my projects:

https://github.com/bashbaug/opencl-extension-loader/blob/main/.github/workflows/cmake.yml#L83

It did require some special-casing that was not trivial, and I'm not 100% sure I've done it in the best way...

presubmit.sh
bashbaug: Would it speed up builds even more to pass `-j` without any specific number?

kpet: It might help when we get 3 cores or more but would reduce the overpressure when we have 2 cores (which is the common case). TBH, with the support for caches I added, it's a micro-optimisation except on Windows (where we should be able to get caching to work eventually).

presubmit.sh
kpet: Yup, sorry @EwanC :). Although, I agree with you that we should really write a decent README. A basic build is easy enough, it's the cross compilation logic that'd require a bit more work (toolchain files handling as a separate step, etc).

presubmit.sh
bashbaug: Works for me, resolving.

bashbaug:
This has two approvals and has been open a while, so merging.

Issue #574, opened 2020-02-04
Get rid of gDeviceType


kpet:
See discussion on https://github.com/KhronosGroup/OpenCL-CTS/pull/560.

PR #1418, opened 2022-04-21
images: Stop checking gDeviceType != CL_DEVICE_TYPE_GPU

jekstrand:
If the device type also advertises CL_DEVICE_TYPE_DEFAULT (which should
be valid), this causes it to be considered a CPU device and the tests
enforce different precision and rounding expectations.

### DIFF
--- a/test_conformance/images/kernel_read_write/test_common.cpp
+++ b/test_conformance/images/kernel_read_write/test_common.cpp
@@ -557,7 +557,7 @@ int test_read_image(cl_context context, cl_command_queue queue,
                                 // Apple requires its CPU implementation to do
                                 // correctly rounded address arithmetic in all
                                 // modes
-                                || gDeviceType != CL_DEVICE_TYPE_GPU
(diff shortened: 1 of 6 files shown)


karolherbst:
I think this issue is a bit bigger than that anyway. I just did a test with our CPU CL impl and treating it like a GPU makes things pass as well

kpet:
Linking https://github.com/KhronosGroup/OpenCL-CTS/issues/574

lakshmih (dismissed):
Qualcomm: Code change looks correct for current interpretation of spec.

jlewis-austin (approved):
Looks good, thanks! One optional cleanup comment; it's not critical since there's a general need for refactoring to cut down on all the duplicated logic.

bashbaug (approved):
LGTM.

IMHO there is a bigger issue here and these tests shouldn't be checking for a specific device type, given that the spec explicitly says:

> The device type is purely informational and has no semantic meaning.

But, at least the GPU device type check is correct now.

PR #1419, opened 2022-04-22
Some fixes for detecting optional features in CL 3.0

karolherbst:
Those are needed if your implementation doesn't support `cl_khr_fp16` or `read_write` images

### DIFF
--- a/test_conformance/half/Test_vLoadHalf.cpp
+++ b/test_conformance/half/Test_vLoadHalf.cpp
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 #include "harness/compat.h"
+#include "harness/deviceInfo.h"
 #include "harness/testHarness.h"
 
 #include <string.h>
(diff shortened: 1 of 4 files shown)


test_conformance/half/Test_vLoadHalf.cpp
```diff
@@ -35,6 +36,8 @@ int Test_vLoadHalf_private( cl_device_id device, bool aligned )
     uint64_t min_time[kVectorSizeCount+kStrangeVectorSizeCount] = {0};
     size_t q;
 
+    if (!is_extension_available(device, "cl_khr_fp16")) return 0;
```
EwanC, line 39: Could you elaborate on why we need these checks for `cl_khr_fp16`? My understanding is that `vload{a}_halfn` and `vstore{a}_halfn` are part of the core OpenCL-C spec rather than the fp16 extension, and so need support even if a device doesn't support this extension.

EwanC (changes requested):

test_conformance/half/Test_vLoadHalf.cpp
karolherbst: seems like you are right. It seems like I was simply hitting an clang bug then not providing those functions unless the extension was enabled.

karolherbst:
uhh, I had another patch here, which should be valid

PR #1021, opened 2020-10-20
Update readme

lakshmih:
Initial patch for a readme for the CTS repo. We filled in the Android sections, would be nice to have someone look at Linux and Windows.
Please also review the general sections.

PR #1427, opened 2022-05-12
Add content to README

EwanC:
Fill in the placeholder readme with some basic information on building and running the project. Information on the conformance submission process and contributing are also included.

Should help close a few issues referenced in https://github.com/KhronosGroup/OpenCL-CTS/issues/1096

I don't think this is all the information we want, but is a starting point from which we can progress. For example, adding the android build instructions from https://github.com/KhronosGroup/OpenCL-CTS/pull/1021

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,115 @@
-# OpenCL-CTS
-The OpenCL Conformance Tests
+# OpenCL Conformance Test Suite (CTS)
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
This looks really good - this is an excellent improvement!

PR #1432, opened 2022-06-01
Added missing tests for integer_dot_product_input_4x8bit and integer_dot_product_input_4x8bit_packed on feature_macro compiler test.

AhmedAmraniAkdi:
Currently, the feature macro test queries the device for the CL_DEVICE_OPENCL_C_FEATURES which returns an array of the optional features supported by the it.

Then does a check one by one to see whether they are supported or not. If they are supported they are appended to an array that gets compared to the one returned by the CL_DEVICE_OPENCL_C_FEATURES query.

Now, devices supporting cl_khr_integer_dot_product will return __opencl_c_integer_dot_product_input_4x8bit and
__opencl_c_integer_dot_product_input_4x8bit_packed but since we never test for them they never get appended and so the comparison fails, and so does the test.

This PR adds the 2 missing tests.

### DIFF
--- a/test_conformance/compiler/test_feature_macro.cpp
+++ b/test_conformance/compiler/test_feature_macro.cpp
@@ -579,22 +579,25 @@ int test_feature_macro_fp64(cl_device_id deviceID, cl_context context,
                                         compiler_status, supported);
 }
 
-int test_feature_macro_integer_dot_product_input_4x8bit_packed(cl_device_id deviceID, cl_context context,
-                             								   std::string test_macro_name, cl_bool& supported)
+int test_feature_macro_integer_dot_product_input_4x8bit_packed(
+    cl_device_id deviceID, cl_context context, std::string test_macro_name,
+    cl_bool& supported)
 {
     cl_int error = TEST_FAIL;
     cl_bool api_status;
     cl_bool compiler_status;
     log_info("\n%s ...\n", test_macro_name.c_str());
-    
-	if (!is_extension_available(deviceID, "cl_khr_integer_dot_product"))
-	{
-		supported = false;
-		return TEST_PASS;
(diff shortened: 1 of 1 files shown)


bashbaug (dismissed):
LGTM.

It seems a little unfortunate that an unknown feature macro will cause this test to fail, especially for e.g. vendor extensions or other non-standard extensions.  Should the test warn on an unknown extension vs. fail?

I certainly think these feature macros should be tested because they're from a standard extension, but I also think this won't always be the case.

test_conformance/compiler/test_feature_macro.cpp
```diff
@@ -703,11 +766,25 @@ int test_consistency_c_features_list(cl_device_id deviceID,
...
         log_info("%s\n", each_f.c_str());
     }
 
+    for (auto each_f : vec_to_cmp)
+    {
+        if (find(vec_device_feature_names.begin(),
+                 vec_device_feature_names.end(), each_f)
+            == vec_device_feature_names.end())
+        {
+            log_info("Comparison list of features - failed\n");
```
bashbaug, line 781: Nice-to-have: Could we print the missing feature here?

bashbaug (dismissed):
This latest version LGTM and matches the logic we discussed in last week's teleconference.  Thanks!

test_conformance/compiler/test_feature_macro.cpp
bashbaug:
```suggestion
            log_info("Comparison list of features - failed - missing %s\n",
                     each_f.c_str());
```

AhmedAmraniAkdi: Done.

PR #1436, opened 2022-06-14
Include headers from the specified directory on Apple

daemyung:
Currently I'm building [CLMTL](https://github.com/daemyung/clmtl) which is OpenCL over Metal. I want to use OpenCL CTS to verify my implementation. But there is no way to include OpenGL headers because including path is embedded to Apple's OpenCL framework. This PR allows to include OpenCL headers.

### DIFF
--- a/test_common/gl/setup.h
+++ b/test_common/gl/setup.h
@@ -20,7 +20,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include "gl_headers.h"
-#ifdef __APPLE__
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/opencl.h>
--- a/test_common/harness/ThreadPool.h
+++ b/test_common/harness/ThreadPool.h
@@ -16,7 +16,7 @@
 #ifndef THREAD_POOL_H
 #define THREAD_POOL_H
 
-#if defined(__APPLE__)
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/cl.h>
--- a/test_common/harness/clImageHelper.h
+++ b/test_common/harness/clImageHelper.h
@@ -16,7 +16,7 @@
 #ifndef test_conformance_clImageHelper_h
 #define test_conformance_clImageHelper_h
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/cl.h>
--- a/test_common/harness/errorHelpers.h
+++ b/test_common/harness/errorHelpers.h
@@ -18,7 +18,7 @@
 
 #include <sstream>
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/opencl.h>
--- a/test_common/harness/kernelHelpers.h
+++ b/test_common/harness/kernelHelpers.h
@@ -31,7 +31,7 @@
 
 #include <string.h>
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/opencl.h>
--- a/test_common/harness/mt19937.h
+++ b/test_common/harness/mt19937.h
@@ -49,7 +49,7 @@
 #ifndef MT19937_H
 #define MT19937_H 1
 
-#if defined(__APPLE__)
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/cl_platform.h>
 #else
 #include <CL/cl_platform.h>
--- a/test_common/harness/threadTesting.h
+++ b/test_common/harness/threadTesting.h
@@ -16,7 +16,7 @@
 #ifndef _threadTesting_h
 #define _threadTesting_h
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/opencl.h>
--- a/test_conformance/api/test_null_buffer_arg.cpp
+++ b/test_conformance/api/test_null_buffer_arg.cpp
@@ -14,7 +14,7 @@
 // limitations under the License.
 //
 #include <stdio.h>
-#if defined(__APPLE__)
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #include <OpenCL/cl_platform.h>
 #else
--- a/test_conformance/math_brute_force/reference_math.h
+++ b/test_conformance/math_brute_force/reference_math.h
@@ -16,7 +16,7 @@
 #ifndef REFERENCE_MATH_H
 #define REFERENCE_MATH_H
 
-#if defined(__APPLE__)
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/cl.h>
--- a/test_conformance/spir/kernelargs.h
+++ b/test_conformance/spir/kernelargs.h
@@ -17,7 +17,7 @@
 #define __KERNELARGS_H
 
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/cl.h>
--- a/test_conformance/spir/run_build_test.cpp
+++ b/test_conformance/spir/run_build_test.cpp
@@ -15,7 +15,7 @@
 //
 #include "harness/compat.h"
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/cl.h>
--- a/test_conformance/spir/run_services.cpp
+++ b/test_conformance/spir/run_services.cpp
@@ -15,7 +15,7 @@
 //
 #include "harness/compat.h"
 
-#ifdef __APPLE__
+#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
 #include <OpenCL/opencl.h>
 #else
 #include <CL/cl.h>


bashbaug:
Hello, thank you for your contribution, and sorry for the delay!  Your project sounds very interesting.

Since we are currently defining `CL_TARGET_OPENCL_VERSION` as `300` unconditionally in our top-level CMakefile I think that this means the OpenCL framework headers will never be used on MacOS - is this intended?

Here's the line in the CMakefile:

```cmake
add_definitions(-DCL_TARGET_OPENCL_VERSION=300)
```

Here's the ifdef added in this PR:

```c
#if defined(__APPLE__) && CL_TARGET_OPENCL_VERSION <= 120
```

As an alternative, would it make more sense to have a conditional build option to either force-disable the framework headers or force-use the upstream OpenCL headers instead (not sure which "polarity" would be better)?  We could do this similar to the current option for "experimental" definitions, e.g.:

```cmake
option(FORCE_CL_HEADERS "Use the OpenCL Headers Unconditionally" OFF)
if(FORCE_CL_HEADERS )
  add_definitions(-DFORCE_CL_HEADERS )
endif(FORCE_CL_HEADERS )
```

and then, guard the include with:

```c
#if defined(__APPLE__) && !defined(FORCE_CL_HEADERS)
#include <OpenCL/opencl.h>
#else
#include <CL/opencl.h>
```

We could probably even add the option only for MacOS builds, if we really wanted to.

Separately: Should we add this scenario (MacOS w/ non-framework headers) to our CI?

daemyung:
@bashbaug Yes exactly. As far as I know, Apple support only OpenCL 120 so I add ifdef like that. Your option is pretty fine so I will update it.

Also I don't think you don't have to add this scenario to your CI. Because Apple deprecated OpenCL on their platforms and Metal can't fully cover OpenCL spec.

bashbaug:
> Also I don't think you don't have to add this scenario to your CI.

To be clear, I was only suggesting to add the build config with this new option to CI, not to run the CTS tests on OSX.

However, your comment prompted me to have a look at how our CI is testing MacOS right now, and it looks like we're already building with the upstream OpenCL headers vs. the framework headers.  See the "For OSX builds" symlink line:

```
      - name: Fetch OpenCL Headers
        shell: bash
        run: |
          git clone https://github.com/KhronosGroup/OpenCL-Headers.git
          cd OpenCL-Headers
          ln -s CL OpenCL # For OSX builds
          cd ..
```

This makes me wonder if it's even worth bothering with the framework headers or if we should just use the upstream OpenCL headers unconditionally.  Is there any way to build the CTS tests with the framework headers right now?

daemyung:
> Is there any way to build the CTS tests with the framework headers right now?

In CI, I think there is no way to see the framework headers right now and On the bare metal, you can use the framework headers with doing nothing.

Also I think `ln -s CL OpenCL # For OSX builds` is W/A to see the upstream OpenCL headers. So I think doing it explicitly will be better than the current implicit way.

kpet:
I've looked at how we manage OpenCL headers throughout the CTS in the past and IIRC, I'd come to the conclusion that the cleanest, most readable solution to this issue and others was to introduce a wrapper header that'd be the only place where we include OpenCL headers and include that wrapper header in all the test and common code instead. It's a fair bit of tedious work but would give us good centralised control over what OpenCL headers are used and how as well as a uniformity guarantees for the whole code base.

bashbaug:
@kpet do you think a wrapper header would still be needed if we switched to use the upstream OpenCL headers unconditionally?

If we use the upstream OpenCL headers unconditionally then I think we would be able to `#include <CL/opencl.h>` in all cases where we need OpenCL headers so the value of a wrapper header would be minimal.  If we still need some kind of conditional inclusion via ifdefs though, I agree a common wrapper header could be useful.

kpet:
>@kpet do you think a wrapper header would still be needed if we switched to use the upstream OpenCL headers unconditionally?

I can't think of an immediate need for it but it'd make it very easy in the future to define macros globally both before and after the inclusion of the OpenCL Headers which can come in very handy for experiments or when developing new features. Plus, switching to using the upstream headers unconditionally requires doing all the tedious work of changing all inclusions anyway so we might as well introduce the indirection then and give ourselves the option.

bashbaug:
OK, I'm convinced, an "OpenCL header wrapper" seems useful.  This could even be used to move some of the macros out of the CMakefile into the header wrapper, which might be a little less confusing.

I think this is more work than is reasonable for this PR, though, which brings me back to my original question: Should we still (try to) support the framework headers?  Or can we remove support for the framework headers - for now at least - and require the upstream headers instead?

Note that some tests - rightly or wrongly - are already including only the upstream headers (e.g. test_device_timer.cpp).

bashbaug:
Trying to un-stick this PR so we can merge the changes in some form.

It looks like this PR modifies some of the places where framework headers are included, but not all.  I think we should fix all of them.  If it's easy enough to do this with an "OpenCL header wrapper" then I think doing so would be ideal, but the total number of affected files doesn't seem to be that big (~20?), so it's not a requirement from my POV.

I'm fine removing support for the framework headers entirely.  At this point I think the upstream headers are effectively required.  If we really want to we can try to get some tests working with the framework headers in a subsequent PR but it doesn't sound like this is a use-case requirement right now.

@kpet does this sound reasonable?

Here are the files I see that are including framework headers that should be updated.  It's still possible this missed a few, but if we can remove the symlink from our CI builds for OSX that should give us confidence we got them all.

```sh
>git grep -B1 -A1 -i "include <OpenCL/"
test_common/gl/setup.h-#ifdef __APPLE__
test_common/gl/setup.h:#include <OpenCL/opencl.h>
test_common/gl/setup.h-#else
--
test_common/harness/ThreadPool.h-#if defined(__APPLE__)
test_common/harness/ThreadPool.h:#include <OpenCL/opencl.h>
test_common/harness/ThreadPool.h-#else
--
test_common/harness/clImageHelper.h-#ifdef __APPLE__
test_common/harness/clImageHelper.h:#include <OpenCL/opencl.h>
test_common/harness/clImageHelper.h-#else
--
test_common/harness/errorHelpers.h-#ifdef __APPLE__
test_common/harness/errorHelpers.h:#include <OpenCL/opencl.h>
test_common/harness/errorHelpers.h-#else
--
test_common/harness/kernelHelpers.h-#ifdef __APPLE__
test_common/harness/kernelHelpers.h:#include <OpenCL/opencl.h>
test_common/harness/kernelHelpers.h-#else
--
test_common/harness/mt19937.h-#if defined(__APPLE__)
test_common/harness/mt19937.h:#include <OpenCL/cl_platform.h>
test_common/harness/mt19937.h-#else
--
test_common/harness/threadTesting.h-#ifdef __APPLE__
test_common/harness/threadTesting.h:#include <OpenCL/opencl.h>
test_common/harness/threadTesting.h-#else
--
test_conformance/api/test_null_buffer_arg.cpp-#if defined(__APPLE__)
test_conformance/api/test_null_buffer_arg.cpp:#include <OpenCL/opencl.h>
test_conformance/api/test_null_buffer_arg.cpp:#include <OpenCL/cl_platform.h>
test_conformance/api/test_null_buffer_arg.cpp-#else
--
test_conformance/conversions/basic_test_conversions.h-#if defined( __APPLE__ )
test_conformance/conversions/basic_test_conversions.h:    #include <OpenCL/opencl.h>
test_conformance/conversions/basic_test_conversions.h-#else
--
test_conformance/events/test_userevents.cpp-#if defined(__APPLE__)
test_conformance/events/test_userevents.cpp:#include <OpenCL/opencl.h>
test_conformance/events/test_userevents.cpp-#include <mach/mach_time.h>
--
test_conformance/half/cl_utils.h-#ifdef __APPLE__
test_conformance/half/cl_utils.h:#include <OpenCL/opencl.h>
test_conformance/half/cl_utils.h-#else
--
test_conformance/math_brute_force/function_list.h-#if defined(__APPLE__)
test_conformance/math_brute_force/function_list.h:#include <OpenCL/opencl.h>
test_conformance/math_brute_force/function_list.h-#else
--
test_conformance/math_brute_force/reference_math.h-#if defined(__APPLE__)
test_conformance/math_brute_force/reference_math.h:#include <OpenCL/opencl.h>
test_conformance/math_brute_force/reference_math.h-#else
--
test_conformance/mem_host_flags/testBase.h-#else
test_conformance/mem_host_flags/testBase.h://#include <OpenCL/cl.h>
test_conformance/mem_host_flags/testBase.h-#endif
--
test_conformance/printf/test_printf.h-#ifdef __APPLE__
test_conformance/printf/test_printf.h:#include <OpenCL/opencl.h>
test_conformance/printf/test_printf.h:#include <OpenCL/cl_platform.h>
test_conformance/printf/test_printf.h-#else
--
test_conformance/select/test_select.h-#ifdef __APPLE__
test_conformance/select/test_select.h:#include <OpenCL/opencl.h>
test_conformance/select/test_select.h-#else
--
test_conformance/spir/kernelargs.h-#ifdef __APPLE__
test_conformance/spir/kernelargs.h:#include <OpenCL/opencl.h>
test_conformance/spir/kernelargs.h-#else
--
test_conformance/spir/run_build_test.cpp-#ifdef __APPLE__
test_conformance/spir/run_build_test.cpp:#include <OpenCL/opencl.h>
test_conformance/spir/run_build_test.cpp-#else
--
test_conformance/spir/run_services.cpp-#ifdef __APPLE__
test_conformance/spir/run_services.cpp:#include <OpenCL/opencl.h>
test_conformance/spir/run_services.cpp-#else
--
test_conformance/vulkan/main.cpp-#else
test_conformance/vulkan/main.cpp:#include <OpenCL/cl.h>
test_conformance/vulkan/main.cpp-#endif
--
test_conformance/vulkan/test_vulkan_api_consistency.cpp-#else
test_conformance/vulkan/test_vulkan_api_consistency.cpp:#include <OpenCL/cl.h>
test_conformance/vulkan/test_vulkan_api_consistency.cpp:#include <OpenCL/cl_ext.h>
test_conformance/vulkan/test_vulkan_api_consistency.cpp-#endif
--
test_conformance/vulkan/vulkan_interop_common/opencl_vulkan_wrapper.hpp-#else
test_conformance/vulkan/vulkan_interop_common/opencl_vulkan_wrapper.hpp:#include <OpenCL/cl.h>
test_conformance/vulkan/vulkan_interop_common/opencl_vulkan_wrapper.hpp:#include <OpenCL/cl_ext.h>
test_conformance/vulkan/vulkan_interop_common/opencl_vulkan_wrapper.hpp-#endif
```

kpet:
@bashbaug Works for me!

bashbaug:
Great.  @daemyung are next steps clear?

daemyung:
@bashbaug Yes. Just for clear, create a warpper header and use it. Right? Could you recommend the file name of a wrapper header?

bashbaug:
Hi @daemyung, at a minimum please update all of the files listed above.  You have many of them already in this PR, but not all of them.

Because we don't expect the CTS tests to work with the OpenCL framework headers it's OK to remove support entirely and to switch to the upstream headers exclusively.  Once we've done this we should be able to remove the symbolic linking from the CI script and the OSX builds will still work:

`ln -s CL OpenCL # For OSX builds`

The header wrapper is a nice-to-have but not a must-have.  If you decide to add the header wrapper, how about naming the file `clHeaderWrapper.h`, in the `test_common/harness` directory?

Please don't hestiate to reach out if you have additional questions - thanks!

bashbaug:
Temporarily removing the "focused review" label - we'll add it back when there are new changes to review.

daemyung:
@bashbaug Sorry for late update. I'm in busy right now. I will update this PR until next month.

PR #1439, opened 2022-06-15
Add tests for cl_khr_subgroup_rotate

StuartDBrady:

### DIFF
--- a/test_conformance/subgroups/subgroup_common_templates.h
+++ b/test_conformance/subgroups/subgroup_common_templates.h
@@ -501,7 +501,31 @@ template <typename Ty, ShuffleOp operation> struct SHF
                     l = (((cl_uint)(genrand_int32(gMTdata) & 0x7fffffff) + 1)
                          % (ns * 2 + 1))
                         - 1;
-                    m[midx] = l;
+                    switch (operation)
+                    {
+                        case ShuffleOp::shuffle:
+                        case ShuffleOp::shuffle_xor:
+                        case ShuffleOp::shuffle_up:
+                        case ShuffleOp::shuffle_down:
+                            // storing information about shuffle index/delta
+                            m[midx] = (cl_int)l;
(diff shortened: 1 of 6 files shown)


test_conformance/subgroups/subgroup_common_templates.h
```diff
@@ -565,6 +589,15 @@ template <typename Ty, ShuffleOp operation> struct SHF
...
                             tr_idx = i + l;
                             break;
+                        // rotate - treat l as delta
+                        case ShuffleOp::rotate:
+                            tr_idx = (i + l) % test_params.subgroup_size;
+                            break;
+                        case ShuffleOp::clustered_rotate: {
+                            tr_idx = ((i & ~(test_params.cluster_size - 1))
+                                      + ((i + l) % test_params.cluster_size));
+                            break;
```
StuartDBrady, line 599: Thanks, @jlewis-austin, for your review under PR #1435.  I've tidied up the code as you suggested.  It was previously:

```
                        // rotate - treat l as delta
                        case ShuffleOp::rotate:
                        case ShuffleOp::clustered_rotate: {
                            if (operation == ShuffleOp::rotate)
                            {
                                tr_idx = (i + l) % test_params.subgroup_size;
                            }
                            else // operation == ShuffleOp::clustered_rotate
                            {
                                tr_idx =
                                    ((i & ~(test_params.cluster_size - 1))
                                     + ((i + l) % test_params.cluster_size));
                            }
                            break;
                        }
```
As you pointed out:
> these two cases don't share any code; it seems cleaner to avoid the fall through and just handle each in its own block

StuartDBrady:
(It turns out that renaming a branch in GitHub automatically closes any PRs raised from it without warning. Apologies for the noise.)

kpet:
Merging as agreed during the 2022/06/28 teleconference.

PR #1449, opened 2022-07-18
fix test kernel attributes when api fcts are failing

rjodinchr:
test_error returns the err given as the first argument. As the
run_test function returns a bool, we end up returning true (meaning
pass) when an api function fails.
Instead return explicitly false (meaning fail).

### DIFF
--- a/test_conformance/api/test_kernel_attributes.cpp
+++ b/test_conformance/api/test_kernel_attributes.cpp
@@ -275,16 +275,16 @@ static bool run_test(cl_context context, cl_device_id deviceID,
         clKernelWrapper kernel;
         cl_int err = create_single_kernel_helper(context, &program, &kernel, 1,
                                                  &kernel_src, "test_kernel");
-        test_error(err, "create_single_kernel_helper");
+        test_error_ret(err, "create_single_kernel_helper", false);
 
         // Get the size of the kernel attribute string returned
         size_t size = 0;
         err = clGetKernelInfo(kernel, CL_KERNEL_ATTRIBUTES, 0, nullptr, &size);
-        test_error(err, "clGetKernelInfo");
+        test_error_ret(err, "clGetKernelInfo", false);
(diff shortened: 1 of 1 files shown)


svenvh (approved):
LGTM, thanks!

bashbaug (approved):
Nice catch!  This looks like an easy mistake to make.  Can we check if other tests made a similar mistake?  Is there a way we can make this less error-prone?

rjodinchr:
> Nice catch! This looks like an easy mistake to make. Can we check if other tests made a similar mistake? Is there a way we can make this less error-prone?

int/bool conversions are quite hard to catch in C. I think `clang-tidy` tried something, but it was just printing too many false positives so they stopped.

I have just run `git grep -p "test_error(" | grep -A 10 bool`, to look for similar things.
Here are the ones that can be problematic:
```
test_common/harness/kernelHelpers.cpp=1784=bool device_supports_double(cl_device_id device)
test_common/harness/kernelHelpers.cpp:1796:        test_error(err,
--
test_conformance/SVM/common.h=43=bool AtomicCompareExchangeStrongExplicit(volatile T *a, T *expected, T desired,
test_conformance/SVM/common.h:64:#define test_error2(error_code, ptr, msg)  { if(error != 0)  { test_error(error_code, msg); } else  { if(NULL == ptr)  {print_null_error(msg); return -1;} } }
--
test_conformance/api/test_kernel_attributes.cpp=264=static bool run_test(cl_context context, cl_device_id deviceID,
test_conformance/api/test_kernel_attributes.cpp:278:        test_error(err, "create_single_kernel_helper");
test_conformance/api/test_kernel_attributes.cpp:283:        test_error(err, "clGetKernelInfo");
test_conformance/api/test_kernel_attributes.cpp:287:        test_error(err, "clGetKernelInfo");
```
Obviously the last one is the one fixed by this PR

Issue #1274, opened 2021-06-11
C11 Atomics: Race condition in atomic_fence


Jeremy-Kemp:
We have identified an issue in the atomic_fence test - specifically `release atomic synchronizes-with acquire fence, local atomic_int, scope_work_group` in the case described here.

@kpet pointed out [Internal Bugzilla 15814](https://www.khronos.org/members/login/bugzilla/show_bug.cgi?id=15814) which appears to be the same underlying issue - that instances _may_ make independent forward progress.

The high level summary is that we think that it is possible for the test to fail if interleaving threads execute their instructions in a particular order which is valid, but leads to an incorrect test failure.

Suppose we have 4 instances 0, 1, 2, and 3 which can each make independent forward progress (with a workgroup size of 4).

The following timeline is possible showing how multiple instances accessing the same address races (showing operations from instances 0, 1, and 2):
```C
localValues[myId*1+myValue] = myValue; // Instance 0 writes localValues[1] = 1
atomic_store_explicit(&destMemory[myId], myValue, memory_order_release, memory_scope_work_group); // Instance 0 writes destMemory[0] = 1
hisAtomicValue = atomic_load_explicit(&destMemory[hisId], memory_order_relaxed, memory_scope_work_group); // Instance 0 reads destMemory[3] which is 0 as instance 3 may not have written it yet
localValues[myId*1+myValue] = myValue; // Instance 2 writes localValues[3] = 1.
atomic_work_item_fence(CLK_LOCAL_MEM_FENCE, memory_order_acquire, memory_scope_work_group); // Instance 0
hisValue = localValues[hisId*1+hisAtomicValue];  // Instance 0 Reads localValues[3] which is 1
if(hisAtomicValue != hisValue)     // Instance 0 - At this point hisAtomicValue != hisValue so test failure
```
We do not yet have a proposal for fixing the kernel. Instead, looking to spark some discussion around the issue and test itself so that we can come up with a proposal to fix (whilst maintaining the original intent).

For reference, the entire kernel is:
```C
__kernel void test_atomic_kernel(uint threadCount, uint numDestItems, __global int *finalDest, __global int *oldValues, volatile __local atomic_int *destMemory, __local int *localValues)
{
    uint tid = get_global_id(0);
 
    // initialize atomics not reachable from host (first thread
    // is doing this, other threads are waiting on barrier)
    if (get_local_id(0) == 0)
[...]

PR #1452, opened 2022-07-27
Minimum 2 non atomic variables per thread for the c11 atomic fence test for embedded profile devices.

AhmedAmraniAkdi:
Having 1 non atomic variable per thread on the c11 atomic fence test would sometimes make it so that one of the threads of the pair that is trying to synchronise will read the memory written by another thread not part of this pair causing a failure.

More info: https://github.com/KhronosGroup/OpenCL-CTS/issues/1274.


paulfradgley (approved):

bashbaug (approved):
Merging as discussed in the September 5th teleconference.

PR #1456, opened 2022-08-02
Casting gptr & lptr to a generic pointer before the comparison

asudarsa:


asudarsa:
Code format failure is in a location which is not touched by this change. Thanks.

bashbaug:
Interesting - I thought this wouldn't be necessary but this is changing how Clang generates code for the comparisons.

@AnastasiaStulova is this expected?

https://godbolt.org/z/PGPEjxj4T

AnastasiaStulova:
My guess is that this change forces comparison to be done using generic address space pointers instead of the named ones... but the default behavior should be a generic address space comparison. However without explicit cast clang seems to convert to a named address space for the comparison: https://godbolt.org/z/rorvWKcGM. 


This seems like a bug to me as embedded C rules say this about the pointer equality operator:

> If the two operands are pointers into different address spaces, one of the address spaces
> encloses the other. The pointer into the enclosed address space is first converted to a pointer to
> the same reference type except with any address-space qualifier removed and any address-
> space qualifier of the other pointer's reference type added. (After this conversion, both
> operands are pointers into the same address space.)

I don't think OpenCL overrules this...

PR #1459, opened 2022-08-02
fix indentation of test_waitlists.cpp

matthiasdiener:
Followup of #1458

### DIFF
--- a/test_conformance/events/test_waitlists.cpp
+++ b/test_conformance/events/test_waitlists.cpp
@@ -29,9 +29,12 @@ int test_waitlist( cl_device_id device, cl_context context, cl_command_queue que
     cl_int error;
 
     if (multiple)
-        log_info("\tExecuting reference event 0, then reference event 1 with reference event 0 in its waitlist, then test event 2 with reference events 0 and 1 in its waitlist.\n");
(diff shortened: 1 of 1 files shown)


matthiasdiener:
This is ready for review. Currently, compilation of this test fails due to `-Wmisleading-indentation`.

StuartDBrady (approved):
LGTM.

StuartDBrady:
@matthiasdiener, am wondering why this wasn't caught by the build checks (GitHub Actions)—do you have any suggestions?

(My best guess right now is that this is caused by differences in what is flagged as misleading indentation by different compilers—the version of GCC that I'm using did not catch this, nor did that used for our automated builds, but other GCC versions may do, as may Clang—so presumably that is what happened here.)

matthiasdiener:
> @matthiasdiener, am wondering why this wasn't caught by the build checks (GitHub Actions)—do you have any suggestions?
> 
> (My best guess right now is that this is caused by differences in what is flagged as misleading indentation by different compilers—the version of GCC that I'm using did not catch this, nor did that used for our automated builds, but other GCC versions may do, as may Clang—so presumably that is what happened here.)

This seems like the most likely explanation - I saw this warning with gcc 11.3.0-5 from Debian testing.

Issue #1445, opened 2022-06-22
test_select always passes


alyssarosenzweig:
check_float says there's an error, but at the end it says PASSED. Is this a CTS bug? Is this a driver bug? Is it both?

```
alyssa@rootfs:~/OpenCL-CTS/build/test_conformance/select$ OCL_ICD_VENDORS=/home/alyssa/ ./test_select select_float_int   
Test binary built Jun 22 2022 13:08:30
 Initializing random seed to 0.
Requesting Default device based on command line for platform index 0 and device index 0
Compute Device Name = Mali-G57 (Panfrost), Compute Device Vendor = Arm, Compute Device Version = OpenCL 3.0, CL C Version = OpenCL C 1.2 
Device latest conformance version passed: v0000-01-01-00
Supports single precision denormals: NO
sizeof( void*) = 8  (host)
sizeof( void*) = 8  (device)
select_float_int...
Building select_float_int(float, float, int)
Building select_float2_int2(float2, float2, int2)
Building select_float3_int3(float3, float3, int3)
Building select_float4_int4(float4, float4, int4)
Building select_float8_int8(float8, float8, int8)
Building select_float16_int16(float16, float16, int16)
Testing...
(check_float) Error for vector size 1 found at 0x00000000 (of 0x00040000):  *0x8c7f0aac vs 0xffffffff
vec_size:1 indx: 0x0000000000000000
select_float_int passed
PASSED sub-test.
PASSED test.
```

alyssarosenzweig:
test-select increments `s_test_fail` to signal fails, but the test harness checks for `gFailCount`, which will always be zero for this subtest. So no matter what the driver does, test-select passes.

svenvh:
@alyssarosenzweig thanks for reporting this and doing the initial analysis!  Definitely a CTS bug, I'll create a PR shortly.

PR #1462, opened 2022-08-04
Fix test_select verification failure reporting

svenvh:
When verification of the computed result fails, the test would still
report as "passed".  This is because `s_test_fail` is only written to
and never read.

Fix the immediate issue by returning a failure value and incrementing
`gFailCount` if any error was detected.  The error handling can be
improved further, but I'm leaving that out of the scope of this fix.

Fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/1445


StuartDBrady (approved):
Looks good to me.

PR #1464, opened 2022-08-10
Fix vulkan test build issue on Intel compiler

gwawiork:
Thich change allow build tests on Windows and Intel compiler:

error : namespace "std" has no member "max"

### DIFF
--- a/test_conformance/vulkan/vulkan_interop_common/vulkan_list_map.cpp
+++ b/test_conformance/vulkan/vulkan_interop_common/vulkan_list_map.cpp
@@ -14,9 +14,8 @@
 // limitations under the License.
 //
 
-#ifdef _WIN32
-#define NOMINMAX
-#endif
+
+#include <algorithm>
 #include "vulkan_list_map.hpp"
 #include "vulkan_utility.hpp"
 #include "vulkan_wrapper.hpp"
--- a/test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
+++ b/test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
@@ -15,10 +15,10 @@
 //
 
 #ifdef _WIN32
-#define NOMINMAX
 #include <Windows.h>
 #include <dxgi1_2.h>
 #include <aclapi.h>
+#include <algorithm>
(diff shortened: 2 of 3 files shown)


test_conformance/vulkan/vulkan_interop_common/vulkan_list_map.cpp
```diff
@@ -16,6 +16,7 @@
 
 #ifdef _WIN32
 #define NOMINMAX
+#include <algorithm>
```
bashbaug, line 19: Can this be moved out of the `_WIN32` block?  It shouldn't be Windows-specific.

Note also that we already define `NOMINMAX` globally via CMake options, so we shouldn't need to define it explicitly here.  This means that the entire `_WIN32` block can be removed, and we can just add the include for `<algorithm>`.

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
```diff
@@ -19,6 +19,7 @@
 #include <Windows.h>
 #include <dxgi1_2.h>
 #include <aclapi.h>
+#include <algorithm>
```
bashbaug, line 22: Same comment here - I think this should be moved out of the `_WIN32` block, and we don't need the explicit define for `NOMINMAX`.

test_conformance/vulkan/vulkan_interop_common/vulkan_list_map.cpp
gwawiork: fixed

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
gwawiork: fixed

bashbaug (commented):
LGTM, needs a non-Intel reviewer.

lakshmih:
We are okay with this change.

Sreelakshmi Haridas Maruthur
Qualcomm

lakshmih (approved):

bashbaug:
Has two reviews including one from a different member company so merging.

PR #1466, opened 2022-08-11
[NFC] Fix missing `double_double.lo` initializer

svenvh:
Fixes a missing-field-initializers warning.  The original intent was
most likely to initialize both fields (similar to other functions in
this file), but a `,` was missed.

### DIFF
--- a/test_conformance/math_brute_force/reference_math.cpp
+++ b/test_conformance/math_brute_force/reference_math.cpp
@@ -2321,7 +2321,7 @@ static inline double_double accum_d(double_double a, double b)
 
 static inline double_double add_dd(double_double a, double_double b)
 {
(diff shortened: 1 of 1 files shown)


StuartDBrady:
Does this suppress compilation warnings?  The second vector component seems to be unused, but it seems cleaner to initialize it to -0.0 as per the other functions in this source, and there's certainly no harm in doing so.  If it suppresses warnings, it might be good to mention that in the commit message.  If it's purely for consistency without affecting warnings, that might also be worth mentioning.  Either way, it might be good to add [NFC] in the commit message to make it clearer that this makes no functional change.

StuartDBrady (approved):
Looks good to me.

EwanC (approved):

PR #1467, opened 2022-08-17
[NFC] Fix sign-compare warnings in math_brute_force

svenvh:
This will enable compiling math_brute_force with `-Wall`, for which I'll create a followup patch.

### DIFF
--- a/test_conformance/math_brute_force/main.cpp
+++ b/test_conformance/math_brute_force/main.cpp
@@ -129,8 +129,9 @@ static int doTest(const char *name)
         const Func *const temp_func = functionList + i;
         if (strcmp(temp_func->name, name) == 0)
         {
-            if ((gStartTestNumber != -1 && i < gStartTestNumber)
-                || i > gEndTestNumber)
+            if ((gStartTestNumber != -1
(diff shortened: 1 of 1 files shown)


EwanC (approved):

bashbaug (approved):
I'm fine with this change as-is, but if we really wanted to eliminate a lot of this casting would it make sense if functionListCount (currently a size_t) had the same type as gStartTestNumber and gEndTestNumber (currently int32_ts)?  I imagine the start and end test numbers are currently int32_ts so they can have the value -1?

svenvh:
> I imagine the start and end test numbers are currently int32_ts so they can have the value -1?

That was my conclusion too.

If we wish to get rid of the static_casts, I suppose we could use something like `~0u` as special value for `gStartTestNumber` and `gEndTestNumber` and turn those into `size_t`s.  I'd prefer doing that in a separate change though as it isn't a single point fix (and ideally we'd first enable some more warning flags!).

StuartDBrady:
Perhaps templated sign-safe comparison functions would help, more generally - but don't think any such templating is available to us, so we'd have to do it ourselves.

PR #1468, opened 2022-08-19
[NFC] Use Unix-style line endings

svenvh:
Use the same line ending style across all source files.

Also fixes a missing EOL at EOF.


svenvh:
As this patch touches every line in the file, should I clang-format the entire file too or can I get a waiver for the code format check failure?

StuartDBrady:
> As this patch touches every line in the file, should I clang-format the entire file too or can I get a waiver for the code format check failure?

I think a waiver might make sense, assuming the scale of the clang-format changes is not too large (and it doesn't seem to be).

Doing this in two steps would make it more practical for anyone to update any patches against the version with CRLF line endings without having to deal with merge conflicts, so I'd be in favor of this approach.

(It's also harder to review code formatting changes mixed in with line ending changes, so I don't think it hurts for these to be done separately.)

StuartDBrady (approved):

StuartDBrady:
I have merged this in spite of the clang-format errors, so as to simplify any rebasing that might be needed.  (I.e. people can rebase on top of this change as an intermediate stage, rather than dealing with line-ending and formatting changes mixed together.)

PR #1470, opened 2022-08-22
Add Python 3 support to run_conformance.py

victzhan:


StuartDBrady:
Could you update run_conformance.py so that it works with both versions of Python, instead, please?

StuartDBrady:
It might also be good to update the existing run_conformance.py to follow PEP-8 before making any further changes.

zhangv16:
Please review the modification, thanks.

test_conformance/run_conformance.py
```diff
@@ -1,302 +1,341 @@
...
+        (test_name_, test_dir_) = t
+        if test_dir_ != previous_test:
+            print("==========   " + test_dir_)
+            log_file.write("========================================================================================\n")
+            log_file.write("========================================================================================\n")
+            log_file.write("(" + get_time() + ")     Running Tests: " + test_dir_ + "\n")
+            log_file.write("========================================================================================\n")
+            log_file.write("========================================================================================\n")
+            previous_test = test_dir_
+        print("(" + get_time() + ")     BEGIN  " + test_name_.ljust(40) + ": "),
```
StuartDBrady, line 249: I notice there's a trailing `,` here which won't behave in Python 3 (or in Python 2 using the print function from futures) in the way that is intended.

I have a fix for this in PR #1492, which I would like for us to apply first before we finish this review.  Please ensure the fix isn't undone in this change after we have closed PR #1492.

test_conformance/run_conformance.py
```diff
@@ -1,302 +1,341 @@
...
+                sys.exit(-1)
+            else:
+                write_screen_log("\nUser chose to continue with other tests. Reporting this test as failed.")
+                result = 1
+        run_time = (time.time() - start_time)
+
+        # Move print the finish status
+        if result == 0:
+            print("(" + get_time() + ")     PASSED " + test_name_.ljust(40) + ": (" + str(int(run_time)).rjust(
+                3) + "s, test " + str(test_number).rjust(3) + os.sep + str(len(tests_)) + ")"),
```
StuartDBrady, line 284: Here, also, there is a trailing `,` which needs to be substituted with `end=''`, addressed in PR #1492.

test_conformance/run_conformance.py
```diff
@@ -1,302 +1,341 @@
...
+        else:
+            print("(" + get_time() + ")     FAILED " + test_name_.ljust(40) + ": (" + str(int(run_time)).rjust(
+                3) + "s, test " + str(test_number).rjust(3) + os.sep + str(len(tests_)) + ")"),
```
StuartDBrady, line 287: Here, also, there is a trailing `,`, addressed in PR #1492.

StuartDBrady (changes requested):
Thanks for your changes.

Unfortunately, it's hard to make sense of the Python 3 changes with the formatting changes mixed together in the same PR.  For this reason, I've committed whitespace changes that address PEP-8 issues separately in PR #1491, and have submitted a small number of other cleanups in PR #1492.

After we have closed PR #1492, could you then please force-push a new commit to your _add-run-conformance-python3_ branch, with the run_conformance.py script updated in PR #1491 deleted and your version copied in its place?  I will comment here to let you know when we are ready.

This way, it will be more practical to review the remaining changes, and once we accept this PR, it will be possible to see the specific changes for Python 3 support in the log of Git commits, if we ever need.

Please also update the commit message and PR description so that they no longer mention adding a new script but instead mention that run_conformance.py is updated to be compatible with both Python 2 and Python 3.

StuartDBrady:
> After we have closed PR #1492, could you then please force-push a new commit to your _add-run-conformance-python3_ branch, with the run_conformance.py script updated in PR #1491 deleted and your version copied in its place? I will comment here to let you know when we are ready.

PR #1492 has now been merged to the main branch.  Please can you check out the main branch again, apply your changes (without any formatting changes) as a new commit, and force-push to your add-run-conformance-python3 branch?

Please can you also update the commit message as described?

Many thanks!

victzhan:
Modification is done, please review. Thanks.

test_conformance/run_conformance.py
```diff
@@ -18,6 +18,9 @@
 import tempfile
 import string
 
+if sys.version_info[0] < 3:
+    import commands
```
StuartDBrady, line 22: Why is `commands` being imported here for Python 2?  I cannot see any use of it.

test_conformance/run_conformance.py
```diff
@@ -63,19 +66,33 @@ def get_tests(filename, devices_to_test):
         comment = re.search("^#.*", line)
         if comment:
             continue
-        device_specific_match = re.search("^\s*(.+?)\s*,\s*(.+?)\s*,\s*(.+?)\s*$", line)
+        if sys.version_info[0] < 3:
+            device_specific_match = re.search("^\s*(.+?)\s*,\s*(.+?)\s*,\s*(.+?)\s*$", line)
+        else:
+            device_specific_match = re.search("^\s*(.+)\s*,\s*(.+)\s*,\s*(.+)\s*", line)
```
StuartDBrady, line 72: Could you explain this change, please?  Is the "$" to match the end of line not important?  Why would it be needed with Python 2 but not with Python 3?

Why the change from the non-greedy `.+?` to the greedy `.+`?  I have tested the non-greedy `.+?` just now and it works fine with Python 3.6.9 and it works fine in my limited test case.  Why does it need changing here?  What sort of input text would "line" contain here, where this would make a difference?

test_conformance/run_conformance.py
```diff
@@ -63,19 +66,33 @@ def get_tests(filename, devices_to_test):
...
         if device_specific_match:
             if device_specific_match.group(1) in devices_to_test:
-                test_path = string.replace(device_specific_match.group(3), '/', os.sep)
-                test_name = string.replace(device_specific_match.group(2), '/', os.sep)
+                if sys.version_info[0] < 3:
+                    test_path = string.replace(device_specific_match.group(3), '/', os.sep)
+                    test_name = string.replace(device_specific_match.group(2), '/', os.sep)
```
StuartDBrady, line 77: str.replace seems to work fine with Python 2.  Is there a reason not to change this for Python 2 as well, to use str.replace instead of string.replace?  (If there is not, then the `sys.version_info[0] < 3` check can be removed.)

test_conformance/run_conformance.py
```diff
@@ -63,19 +66,33 @@ def get_tests(filename, devices_to_test):
...
+                else:
+                    test_path = str.replace(device_specific_match.group(3), '/', os.sep)
+                    test_name = str.replace(device_specific_match.group(2), '/', os.sep)
                 tests.append((test_name, test_path))
             else:
                 print("Skipping " + device_specific_match.group(2) + " because " + device_specific_match.group(1) + " is not in the list of devices to test.")
             continue
-        match = re.search("^\s*(.+?)\s*,\s*(.+?)\s*$", line)
+        if sys.version_info[0] < 3:
+            match = re.search("^\s*(.+?)\s*,\s*(.+?)\s*$", line)
```
StuartDBrady, line 86: As above, I'm confused by this regex change.  Could you explain it please?

test_conformance/run_conformance.py
```diff
@@ -63,19 +66,33 @@ def get_tests(filename, devices_to_test):
...
+        else:
+            match = re.search("^\s*(.+)\s*,\s*(.+)\s*", line)
         if match:
-            test_path = string.replace(match.group(2), '/', os.sep)
-            test_name = string.replace(match.group(1), '/', os.sep)
+            if sys.version_info[0] < 3:
+                test_path = string.replace(match.group(2), '/', os.sep)
+                test_name = string.replace(match.group(1), '/', os.sep)
```
StuartDBrady, line 92: As above, would str.replace work here?

StuartDBrady (changes requested):

test_conformance/run_conformance.py
victzhan: The original script imported commands, although it seems unnecessary. I just imported it to maintain what was like. I believe this can be removed.

test_conformance/run_conformance.py
victzhan: I'm not the developer that made this change and it's impossible to track who made the changes now. but seems that if I didn't change the regex, it would run into an error. See the following (the line numbers are not accurate as it's in a older version of this file.):

oem@payam-linux-nv32:~/nv32/conformance/2.0$ python3 run_conformance_kr.py svm.csv CL_DEVICE_TYPE_GPU
Testing on: ['CL_DEVICE_TYPE_GPU']
Traceback (most recent call last):
  File "run_conformance_kr.py", line 336, in <module>
    tests = get_tests(sys.argv[1], devices_to_test)
  File "run_conformance_kr.py", line 87, in get_tests
    test_path = str.replace(device_specific_match.group(3), '/', os.sep)
AttributeError: 'NoneType' object has no attribute 'group'
oem@payam-linux-nv32:~/nv32/conformance/2.0$

test_conformance/run_conformance.py
victzhan: Again, I did not make the changes, if it works fine then it should be okay to remove it. I think str replace and string.replace is analogous in python 2.

test_conformance/run_conformance.py
victzhan: See above.

test_conformance/run_conformance.py
victzhan: See above.

test_conformance/run_conformance.py
StuartDBrady: If it's not needed, please leave it removed.

test_conformance/run_conformance.py
StuartDBrady: Please use only `str.replace`, if it works with both Python 2 and Python 3.

test_conformance/run_conformance.py
StuartDBrady: Please use only `str.replace`, if it works with both Python 2 and Python 3.

test_conformance/run_conformance.py
StuartDBrady: Would you be able to share what the content of `line` was, here?

victzhan: [svm.csv](https://github.com/KhronosGroup/OpenCL-CTS/files/9609667/svm.csv)

This should be it, as line is read from the csv file.

StuartDBrady (changes requested):

test_conformance/run_conformance.py
StuartDBrady: Now done.

test_conformance/run_conformance.py
StuartDBrady: Now done.

test_conformance/run_conformance.py
StuartDBrady: Now done.

test_conformance/run_conformance.py
victzhan: Might be the variable device_specific_match that causes the error in python 3 if we strip the whitespace. I double checked python documentation that every character in this search is exactly the same for python 2 and python 3. Will look into that.

victzhan: You are right - I didn't see why we need to change the regex search, thus I removed those lines. In addition I removed import string as we didn't use it. Please see my new commit. Thanks.

StuartDBrady: Thanks for looking into this!

test_conformance/run_conformance.py
StuartDBrady: Regex changes removed.

StuartDBrady (approved):

StuartDBrady:
Merged as there has been no objection to any of the remaining changes since this PR was raised over a month ago.

Many thanks for this!

PR #1477, opened 2022-08-25
[NFC] Enable -Wall for math_brute_force

svenvh:
math_brute_force compiles cleanly with `-Wall` currently, so avoid
regressing from that state.  Ideally we would enable `-Wall` in the
top-level CMakeLists.txt, but other tests do not compile cleanly with
`-Wall` yet.

### DIFF
--- a/test_conformance/math_brute_force/CMakeLists.txt
+++ b/test_conformance/math_brute_force/CMakeLists.txt
@@ -40,4 +40,14 @@ set(${MODULE_NAME}_SOURCES
     utility.h
 )
 
+# math_brute_force compiles cleanly with -Wall but other tests not (yet), so
+# enable -Wall locally.
+if(CMAKE_COMPILER_IS_GNUCC OR "${CMAKE_CXX_COMPILER_ID}" MATCHES "(Apple)?Clang")
+    SET_SOURCE_FILES_PROPERTIES(
+        ${${MODULE_NAME}_SOURCES}
+        PROPERTIES
+        COMPILE_FLAGS "-Wall -Wno-format -Wno-strict-aliasing -Wno-unknown-pragmas"
+    )
+endif()
+
 include(../CMakeCommon.txt)


StuartDBrady (approved):
Looks good to me.

test_conformance/math_brute_force/CMakeLists.txt
```diff
@@ -40,4 +40,14 @@ set(${MODULE_NAME}_SOURCES
     utility.h
 )
 
+# math_brute_force compiles cleanly with -Wall but other tests not (yet), so
+# enable -Wall locally.
+if(CMAKE_COMPILER_IS_GNUCC)
```
EwanC, line 45: In other places in CMake we have `if(CMAKE_COMPILER_IS_GNUCC OR "${CMAKE_CXX_COMPILER_ID}" MATCHES "(Apple)?Clang")`. Is there a reason you've left out clang here? Are there still -Wall warnings in brute force with clang?

svenvh: Good point; it makes sense to be consistent with other places so I'll fix that.

StuartDBrady: Would it be possible for us to factor this logic out, either now or at some point in the future?  I can imagine it getting repeated in quite a few places in the medium term.

svenvh: ... except that the code base doesn't seem ready for it yet!  I'll see if I can fix the new issues.

EwanC: > ... except that the code base doesn't seem ready for it yet! I'll see if I can fix the new issues.

If sorting out any new issues revealed by CI clang jobs is a bunch more work, I'm not opposed to merging this with the GCC only condition and then tackling the clang fails and refactored compiler check in a follow-up MR

svenvh: > Would it be possible for us to factor this logic out

Certainly, I can do that in a followup PR.

> If sorting out any new issues revealed by CI clang jobs is a bunch more work, I'm not opposed to merging this with the GCC only condition and then tackling the clang fails and refactored compiler check in a follow-up MR

It seems some builds behave slightly differently as `-Wall` re-enables `-Wformat` again (which is disabled in the top-level `CMakeLists.txt`).  For now I'll try including clang too, but (re-)disable any warnings that show up.  We can then address those warnings in future PRs.

StuartDBrady: Sounds reasonable to me.

StuartDBrady (approved):
Looks good to me.

EwanC (approved):

PR #1480, opened 2022-08-31
Update known extensions in compiler define test

EwanC:
Add [cl_khr_command_buffer_mutable_dispatch](https://github.com/KhronosGroup/OpenCL-Docs/pull/819) and [cl_khr_extended_async_copies](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_extended_async_copies) to the list of known extensions used in `test_compiler_defines_for_extensions`

### DIFF
--- a/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
+++ b/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
@@ -20,7 +20,7 @@
 #include <unistd.h>
 #endif
 
-
+// List should follow order in the extension spec
 const char *known_extensions[] = {
     "cl_khr_byte_addressable_store",
     "cl_khr_3d_image_writes",
@@ -42,6 +42,7 @@ const char *known_extensions[] = {
     "cl_khr_mipmap_image_writes",
     "cl_khr_srgb_image_writes",
     "cl_khr_subgroup_named_barrier",
+    "cl_khr_extended_async_copies",
     "cl_khr_subgroup_extended_types",
     "cl_khr_subgroup_non_uniform_vote",
     "cl_khr_subgroup_ballot",
@@ -51,6 +52,7 @@ const char *known_extensions[] = {
     "cl_khr_subgroup_clustered_reduce",
     "cl_khr_extended_bit_ops",
     "cl_khr_integer_dot_product",
+    "cl_khr_subgroup_rotate",
     // API-only extensions after this point.  If you add above here, modify
     // first_API_extension below.
     "cl_khr_icd",
@@ -82,10 +84,11 @@ const char *known_extensions[] = {
     "cl_khr_command_buffer",
     "cl_khr_external_memory",
     "cl_khr_external_memory_opaque_fd",
+    "cl_khr_command_buffer_mutable_dispatch",
 };
 
 size_t num_known_extensions = ARRAY_SIZE(known_extensions);
-size_t first_API_extension = 29;
+size_t first_API_extension = 31;
 
 const char *known_embedded_extensions[] = {
     "cles_khr_int64",


test_conformance/compiler/test_compiler_defines_for_extensions.cpp
```diff
@@ -80,8 +80,10 @@ const char *known_extensions[] = {
     "cl_khr_external_semaphore_opaque_fd",
     "cl_khr_external_semaphore_sync_fd",
     "cl_khr_command_buffer",
+    "cl_khr_command_buffer_mutable_dispatch",
     "cl_khr_external_memory",
     "cl_khr_external_memory_opaque_fd",
+    "cl_khr_extended_async_copies",
```
bashbaug, line 86: I think [cl_khr_extended_async_copies](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_extended_async_copies) shouldn't be considered an API-only extension.

bashbaug (commented):
Mostly LGTM but I have one question.

If you're feeling up to it you could add [cl_khr_subgroup_rotate](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_subgroup_rotate) also.

test_conformance/compiler/test_compiler_defines_for_extensions.cpp
EwanC: ahh, I missed the comment on line 54 differentiating the types of extensions. Will update

test_conformance/compiler/test_compiler_defines_for_extensions.cpp
```diff
@@ -49,8 +49,10 @@ const char *known_extensions[] = {
     "cl_khr_subgroup_shuffle",
     "cl_khr_subgroup_shuffle_relative",
     "cl_khr_subgroup_clustered_reduce",
+    "cl_khr_subgroup_rotate"
```
svenvh, line 52: Nitpick: I believe this list follows the order of the [extension spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html). cl_khr_subgroup_rotate is in Section 48, so I would put it after cl_khr_integer_dot_product (in Section 42).

test_conformance/compiler/test_compiler_defines_for_extensions.cpp
```diff
@@ -49,8 +49,10 @@ const char *known_extensions[] = {
...
     "cl_khr_extended_bit_ops",
     "cl_khr_integer_dot_product",
+    "cl_khr_extended_async_copies",
```
svenvh, line 55: Likewise, cl_khr_extended_async_copies is in Section 34 so I guess it could come just after cl_khr_subgroup_named_barrier (Section 33).

test_conformance/compiler/test_compiler_defines_for_extensions.cpp
EwanC: that ordering makes sense, updated and left a comment above the array about this.

svenvh: Thanks, good to have a comment indeed.

svenvh (approved):

test_conformance/compiler/test_compiler_defines_for_extensions.cpp
```diff
@@ -42,6 +42,7 @@ const char *known_extensions[] = {
     "cl_khr_mipmap_image_writes",
     "cl_khr_srgb_image_writes",
     "cl_khr_subgroup_named_barrier",
+    "cl_khr_extended_async_copies",
```
StuartDBrady, line 45: Should cl_khr_extended_async_copies have an extension macro defined in OpenCL C? I.e. should it be in the this uppermost "language extensions" part of the known_extensions array?  It's not clear to me from the extension spec whether or not it should.

StuartDBrady: I've since been made aware that in the OpenCL Extension Specification § 1.2. Compiler Directives for Optional Extensions, the following is stated:

> Every extension which affects the OpenCL language semantics, syntax or adds built-in functions to the language must create a preprocessor `#define` that matches the extension name string. This `#define` would be available in the language if and only if the extension is supported on a given implementation.

There are three extensions for which the OpenCL Extension Specification explicitly mentions the preprocessor macro that needs to be defined, which may give the false impression that these extensions are special in that regard.  There is no such mention made for cl_khr_extended_async_copies.

bashbaug: Right, I think that cl_khr_extended_async_copies should have an OpenCL C preprocessor define because it adds new built-in functions, so it's correct to include it in this section.

I don't have a very strong opinion whether we list the OpenCL C preprocessor define for all language extensions or whether the general statement in section 1.2 is sufficient, though I do agree listing it explicitly for some extensions but not others could be a cause for confusion.  Feel free to file an OpenCL-Docs issue if you'd like and we can discuss further.

bashbaug (approved):

PR #1482, opened 2022-09-05
External sharing new updates

nikhiljnv:
Updates to external sharing tests:
1. Use SPIR-V shaders as a standard way to use Vulkan shaders instead of GLSL to Vulkan shaders.
2. Replace clReleaseSemaphoreObjectKHR with clReleaseSemaphoreKHR
3. Minor fixes to take care of low memory usages for lower end GPUs
4. Additional checks for Vulkan support before continuing on test execution or skip the test otherwise.


bcalidas:
Qualcomm is building and testing with this PR. We have some edits that we would like to make.  Would it be possible to squash the commits on nikhiljv-external-sharing-new-updates and bring it over to main as a single commit.

Thanks,
Balaji Calidas

bcalidas:
We had to make some changes to the external sharing tests to get them to build and we are still debugging some failures. However, the edits are independent of this PR  We are ok with this PR being merged.

Thanks,
Balaji Calidas,
Qualcomm

bashbaug (approved):
There are additional changes we could make but this is a nice improvement we should merge to unblock progress.

bashbaug:

PR #1483, opened 2022-09-05
[NFCI] Remove unused variables and enable -Wunused-variable

svenvh:
Remove unused variables throughout the code base and enable the
`-Wunused-variable` warning flag globally to prevent new unused
variable issues being introduced in the future.

This is mostly a non-functional change, with one exception:

 - In `test_conformance/api/test_kernel_arg_info.cpp`, an error check
   of the clGetDeviceInfo return value was added.

Signed-off-by: Sven van Haastregt <sven.vanhaastregt@arm.com>

PR #1486, opened 2022-09-06
Include release builds in GitHub Actions

StuartDBrady:
The "Ninja" CMake generator does not support multiple configurations, i.e. it does not support use of the '--config' option when running 'cmake --build'.  As such, the default configuration (i.e. Debug) was getting used for all builds.
    
Use the CMAKE_BUILD_TYPE variable instead, so that we do release builds, but change one build (ubuntu-20.04 aarch64) to use Debug as its build type, to keep some build coverage for asserts, etc.
    
For Vulkan-Loader and OpenCL-ICD-Loader, we do release builds unconditionally, as we assume there is no need in the CI workflow to actually run the binaries that are built, and therefore no need for any additional debug info.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -3,11 +3,12 @@ on: [push, pull_request]
 
 jobs:
   build:
-    name: Build ${{ matrix.os }} ${{ matrix.arch }}
+    name: Build ${{ matrix.os }} ${{ matrix.arch }}${{ matrix.extra }}
     runs-on: ${{ matrix.os }}
     env:
       JOB_ARCHITECTURE: ${{ matrix.arch }}
       JOB_ENABLE_GL: ${{ matrix.gl }}
+      JOB_ENABLE_DEBUG: ${{ matrix.debug }}
     strategy:
       fail-fast: false
       matrix:
@@ -17,12 +18,15 @@ jobs:
           - os: ubuntu-20.04
             mainmatrix: true
             gl: 1
+            extra: " gl"
           - os: ubuntu-20.04
             mainmatrix: false
             arch: arm
           - os: ubuntu-20.04
             mainmatrix: false
             arch: aarch64
+            debug: 1
+            extra: " debug"
     steps:
       - uses: actions/checkout@v2
       - name: Setup Ninja
--- a/presubmit.sh
+++ b/presubmit.sh
@@ -14,6 +14,9 @@ TOOLCHAIN_FILE=${TOP}/toolchain.cmake
 touch ${TOOLCHAIN_FILE}
 BUILD_OPENGL_TEST="OFF"
 
+cmake --version
+echo
+
 # Prepare toolchain if needed
 if [[ ${JOB_ARCHITECTURE} != "" && ${RUNNER_OS} != "Windows" ]]; then
     TOOLCHAIN_URL_VAR=TOOLCHAIN_URL_${JOB_ARCHITECTURE}
@@ -40,6 +43,12 @@ if [[ ( ${JOB_ARCHITECTURE} == "" && ${JOB_ENABLE_GL} == "1" ) ]]; then
     BUILD_OPENGL_TEST="ON"
 fi
 
+if [[ ${JOB_ENABLE_DEBUG} == 1 ]]; then
+    BUILD_CONFIG="Debug"
+else
+    BUILD_CONFIG="Release"
+fi
+
 #Vulkan Headers
 git clone https://github.com/KhronosGroup/Vulkan-Headers.git
 
@@ -48,8 +57,11 @@ git clone https://github.com/KhronosGroup/OpenCL-ICD-Loader.git
 cd ${TOP}/OpenCL-ICD-Loader
 mkdir build
 cd build
-cmake .. -G Ninja -DCMAKE_TOOLCHAIN_FILE=${TOOLCHAIN_FILE} -DOPENCL_ICD_LOADER_HEADERS_DIR=${TOP}/OpenCL-Headers/
-cmake --build . -j2 --config Release
+cmake .. -G Ninja \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_TOOLCHAIN_FILE=${TOOLCHAIN_FILE} \
+      -DOPENCL_ICD_LOADER_HEADERS_DIR=${TOP}/OpenCL-Headers/
+cmake --build . -j2
 
 #Vulkan Loader
 cd ${TOP}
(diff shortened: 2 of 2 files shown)


presubmit.sh
```diff
@@ -40,6 +40,12 @@ if [[ ( ${JOB_ARCHITECTURE} == "" && ${JOB_ENABLE_GL} == "1" ) ]]; then
     BUILD_OPENGL_TEST="ON"
 fi
 
+if [[ JOB_ENABLE_DEBUG == 1 ]]; then
```
kpet, line 43: Missing $?

StuartDBrady: Thanks, fixed.

.github/workflows/presubmit.yml
```diff
@@ -23,6 +24,7 @@ jobs:
           - os: ubuntu-20.04
             mainmatrix: false
             arch: aarch64
+            debug: 1
```
MeyeChris, line 27: What does this do?

I'm not conversant with the CI but how does this change stay scoped to just the Debug build?

StuartDBrady: I'm not sure I really understand the question, but I will try my best to explain.

There are five builds specified here:
* ubuntu-20.04 (x86-64)
* ubuntu-20.04 arm
* ubuntu-20.04 aarch64
* macos-latest
* windows-latest

The "include" section is what causes the CTS to be built with additional architectures for ubuntu-20.04.

"debug: 1" only applies to ubuntu-20.04 with the arch being "aarch64".

The "ubuntu-20.04 arm" and "ubuntu-20.04" (x86-64) builds do not contain "debug: 1" in their configuration.  In these cases, JOB_ENABLE_DEBUG will not be "1".  (We could perhaps add a default of 0, but there is no strict need, and the same would apply for gl - so it seems best to keep them consistent for now.)

See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-expanding-configurations for documentation regarding workflows in GitHub Actions.

MeyeChris: Thanks you, completely clear now - I should have looked more closely and rtfm.

So in summary we only do a debug build on ubuntu-20.04-aarch64.

This does mean - if I understand correctly - that we have not just added a debug build option, but changed the ubuntu-20.04-aarch64 from release to debug?

Sorry to be a pedant, and I can be :-), but should the commit message not say something along Add debug build option and change ubuntu-20.04-aarch64 from Release to Debug?

presubmit.sh
```diff
@@ -84,6 +90,6 @@ cmake .. -G Ninja \
       -DGL_IS_SUPPORTED=${BUILD_OPENGL_TEST} \
       -DVULKAN_INCLUDE_DIR=${TOP}/Vulkan-Headers/include/ \
       -DVULKAN_LIB_DIR=${TOP}/Vulkan-Loader/build/loader/
-cmake --build . -j3 --config Release
+cmake --build . -j3 --config "${BUILD_CONFIG}"
```
svenvh, line 93: I'm not sure if this does what's intended.  Before this change, we'd expect all builds to be release builds, but [this comment](https://github.com/KhronosGroup/OpenCL-CTS/pull/1483#issuecomment-1241886877) makes me think they're actually all debug builds?

As far as I know, the build type is typically set in the configuration step using `-D CMAKE_BUILD_TYPE=Release`.

StuartDBrady: I think you might be right.  It turns out that the "Ninja" generator _does not support_ multiple configurations using `--config` (as listed in CMAKE_CONFIGURATION_TYPES, and defaulting to CMAKE_DEFAULT_BUILD_TYPE if set, or otherwise the first configuration type in the list).  There is a special "Ninja Multi-Config" generator that supports multiple configurations, but it was added in CMake 3.17.  The version of CMake included in Ubuntu 20.04 is 3.16, so it is too old to use "Ninja Multi-Config".

It's probably best if we use `CMAKE_BUILD_TYPE=Release`.

.github/workflows/presubmit.yml
StuartDBrady: That would have been the case, except we've discovered that we have _only_ doing debug builds so far, and that we should include release builds.

"Add debug build option and change ubuntu-20.04-aarch64 from Release to Debug?" might have been a little on the verbose side.  I will go with "Include release builds in GitHub Actions" as the commit title and elaborate below, unless there's a more accurate title I could use that is also concise.

svenvh:
The list of checks for this PR contains the following entry (which looks different from the other Presubmit entries):
```
Build ubuntu-20.04 Expected — Waiting for status to be reported 
```
Not sure if there is something in this change that's causing this stale(?) entry?

StuartDBrady:
> The list of checks for this PR contains the following entry (which looks different from the other Presubmit entries):
> 
> ```
> Build ubuntu-20.04 Expected — Waiting for status to be reported 
> ```
> 
> Not sure if there is something in this change that's causing this stale(?) entry?

Yes, the GitHub config for the KhronosGroup/OpenCL-CTS repo waits for a number of builds to finish, one of which being "ubuntu-20.04".  However, I changed this to "ubuntu-20.04 gl" at the same time as using the name "ubuntu-20.04 aarch64 debug".  Once we merge this change, I can update the GitHub config after this change has been merged.  Alternatively, I could remove the part of the change that sets " gl" as the extra info for the default-arch "ubuntu-20.04" build.

.github/workflows/presubmit.yml
```diff
@@ -3,11 +3,12 @@ on: [push, pull_request]
 
 jobs:
   build:
-    name: Build ${{ matrix.os }} ${{ matrix.arch }}
+    name: Build ${{ matrix.os }} ${{ matrix.arch }}${{ matrix.extra }}
```
svenvh, line 6: What's the reason for not doing
```suggestion
    name: Build ${{ matrix.os }} ${{ matrix.arch }} ${{ matrix.extra }}
```
and avoiding the leading space in the `extra:` specifications below?  I fear the leading space might be easily overlooked/forgotten.

svenvh:
> Yes, the GitHub config for the KhronosGroup/OpenCL-CTS repo waits for a number of builds to finish, one of which being "ubuntu-20.04". However, I changed this to "ubuntu-20.04 gl" at the same time as using the name "ubuntu-20.04 aarch64 debug". Once we merge this change, I can update the GitHub config after this change has been merged.

Okay, if this is expected then that sounds okay to me.

.github/workflows/presubmit.yml
StuartDBrady: To avoid including a trailing space in the name where there matrix.extra is unset.

svenvh:
LGTM, but would be good to have approval from someone else.

kpet (approved):
LGTM!

@StuartDBrady If you want to keep working on the CI, you might want to consider continuing the work I started to get rid of `presubmit.sh`.

PR #1465, opened 2022-08-10
Fix build issue of 32 bitness binaries in vulcan test when using Intel compiler.

gwawiork:
This fix solves error messages:

>C:\Repos\_CTS_KHR\OpenCL-CTS\test_conformance\vulkan\test_vulkan_interop_buffer.cpp(787,44): : error : use of overloaded operator '[]' is ambiguous (with operand types 'VulkanBufferList' and 'int')
2>                    (uint32_t)(vkBufferList[0].getSize());
2>                               ~~~~~~~~~~~~^~
2>C:\Repos\_CTS_KHR\OpenCL-CTS\test_conformance\vulkan\.\vulkan_interop_common/vulkan_list_map.hpp(39,28): note: candidate function
2>    virtual VulkanWrapper &operator[](size_t idx);
2>                           ^
2>C:\Repos\_CTS_KHR\OpenCL-CTS\test_conformance\vulkan\.\vulkan_interop_common/vulkan_list_map.hpp(38,34): note: candidate function
2>    virtual const VulkanWrapper &operator[](size_t idx) const;
C:\Repos\_CTS_KHR\OpenCL-CTS\test_conformance\vulkan\test_vulkan_interop_buffer.cpp(787,44): note: built-in candidate operator[](const unsigned long long *, int)
2>                    (uint32_t)(vkBufferList[0].getSize());
2>                                           ^
2>C:\Repos\_CTS_KHR\OpenCL-CTS\test_conformance\vulkan\test_vulkan_interop_buffer.cpp(787,44): note: built-in candidate operator[](const volatile unsigned long long *, int)
2>C:\Repos\_CTS_KHR\OpenCL-CTS\test_conformance\vulkan\test_vulkan_interop_buffer.cpp(828,28): : warning : unused variable 'buffer_size' [-Wunused-variable]
2>                    size_t buffer_size = vkBufferList[bIdx].getSize();
2>                           ^
2>C:\Repos\_CTS_KHR\OpenCL-CTS\test_conformance\vulkan\test_vulkan_interop_buffer.cpp(1129,44): : error : use of overloaded operator '[]' is ambiguous (with operand types 'VulkanBufferList' and 'int')
2>                    (uint32_t)(vkBufferList[0].getSize());
2>                               ~~~~~~~~~~~~^~
2>C:\Repos\_CTS_KHR\OpenCL-CTS\test_conformance\vulkan\.\vulkan_interop_common/vulkan_list_map.hpp(39,28): note: candidate function
[...]

PR #1488, opened 2022-09-06
remove implicit conversion to pointer to fix 32-bit compile

bashbaug:
This is an alternative proposal to fix 32-bit compiles, see #1465.

This version removes the implicit conversion to a pointer and relies on an explicit conversion instead, by overloading `operator()`.  Note that this is a similar pattern used by the OpenCL C++ bindings:

https://github.com/KhronosGroup/OpenCL-CLHPP/blob/main/include/CL/opencl.hpp#L1913

By removing the implicit conversion there is only one overload for both 32-bit and 64-bit builds.


nikhiljnv (approved):

bashbaug:

PR #1490, opened 2022-09-07
Improve MTdataHolder design and use it in math_brute_force

svenvh:
Improve the design of the MTdataHolder wrapper:

 * Make it a class instead of a struct with a private member, to clarify that there is no direct access to the MTdata member.

 * Make the 1-arg constructor `explicit` to avoid unintended conversions.

 * Forbid copy construction/assignment as MTdataHolder is never initialised from an MTdataHolder object in the codebase.

 * Define move construction/assignment as per the "rule of five".

Use the MTdataHolder class throughout math_brute_force, to simplify code by avoiding manual resource management.

### DIFF
--- a/test_conformance/math_brute_force/binary_double.cpp
+++ b/test_conformance/math_brute_force/binary_double.cpp
@@ -134,7 +134,7 @@ struct ThreadInfo
         maxErrorValue; // position of the max error value (param 1).  Init to 0.
     double maxErrorValue2; // position of the max error value (param 2).  Init
                            // to 0.
-    MTdata d;
+    MTdataHolder d;
 
     // Per thread command queue to improve performance
     clCommandQueueWrapper tQueue;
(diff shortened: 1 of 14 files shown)


test_conformance/math_brute_force/binary_double.cpp
```diff
@@ -691,9 +691,11 @@ int TestFunc_Double_Double_Double(const Func *f, MTdata d, bool relaxedMode)
         test_info.k[i].resize(test_info.threadCount, nullptr);
     }
 
-    test_info.tinfo.resize(test_info.threadCount, ThreadInfo{});
+    test_info.tinfo.reserve(test_info.threadCount);
```
EwanC, line 694: Did you see an advantage to doing `reserve` then the individual `push_back`s rather than a single `test_info.tinfo.resize(test_info.threadCount)`?

EwanC (approved):

test_conformance/math_brute_force/binary_double.cpp
svenvh: The 1-argument overload requires [MoveInsertable](https://en.cppreference.com/w/cpp/named_req/MoveInsertable) and [DefaultInsertable](https://en.cppreference.com/w/cpp/named_req/DefaultInsertable), so yes, that should actually work; thanks for the suggestion!

PR #1493, opened 2022-09-08
[NFC] Declare format tables as const

svenvh:
Without `const`, these variables would be flagged up by `-Wunused-variable`.

Preserve constness across functions that use these variables.

### DIFF
--- a/test_conformance/gl/common.h
+++ b/test_conformance/gl/common.h
@@ -32,12 +32,8 @@ struct format {
 };
 
 // These are the typically tested formats.
-// TODO: These variables should be made const; until then, suppress unused
-// variable warnings as not every translation unit including this header uses
-// all variables.
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunused-variable"
-static struct format common_formats[] = {
+// clang-format off
+static const format common_formats[] = {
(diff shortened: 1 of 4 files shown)


test_conformance/gl/common.h
```diff
@@ -32,12 +32,8 @@ struct format {
...
 
 // These are the typically tested formats.
-// TODO: These variables should be made const; until then, suppress unused
-// variable warnings as not every translation unit including this header uses
-// all variables.
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wunused-variable"
-static struct format common_formats[] = {
+// clang-format off
+static const struct format common_formats[] = {
```
EwanC, line 36: A potential further clean-up if you feel like it could be to drop the `struct` keyword from `const struct format` everywhere in this change and just have `const format`. I think the `struct` keyword in these places should only be needed for a `C` code rather than `C++` .

EwanC (approved):

test_conformance/gl/common.h
svenvh: Good idea, I've pushed a new version with `struct` removed.

EwanC (approved):

PR #1495, opened 2022-09-09
gles: Limit variable definition to the same scope as usage

lakshmih:
Fix unused-variable errors by limiting variable definition to the case that would use it


bashbaug (approved):
We could consider moving this into the same #ifdef also, but this is fine as-is.

Issue #1421, opened 2022-04-28
change allocation size from cl_int to size_t in the remaining SVM tests


bashbaug:
#1381 describes a problem in the coarse grain SVM tests, but the same problem also seems to be in the fine-grain SVM tests also.  Note the mismatch between sizeof(cl_int) and size_t*.

```c
  // this allocation holds an index into the nodes buffer, it is used for node allocation
  size_t* pAllocator = (size_t*) align_malloc(sizeof(cl_int), 128);
```

StuartDBrady:
Fixed by PR #1496.

PR #1496, opened 2022-09-10
Use correct size for memory allocation in SVM test

niranjanjoshi121:
Memory is allocated for cl_int,
but mapped as size_t.
Use size_t instead of cl_int during
allocation and mapping for consistency
Fixes OpenCL-CTS #1421


bashbaug (approved):
Thanks!

EwanC (approved):

PR #1497, opened 2022-09-12
[NFC] Reformat code in events test

StuartDBrady:

### DIFF
--- a/test_conformance/events/action_classes.cpp
+++ b/test_conformance/events/action_classes.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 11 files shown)


gwawiork (approved):
LGTM

EwanC (dismissed):

test_conformance/events/action_classes.cpp
```diff
@@ -101,25 +111,25 @@ cl_int    Action::IGetPreferredImageSize3D( cl_device_id device, size_t &outWidt
...
+    log_info("\tImage size: %d x %d x %d (%gMB)\n", (int)outWidth,
+             (int)outHeight, (int)outDepth,
+             (double)((int)outWidth * (int)outHeight * (int)outDepth * 4)
+                 / (1024.0 * 1024.0));
 
     return CL_SUCCESS;
 }
 
-#pragma mark -------------------- Execution Sub-Classes -------------------------
+#pragma mark-------------------- Execution Sub-Classes -------------------------
```
svenvh, line 129: This change doesn't look right; occurs in other places in this PR too.

StuartDBrady: Thanks - I have fixed this.

We should probably remove these #pragmas, as they're Xcode-specific and do nothing to improve the organization of the code for those not using these IDEs.  Indeed, I would more generally say that pragmas are not the proper solution for organizing code so that it is easy to search and navigate.

However, it seems best to leave their removal for another change, and leave this PR purely as a reformatting change.

svenvh:
LGTM; presumably this is an NFC commit, could you please tag it as such?

StuartDBrady:
> LGTM; presumably this is an NFC commit, could you please tag it as such?

Done – thanks!

svenvh (approved):

Issue #1022, opened 2020-10-21
test_allocation clamps output of GLOBAL_MEM_SIZE query to SIZE_MAX, but not for MAX_MEM_ALLOC_SIZE


nikhiljnv:
test_allocation clamps output of GLOBAL_MEM_SIZE query to SIZE_MAX, but not for MAX_ALLOC_SIZE.
Moreover, GLOBAL_MEM_SIZE is cl_ulong, but SIZE_MAX depends on platform bitness. So, we are capping anything larger than 4 GB to 4 GB. This leads to MAX_MEM_ALLOC_SIZE larger than capped GLOBAL_MEM_SIZE resulting in test failure.

https://github.com/KhronosGroup/OpenCL-CTS/blob/6adf4ead5cc39f405568ff4f30095dcccaecd9a4/test_conformance/allocations/main.cpp#L61

A couple of solutions -

1. Always use 64-bit SIZE_MAX limit irrespective of platform bitness to be in sync with cl_ulong size.

2. Cap MAX_MEM_ALLOC_SIZE accordingly to make sure it is within limit of capped GLOBAL_MEM_SIZE.

Jeremy-Kemp:
Instead of `SIZE_MAX`, we could clamp to `CL_ULONG_MAX` instead.

I think we should also make sure that `MAX_MEM_ALLOC_SIZE  <= GLOBAL_MEM_SIZE`

nikhiljnv:
Thanks Jeremy !! How does this affect 32-bit tests on 64-bit platforms? Do we want to test alloc sizes > SIZE_MAX?
Just want to make sure that's ok with everyone.

MathiasMagnus:
As strange as the situation is (a target device having more RAM than the host system does), moreover the host not even being able to address that much memory, I'd vote for capping at the host system bitness. My rationale is that should someone actually using such a system and go beyond system addressable limits with allocations on the target device, they're going to hit hard errors around `CL_MEM_ALLOC_HOST_PTR`, moreover trying to map that buffer to a host ptr. That is a situation where the API will malfunction when going over host addressable sizes. Limiting to `SIZE_MAX`, whatever that means for the host may be a fail-safe solution.

nikhiljnv:
Agree that limiting to SIZE_MAX seems safer to avoid going beyond system addressable ranges.

rjodinchr:
Looking at the current state of the CTS, everything seems to have been addressed.
[...]

PR #1501, opened 2022-09-14
Cap CL_DEVICE_MAX_MEM_ALLOC_SIZE to SIZE_MAX

nikhiljnv:
Cap CL_DEVICE_MAX_MEM_ALLOC_SIZE to SIZE_MAX
when CL_DEVICE_GLOBAL_MEM_SIZE is capped with SIZE_MAX.
test_allocation caps the value of GLOBAL_MEM_SIZE to SIZE_MAX
if it exceeds the value of SIZE_MAX(value depends on platform bitness),
but doesn’t modify MAX_ALLOC_SIZE the same way.
Due to this MAX_ALLOC_SIZE becomes greater than GLOBAL_MEM_SIZE
and the test fails.

Modify MAX_MEM_ALLOC_SIZE as GLOBAL_MEM_SIZE when it exceeds SIZE_MAX

OpenCL-CTS https://github.com/KhronosGroup/OpenCL-CTS/issues/1022


bashbaug:
I could still be reading this wrong, but what I see for [CL_DEVICE_MAX_MEM_ALLOC_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_MAX_MEM_ALLOC_SIZE) is:

> The minimum value is max(min(1024 × 1024 × 1024, 1/4th of [CL_DEVICE_GLOBAL_MEM_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_GLOBAL_MEM_SIZE)), 32 × 1024 × 1024) for devices that are not of type [CL_DEVICE_TYPE_CUSTOM](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_TYPE_CUSTOM).

So, if you had a hypothetical device where CL_DEVICE_GLOBAL_MEM_SIZE were 100GB you could still return CL_DEVICE_MAX_MEM_ALLOC_SIZE == (1024 x 1024 x 1024) == 1GB.  You would not be required to return CL_DEVICE_MAX_MEM_ALLOC_SIZE == 25GB == 1/4 of 100GB because min(1GB, 25GB) == 1GB.

Let's say that the same hypothetical device returned 25GB for CL_DEVICE_MAX_MEM_ALLOC_SIZE.  Are there any circumstances where a 32-bit driver could support a memory allocation this big?  I suppose it _might_ be possible using CL_MEM_NO_HOST_ACCESS, but that's a bit of a stretch.  Note: It'd also need to be an image, because the size of a buffer is specified via a size_t, and hence the largest buffer is effectively 4GB for a 32-bit driver (and quite likely smaller in practice).

The more I think about this the more it seems like a corner case and we should just merge these changes as-is.

bcalidas:
We agree with Ben's analysis. It seems like implementations should not return a CL_DEVICE_MAX_MEM_ALLOC_SIZE that exceeds the  addressable memory range for that process. We would be interested in learning if a need for exceeding the addressable memory range exist.

The change itself seems safe, but it would be good to discuss if such corner cases need to be considered.

Thanks,
Balaji Calidas
Qualcomm

bashbaug (approved):

bashbaug:

PR #1504, opened 2022-09-16
Remove invalid assumptions from sub-group tests

FranklandJack:
The sub-group dispatch tests were making two invalid assumptions about the behaviour of the `clGetKernelSubGroupInfo` API:

1. That for the `CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT` query the actual size of the local size array being queried via the `param_value_size_ret` must be equal to the size of the out parameter used to query the local size `param_value_size`. If this is not the case it wouldn't be possible to query the required size via `param_value_size_ret` without passing `param_value_size` at the same time, which is the very thing you are querying to find out.

2. That querying the local size (via `CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT`) for a sub-group count obtained from a `CL_KERNEL_SUB_GROUP_COUNT_FOR_ND_RANGE` with local size `{x, y, z}` will result in the same local size i.e. that `CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT` is the inverse of `CL_KERNEL_SUB_GROUP_COUNT_FOR_ND_RANGE`. To take a trivial example consider the case that there is only ever one sub-group per work group(i.e. sub-group == work-group). Then regardless of the local size passed to `CL_KERNEL_SUB_GROUP_COUNT_FOR_ND_RANGE` the sub-group count will always be `1`, and `CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT` can return any valid local size since all local sizes result in `1` sub-group.

### DIFF
--- a/test_conformance/api/test_sub_group_dispatch.cpp
+++ b/test_conformance/api/test_sub_group_dispatch.cpp
@@ -127,7 +127,9 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
     log_info("The CL_KERNEL_SUB_GROUP_COUNT_FOR_NDRANGE for the kernel is %d.\n", (int)kernel_subgroup_count);
 
     if (realSize != sizeof(kernel_subgroup_count)) {
-        log_error( "ERROR: Returned size of sub group count not valid! (Expected %d, got %d)\n", (int)sizeof(kernel_subgroup_count), (int)realSize );
+        log_error("ERROR: Returned size of sub group count not valid! "
+                  "(Expected %d, got %d)\n",
+                  (int)sizeof(kernel_subgroup_count), (int)realSize);
         return -1;
     }
 
--- a/test_conformance/api/test_sub_group_dispatch.cpp
+++ b/test_conformance/api/test_sub_group_dispatch.cpp
@@ -135,18 +135,11 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
     for (size_t i = kernel_subgroup_count; i > 0; --i)
     {
         // test all 3 different dimention of requested local size
-        size_t expect_size = kernel_max_subgroup_size * i;
         size_t kernel_ret_size = 0;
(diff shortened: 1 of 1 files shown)


bcalidas:
Regarding Item 1 of this change, we believe that the current test code is correct. The spec effectively forces the application to pass a multiple ( either 1x or 2x or 3x )  of sizeof(size_t) as the param_value_size .  The implementation cannot possibly access any more memory than passed in through param_value_size and therefore must return exactly param_value_size in param_value_size_ret. Our reading of the spec is that param_value_size can never passed in with a zero value.

Regarding Item 2,  the implementation could use a different subgroup size for the CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT query compared to the subgroup size returned by CL_KERNEL_MAX_SUB_GROUP_SIZE_FOR_NDRANGE. Therefore checking the flattened expected local work size generated from these 2 queries seems problematic. This needs more discussion.

FranklandJack:
Thanks for your review Balaji, it's much appreciated :)

For point 1. I think I follow what you are saying. I guess I interpreted the spec slightly differently. The way I read it was that `param_value_size_ret` would query the size of the underlying data being queried and it would be valid for the user to query this data with a smaller `param_value_size` and only get part of the array. I think in this case though, because the return value size is inferred from the input size, the utility of the `param_value_size_ret` is questionable.

For point 2. are you happy to discuss this on here or should we wait for the WG call?

Cheer,

Jack.

bcalidas:
For 2. , let's discuss further in the WG call.

lakshmih:
Minor comment to fix logging.

Sreelakshmi Haridas Maruthur
Qualcomm

bashbaug:
Is this one ready to go?  Needs a review approval before merging.

lakshmih (approved):
Looks good.

Sreelakshmi Haridas Maruthur
Qualcomm

bashbaug:
Has two review approvals - merging.

PR #1505, opened 2022-09-16
[NFC] Avoid mixing signed and unsigned in subhelpers run

svenvh:
Fix a `-Wsign-compare` warning in the `run()` function, which resulted in many repeated warnings when compiling with `-Wall` due to the many template instantiations.

Both `clGetKernelSubGroupInfo` queries return a `size_t`, so it is unclear why the results of these queries were being cast to `int`.

The `dynsc` uses don't seem to work with negative values, so make the field unsigned.


EwanC (approved):

bashbaug:
Merging as discussed on the September 27th teleconference.

PR #1506, opened 2022-09-19
Remove __DATE__ and __TIME__ usage.

tweenk:
These macros make the build non-deterministic.

### DIFF
--- a/test_conformance/contractions/contractions.cpp
+++ b/test_conformance/contractions/contractions.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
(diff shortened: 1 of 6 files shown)


bashbaug (approved):
Does anyone remember why this was added?

At some point we want to add something better (and more deterministic, like the git hash), but that can be done as a separate PR if desired.

bashbaug:
> In my opinion, `-Werror=date-time` should generally be used, unless there is a specific reason not to.

+1

@StuartDBrady are you OK doing this as a separate PR?  OK to merge this as-is?

StuartDBrady:
> @StuartDBrady are you OK doing this as a separate PR? OK to merge this as-is?

I'm happy to do it separately – I don't mind either way.

PR #1507, opened 2022-09-20
 Update extension list of test_compiler

nikhiljnv:
Upate extension list of test_compiler
with missing external memory and semaphore
extensions.


bashbaug (approved):
Merging as discussed on the September 20th teleconference.

PR #1513, opened 2022-09-26
Fix memory oob problem in test conversions

niranjanjoshi121:
Allocate memory for argc arguments
instead of argc - 1.

### DIFF
--- a/test_conformance/conversions/test_conversions.cpp
+++ b/test_conformance/conversions/test_conversions.cpp
@@ -343,7 +343,7 @@ int main (int argc, const char **argv )
 static int ParseArgs( int argc, const char **argv )
 {
     int i;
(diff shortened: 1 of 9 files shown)


bashbaug (approved):

bcalidas:
Given this loop "for( i = 1; i < argc; i++ )" I would think that argCount cannot exceed argc-1 and therefore allocating for argc -1 arguments should be sufficient. Could you provide details on the conditions where the argc-1 limit gets exceeded.

Thanks,
Balaji Calidas
Qualcomm.

jlewis-austin:
I might be having a brain cramp, but I can't see what purpose `argList` serves, and where its allocation is freed? I don't see any similar allocation in the parsing routine for math_brute_force.

bashbaug:

PR #1515, opened 2022-09-27
Fix image test image2d_from_buffer_positive

gwawiork:
This test is failing in case when extensions `cl_ext_image_requirements_info` is not supported and `cl_khr_image2d_from_buffer` is supported. In function `get_image_requirement_alignment` there is macro run `GET_EXTENSION_FUNC` with function `clGetImageRequirementsInfoEXT`. Then test should check if extension `cl_ext_image_requirements_info` is supported


kpet (approved):

bcalidas:
This change looks reasonable. We would like to test and confirm.

bashbaug:

PR #1520, opened 2022-09-30
Add missing type declaration

FranklandJack:
Add a missing type declaration to OpenCL C code strings in 2D async copy tests.

### DIFF
--- a/test_conformance/basic/test_async_copy2D.cpp
+++ b/test_conformance/basic/test_async_copy2D.cpp
@@ -53,7 +53,7 @@ __kernel void test_fn(const __global %s *src, __global %s *dst,
 
   for (int i = 0; i < lineCopiesPerWorkItem; i++) {
     for (int j = 0; j < numElementsPerLine; j++) {
-      const local_index = (get_local_id(0) * lineCopiesPerWorkItem + i) * dstStride + j;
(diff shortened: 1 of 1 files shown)


FranklandJack:
Interestingly this compiles with clang 14 but fails on clang 15 with: 
```
error: type specifier missing, defaults to 'int'; ISO C99 and later do not support implicit int
```
since CL C is based on C99 in which doesn't allow implicit `int` type declarations I'm guessing this is a bug in the CTS rather than in clang (although I'm not a language lawyer, perhaps the CL C spec does explicitly allow this, or is based on an earlier C version).

For reference: https://godbolt.org/z/qE7YW5zqn which uses implicit `int` for `std=c99` and clang 14 compiles with a warning, while https://godbolt.org/z/8M5fzETb1 which is the same but with clang 15 has a compilation error.

jlewis-austin (approved):
LGTM

svenvh (approved):
LGTM.

The clang behaviour has changed with llvm-project commit `2cb2cd242ca0 ("Change the behavior of implicit int diagnostics", 2022-05-04)`:

> This patch [...] defaults the warning to an error in C99 through C17

StuartDBrady:
Merged without any further discussion as this seems fairly trivial.

PR #1521, opened 2022-10-01
Produce JSON results even when a suite's init function reports SKIP or FAIL

kpet:
Also tidy-up some surrounding code.


bcalidas:
We are ok with this change.
Thanks,
Balaji Calidas
Qualcomm

bashbaug (approved):

PR #1522, opened 2022-10-03
pipes: Fix readwrite verification function for fp64

svenvh:
Use the appropriate function for verifying double precision values in the `pipe_readwrite_double` test.

There doesn't seem to be a significant functional difference between `verify_readwrite_double` and `verify_readwrite_long` as both functions use 64-bit integer data types.

Found by compiling with `-Wunused-function`, which flagged `verify_readwrite_double` as unused.

### DIFF
--- a/test_conformance/pipes/test_pipe_read_write.cpp
+++ b/test_conformance/pipes/test_pipe_read_write.cpp
@@ -414,9 +414,9 @@ static int verify_readwrite_ulong(void *ptr1, void *ptr2, int n)
 static int verify_readwrite_double(void *ptr1, void *ptr2, int n)
 {
     int                i;
-    long long int    sum_input = 0, sum_output = 0;
-    long long int    *inptr = (long long int *)ptr1;
-    long long int    *outptr = (long long int *)ptr2;
+    cl_long sum_input = 0, sum_output = 0;
+    cl_long *inptr = (cl_long *)ptr1;
+    cl_long *outptr = (cl_long *)ptr2;
 
     for(i = 0; i < n; i++)
     {
@@ -1246,7 +1246,7 @@ int test_pipe_readwrite_double( cl_device_id deviceID, cl_context context, cl_co
 
     size_t  min_alignment = get_min_alignment(context);
 
-    foo = verify_readwrite_long;
+    foo = verify_readwrite_double;
 
     ptrSizes[0] = sizeof(cl_double);
     ptrSizes[1] = ptrSizes[0] << 1;


StuartDBrady:
Looks good to me.

One could argue that we could _in principle_ have one verify function for char and uchar, another for short, ushort and half, another for int, uint and float, and one more for long, ulong and double – but of course, we should avoid the type-punning of pointers that we currently do for float and double (if I've understood things correctly).  Using the verification function for double certainly makes the handling of doubles _no worse_ than that of floats, so seems like the right thing to do – at least then, if we change verify_readwrite_double to avoid the bad pointer casts, the change would actually take effect.

bcalidas:
This change looks reasonable . We would like to test it before posting final feedback.

StuartDBrady:
@bcalidas, have you understood that there is no difference between `verify_readwrite_double` and `verify_readwrite_long`, given that `long long int` and `cl_long` are equivalent?

bashbaug (commented):
This change itself is fine and this is almost certainly what the test was intending to do, but the entire strategy for verifying results for these tests is rather dodgy: instead of comparing each item of the input and output buffers the verification functions instead sum up each of the elements of the input and output buffers (as integer types) and compares the sum.  Aside from concerns about integer wrapping, this also means that if every even element is one bit too big and every odd element were one bit too small the test would still "pass".

The "float" and "double" verification functions also use the standard "int" and "long long int" types, which AFAIK are not guaranteed to match the size of "float" or "double" respectively.

Fixing everything properly is obviously beyond the scope of this PR, but I would recommend changing to types with a known size (such as `cl_int` and `cl_long`) vs. the standard types at a minimum, to avoid possible problems in the future.

bashbaug:
> For `int` and `float` there's a potential mismatch indeed, as `int` can be 16-bit in some data models. For `double` and `long long int` I'm not too worried: `long long int` is [at least 64 bits wide](https://en.cppreference.com/w/cpp/language/types#Data_models) in all data models.

Yeah, the case I'm worried about (which is still theoretical, as far as I know) is where `long long int` is greater than 64-bits.  If this were to occur then I think the test would read out-of-bounds:

```c
    long long int    sum_input = 0, sum_output = 0;
    long long int    *inptr = (long long int *)ptr1;
    long long int    *outptr = (long long int *)ptr2;

    for(i = 0; i < n; i++)
    {
        sum_input += inptr[i];
        sum_output += outptr[i];
    }
```

I think at least the `long long int` usage should be addressed as part of this PR because `long long int` is only used in `verify_readwrite_double` and hence switching to this function could expose this issue (`verify_readwrite_long` uses `cl_long` and is hence guaranteed to match `sizeof(double)`).  I'm OK addressing the `int` usage in `verify_readwrite_float` in a subsequent PR, if desired, though it'd be nice to fix it at the same time.

svenvh:
Thanks for the clarification, I get your concern now.  This PR could indeed break double verification on some platforms, so I'll address it as part of this.

StuartDBrady:
Note that there around around 80 other occurrences of `long long` in the CTS.  I expect that most of these will need to be changed, too.

bashbaug (approved):
Thanks, latest version LGTM!

bashbaug:
> Note that there around around 80 other occurrences of `long long` in the CTS. I expect that most of these will need to be changed, too.

Agreed, we should file an issue to track this.

svenvh:
> Agreed, we should file an issue to track this.

I have opened #1526 for this.

bcalidas:
We were able to test this change. We are good with merging it.

Thanks,
Balaji Calidas
Qualcomm

PR #1524, opened 2022-10-04
test_bufferreadwriterect - Limit memory allocation on host side

gwawiork:
Usually device memory is almost the same or smaller than host memory. Looks like this test was written as it always true. However in condition when device memory is much bigger than host memory buffers allocation on host side (for verification) cause out of memory on the system. This code prevents this situation.
We can discuss how much `max_mem_alloc_size` should be limited. In my proposal is set to `host_max_memory / 2`

### DIFF
--- a/test_common/CMakeLists.txt
+++ b/test_common/CMakeLists.txt
@@ -12,6 +12,7 @@ set(HARNESS_SOURCES
     harness/imageHelpers.cpp
     harness/kernelHelpers.cpp
     harness/deviceInfo.cpp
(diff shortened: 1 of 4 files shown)


gwawiork:
Can anybody propose `sysinfo` equivalent for mac OS

bashbaug:
Consider addressing https://github.com/KhronosGroup/OpenCL-CTS/issues/1529 as part of this PR also, if it is easy enough to do.

PR #1511, opened 2022-09-23
math_brute_force: Factor out GetTernaryKernel

svenvh:
Use a common function to create the kernel source code for testing 3-argument math builtins.  This reduces code duplication.  1-argument and 2-argument math kernel construction will be factored out in a future PR.

Change the kernels to use preprocessor defines for argument types and undef values, to make the CTS code easier to read.

Co-authored-by: Marco Antognini <marco.antognini@arm.com>
Signed-off-by: Marco Antognini <marco.antognini@arm.com>
Signed-off-by: Sven van Haastregt <sven.vanhaastregt@arm.com>

PR #1525, opened 2022-10-05
math_brute_force: Factor out GetUnaryKernel and GetBinaryKernel

svenvh:
Use common functions to create the kernel source code for testing 1-argument and 2-argument math builtins.  This reduces code duplication.

Use appropriate patterns to initialise variables to their full bit widths.  For example, `0xdead` was previously used to initialise 32-bit integers, while now a larger number spanning all bytes is used.

Co-authored-by: Marco Antognini <marco.antognini@arm.com>


svenvh:
Continuation of the work in #1511 .

lakshmih:
We are okay with this change.

Sreelakshmi Haridas Maruthur
Qualcomm

lakshmih (approved):

bashbaug:
Has two reviews, merging.  (Thanks for the nice cleanup!)

PR #1530, opened 2022-10-11
Make building the Vulkan interop tests optional

kbenzie:
Allow enabling/disabling building of the Vulkan interop tests via CMake option without modifying the CMake. This matches the optionality of the D3D10, D3D11, GL, and GLES interop tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,7 @@ option(D3D10_IS_SUPPORTED "Run DirectX 10 interop tests" OFF)
 option(D3D11_IS_SUPPORTED "Run DirectX 11 interop tests" OFF)
 option(GL_IS_SUPPORTED "Run OpenGL interop tests" OFF)
 option(GLES_IS_SUPPORTED "Run OpenGL ES interop tests" OFF)
+option(VULKAN_IS_SUPPORTED "Run Vulkan interop tests" OFF)
 
 
 #-----------------------------------------------------------
--- a/test_conformance/CMakeLists.txt
+++ b/test_conformance/CMakeLists.txt
@@ -52,7 +52,9 @@ add_subdirectory( pipes )
 add_subdirectory( device_timer )
 add_subdirectory( spirv_new )
 add_subdirectory( spir )
-add_subdirectory( vulkan )
+if(VULKAN_IS_SUPPORTED)
+    add_subdirectory( vulkan )
+endif()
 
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ option(D3D10_IS_SUPPORTED "Run DirectX 10 interop tests" OFF)
 option(D3D11_IS_SUPPORTED "Run DirectX 11 interop tests" OFF)
 option(GL_IS_SUPPORTED "Run OpenGL interop tests" OFF)
 option(GLES_IS_SUPPORTED "Run OpenGL ES interop tests" OFF)
+option(VULKAN_IS_SUPPORTED "Run Vulkan interop tests" OFF)
```
kbenzie, line 33: This is a behavior change, should it be `ON` be default?

bcalidas:
We are ok with this change.
Thanks,
Balaji Calidas

bashbaug (approved):

lakshmih (approved):

nikhiljnv:
Given that these tests are likely to see more changes coming in, turning them off in the builds will prevent us from catching build issues on new PRs touching these tests. Possibly, we will need to turn them on temporarily to get CI builds passing with new PRs and turn it off when merging PRs.

EwanC:
> Given that these tests are likely to see more changes coming in, turning them off in the builds will prevent us from catching build issues on new PRs touching these tests. Possibly, we will need to turn them on temporarily to get CI builds passing with new PRs and turn it off when merging PRs.

We could set this flag to ON in the CI build script - https://github.com/KhronosGroup/OpenCL-CTS/blob/main/presubmit.sh#L95 So that the tests are always enabled in CI, but the default is still OFF

nikhiljnv:
> > Given that these tests are likely to see more changes coming in, turning them off in the builds will prevent us from catching build issues on new PRs touching these tests. Possibly, we will need to turn them on temporarily to get CI builds passing with new PRs and turn it off when merging PRs.
> 
> We could set this flag to ON in the CI build script - https://github.com/KhronosGroup/OpenCL-CTS/blob/main/presubmit.sh#L95 So that the tests are always enabled in CI, but the default is still OFF

Agree. Realized after commenting here. Mentioned the same here https://github.com/KhronosGroup/OpenCL-CTS/pull/1559#issuecomment-1307522266. Will make the change in build script.

PR #1531, opened 2022-10-11
Add explicit flush to clEnqueueMapBuffer test

FranklandJack:


bcalidas:
We have concerns about this change given that the queue is in-order. Our understanding of the spec is that command B enqueued to an in-order queue should execute as if the previously enqueued command A had fully completed. In this context, interference between an EnqueueUmapMemObject command and subsequent EnqueueMapBuffer commands would point to an implementation issue.

Thanks,
Balaji Calidas

PR #1541, opened 2022-10-13
Use CTS type wrappers for test_basic test_loop

JohnKesapidesARM:

### DIFF
--- a/test_conformance/basic/test_loop.cpp
+++ b/test_conformance/basic/test_loop.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


test_conformance/basic/test_loop.cpp
```diff
@@ -21,165 +21,117 @@
...
+        dst[tid] += src[j];
+    }
+}
+)";
+
+
+int verify_loop(std::vector<cl_int> inptr, std::vector<cl_int> loopindx,
+                std::vector<cl_int> loopcnt, std::vector<cl_int> outptr, int n)
+{
+    int r, i, j, k;
```
EwanC, line 50: Can we move these variable declaration to where they are initialized? e.g `for (int i = 0;...`. This is quite a nice clean-up when you're done elsewhere.

EwanC (approved):

svenvh (approved):

PR #1545, opened 2022-10-13
Use CTS type wrappers for test_global_linear_id.

JohnKesapidesARM:

### DIFF
--- a/test_conformance/basic/test_global_linear_id.cpp
+++ b/test_conformance/basic/test_global_linear_id.cpp
@@ -74,7 +74,7 @@ int test_global_linear_id(cl_device_id device, cl_context context,
                                    "test_global_linear_id_2d" };
     const char *kernel_code[] = { global_linear_id_1d_code,
                                   global_linear_id_2d_code };
-    int err;
+    int err = CL_SUCCESS;
 
     num_elements = static_cast<int>(sqrt(static_cast<float>(num_elements)));
     int length = 1;
--- a/test_conformance/basic/test_global_linear_id.cpp
+++ b/test_conformance/basic/test_global_linear_id.cpp
@@ -53,22 +53,20 @@ __kernel void test_global_linear_id_1d(global int *dst)
 
 int verify_global_linear_id(std::vector<cl_int> &result, int n)
 {
-
     if (std::any_of(result.begin(), result.begin() + n,
                     [](cl_int value) { return 0 == value; }))
     {
         log_error("get_global_linear_id failed\n");
-        return -1;
+        return TEST_FAIL;
     }
     log_info("get_global_linear_id passed\n");
-    return 0;
+    return TEST_PASS;
 }
 }
 
 int test_global_linear_id(cl_device_id device, cl_context context,
                           cl_command_queue queue, int num_elements)
 {
-    clMemWrapper streams;
     clProgramWrapper program[2];
     clKernelWrapper kernel[2];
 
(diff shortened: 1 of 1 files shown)


test_conformance/basic/test_global_linear_id.cpp
```diff
@@ -22,105 +22,93 @@
...
+    for (int i = 0; i < ARRAY_SIZE(program); i++)
+    {
+        err = create_single_kernel_helper(context, &program[i], &kernel[i], 1,
+                                          &kernel_code[i], kernel_names[i]);
+        test_error(err, "create_single_kernel_helper failed");
+
+        err = clSetKernelArg(kernel[i], 0, sizeof streams, &streams);
+        test_error(err, "clSetKernelArgs failed.");
+
+        err = clEnqueueNDRangeKernel(queue, kernel[i], i + 1, nullptr, threads,
```
EwanC, line 101: In the old code we had

* line 100 - `err = clEnqueueNDRangeKernel(queue, kernel[1], 2,...`
* line 109 - `err = clEnqueueNDRangeKernel(queue, kernel[1], 1,..`

So they were both using `kernel[1]`, and `kernel[0]` never seemed to be enqueued, was this a bug?

JohnKesapidesARM: yeap indeed it was ;).

EwanC (approved):

test_conformance/basic/test_global_linear_id.cpp
kpet: Hmm, I think the test as refactored is broken. The first kernel will only write to the first `num_elements` value of the buffer (that is not pre-initialised) and yet the verification code will always check `num_elements * num_elements` values. Am I missing something?

test_conformance/basic/test_global_linear_id.cpp
```diff
@@ -22,105 +22,93 @@
...
+                                     nullptr, 0, nullptr, nullptr);
+        test_error(err, "clEnqueueNDRangeKernel failed.");
+
+        err = clEnqueueReadBuffer(queue, streams, CL_TRUE, 0,
+                                  length * sizeof(int), output.data(), 0,
+                                  nullptr, nullptr);
+        test_error(err, "clEnqueueReadBuffer failed.");
+
+        err = verify_global_linear_id(output, length);
```
kpet, line 110: Verification probably fails in the first iteration of the loop (see comment above) but the status isn't checked and in the second iteration the error gets overwritten.

kpet (changes requested):

test_conformance/basic/test_global_linear_id.cpp
JohnKesapidesARM: hmm, looks like you are right. I replaced one bug with another ;)

test_conformance/basic/test_global_linear_id.cpp
```diff
@@ -22,105 +22,93 @@
...
-    clReleaseProgram(program[0]);
-    clReleaseProgram(program[1]);
-    free(output_ptr);
+
+    num_elements = static_cast<int>(sqrt(static_cast<float>(num_elements)));
+    int length = 1;
+    size_t threads[] = { static_cast<size_t>(num_elements),
+                         static_cast<size_t>(num_elements) };
+
+    for (int i = 0; i < ARRAY_SIZE(program) && !err; i++)
```
EwanC, line 84: we are checking `&& !err` here, but `int err;` on line 77 isn't getting initialized before the first loop iteration. Think line 77 should be `int err = CL_SUCCESS;`

JohnKesapidesARM: yeap you are right, thanks for spotting that. I'll fix asap

EwanC (approved):

lakshmih (approved):

bashbaug (approved):

kpet (approved):

PR #1549, opened 2022-10-14
subgroups: Fix Wformat warnings

svenvh:
The main source of warnings was the use of `%d` for printing a templated type `T`, where `T` could be any cl_ scalar or vector type.

Introduce `print_expected_obtained`.  It takes const references to handle alignment of the cl_ types.

Define `operator<<` for all types used by the subgroup tests.  Ideally those would be template functions enabled by TypeManager data, but that requires some more work on the TypeManager (which we'd ideally do after more warnings have been enabled).  So for now, define the `operator<<` instances using preprocessor defines.

Also fix a few instances where the wrong format specifier was used for `size_t` types.

### DIFF
--- a/test_conformance/subgroups/subhelpers.cpp
+++ b/test_conformance/subgroups/subhelpers.cpp
@@ -1,0 +1,62 @@
+//
+// Copyright (c) 2022 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "subhelpers.h"
+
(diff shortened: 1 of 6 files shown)


test_conformance/subgroups/subhelpers.cpp
```diff
@@ -0,0 +1,62 @@
...
+// limitations under the License.
+//
+
+#include "subhelpers.h"
+
+// Define operator<< for cl_ types, accessing the .s member.
+#define OP_OSTREAM(Ty, VecSize)                                                \
+    std::ostream& operator<<(std::ostream& os, const Ty##VecSize& val)         \
+    {                                                                          \
+        os << +val.s[0];                                                       \
```
EwanC, line 23: I'm probably missing something, but why do we need the `+` before `val.s[x]`?

EwanC (dismissed):

test_conformance/subgroups/subhelpers.cpp
svenvh: It's to ensure `char`s get printed as a number instead of being rendered as a character.  Perhaps I should add a comment to explain that.

EwanC: ah that makes sense, a quick comment would be great

svenvh: Added a comment.

EwanC (approved):

lakshmih (approved):
Looks good.

Sreelakshmi Haridas Maruthur
Qualcomm

Issue #1266, opened 2021-05-26
Data races and thread leaks


mantognini:
ThreadSanitizer detects some data races and thread leaks, at least when running math_brute_force. They happen in the harness code. #1265 fixes _some_ data races but:

- many suspicious use of `volatile` remain (which is not appropriate to prevent data races*) in the harness and CTS as a whole;
- leaks are not fixed.

Example of leaks reported by ThreadSanitizer (`-fsanitizer=thread`) on a Ubuntu system:
```
WARNING: ThreadSanitizer: thread leak (pid=16188)                                                                                    
  Thread T58 (tid=16247, finished) created by main thread at:                                                                        
    #0 pthread_create <null> (test_bruteforce+0x42c1bb)                                                                              
    #1 ThreadPool_Init() test_common/harness/ThreadPool.cpp:635:15 (test_bruteforce+0x53041a)                                                                                                               
    #2 GetThreadCount() test_common/harness/ThreadPool.cpp:945:18 (test_bruteforce+0x53074f)                                                                                                                
    #3 runTestHarnessWithCheck(int, char const**, int, test_definition*, int, unsigned long, test_status (*)(_cl_device_id*)) test_common/harness/testHarness.cpp:432:30 (test_bruteforce+0x52d01f)         
    #4 __libc_start_main /build/glibc-S9d2JN/glibc-2.27/csu/../csu/libc-start.c:310 (libc.so.6+0x21bf6)                                  
                                                                    
  And 55 more similar thread leaks.                                 
                                                                    
SUMMARY: ThreadSanitizer: thread leak (test_bruteforce+0x42c1bb) in pthread_create
```

I suspect most of the existing platform-specific code can be replaced with standard C++11 alternatives from the STL.

---

\* quoting http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1152r0.html

> [...] volatile doesn’t imply any observable ordering in terms of the C++ memory model. Atomic instructions guarantee sequential consistency for data-race free programs (data races are otherwise undefined behavior) [BATTY]. volatile has no such guarantee and doesn’t imply a memory ordering or any fencing, though some implementations provide stronger guarantees [...]

PR #1553, opened 2022-10-28
Fix thread leaks in the thread pool

kbenzie:
While testing an OpenCL driver with ThreadSanitizer enabled the OpenCL-CTS suffers from thread leaks in conversions and bruteforce on posix systems. This is because `pthread_join` is never called in `ThreadPool_Exit` for the `pthread_t`s created by the thread pool. Instead, the threads are only informed to stop waiting on the condition variable which unblocks the worker thread but does not clean up after itself.

```
ThreadPool: thread 1 exiting.
ThreadPool: thread 5 exiting.
ThreadPool: thread 4 exiting.
ThreadPool: thread 2 exiting.
ThreadPool: thread 7 exiting.
ThreadPool: thread 0 exiting.
ThreadPool: thread 3 exiting.
ThreadPool: thread 6 exiting.
Thread pool exited in a orderly fashion.
==================
WARNING: ThreadSanitizer: thread leak (pid=2292842)
  Thread T9 (tid=2292855, finished) created by main thread at:
    #0 pthread_create ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:969 (libtsan.so.0+0x5ad75)
    #1 ThreadPool_Init() <null> (test_conversions+0x35b2c)
    #2 pthread_once ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:1449 (libtsan.so.0+0x4057c)
    #3 GetThreadCount() <null> (test_conversions+0x36262)
    #4 DoTest(_cl_device_id*, Type, Type, SaturationMode, RoundingMode, _MTdata*) [clone .isra.0] <null> (test_conversions+0x10555)
    #5 test_conversions(_cl_device_id*, _cl_context*, _cl_command_queue*, int) <null> (test_conversions+0x13226)
    #6 callSingleTestFunction(test_definition, _cl_device_id*, int, int, unsigned long) <null> (test_conversions+0x2e66d)
    #7 parseAndCallCommandLineTests(int, char const**, _cl_device_id*, int, test_definition*, int, unsigned long, int) <null> (test_conversions+0x2fb3a)
    #8 runTestHarnessWithCheck(int, char const**, int, test_definition*, int, unsigned long, test_status (*)(_cl_device_id*)) <null> (test_conversions+0x349d8)
    #9 main <null> (test_conversions+0xd725)

  And 7 more similar thread leaks.

SUMMARY: ThreadSanitizer: thread leak (OpenCL-CTS/buildbin/conversions/test_conversions+0x35b2c) in ThreadPool_Init()
```

This patch adds global state to keep track of the `pthread_t`s created by `pthread_create` in `ThreadPool_Init`. The list of `pthread_t`s is then used by `ThreadPool_Exit` to call `pthread_join` to cleanup the `pthread_t`s correctly.

A near identical example, and additional explanation, can be found on [stackoverflow](https://stackoverflow.com/questions/72435574/thread-leak-detected-when-using-condition-variable-instead-of-join-with-pthrea).

On the Windows path, a similar change is not necessary because `_beginthread` is used which automatically cleans up after itself when the worker thread function returns.

### DIFF
--- a/test_common/harness/ThreadPool.cpp
+++ b/test_common/harness/ThreadPool.cpp
@@ -23,6 +23,7 @@
 // or any other POSIX system
 
 #include <atomic>
+#include <vector>
 
 #if defined(_WIN32)
 #include <windows.h>
@@ -56,6 +57,12 @@ CRITICAL_SECTION gAtomicLock;
 #elif defined(__GNUC__) || defined(_MSC_VER)
 #else
 pthread_mutex_t gAtomicLock;
+#endif
+
+#if !defined(_WIN32)
+// Keep track of pthread_t's created in ThreadPool_Init() so they can be joined
+// in ThreadPool_Exit() and avoid thread leaks.
+static std::vector<pthread_t> pthreads;
 #endif
 
 // Atomic add operator with mem barrier.  Mem barrier needed to protect state
@@ -642,6 +649,9 @@ void ThreadPool_Init(void)
             gThreadCount = i;
             break;
         }
+#if !defined(_WIN32)
+        pthreads.push_back(tid);
+#endif // !_WIN32
     }
 
     atexit(ThreadPool_Exit);
@@ -721,7 +731,20 @@ void ThreadPool_Exit(void)
                   "still active.\n",
                   gThreadCount.load());
     else
+    {
+#if !defined(_WIN32)
+        for (pthread_t pthread : pthreads)
+        {
+            if (int err = pthread_join(pthread, nullptr))
+            {
+                log_error("Error from %d from pthread_join. Unable to join "
+                          "work threads. ThreadPool_Exit failed.\n",
+                          err);
+            }
+        }
+#endif
         log_info("Thread pool exited in a orderly fashion.\n");
+    }
 }
 
 


b-sumner:
Is this preferable to creating the threads detached?

kbenzie:
> Is this preferable to creating the threads detached?

For me yes, since this resolves the ThreadSanitizer issue I came across.

I can't speak to the original intent of the thread pool implementation though. Having looked at the code I don't really see a good reason to make these threads detached. This cleanup is happening after `main` exists in an `atexit` handler and all work should already be complete, joining seems like the sensible thing to do to me.

bashbaug (dismissed):
LGTM.

Longer term it would be nice to shift this thread pool stuff to use std::threads rather than OS-specific codepaths.

bashbaug:
Note: possibly related to #1266?

kbenzie:
> Longer term it would be nice to shift this thread pool stuff to use std::threads rather than OS-specific codepaths.

Fully agree, although that's a larger task than I have bandwidth for.

> Note: possibly related to https://github.com/KhronosGroup/OpenCL-CTS/issues/1266?

It does look related, yeah. I haven't come across any data races though because in my situation the OpenCL-CTS itself was not compiled with `-fsanitize=thread` only the application using it. That still means the posix threading functions are intercepted though, which is why I found these thread leaks.

test_common/harness/ThreadPool.cpp
```diff
@@ -634,6 +641,7 @@ void ThreadPool_Init(void)
         pthread_t tid = 0;
         err = pthread_create(&tid, NULL, ThreadPool_WorkerFunc,
                              (void *)&threadID);
+        pthreads.push_back(tid);
```
svenvh, line 644: I'd suggest to do this after checking `err` to avoid storing garbage into the vector in case `pthread_create` fails.

kbenzie: Good spot.

kbenzie: Fixed.

svenvh (approved):

bashbaug:
Has two approvals, merging.

PR #1554, opened 2022-11-01
Half

gpx1000:


bcalidas:
We  are in the process of reviewing this change and will provide feedback once our review is completed.

Thanks,
Balaji Calidas
Qualcomm

EwanC (changes requested):
When I try out this branch I see
```
vstore_half...
                FAILED -- clBuildProgramExecutable() failed:
Log: Compilation started
37:5:13: error: assigning directly to pointer to type 'half' requires cl_khr_fp16. Use vector data store builtin functions instead
   *(f + i) = 0xdeaddead;}
            ^
Compilation failed

Source :
__kernel void reset( __global float *p, __global half *f,
                   uint extra_last_thread)
{
   size_t i = get_global_id(0);
   *(f + i) = 0xdeaddead;}

vstore_half FAILED
FAILED sub-test.
FAILED test.
```

bcalidas:
This PR is missing   "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n"  in the kernel source. I believe this is still required.
This leads to a kernel build failure.

test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -562,151 +632,227 @@ int Test_vStoreHalf_private( cl_device_id device, f2h referenceFunc, d2h doubleR
...
+        "__kernel void reset( __global float *p, __global half *f,\n"
+        "                   uint extra_last_thread)\n"
+        "{\n"
+        "   size_t i = get_global_id(0);\n"
+        "   *(f + i) = 0xdeaddead;"
+        "}\n"
+    };
+
+    resetProgram = MakeProgram(device, reset, sizeof(reset) / sizeof(reset[0]));
+    resetKernel = clCreateKernel(resetProgram, "reset", &error);
```
bashbaug, line 837: Could this use `clEnqueueFillBuffer` to do the on-device reset instead of a kernel, similar to the changes in #1561?

bashbaug:
I don't think these tests should require `cl_khr_fp16`.  Even though they're "half" tests, these specific tests are only testing `vload_half` and `vstore_half` and hence require no extensions.

I think the best solution is to use `clEnqueueFillBuffer` to do the on-device reset, see review comment above.  If we absolutely need to use a kernel for some reason though, I think we can still do so by initializing this buffer as a 16-bit integer `ushort` rather than a 16-bit floating-point `half`.  This is probably what is intended anyhow, because otherwise `0xdeaddead` exceeds the maximum value of a `half`.

bcalidas:
EnqueueFillBuffer might not always be optimally implemented.  Using a kernel to reset the values in the output buffer seems like a safe approach. We'd like also to ask if the reset kernel has been consistently leveraged throughout the test. For example, is there any reason to use EnqueueWriteBuffer.

The half data type appears in the reset kernel.

bashbaug:
> EnqueueFillBuffer might not always be optimally implemented. Using a kernel to reset the values in the output buffer seems like a safe approach.

Even if this is true in practice wouldn't it be best for these implementations to optimize clEnqueueFillBuffer vs. falling back to a kernel?  A pattern size of 16-bits is fairly common so it's not like this is some weird special case.

It seems odd that we would have a API specifically dedicated to fill a buffer but we wouldn't use it when we need to fill (reset) a buffer.  Note that some implementations may perform better with clEnqueueFillBuffer vs. a kernel if they know this is a fill operation, especially if if they have dedicated hardware to fill (or reset, or "clear") a buffer.

bcalidas:
The issue of whether the reset kernel needs the cl_khr_fp16 pragma needs further discussion in the next WG meeting.

bashbaug:
I think as-written the kernel **does** need the cl_khr_fp16 pragma, but I don't think adding it is the right thing to do.

I think we have two viable paths forward:

1. We switch from a reset kernel to `clEnqueueFillBuffer`, in which case we don't need a kernel at all.
2. We switch the reset kernel from writing `half` data to writing some other data type (probably `ushort`, though it could also be a larger datatype like `uint`).  We can't use `half` directly because this requires `cl_khr_fp16` and not all devices that run this test support the `cl_khr_fp16` extension, and we shouldn't use `vstore_half` to do the reset because `vstore_half` is what is being tested.

I still think (1) is the right thing to do but I could live with (2) to unblock progress.

lakshmih:
Thanks for the update, we are reviewing the latest patch. 

Sreelakshmi Haridas Maruthur
Qualcomm

test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -746,7 +892,21 @@ int Test_vStoreHalf_private( cl_device_id device, f2h referenceFunc, d2h doubleR
     dchk.lim = blockCount;
     dchk.count = (blockCount + threadCount - 1) / threadCount;
 
-    for( i = 0; i < lastCase; i += stride )
+    cl_uint pattern = 0xdeaddead;
+    memset_pattern4(gOut_half, &pattern, BUFFER_SIZE / 2);
+
+    error =
+        clEnqueueWriteBuffer(gQueue, gOutBuffer_half, CL_FALSE, 0,
+                             count * sizeof(cl_half), gOut_half, 0, NULL, NULL);
```
bashbaug, line 900: Can you please check if these added calls to clEnqueueWriteBuffer are needed?  It seems like the buffer (gOutBuffer_half) should be reset via the kernel exclusively now (inside of the loop), so this call to clEnqueueWriteBuffer is unnecessary.

test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -755,556 +915,771 @@ int Test_vStoreHalf_private( cl_device_id device, f2h referenceFunc, d2h doubleR
...
+                                  gOutBuffer_half, count, 0);
+                if (error)
+                {
                     gFailCount++;
                     goto exit;
                 }
 
-                error = RunKernel(device, kernels[vectorSize][addressSpace], gInBuffer_single, gOutBuffer_half,
-                                       numVecs(count, vectorSize, aligned) ,
+                memset_pattern4(gOut_half, &pattern, BUFFER_SIZE / 2);
```
bashbaug, line 964: This may not matter much in the grand scheme of things, but gOut_half is only used as a destination for the call to clEnqueueReadBuffer below now, so memset'ing the pattern each time before calling clEnqueueReadBuffer isn't necessary.

In order, I think my preference would be:

1. Remove it completely because it isn't needed.
2. If we can't remove it completely, move it immediately before the call to clEnqueueReadBuffer, perhaps with a comment why it is there.
3. If we don't want to remove it or move it, I suppose we can keep it as-is, because it's functionally correct now even if it is confusing.

bashbaug (commented):
Getting closer, thanks.  I have a few remaining nice-to-have fixes.

JohnKesapidesARM:
Same comment here for the on-device reset as https://github.com/KhronosGroup/OpenCL-CTS/pull/1561#issuecomment-1339740910
Could we make this behavior controllable via a command line option? If this is executed on device, on real silicon it might be faster, but a lot of us work on FPGAs. Unless we say there is test coverage benefit this will slow down regression testing a lot.

cdavis5e:
Sorry about the delay. Andrew and Taet have been working on other important projects, and Steve has been battling cancer--a battle he has hopefully now won 🙏 .

test_conformance/half/Test_vStoreHalf.cpp
cdavis5e: Doesn't appear to be necessary after all. I've removed it.

test_conformance/half/Test_vStoreHalf.cpp
cdavis5e: Removed.

cdavis5e:
> Could we make this behavior controllable via a command line option?

Done. The new `-r` option will cause resets to occur on host instead of on device.

test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -562,151 +632,230 @@ int Test_vStoreHalf_private( cl_device_id device, f2h referenceFunc, d2h doubleR
...
+        "   *(f + i) = 0xdeaddead;"
+        "}\n"
+    };
+
+    if (!gHostReset)
+    {
+        resetProgram =
+            MakeProgram(device, reset, sizeof(reset) / sizeof(reset[0]));
+        resetKernel = clCreateKernel(resetProgram, "reset", &error);
+        if (NULL == resetProgram || NULL == resetKernel)
```
EwanC, line 841: I think we should check the `NULL == resetProgram` error case before calling `clCreateKernel()` , rather than afterwards, as is the convention in the calls to `MakeProgram` / `clCreateKernel` before this.

test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -562,151 +632,230 @@ int Test_vStoreHalf_private( cl_device_id device, f2h referenceFunc, d2h doubleR
...
             }
         }
     } // end for vector size
 
+    const char *reset[] = {
+        "__kernel void reset( __global float *p, __global ushort *f,\n"
+        "                   uint extra_last_thread)\n"
+        "{\n"
+        "   size_t i = get_global_id(0);\n"
+        "   *(f + i) = 0xdeaddead;"
```
EwanC, line 832: I'm seeing the following warning with on this line of kernel code
```
kernel.opencl:5:15: warning: implicit conversion from 'unsigned int' to 'ushort' (aka 'unsigned short') changes value from 3735936685 to 57005
````

cdavis5e: Fixed.

I wonder if this could be sped up even more by storing 32-bit ints instead of 16-bit ints...

test_conformance/half/Test_vStoreHalf.cpp
cdavis5e: Done.

EwanC (commented):
Looks like there's still some clang-formatting needed from the recent changes to appease CI, but apart from that this change now looks fine to me

cdavis5e:
> Looks like there's still some clang-formatting needed from the recent changes to appease CI, but apart from that this change now looks fine to me

If I make those changes, I'll also need to alter the formatting of the surrounding code, or it'll be inconsistent. I kinda want to avoid that, because it affects e.g. `git blame`.

EwanC (approved):

lakshmih (approved):

bashbaug:

PR #1556, opened 2022-11-01
Improve logging of unknown extensions

StuartDBrady:
Add missing newlines, and log all unknown extensions instead of logging only the first unknown extension found.

### DIFF
--- a/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
+++ b/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
@@ -164,6 +164,7 @@ int test_compiler_defines_for_extensions(cl_device_id device, cl_context context
 
     memset( extension_type, 0, sizeof( extension_type) );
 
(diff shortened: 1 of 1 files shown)


StuartDBrady:
Note: this test has too many issues to deal with in one go.  My hope is to initially decrease the amount of duplication in the C code used here with trivially-safe changes, and once that is done, rewrite it to use std::vector, std::string etc.

bashbaug (approved):
LGTM, tested all of these conditions by injecting a device extension string with the OpenCL intercept layer:

```sh
$ CLI_DeviceExtensions="cl_khr_foobar cl_khr_baz cles_khr_int64 myext_is_awesome cl_ cl__myextension cl_myextension " cliloader ./test_conformance/compiler/test_compiler compiler_defines_for_extensions
```

All of the errors were reported as expected.

EwanC (approved):

PR #1557, opened 2022-11-02
Expand math_brute_force in the conformance lists to list each subtest.

anholt:
If you're using a test runner to parse the .csv and shard execution across multiple threads, spawning the test per thread helps balance across CPUs.  On my workstation, running wimpy test_bruteforce on a radeon with no args takes 5:21, while with parallelization it completes in 4:21.  Given the duration of each subtest (generally 3+ seconds), we don't care about the overhead of spinning up the driver extra times.


anholt:
List generated using the `-p` argument.

Karol Herbst has a parallel runner that parses the output of various tests with specific flags and regexes to print and parse their subtests.  I would rather we either had this, or a standardized subtest printing in the tests, so that we don't need multiple parsers.  I'm happy to pursue either route, but wanted to start with this MR before I go generating and pasting subtest lists for other tests.

anholt:
List generated using the `-p` argument.

Karol Herbst has a parallel runner that parses the output of various tests with specific flags and regexes to print and parse their subtests.  I would rather we either had this, or a standardized subtest printing in the tests, so that we don't need multiple parsers.  I'm happy to pursue either route, but wanted to start with this MR before I go generating and pasting subtest lists for other tests.

PR #1559, opened 2022-11-03
Vulkan: Use linear tiling

lakshmih:
The external memory sharing spec does not require
OpenCL to deduce if a Vulkan image is tiled or linear. Use tiled images.

Additionally, when selecting a queue family, the Vulkan implementation is not required to report the transfer bit for compute or graphics queues, it is implied.

### DIFF
--- a/test_conformance/vulkan/main.cpp
+++ b/test_conformance/vulkan/main.cpp
@@ -143,6 +143,7 @@ bool useDeviceLocal = false;
 bool disableNTHandleType = false;
 bool enableOffset = false;
 bool non_dedicated = false;
+bool useOptimalTiling = false;
 
 static void printUsage(const char *execName)
 {
@@ -192,6 +193,10 @@ size_t parseParams(int argc, const char *argv[], const char **argList)
             if (!strcmp(argv[i], "--non_dedicated"))
             {
                 non_dedicated = true;
+            }
+            if (!strcmp(argv[i], "--useOptimalTiling"))
+            {
+                useOptimalTiling = true;
             }
             if (strcmp(argv[i], "-h") == 0)
             {
--- a/test_conformance/vulkan/vulkan_interop_common/vulkan_utility.hpp
+++ b/test_conformance/vulkan/vulkan_interop_common/vulkan_utility.hpp
@@ -33,7 +33,8 @@
 const VulkanInstance& getVulkanInstance();
 const VulkanPhysicalDevice& getVulkanPhysicalDevice();
 const VulkanQueueFamily&
-getVulkanQueueFamily(uint32_t queueFlags = VULKAN_QUEUE_FLAG_MASK_ALL);
+getVulkanQueueFamily(uint32_t queueFlags = VULKAN_QUEUE_FLAG_COMPUTE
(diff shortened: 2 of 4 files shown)


nikhiljnv:
#1530 has disabled these tests in CMake by default.
So, passing CI builds does not guarantee successful builds with the changes.
While the changes themselves are small and unlikely to cause any build breaks, I would encourage that we enable  these tests in CMake temporarily for future PRs to ensure we don't break the builds and turn off it again in CMake before merging the change. Alternatively, we can update presubmit  script to have VULKAN_IS_SUPPORTED option enabled by default so that CI builds don't skip these tests. I prefer latter.

bcalidas:
HI Nikhil,

We do want to build and run these tests. Our recommendation is that supporting interop of Linear tiled Vulkan images with OpenCL is sufficient for an implementation to be conformant.

Thanks,
Balaji

nikhiljnv:
> HI Nikhil,
> 
> We do want to build and run these tests. Our recommendation is that supporting interop of Linear tiled Vulkan images with OpenCL is sufficient for an implementation to be conformant.
> 
> Thanks, Balaji

Hi Balaji,
I am fine to keep conformance requirements minimal - Passing the test only for required (linear) tiling support should be sufficient.
However, not having test coverage for other tiling cases will prevent us not covering the underlying issues. In near future, we may want to address any special requirements for these cases in the spec. And to get there, it is important to have test coverage for different Vulkan formats and settings.
Any concerns keeping all the tiling types in the test and have an option to control which types to exercise with default being only linear tiling. Can think of similar needs for different memory flags. (Think of image sampler tests where certain addressing/ filtering modes are not required for conformance, but conformance tests cover the other non-mandatory modes as well that can be optionally run).

bcalidas:
Hi Nikhil,

We are ok with adding an option for testing optimally tiled images as long as the default is linear. Should we update the PR with that option?

Thanks,
Balaji

nikhiljnv:
> Hi Nikhil,
> 
> We are ok with adding an option for testing optimally tiled images as long as the default is linear. Should we update the PR with that option?
> 
> Thanks, Balaji

If it is a small change, prefer to do it as part of this PR. Otherwise, we can file a separate issue and/or fix it in a subsequent PR.

bashbaug:
We don't say anything about tiling requirements in the external memory extension specification.  Should we?

nikhiljnv:
> We don't say anything about tiling requirements in the external memory extension specification. Should we?

I have filed https://github.com/KhronosGroup/OpenCL-Docs/issues/861 to track the spec discussions and required changes to the spec.

bashbaug:
Removing "focused review" while we figure out https://github.com/KhronosGroup/OpenCL-Docs/issues/861.

lakshmih:
We will update this test as per final resolution of https://github.com/KhronosGroup/OpenCL-Docs/issues/861

nikhiljnv:
@lakshmih 
Is this already taken care by #1676 ? Or do we still need to update this after latest spec changes?

nikhiljnv:
Ping @lakshmih !! Is this still needed? If not, can we close?

PR #1561, opened 2022-11-04
Enqueue fill buffer

gpx1000:


CMakeLists.txt
```diff
@@ -50,6 +50,16 @@ set(CONFORMANCE_SUFFIX "" )
...
 # CL_INCLUDE_DIR - path to dir with OpenCL headers
+find_package(OpenCL REQUIRED)
+if(NOT (CL_INCLUDE_DIR AND CL_LIB_DIR) AND OpenCL_FOUND)
+    set(CL_INCLUDE_DIR OpenCL_INCLUDE_DIRS)
+    set(CL_LIB_DIR OpenCL_LIBRARY)
+    message(STATUS "Using system OpenCL: Include DIRS = ${CL_INCLUDE_DIR}")
+else()
+    message(STATUS "Using user defined OpenCL: Include Dirs = ${CL_INCLUDE_DIR}")
+    message(STATUS "Using user defined OpenCL: lib Dir = ${CL_LIB_DIR}")
+endif()
```
EwanC, line 61: This CMake change looks unrelated to the `clEnqueueFillBuffer` changes, I think it should be a standalone PR if this is a change to the build system you'd like to see.

test_conformance/math_brute_force/binary_double.cpp
```diff
@@ -317,7 +317,16 @@ cl_int Test(cl_uint job_id, cl_uint thread_id, void *data)
         // Fill the result buffer with garbage, so that old results don't carry
         // over
         uint32_t pattern = 0xffffdead;
-        memset_pattern4(out[j], &pattern, buffer_size);
+
+        if ((error = clEnqueueFillBuffer(tinfo->tQueue, tinfo->outBuf[j],
+                                         &pattern, sizeof(pattern), 0,
+                                         buffer_size, 0, NULL, NULL)))
```
bashbaug, line 323: I think the preceding call to clEnqueueMapBuffer and the call to clEnqueueUnmapMemObject immediately following should be removed after adding the calls to clEnqueueFillBuffer because the intialization is being done entirely on the device now, correct?

test_conformance/math_brute_force/binary_two_results_i_double.cpp
```diff
@@ -147,7 +147,16 @@ int TestFunc_DoubleI_Double_Double(const Func *f, MTdata d, bool relaxedMode)
         for (auto j = gMinVectorSizeIndex; j < gMaxVectorSizeIndex; j++)
         {
             uint32_t pattern = 0xffffdead;
-            memset_pattern4(gOut[j], &pattern, BUFFER_SIZE);
+
+            if ((error = clEnqueueFillBuffer(gQueue, gOutBuffer[j],
+                                             &pattern, sizeof(pattern), 0,
+                                             BUFFER_SIZE, 0, NULL, NULL)))
```
bashbaug, line 153: As before, I think the subsequent call to clEnqueueWriteBuffer should be removed now because the initialization is being performed via clEnqueueFillBuffer instead, correct?

bashbaug (approved):
Can you please check that all of the calls to map/unmap or write data to the buffer are properly removed after switching to `clEnqueueFillBuffer`?

bashbaug (changes requested):
Sorry, my previous review should have been tagged "request changes".

CMakeLists.txt
lakshmih: Agree that build changes should be separated out into their own PR

Sreelakshmi Haridas Maruthur
Qualcomm

JohnKesapidesARM:
For replacing  memset with clEnqueueFillbuffer. Could we make this behavior controllable via a command line option? If this is executed on device, on real silicon it might be faster, but a lot of us work on FPGAs. Unless we say there is test coverage benefit this will slow down regression testing a lot.

bashbaug:
My specific issues appear to be fixed - thanks!

We still need to decide what to do about the unrelated build system changes.  I'd recommend moving them into a separate PR unless there's a strong reson to include them here.

lakshmih:
Looks fine except for needing to separate out build changes. 

Sreelakshmi Haridas Maruthur
Qualcomm

CMakeLists.txt
cdavis5e: I've removed it for now.

cdavis5e:
> For replacing memset with clEnqueueFillbuffer. Could we make this behavior controllable via a command line option?

Done. The new `-f` option will perform fills on the host, just as before.

cdavis5e:
Ugh. Is there some way to force the CI to cycle again? The failure of `ubuntu-20.04 gl` was a fluke caused by a random network error.

lakshmih (approved):

bashbaug (approved):
Build system changes are removed - thanks!

bashbaug:

PR #1562, opened 2022-11-07
math_brute_force: Use clKernelWrapper in threaded tests

svenvh:
Simplify code by relying on RAII to free resources.  Reduce code duplication.

This commit only affects tests that use `BuildKernelInfo`, which are the multi-threaded tests.  Another patch will deal with the single-threaded tests, i.e., those using `BuildKernelInfo2`.

### DIFF
--- a/test_conformance/math_brute_force/common.cpp
+++ b/test_conformance/math_brute_force/common.cpp
@@ -85,6 +85,28 @@ void EmitEnableExtension(std::ostringstream &kernel, ParameterType type)
     }
 }
 
+std::string GetBuildOptions(bool relaxed_mode)
+{
+    std::ostringstream options;
+
+    if (gForceFTZ)
+    {
+        options << " -cl-denorms-are-zero";
+    }
+
+    if (gFloatCapabilities & CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT)
+    {
+        options << " -cl-fp32-correctly-rounded-divide-sqrt";
+    }
+
+    if (relaxed_mode)
+    {
+        options << " -cl-fast-relaxed-math";
+    }
+
+    return options.str();
+}
+
 } // anonymous namespace
 
 std::string GetKernelName(int vector_size_index)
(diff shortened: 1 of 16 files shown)


test_conformance/math_brute_force/common.cpp
```diff
@@ -530,3 +552,46 @@ __kernel void )", kernel_name.c_str(), R"((__global RETTYPE_SCALAR* out,
...
+    // one for every thread
+    auto &kernels = info.kernels[vector_size_index];
+    assert(kernels.empty() && "Dirty BuildKernelInfo");
+    kernels.resize(info.threadCount);
+    for (auto &kernel : kernels)
+    {
+        kernel = clCreateKernel(program, kernel_name.c_str(), &error);
+        if (!kernel || error != CL_SUCCESS)
+        {
+            char buffer[2048];
```
EwanC, line 587: I know the old code hardcoded `2048`, but we could take this opportunity to make the code slightly better by querying the log size instead. Something like:
```cpp
size_t log_size;
clGetProgramBuildInfo(program, gDevice, CL_PROGRAM_BUILD_LOG, 0, nullptr, &size);
std::vector<char> buffer(log_size);
clGetProgramBuildInfo(program, gDevice, CL_PROGRAM_BUILD_LOG, log_size, buffer.data(), NULL);
```

EwanC (dismissed):
Looks like a nice improvment, only 1 minor comment

test_conformance/math_brute_force/common.cpp
```diff
@@ -530,3 +552,50 @@ __kernel void )", kernel_name.c_str(), R"((__global RETTYPE_SCALAR* out,
...
+        if (!kernel || error != CL_SUCCESS)
+        {
+            vlog_error("\t\tFAILED -- clCreateKernel() failed: (%d)\n", error);
+            size_t log_size;
+            clGetProgramBuildInfo(program, gDevice, CL_PROGRAM_BUILD_LOG, 0,
+                                  nullptr, &log_size);
+            std::string buffer;
+            buffer.resize(log_size + 1);
+            clGetProgramBuildInfo(program, gDevice, CL_PROGRAM_BUILD_LOG,
+                                  log_size, &buffer[0], NULL);
```
bashbaug, line 594: Definitely not a must-fix since this was in the original code as well, but on our implementation at least the build log would only be useful when the program failed to build and wouldn't be useful when clCreateKernel failed.

bashbaug (approved):
This is a nice cleanup!

test_conformance/math_brute_force/common.cpp
svenvh: Good observation; I suppose we can remove the `clGetProgramBuildInfo` queries here, as the spec says this will "Return the build, compile or link log for `clBuildProgram`, `clCompileProgram` or `clLinkProgram`", i.e., nothing about `clCreateKernel`.

I'll add it to my list of future cleanups.

StuartDBrady:
@bashbaug, do you think this can be merged?  (I gather there are other math test cleanups waiting that are dependent on this one.)

EwanC (approved):

bashbaug:
> do you think this can be merged?

Yup, all good on my end!  This has two reviews now so I'll just merge it.

bcalidas:

svenvh:
> Qualcomm would like some more time to review this change.

@bcalidas this has been merged already.  Please let me know if you discover any problems.

PR #1563, opened 2022-11-07
Test that cloned kernels can't be enqueued before all args are set

callumfare:


lakshmih (approved):
Looks good.

Sreelakshmi Haridas Maruthur
Qualcomm

bashbaug (approved):
LGTM

PR #1567, opened 2022-11-09
relationals: Fix vector component accessor selection

svenvh:
It seems the intent was to use `.xyzw` vector component accessors in
the kernel whenever `vecSize` is 2, 3, or 4.  The condition was wrong
(i.e., always true for any value of `vecSize`), causing it to always
create a kernel with numeric (`.s0123...`) vector component accessors.

The restriction on the use of `.xyzw` for 8 and 16-component vectors
has been lifted a while ago (see
https://github.com/KhronosGroup/OpenCL-Docs/issues/549) so we do not
need this check at all.

Also, fix the `vecSize` argument of two calls of `get_order_string`
when getting the order string for the "in" kernel argument.  Pass
`inVecSize` in these cases, not `outVecSize`.

### DIFF
--- a/test_conformance/relationals/test_shuffles.cpp
+++ b/test_conformance/relationals/test_shuffles.cpp
@@ -216,11 +216,8 @@ const char *get_order_string( ShuffleOrder &order, size_t vecSize, cl_uint lengt
     if (!byNumber) {
         byNumber = (useNumbersFlip++)%2;
     }
-    // Do not use xyzw for vectors whose length is not 2 or 4 per the spec.
-    if (vecSize != 2 || vecSize != 4 || vecSize != 3)
-        byNumber = 1;
 
-    if( byNumber || vecSize > 4 )
+    if (byNumber)
     {
         idx = 0;
         // Randomly chose upper and lower case S
@@ -251,7 +248,9 @@ char * get_order_name( ExplicitType vecType, size_t inVecSize, size_t outVecSize
     if( inVecSize == 1 )
         inOrderStr[ 0 ] = 0;
     else
-        sprintf( inOrderStr, "%d.%s", (int)inVecSize, get_order_string( inOrder, outVecSize, lengthToUse, inUseNumerics, d ) );
+        sprintf(inOrderStr, "%d.%s", (int)inVecSize,
+                get_order_string(inOrder, inVecSize, lengthToUse, inUseNumerics,
+                                 d));
     if( outVecSize == 1 )
         outOrderStr[ 0 ] = 0;
     else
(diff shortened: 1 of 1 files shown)


test_conformance/relationals/test_shuffles.cpp
```diff
@@ -217,8 +217,10 @@ const char *get_order_string( ShuffleOrder &order, size_t vecSize, cl_uint lengt
         byNumber = (useNumbersFlip++)%2;
     }
     // Do not use xyzw for vectors whose length is not 2 or 4 per the spec.
-    if (vecSize != 2 || vecSize != 4 || vecSize != 3)
+    if (vecSize != 2 && vecSize != 4 && vecSize != 3)
```
EwanC, line 220: Would say that `if (vecSize < 2 && vecSize > 4)` is slightly more readable, but it's a bit of a personal preference.

bashbaug:
Should this condition be removed entirely?

https://github.com/KhronosGroup/OpenCL-Docs/issues/549

test_conformance/relationals/test_shuffles.cpp
svenvh: I went for the minimal fix; would have applied your suggestion but following Ben's comment I believe we can drop this altogether.

svenvh:
> Should this condition be removed entirely?
> 
> [KhronosGroup/OpenCL-Docs#549](https://github.com/KhronosGroup/OpenCL-Docs/issues/549)

Nice catch, I think we can.  I'll update the patch to remove the `if` altogether.

svenvh:
> I'll update the patch to remove the if altogether.

Done.  I then noticed that the `vecSize` arguments were wrong in some cases, so have fixed that in the same commit.

EwanC (dismissed):

lakshmih:

lakshmih (approved):
Looks good

Sreelakshmi Haridas Maruthur
Qualcomm

test_conformance/relationals/test_shuffles.cpp
```diff
@@ -216,9 +216,6 @@ const char *get_order_string( ShuffleOrder &order, size_t vecSize, cl_uint lengt
     if (!byNumber) {
         byNumber = (useNumbersFlip++)%2;
     }
-    // Do not use xyzw for vectors whose length is not 2 or 4 per the spec.
-    if (vecSize != 2 || vecSize != 4 || vecSize != 3)
-        byNumber = 1;
 
     if( byNumber || vecSize > 4 )
```
bashbaug, line 220: I think if we really want to test all vector sizes then we should remove this `vecSize > 4` condition also.  We can do this safely because we've already checked for indices greater than three, above.

(Aside: you didn't break this either, but it would be nice to remove the incorrect second "randomly choose upper and lower case" comment below, because the .xyzw selection is exclusively lower-case.)

bashbaug (approved):
Mostly LGTM, this is a definite improvement.  One more possible comment to consider.

test_conformance/relationals/test_shuffles.cpp
svenvh: > I think if we really want to test all vector sizes then we should remove this vecSize > 4 condition also.

Yes that seems safe; I've made the change and tested it.

> (Aside: you didn't break this either, but it would be nice to remove the incorrect second "randomly choose upper and lower case" comment below, because the .xyzw selection is exclusively lower-case.)

Yes I noticed that misleading comment too and already have a followup change to tidy up the comments in this function.  I'll push that after this PR has landed.

bashbaug (approved):

EwanC (approved):

bashbaug:
Has two approvals and another thumbs up, merging.

PR #1569, opened 2022-11-10
Image2d_from_buffer_positive comparison fix

JohnKesapidesARM:
Inverted the comparisons as we expect the alignments reported from the 
extension query to always be greater or equal to what is reported from 
the device in pixels through 
CL_DEVICE_IMAGE_PITCH_ALIGNMENT/CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT.

### DIFF
--- a/test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
+++ b/test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
@@ -18,6 +18,8 @@
 #include "../common.h"
 #include "test_cl_ext_image_buffer.hpp"
 
+static inline bool is_power_of_two(size_t num) { return !(num & (num - 1)); }
+
 static int get_image_requirement_alignment(
     cl_device_id device, cl_context context, cl_mem_flags flags,
     const cl_image_format* image_format, const cl_image_desc* image_desc,
@@ -79,11 +81,17 @@ int image2d_from_buffer_positive(cl_device_id device, cl_context context,
         return TEST_SKIPPED_ITSELF;
     }
 
-    std::vector<cl_mem_object_type> imageTypes{
-        CL_MEM_OBJECT_IMAGE1D,       CL_MEM_OBJECT_IMAGE2D,
-        CL_MEM_OBJECT_IMAGE3D,       CL_MEM_OBJECT_IMAGE1D_BUFFER,
-        CL_MEM_OBJECT_IMAGE1D_ARRAY, CL_MEM_OBJECT_IMAGE2D_ARRAY
-    };
+    cl_uint row_pitch_alignment_2d = 0;
+    cl_int err = clGetDeviceInfo(device, CL_DEVICE_IMAGE_PITCH_ALIGNMENT,
+                                 sizeof(row_pitch_alignment_2d),
+                                 &row_pitch_alignment_2d, nullptr);
+    test_error(err, "Error clGetDeviceInfo");
+
+    cl_uint base_address_alignment_2d = 0;
+    err = clGetDeviceInfo(device, CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT,
+                          sizeof(base_address_alignment_2d),
+                          &base_address_alignment_2d, nullptr);
+    test_error(err, "Error clGetDeviceInfo");
 
     std::vector<cl_mem_flags> flagTypes{ CL_MEM_READ_ONLY, CL_MEM_WRITE_ONLY,
                                          CL_MEM_READ_WRITE,
@@ -91,55 +99,43 @@ int image2d_from_buffer_positive(cl_device_id device, cl_context context,
 
     for (auto flagType : flagTypes)
     {
-        for (auto imageType : imageTypes)
+
+        /* Get the list of supported image formats */
+        std::vector<cl_image_format> formatList;
+        if (TEST_PASS
+                != get_format_list(context, CL_MEM_OBJECT_IMAGE2D, formatList,
+                                   flagType)
+            || formatList.size() == 0)
         {
-            /* Get the list of supported image formats */
-            std::vector<cl_image_format> formatList;
-            if (TEST_PASS
-                    != get_format_list(context, imageType, formatList, flagType)
-                || formatList.size() == 0)
+            test_fail("Failure to get supported formats list\n");
+        }
+
+        for (auto format : formatList)
+        {
+            cl_image_desc image_desc = { 0 };
+            image_desc_init(&image_desc, CL_MEM_OBJECT_IMAGE2D);
+
+            cl_mem_flags flag = (flagType == CL_MEM_KERNEL_READ_AND_WRITE)
+                ? CL_MEM_READ_WRITE
+                : flagType;
+
+            size_t row_pitch_alignment = 0;
+            size_t base_address_alignment = 0;
+
+            int get_error = get_image_requirement_alignment(
+                device, context, flag, &format, &image_desc,
+                &row_pitch_alignment, nullptr, &base_address_alignment);
+            if (TEST_PASS != get_error)
             {
-                test_fail("Failure to get supported formats list\n");
+                return get_error;
             }
 
-            cl_uint row_pitch_alignment_2d = 0;
-            cl_int err =
-                clGetDeviceInfo(device, CL_DEVICE_IMAGE_PITCH_ALIGNMENT,
-                                sizeof(row_pitch_alignment_2d),
-                                &row_pitch_alignment_2d, nullptr);
-            test_error(err, "Error clGetDeviceInfo");
+            const size_t element_size =
+                get_format_size(context, &format, CL_MEM_OBJECT_IMAGE2D, flag);
 
-            cl_uint base_address_alignment_2d = 0;
-            err =
-                clGetDeviceInfo(device, CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT,
-                                sizeof(base_address_alignment_2d),
-                                &base_address_alignment_2d, nullptr);
-            test_error(err, "Error clGetDeviceInfo");
-
-            for (auto format : formatList)
+            if (is_power_of_two(element_size))
             {
-                cl_image_desc image_desc = { 0 };
-                image_desc_init(&image_desc, imageType);
-
-                cl_mem_flags flag = (flagType == CL_MEM_KERNEL_READ_AND_WRITE)
-                    ? CL_MEM_READ_WRITE
-                    : flagType;
-
-                size_t row_pitch_alignment = 0;
-                size_t base_address_alignment = 0;
-
-                int get_error = get_image_requirement_alignment(
-                    device, context, flag, &format, &image_desc,
-                    &row_pitch_alignment, nullptr, &base_address_alignment);
-                if (TEST_PASS != get_error)
-                {
-                    return get_error;
-                }
-
-                const size_t element_size =
-                    get_format_size(context, &format, imageType, flag);
-
-                /*  Alignements in pixels vs bytes */
+                /*  Alignments in pixels vs bytes */
                 if (base_address_alignment
                     > base_address_alignment_2d * element_size)
                 {


lakshmih:

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
```diff
@@ -139,14 +139,14 @@ int image2d_from_buffer_positive(cl_device_id device, cl_context context,
...
+                    < base_address_alignment_2d * element_size)
                 {
                     test_fail("Unexpected base_address_alignment\n");
                 }
 
-                if (row_pitch_alignment > row_pitch_alignment_2d * element_size)
+                if (row_pitch_alignment < row_pitch_alignment_2d * element_size)
                 {
                     test_fail("Unexpected row_pitch_alignment\n");
                 }
```
bashbaug, line 152: Can you say a little more what these conditions are designed to check for?

As far as I can tell these are checking the values queried for `CL_IMAGE_REQUIREMENTS_BASE_ADDRESS_ALIGNMENT_EXT` and `CL_IMAGE_REQUIREMENTS_ROW_PITCH_ALIGNMENT_EXT` for all image types (meaning: `1D`, `2D`, `3D`, etc.), then checking against the values queried for `CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT` and `CL_DEVICE_IMAGE_PITCH_ALIGNMENT`, which are implicitly for 2D images.  Are these two values comparable in the general case?  Or can we only compare the values queried for 2D images?

JohnKesapidesARM: That's a good question. Let me discuss this internally, because I wasn't involved on the implementation, so I may be missing context.

bashbaug: I'm going to take off "focused review" while this is under discussion.  Feel free to add it back when this is ready to go again - thanks!

bcalidas:
We think this comparison is flawed and should be removed altogether. The root cause of this issue is that CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT returns the alignment in pixels which is highly problematic. It should have returned the alignment in bytes which is how most implementations work. Since there is no information about the actual image format in this query implementations have to make very conservative assumptions.  These assumptions will not always match up with CL_IMAGE_REQUIREMENTS_BASE_ADDRESS_ALIGNMENT_EXT (after conversion to bytes) which is specific to the image format. It is probably too late to correct CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT in the spec, but I think the solution going forward will be to add clarifications to the image_requirements_info_ext spec.

Thanks,
Balaji Calidas
Qualcomm

JohnKesapidesARM:
> We think this comparison is flawed and should be removed altogether. The root cause of this issue is that CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT returns the alignment in pixels which is highly problematic. It should have returned the alignment in bytes which is how most implementations work. Since there is no information about the actual image format in this query implementations have to make very conservative assumptions. These assumptions will not always match up with CL_IMAGE_REQUIREMENTS_BASE_ADDRESS_ALIGNMENT_EXT (after conversion to bytes) which is specific to the image format. It is probably too late to correct CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT in the spec, but I think the solution going forward will be to add clarifications to the image_requirements_info_ext spec.
> 
> Thanks, Balaji Calidas Qualcomm
I don't disagree about the unnecessary restrictions imposed by the legacy queries being in pixels, but since we have to adhere to the restrictions the check the test is doing should hold (unless I misunderstood your point).  I recommend we merge the fix for now so the test works assuming the spec as is, and move the discussions higher into the workgroup.

JohnKesapidesARM:
@bashbaug looks like I don't have permissions to add the focused review label back on. Can we resume the review please?

JohnKesapidesARM:
why was the focused review label removed from this one?

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
```diff
@@ -79,77 +79,69 @@ int image2d_from_buffer_positive(cl_device_id device, cl_context context,
         return TEST_SKIPPED_ITSELF;
     }
 
-    std::vector<cl_mem_object_type> imageTypes{
-        CL_MEM_OBJECT_IMAGE1D,       CL_MEM_OBJECT_IMAGE2D,
-        CL_MEM_OBJECT_IMAGE3D,       CL_MEM_OBJECT_IMAGE1D_BUFFER,
-        CL_MEM_OBJECT_IMAGE1D_ARRAY, CL_MEM_OBJECT_IMAGE2D_ARRAY
-    };
+    cl_mem_object_type imageType = CL_MEM_OBJECT_IMAGE2D;
```
bashbaug, line 82: Minor: since this value never changes I think it help understanding if it were "const".

Or, even better, just remove it and pass around CL_MEM_OBJECT_IMAGE2D explicitly.

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
```diff
@@ -79,77 +79,69 @@ int image2d_from_buffer_positive(cl_device_id device, cl_context context,
...
-            err =
-                clGetDeviceInfo(device, CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT,
-                                sizeof(base_address_alignment_2d),
-                                &base_address_alignment_2d, nullptr);
-            test_error(err, "Error clGetDeviceInfo");
+        cl_uint base_address_alignment_2d = 0;
+        err = clGetDeviceInfo(device, CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT,
+                              sizeof(base_address_alignment_2d),
+                              &base_address_alignment_2d, nullptr);
+        test_error(err, "Error clGetDeviceInfo");
```
bashbaug, line 110: Minor, these queries dont' depend on the "flagType", so they could be moved outside of the loop.

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
```diff
@@ -79,77 +79,69 @@ int image2d_from_buffer_positive(cl_device_id device, cl_context context,
...
-                    > base_address_alignment_2d * element_size)
-                {
-                    test_fail("Unexpected base_address_alignment\n");
-                }
+            /*  Alignments in pixels vs bytes */
+            if (base_address_alignment
+                < base_address_alignment_2d * element_size)
+            {
+                test_fail("Unexpected base_address_alignment\n");
+            }
```
bashbaug, line 140: At least according to my understanding this check is inverted, and I think this should be failing if the queried base_address_alignment is **greater than** the generic base_address_aligment_2d.  This is why the image requirements info extension is valuable - instead of having a single, generic per-device query that must return the worst-case scenario an application can instead query the specific requirements for an image.

```suggestion
            if (base_address_alignment
                > base_address_alignment_2d * element_size)
            {
                test_fail("Unexpected base_address_alignment\n");
            }
```

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
```diff
@@ -79,77 +79,69 @@ int image2d_from_buffer_positive(cl_device_id device, cl_context context,
...
 
-                if (row_pitch_alignment > row_pitch_alignment_2d * element_size)
-                {
-                    test_fail("Unexpected row_pitch_alignment\n");
-                }
+            if (row_pitch_alignment < row_pitch_alignment_2d * element_size)
+            {
+                test_fail("Unexpected row_pitch_alignment\n");
```
bashbaug, line 144: This check appears to be inverted as well, for the same reason.

```suggestion
            if (row_pitch_alignment > row_pitch_alignment_2d * element_size)
            {
                test_fail("Unexpected row_pitch_alignment\n");
```

bashbaug (commented):
Thanks, the change to only query image requirements for 2D images looks good.

I have a few other comments and suggestions now.

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
JohnKesapidesARM: will do thank you

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
JohnKesapidesARM: thank I'll amend the code

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
JohnKesapidesARM: hmmm. It looks like I misunderstood the extension, thank you for clearing it up. But with this interpretation it looks like @bcalidas was correct in pointing out that the comparison is flawed to begin with. 
The issue is that base_address_alignment returned by the extension is in bytes and has to be a power of 2. The right-hand part of the comparison is a calculation yielding a value not round up to a power of 2.
So say for a device where base_address_alignment_2d=32 pixels
for this format  ( CL_RGB, CL_UNORM_INT8) for instance 
* The extension could return  base_address_alignment = round_up_pow_of_2(32*3) = 128 
* The generic device would return base_address_alignment_2d * element_size=32*3 = 96

Should I round up (base_address_alignment_2d * element_size) to the closest power of 2? or drop the comparison

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
JohnKesapidesARM: same question as above?

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
bashbaug: Maybe to level set what we are comparing:

`base_address_alignment_2d` is the result of the query for [`CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT`](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT), which is:

> This query specifies the minimum alignment in pixels of the host_ptr specified to clCreateBuffer or clCreateBufferWithProperties when a 2D image is created from a buffer which was created using CL_MEM_USE_HOST_PTR. The value returned must be a power of 2.

`base_address_alignment` is the result of the query for [`CL_IMAGE_REQUIREMENTS_BASE_ADDRESS_ALIGNMENT_EXT`](https://registry.khronos.org/OpenCL/extensions/ext/cl_ext_image_requirements_info.html#clGetImageRequirementsInfoEXT), which is:

> Returns the minimum alignment in bytes required for the data store backing an image created using the parameters passed to clGetImageRequirementsInfoEXT. image_format or image_desc are allowed to be NULL. When either or both is NULL the implementation will return an alignment that would be sufficient for all possible values of the missing argument.
The value returned is a power of two.

So, both queries have a power of two requirement, but one is in units of pixels and another is in units of bytes.  I could see how this might be a problem for image formats like `CL_RGB` where the image element size is not a power of two.

@kpet, I don't think this was intended.  Is the fix to clarify that the computed alignment using the core `CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT` query must be a power of two after converting from pixels to bytes?  The extension text seems to be saying what we want it to say, so I don't think any changes would occur there.

@JohnKesapidesARM, if you need a short-term fix while we work this out could we perhaps perform this check only when the element size is a power of two?  I think the only questionable cases are when the element size is non-power of two.

JohnKesapidesARM: @bashbaug I have updated the test based on the comments. The test is now only doing the checks checking when element_size is a power of two. 
Unfortunately, I had to force push the update to amend the commit message.

bashbaug (approved):
LGTM, we should file an issue so we don't lose track of the non-power-of-two image element size issue.  Thanks!

JohnKesapidesARM:
> LGTM, we should file an issue so we don't lose track of the non-power-of-two image element size issue. Thanks!
Issue raised https://github.com/KhronosGroup/OpenCL-CTS/issues/1832

JohnKesapidesARM:
@bashbaug Can we please merge this one?

lakshmih (approved):

bashbaug:
Merging as discussed in the November 7th teleconference.

Issue #1198, opened 2021-03-26
Subgroups tests - adjust test non_uniform_broadcast and broadcast_first to use mask for active workitems


gwawiork:
* Remove variable NR_OF_ACTIVE_WORK_ITEMS
* use masks as a pattern for active workitems (like other ballot functions)

karolherbst:
Hey, I tried to implement non-uniform workgroups and was running into this issue as well.

And I simply fixed it here: https://github.com/KhronosGroup/OpenCL-CTS/pull/2709

PR #1574, opened 2022-11-18
Fix sub_group_broadcast_first expected value

### LINKED ISSUE #1573: sub_group_broadcast_first failures
I'm finding that our platform is failing all the `sub_group_broadcast_first` tests for work items that have  
`get_sub_group_local_id() >= NR_OF_ACTIVE_WORK_ITEMS`.
These work items have the same result as the lowest active instance (0) in their subgroup, which seems like the correct result to me, based on the shader source from test_subgroup_ballot.cpp:
```
__kernel void test_sub_group_broadcast_first(const __global Type *in, __global int4 *xy, __global Type *out) {
    int gid = get_global_id(0);
    XY(xy,gid);
    Type x = in[gid];
    if (xy[gid].x < NR_OF_ACTIVE_WORK_ITEMS) {
        out[gid] = sub_group_broadcast_first(x);;
    } else {
        out[gid] = sub_group_broadcast_first(x);;
    }
}
```
but that's not what the CTS is checking.

I find that the following change to the CTS causes our platform to pass these tests, and it seems like a correct change to me.

```
$ git diff
diff --git a/test_conformance/subgroups/subgroup_common_templates.h b/test_conformance/subgroups/subgroup_common_templates.h
index b2648c3..1e73b60 100644
--- a/test_conformance/subgroups/subgroup_common_templates.h
+++ b/test_conformance/subgroups/subgroup_common_templates.h
@@ -221,9 +221,7 @@ template <typename Ty, SubgroupsBroadcastOp operation> struct BC
                     for (i = 0; i < n; ++i)
                     {
 
-                        lowest_active_id = i < NR_OF_ACTIVE_WORK_ITEMS
-                            ? 0
-                            : NR_OF_ACTIVE_WORK_ITEMS;
+                        lowest_active_id = 0;
                         //  findout if broadcasted
                         //  value is the same
                         tr = mx[ii + lowest_active_id];
```

The worrying thing is that other platforms must be passing these CTS tests as they are, and I don't understand how that can be possible, so maybe I am misunderstanding something in the CTS.

stevec611:
All subgroup members should return the value of the first member.

Fixes #1573

### DIFF
--- a/test_conformance/subgroups/subgroup_common_templates.h
+++ b/test_conformance/subgroups/subgroup_common_templates.h
@@ -217,13 +217,9 @@ template <typename Ty, SubgroupsBroadcastOp operation> struct BC
                 // Check result
                 if (operation == SubgroupsBroadcastOp::broadcast_first)
                 {
-                    int lowest_active_id = -1;
+                    int lowest_active_id = 0;
                     for (i = 0; i < n; ++i)
                     {
-
-                        lowest_active_id = i < NR_OF_ACTIVE_WORK_ITEMS
-                            ? 0
-                            : NR_OF_ACTIVE_WORK_ITEMS;
                         //  findout if broadcasted
                         //  value is the same
                         tr = mx[ii + lowest_active_id];


StuartDBrady:
This may well make the test pass, but it is not sufficient to test that values are broadcast only from work items with a subgroup local ID of 0.

If the work item with subgroup local ID 0 within the subgroup does not participate in the _sub_group_non_uniform_broadcast_first_ operation, then the value from the work item with subgroup local ID will be used, assuming it does participate.  This is the entire point of the function – otherwise, you'd use _sub_group_broadcast_(_x_, 0).

_sub_group_broadcast_first_(_x_) is equivalent to _sub_group_broadcast_(_x_, _sub_group_ballot_find_lsb_(_sub_group_ballot_(1))).

We should be counting trailing bits set in _test_params_._work_items_mask_, and passing in a divergence mask vector to the test kernel, as per the testing for _sub_group_elect_ in _test_subgroup_non_uniform_vote.cpp_.

stevec611:
Are you agreeing that the test, in it's current state, is broken?

StuartDBrady:
> Are you agreeing that the test, in it's current state, is broken?

No, I'm afraid I am not.

Please read _sub_group_broadcast_first_source_, and you will see that it performs two ballot operations: one with work items with subgroup local IDs 0..3 active, and a second one with the remaining work items active.  The logic for determining _lowest_active_id_ is consistent with this.

stevec611:
>Please read sub_group_broadcast_first_source, and you will see that it performs two ballot operations: one with work items with subgroup local IDs 0..3 active, and a second one with the remaining work items active.

Ah ok, thank you for clearing up my misunderstanding of the test.

StuartDBrady:
Note that in its current form, this test checks that:

* for GPUs with a subgroup size of 4 or less, _sub_group_broadcast_first_(_x_) behaves in the same way as _sub_group_broadcast_(_x_, 0) when all work items in the subgroup are active
* for GPUs with a subgroup size greater than 4, _sub_group_broadcast_first_(_x_) behaves in the same way as _sub_group_broadcast_(_x_, 0) when the low four work items in the subgroup are active, and _sub_group_broadcast_first_(_x_) behaves in the same way as _sub_group_broadcast_(_x_, 4) when all work items _except_ the low four are active within the subgroup.

StuartDBrady:
For anyone who may need it, see also:
* Issue #1198 – for sub_group_non_uniform_broadcast and sub_group_broadcast_first.
* Issue #1206 – for the more general issue of the (lack of) appropriateness of NR_OF_ACTIVE_WORK_ITEMS.

PR #1575, opened 2022-11-21
Fix generic address space OpenCL 2.0 assumption

kbenzie:
Conformance tests should not assume support for OpenCL 2.0, or specific features. This patch removes uses of
`__opencl_c_program_scope_global_variables` for devices which do not support it in the generic address space conformance tests. Additionally, the `clCompileProgram` options have been updated to select between `CL2.0` and `CL3.0` depending on which the device under test supports.

### DIFF
--- a/test_conformance/generic_address_space/advanced_tests.cpp
+++ b/test_conformance/generic_address_space/advanced_tests.cpp
@@ -107,7 +107,32 @@ public:
             program = clCreateProgramWithSource(context, 1, &srcPtr, NULL, &error);
             test_error(error, "clCreateProgramWithSource failed");
 
-            error = clCompileProgram(program, 1, &deviceID, "-cl-std=CL2.0", 0, NULL, NULL, NULL, NULL);
+            // Use the latest OpenCL-C version supported by the device. This
+            // allows calling code to force a particular CL C version if it is
+            // required, but also means that callers need not specify a version
+            // if they want to assume the most recent CL C.
+
+            auto version = get_max_OpenCL_C_for_context(context);
+
+            const char* cl_std = nullptr;
+            if (version >= Version(3, 0))
+            {
+                cl_std = "-cl-std=CL3.0";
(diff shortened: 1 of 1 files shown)


test_conformance/generic_address_space/advanced_tests.cpp
```diff
@@ -276,31 +304,24 @@ int test_library_function(cl_device_id deviceID, cl_context context, cl_command_
...
-        NL "    failures += helperFunction(lfloatp, lfloat) ? 0 : 1;"
-        NL "    failures += helperFunction(pfloatp, pfloat) ? 0 : 1;"
-        NL
-        NL "    results[tid] = failures == 0;"
-        NL "}"
-        NL;
+    const std::string KERNEL_FUNCTION = NL NL
+        "extern bool helperFunction(float *floatp, float val);" NL NL
+        "#ifdef __opencl_c_program_scope_global_variables" NL
+        "__global float gfloat = 1.0f;" NL "#endif" NL NL
```
svenvh, line 310: The formatting changes don't make this very readable now.  I think you should be able to keep the original layout if you use a C++ raw string literal (and then you won't need `NL` any longer).

lakshmih:

bashbaug (approved):
At some point it would be nice to factor the code to create a program with source and compile it via `clCompileProgram` into a separate helper function, but I think this is fine for now.

Note that there are not a lot of other places where we call `clCompileProgram` and pass a specific OpenCL C version right now.

test_conformance/generic_address_space/advanced_tests.cpp
kbenzie: I've made the change to raw string literally for this kernel.

svenvh (approved):

lakshmih (approved):
Looks good

Sreelakshmi Haridas Maruthur
Qualcomm

PR #1577, opened 2022-11-23
Partial clean up of test_compiler_defines_for_extensions

StuartDBrady:
Replace repetitions of extensions_supported[num_of_supported_extensions] with a single pointer variable.

Replace character-by-character comparisons with a new string_has_prefix function.

### DIFF
--- a/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
+++ b/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
@@ -117,6 +117,11 @@ const char *kernel_strings[] = {
     "}\n"
 };
 
+bool string_has_prefix(const char *str, const char *prefix)
+{
+    return strncmp(str, prefix, strlen(prefix)) == 0;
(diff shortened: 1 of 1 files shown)


StuartDBrady:
I am of the opinion that this should code should be converted over to use C++ strings, vectors, etc. in due course, and that the log messages could be improved, but I have decided to limit this change to making better use of C strings for now, so that it is easier to review.

StuartDBrady:
Note that the strictly speaking, the comparison `memcmp(extensions_supported[num_of_supported_extensions], "cles_khr_", 9)` could go out of bounds, if the extension name is allocated towards the the end of a page and is shorter than "cles_khr_" – this is _very_ unlikely, to the point of being mostly theoretical, so I have chosen not to mention this in the commit message.

EwanC (approved):

bcalidas:

bashbaug:

PR #1583, opened 2022-11-25
Fix -Wreorder warnings

svenvh:
Initialize class members in the order in which they are declared in the class.

### DIFF
--- a/test_conformance/basic/test_progvar.cpp
+++ b/test_conformance/basic/test_progvar.cpp
@@ -81,20 +81,20 @@ class TypeInfo {
 
 public:
     TypeInfo()
-        : name(""), m_buf_elem_type(""), m_is_vecbase(false),
+        : name(""), m_elem_type(0), m_num_elem(0), m_is_vecbase(false),
           m_is_atomic(false), m_is_like_size_t(false), m_is_bool(false),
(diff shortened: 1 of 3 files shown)


EwanC:
Can we add `add_cxx_flag_if_supported(-Wreorder)` to the CMake to catch this going forward?

svenvh:
> Can we add `add_cxx_flag_if_supported(-Wreorder)` to the CMake to catch this going forward?

Surely we could, but I'm hoping to enable `-Wall` soon (and I don't expect many new `Wreorder` errors to be introduced before then as new classes aren't being introduced that often).

I can still add it to CMakeLists if you wish, but in other PRs that fixed warnings I've only done so if I felt that new errors could be introduced easily (such as unused variables).

EwanC (approved):

EwanC:
> Surely we could, but I'm hoping to enable -Wall soon (and I don't expect many new Wreorder errors to be introduced before then as new classes aren't being introduced that often).
>
> I can still add it to CMakeLists if you wish, but in other PRs that fixed warnings I've only done so if I felt that new errors could be introduced easily (such as unused variables).

Sound sensible, let's leave out modifying the CMake in this PR then

bashbaug:

Issue #1369, opened 2021-12-17
Command-buffer extension testing


EwanC:
In order for the [cl_khr_command_buffer extension](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_command_buffer) to become fully ratified from its provisional status it needs CTS testing. This issue is a plan of areas that need to be covered to fully test the extension. By having this breakdown, we should be able to collaborate on the tasks needing to be done.

## Device Info Queries

Verify the following queries can be performed successfully and return
defined values:

- [x] `CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR`
- [x] `CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR`

EDIT: Done in test fixture

### 1.1 Individual command entry-points

For each command which can be recorded into a command-buffer, check that the
follow steps can be performed successfully:

* Create a new command-buffer targeting a single queue
* Record <command>
* Finalize
* Enqueue
* clFinish & verify result.

- [x] `clCommandBarrierWithWaitListKHR`
- [x] `clCommandCopyBufferKHR`
- [x] `clCommandCopyBufferRectKHR`
- [x] `clCommandCopyBufferToImageKHR`
- [x] `clCommandCopyImageKHR`
- [x] `clCommandCopyImageToBufferKHR`
- [x] `clCommandFillBufferKHR`
- [x] `clCommandFillImageKHR`
- [x] `clCommandNDRangeKernelKHR`

EDIT: Done in https://github.com/KhronosGroup/OpenCL-CTS/pull/1594

### 1.2 Interleave command-buffer enqueues with regular enqueues

> *Only* when the device supports optional Simultaneous Use

- [x] Test case for interleaving command-buffer enqueues with enqueues of core
      OpenCL commands:

* Create a new in-order command-buffer targeting a single queue
* Record `clCommandNDRangeKernel` command
* Finalize
* Enqueue command-buffer
* Enqueue an `clEnqueueNDRangeKernel` operation to the same queue
* Enqueue command-buffer again
* clFinish & verify through result that ordering of enqueues was preserved.

 **DONE**: Scenario tested by `InterleavedEnqueueTest` in https://github.com/KhronosGroup/OpenCL-CTS/pull/1368

#### 1.3.1 

- [x] substitution on queue without properties 

* Create a command-buffer recorded to a single queue with no (or required) properties
[...]

PR #1584, opened 2022-11-28
Command buffer queue substitution

shajder:
Introduce series of queue substitution tests for [cl_khr_command_buffer](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_command_buffer) extension according to point 1.3 of [issue](https://github.com/KhronosGroup/OpenCL-CTS/issues/1369) description.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
@@ -13,174 +13,131 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#include "command_buffer_test_base.h"
+#include "basic_command_buffer.h"
 #include "procs.h"
-#include "harness/typeWrappers.h"
 
 #include <algorithm>
 #include <cstring>
 #include <vector>
 
-#define ADD_PROP(prop)                                                         \
-    {                                                                          \
-        prop, #prop                                                            \
+
+BasicCommandBufferTest::BasicCommandBufferTest(cl_device_id device, cl_context context,
+                       cl_command_queue queue)
+    : CommandBufferTestBase(device), context(context), queue(queue),
+      num_elements(0),
+      command_buffer(this),
+      simultaneous_use(false),
+      out_of_order_support(false),
+      simultaneous_use_requested(true), // try to use simultaneous path by default
+      double_buffers_size(false) // due to simultaneous case extend buffer size
+
+{}
+
+bool BasicCommandBufferTest::Skip()
+{
+    cl_command_queue_properties required_properties;
+    cl_int error = clGetDeviceInfo(
+        device, CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR,
+        sizeof(required_properties), &required_properties, NULL);
+    test_error(error,
+               "Unable to query "
+               "CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR");
+
+    cl_command_queue_properties queue_properties;
+
+    error = clGetCommandQueueInfo(queue, CL_QUEUE_PROPERTIES,
+                                  sizeof(queue_properties),
+                                  &queue_properties, NULL);
+    test_error(error, "Unable to query CL_QUEUE_PROPERTIES");
+
+    // Skip if queue properties don't contain those required
+    return required_properties != (required_properties & queue_properties);
+}
+
+cl_int BasicCommandBufferTest::SetUp(int elements)
+{
+    cl_int error = init_extension_functions();
+    if (error != CL_SUCCESS)
+    {
+        return error;
     }
 
-#define CHECK_VERIFICATION_ERROR(reference, result, index)                     \
-    {                                                                          \
-        if (reference != result)                                               \
-        {                                                                      \
-            log_error("Expected %d was %d at index %u\n", reference, result,   \
-                      index);                                                  \
-            return TEST_FAIL;                                                  \
-        }                                                                      \
+    // Query if device supports simultaneous use
+    cl_device_command_buffer_capabilities_khr capabilities;
+    error =
+        clGetDeviceInfo(device, CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR,
+                        sizeof(capabilities), &capabilities, NULL);
+    test_error(error,
+               "Unable to query CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR");
+    simultaneous_use = simultaneous_use_requested &&
+        (capabilities & CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR) != 0;
+    out_of_order_support =
+        capabilities & CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR;
+
+    if (elements <= 0)
+    {
+        return CL_INVALID_VALUE;
     }
+    num_elements = static_cast<size_t>(elements);
+
+    // Kernel performs a parallel copy from an input buffer to output buffer
+    // is created.
+    const char *kernel_str =
+        R"(
+    __kernel void copy(__global int* in, __global int* out, int offset) {
+        size_t id = get_global_id(0);
+        int ind = offset + id;
+        out[ind] = in[ind];
+    })";
+
+    error = create_single_kernel_helper_create_program(context, &program, 1,
+                                                       &kernel_str);
+    test_error(error, "Failed to create program with source");
+
+    error = clBuildProgram(program, 1, &device, nullptr, nullptr, nullptr);
+    test_error(error, "Failed to build program");
+
+    in_mem = clCreateBuffer(context, CL_MEM_READ_ONLY,
+        sizeof(cl_int) * num_elements * (double_buffers_size?2:1), nullptr, &error);
+    test_error(error, "clCreateBuffer failed");
+
+    out_mem = clCreateBuffer(context, CL_MEM_WRITE_ONLY,
+        sizeof(cl_int) * num_elements * (double_buffers_size?2:1), nullptr, &error);
+    test_error(error, "clCreateBuffer failed");
+
+    kernel = clCreateKernel(program, "copy", &error);
+    test_error(error, "Failed to create copy kernel");
+
+    error = clSetKernelArg(kernel, 0, sizeof(in_mem), &in_mem);
+    test_error(error, "clSetKernelArg failed");
+
+    error = clSetKernelArg(kernel, 1, sizeof(out_mem), &out_mem);
+    test_error(error, "clSetKernelArg failed");
+
+    const cl_int offset=0;
+    error = clSetKernelArg(kernel, 2, sizeof(cl_int), &offset);
+    test_error(error, "clSetKernelArg failed");
+
+    if (simultaneous_use)
+    {
+        cl_command_buffer_properties_khr properties[3] = {
+            CL_COMMAND_BUFFER_FLAGS_KHR,
+            CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR, 0
+        };
+        command_buffer =
+            clCreateCommandBufferKHR(1, &queue, properties, &error);
+    }
+    else
+    {
+        command_buffer =
+            clCreateCommandBufferKHR(1, &queue, nullptr, &error);
+    }
+    test_error(error, "clCreateCommandBufferKHR failed");
+
+    return CL_SUCCESS;
+}
 
 namespace {
-
-// Helper test fixture for constructing OpenCL objects used in testing
-// a variety of simple command-buffer enqueue scenarios.
-struct BasicCommandBufferTest : CommandBufferTestBase
-{
-
-    BasicCommandBufferTest(cl_device_id device, cl_context context,
-                           cl_command_queue queue)
-        : CommandBufferTestBase(device), context(context), queue(queue),
-          num_elements(0),
-          command_buffer(this),
-          simultaneous_use(false),
-          out_of_order_support(false),
-          simultaneous_use_requested(true), // try to use simultaneous path by default
-          double_buffers_size(false) // due to simultaneous case extend buffer size
-
-    {}
-
-    virtual bool Skip()
-    {
-        cl_command_queue_properties required_properties;
-        cl_int error = clGetDeviceInfo(
-            device, CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR,
-            sizeof(required_properties), &required_properties, NULL);
-        test_error(error,
-                   "Unable to query "
-                   "CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR");
-
-        cl_command_queue_properties queue_properties;
-
-        error = clGetCommandQueueInfo(queue, CL_QUEUE_PROPERTIES,
-                                      sizeof(queue_properties),
-                                      &queue_properties, NULL);
-        test_error(error, "Unable to query CL_QUEUE_PROPERTIES");
-
-        // Skip if queue properties don't contain those required
-        return required_properties != (required_properties & queue_properties);
-    }
-
-    virtual cl_int SetUp(int elements)
-    {
-        cl_int error = init_extension_functions();
-        if (error != CL_SUCCESS)
-        {
-            return error;
-        }
-
-        // Query if device supports simultaneous use
-        cl_device_command_buffer_capabilities_khr capabilities;
-        error =
-            clGetDeviceInfo(device, CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR,
-                            sizeof(capabilities), &capabilities, NULL);
-        test_error(error,
-                   "Unable to query CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR");
-        simultaneous_use = simultaneous_use_requested &&
-            (capabilities & CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR) != 0;
-        out_of_order_support =
-            capabilities & CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR;
-
-        if (elements <= 0)
-        {
-            return CL_INVALID_VALUE;
-        }
-        num_elements = static_cast<size_t>(elements);
-
-        // Kernel performs a parallel copy from an input buffer to output buffer
-        // is created.
-        const char *kernel_str =
-            R"(
-        __kernel void copy(__global int* in, __global int* out, int offset) {
-            size_t id = get_global_id(0);
-            int ind = offset + id;
-            out[ind] = in[ind];
-        })";
-
-        error = create_single_kernel_helper_create_program(context, &program, 1,
-                                                           &kernel_str);
-        test_error(error, "Failed to create program with source");
-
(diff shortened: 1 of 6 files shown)


test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,349 @@
...
+        {
+          out_of_order_queue = true;
+        }
+
+        // record command buffer with primary queue
+        error = RecordCommandBuffer();
+        test_error(error, "RecordCommandBuffer failed");
+
+        // create substitute queue
+        cl_command_queue new_queue = nullptr;
```
EwanC, line 129: Can use `clCommandQueueWrapper` here, it reference counts and automatically calls `clReleaseCommandQueue` when the last reference gets deleted.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,349 @@
...
+        }
+        return CL_SUCCESS;
+    }
+
+    //--------------------------------------------------------------------------
+    cl_int RecordCommandBuffer()
+    {
+      cl_int error = CL_SUCCESS;
+
+      if(out_of_order_queue)
```
EwanC, line 169: Could this condition be `if (!simultaneous_use)` instead?

Thanks to the sync point, the below command-buffer would have the correct ordering if a queue was either in-order or out-of-order. And could then remove
```
  if (!out_of_order_queue)
      {
        error = clEnqueueFillBuffer(q, in_mem, &pattern_pri, sizeof(cl_int),
                                    0, data_size(), 0, nullptr, nullptr);
        test_error(error, "clEnqueueFillBuffer failed");
      }
```
From `RunSingle()`

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.h
```diff
@@ -0,0 +1,95 @@
...
+// limitations under the License.
+
+#ifndef _CL_KHR_BASIC_COMMAND_BUFFER_H
+#define _CL_KHR_BASIC_COMMAND_BUFFER_H
+
+#include "command_buffer_test_base.h"
+#include "harness/typeWrappers.h"
+
+// temporary flag to be removed once simultaneous test will be sorted out
+#define USE_COMMAND_BUF_KENEL_ARG 0
```
EwanC, line 23: A note that I think the approach with this flag set is incorrect, but it passed on one implementation, so the code is in this draft PR so that we can reference it on the WG teleconference

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,349 @@
...
+    cl_int SetUp(int elements) override
+    {
+        // By default command queue is created without properties,
+        // if test requires queue with properties default queue must be replaced.
+        if (properties_use_requested)
+        {
+            // due to the skip condition
+            queue = nullptr;
+            if ( CreateCommandQueueWithProperties(queue) != CL_SUCCESS )
+              return CL_SUCCESS;
```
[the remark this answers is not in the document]
shajder: You're right, according to spec:

> The mandated minimum capability is: [CL_QUEUE_PROFILING_ENABLE]

which means the test should fail if queue with properties wasn't created.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,349 @@
...
+      {
+          CHECK_VERIFICATION_ERROR(pattern_pri, output_data[i], i);
+      }
+
+      return CL_SUCCESS;
+    }
+
+    //--------------------------------------------------------------------------
+    // tuple order: pattern, offset, queue, output-buffer
+    using SimulPassData =
```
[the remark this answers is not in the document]
shajder: Done.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.h
```diff
@@ -0,0 +1,95 @@
...
+    cl_context context;
+    cl_command_queue queue;
+    clCommandBufferWrapper command_buffer;
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+    clMemWrapper in_mem, out_mem, off_mem;
+    size_t num_elements;
+
+    // Device support query results
+    bool simultaneous_use;
```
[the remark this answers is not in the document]
shajder: Done.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.h
```diff
@@ -0,0 +1,95 @@
...
+    bool out_of_order_support;
+
+    // user request for simultaneous use
+    bool simultaneous_use_requested;
+    bool double_buffers_size;
```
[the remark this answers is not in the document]
shajder: Done.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,349 @@
...
+    //--------------------------------------------------------------------------
+    const cl_int pattern_pri = 0xB;
+    const cl_int pattern_sec = 0xC;
+
+    bool properties_use_requested = false;
+    bool out_of_order_queue = false;
+    clEventWrapper user_event = nullptr;
+};
+
+//#undef CHECK_VERIFICATION_ERROR
```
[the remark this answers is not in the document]
shajder: Done.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -525,26 +519,6 @@ struct OutOfOrderTest : public BasicCommandBufferTest
 
 #undef CHECK_VERIFICATION_ERROR
```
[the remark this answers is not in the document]
shajder: Done.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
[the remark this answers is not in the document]
shajder: Done.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
shajder: Done, there was one more malfunction of queue with properties test scenario. As result of this correction flag "out_of_order_queue" became unnecessary.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.h
shajder: Correct, just for completeness, setting this flag to 1 enables the path which calls clSetKernelArg between queuing simultaneous command buffers which according to specification is not correct approach but makes the test scenario work with POCL library.

shajder: Mentioned flag was removed due to narrowed scope of test scenario which works now with both POCL and Command Buffer Emulation Layer.

EwanC (dismissed):

EwanC:
@aharon-abramson Could you try this out on your implementation

aharon-abramson:
Yes, I'll give it a try.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,321 @@
...
+        else
+        {
+            // enque single command-buffer with substitute queue
+            error = RunSingle(new_queue);
+            test_error(error, "RunSingle failed");
+        }
+
+        if (properties_use_requested)
+        {
+            clReleaseCommandQueue(queue);
```
bashbaug, line 161: I'm not quite following why the `queue` needs to be explicitly released here.  Is this needed?  If so, can we use a queue wrapper for it instead so an explicit release is not required?

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,321 @@
...
+    {
+        cl_int error = clEnqueueFillBuffer(
+            pd.queue, in_mem, &pd.pattern, sizeof(cl_int),
+            pd.offset * sizeof(cl_int), data_size(), 0, nullptr, nullptr);
+        test_error(error, "clEnqueueFillBuffer failed");
+
+        error =
+            clEnqueueFillBuffer(pd.queue, off_mem, &pd.offset, sizeof(cl_int),
+                                0, sizeof(cl_int), 0, nullptr, nullptr);
+        test_error(error, "clEnqueueFillBuffer failed");
```
bashbaug, line 232: Checking: I think this is using the same "offset" buffer `off_mem` for each pass?  If so is there a race condition if each pass really does execute simultaneously?  This may be another reason to eliminate the "offset" buffer (see comment above).

bashbaug (commented):
FYI, I tried these new tests with my command buffer emulation layer and they all passed, pretty cool!

```
queue_substitution...
queue_substitution passed
properties_queue_substitution...
Queue property CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE supported. Testing ...
Queue property CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE supported. Testing ...
properties_queue_substitution passed
simultaneous_queue_substitution...
simultaneous_queue_substitution passed
PASSED sub-test.
PASSED 9 of 9 tests.
```

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -13,157 +13,139 @@
...
+
+    // Kernel performs a parallel copy from an input buffer to output buffer
+    // is created.
+    const char *kernel_str =
+        R"(
+    __kernel void copy(__global int* in, __global int* out, __global int* offset) {
+        size_t id = get_global_id(0);
+        int ind = offset[0] + id;
+        out[ind] = in[ind];
+    })";
```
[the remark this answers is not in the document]
shajder: Thanks for feedback. Please consider this: kernel is already recorded to command buffer and according to spec we cant use clSetKernelArg to change `int offset`. Next, we want to launch two command-buffers simultaneously keeping in mind both passes contains the same kernel with the same arguments. Without offset those kernel will try to read/write the same memory area, right ? Conclusion is that we need any mechanism to inform kernel about offset.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,321 @@
...
+        std::vector<PropPair> props = {
+            ADD_PROP(CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE),
+            ADD_PROP(CL_QUEUE_PROFILING_ENABLE)
+        };
+
+        for (auto&& prop : props)
+        {
+            check_property(prop);
+            test_error_ret(error, "clCreateCommandQueue failed", ret_queue);
+            if (ret_queue != nullptr) return ret_queue;
```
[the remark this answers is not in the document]
shajder: according to issue description:

>Create a command-buffer recorded to a single queue with additional properties (e.g profiling enabled) set

so this code will try to create either out-of-order or profiling queue depending on device capabilities

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
shajder: correct, just for completeness, it also uses the same `in_mem` and `out_mem` buffers which brings the same concerns I posted above

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
EwanC: I think this is the queue coming from the main test entry-point, which is then stored in `BasicCommandBufferTest` as a `cl_command_queue`. We don't use a wrapper if I recall because the common test fixture code which created that original queue will do the release after the test returns. 

This `BasicCommandBufferTest::queue` member variable is then overwritten in `SubstituteQueueTest::SetUp` before calling `BasicCommandBufferTest::SetUp` to create a command-buffer from a queue with properties set. So here we're releasing that new queue we created in `SubstituteQueueTest::SetUp` 

I agree this is confusing, and the way that comes to mind for improving this is by creating a `SubstituteQueueTest` command-buffer member variable in `SubstituteQueueTest::SetUp` which we can use in the tests, rather than using the base class command-buffer variable. We do this already in `OutOfOrderTest ` with the `out_of_order_command_buffer` variable, so can use that for reference.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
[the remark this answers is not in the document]
EwanC: To reinforce Marcin's comment, we tried that approach previously and ran into the issue described about `clSetKernelArg` not having an effect after a command-buffer command is recorded.

It's still in the commit history for reference - https://github.com/KhronosGroup/OpenCL-CTS/pull/1584/commits/d6c2b7bda79a356d0d41f7e27efbd6edb95ba008

I think we'd need the mutable-dispatch extension for this approach, and I actually added this scenario to the test plan for that extension.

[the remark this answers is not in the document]
bashbaug: Ah, gotcha, that makes sense.  Yes, we cannot modify the kernel argument or the clEnqueueNDRangeKernel global offset after the kernel is recorded into the command buffer, so we have to pass it "indirectly" like this.  Might be worth adding a comment just so people like me aren't confused in the future, but I agree this is correct.  Thanks for the explanation!

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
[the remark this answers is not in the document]
bashbaug: If we're only going to test one of the queue properties (which is fine) then I have a slight preference for (A), but only a slight preference.  The nice thing about (A) is that if somebody modifies this test in the future then it can be tested on any implementation, whereas (B) would require testing on implementations with and without out-of-order queue support.

This does assume that we will have tests for out-of-order queues elsewhere so we won't need them here but I think that's a safe assumption.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
bashbaug: OK, yes, I get it now.

Could we do this though?

1. Use a clCommandQueueWrapper instead of a cl_command_queue for BasicCommandBufferTest::queue.
2. Explicitly "retain" the command queue in the BasicCommandBufferTest constructor.  This is needed because we're using the command queue from the harness and we don't necessarily want the command queue to be destroyed when the wrapper is destroyed, we just want to release it (decrement the refcount by one).
3. Now that we're using a command queue wrapper we can remove this explicit release.

I just gave this a try and it seemed to work for both this "properties" test and the other command buffer tests that do not replace the queue, with no leaks reported.

EwanC: That's a good idea, using an explicit `clRetainCommandQueue` and then holding the queue in a wrapper I agree would be a better solution.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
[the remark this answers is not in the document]
EwanC: Cool, let's go with A to keep things simple since we do indeed have out-of-order queue testing elsewhere

EwanC (dismissed):

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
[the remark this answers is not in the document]
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
shajder: Done

bashbaug (approved):

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -13,157 +13,146 @@
...
+    error = clSetKernelArg(kernel, 0, sizeof(in_mem), &in_mem);
+    test_error(error, "clSetKernelArg failed");
+
+    error = clSetKernelArg(kernel, 1, sizeof(out_mem), &out_mem);
+    test_error(error, "clSetKernelArg failed");
+
+    error = clSetKernelArg(kernel, 2, sizeof(off_mem), &off_mem);
+    test_error(error, "clSetKernelArg failed");
+
+    if (simultaneous_use_support)
```
aharon-abramson, line 137: It's problematic that you try to create a command buffer before checking CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR in the Skip() method. It should be the other way around.

shajder: Looks like a good point to me. If queue will not meet the minimum properties specified by `CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR` then `clCreateCommandBufferKHR` will likely fail (`CL_INCOMPATIBLE_COMMAND_QUEUE_KHR`) instead of skipping the test.

My proposition is to move all `clGetDeviceInfo` queries to `BasicCommandBufferTest::Skip` method and change the order of `Skip` and `SetUp` calls in `MakeAndRunTest` function. I verified that approach with previous and new tests introduced in this PR - it needs some additional refactoring (and reviewing all remaining `cl_khr_command_buffer` related PRs) but it looks promising.

Please let me know if you have other propositions on your mind, thanks.

aharon-abramson: No, this approach looks good to me.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -0,0 +1,278 @@
...
+    {
+        if (properties_use_requested)
+        {
+            Version version = get_device_cl_version(device);
+            const cl_device_info host_queue_query = version >= Version(2, 0)
+                ? CL_DEVICE_QUEUE_ON_HOST_PROPERTIES
+                : CL_DEVICE_QUEUE_PROPERTIES;
+
+            cl_queue_properties host_queue_props = 0;
+            int error = clGetDeviceInfo(device, host_queue_query,
```
EwanC, line 52: `int error` -> `cl_int error`

EwanC (approved):
One very minor comment, but the change so that `BasicCommandBufferTest::Skip` happens before `BasicCommandBufferTest::SetUp` makes sense to me.

aharon-abramson:
Looks OK to me

bashbaug:

PR #1586, opened 2022-11-28
events: Remove unused BufferAction::Setup parameter

svenvh:
`BufferAction::Setup` has an additional parameter `allocate` which causes the virtual function `Action::Setup` to be hidden.  The additional parameter was unused anyway, so remove it.

Fixes a `-Woverloaded-virtual` warning.

### DIFF
--- a/test_conformance/events/action_classes.cpp
+++ b/test_conformance/events/action_classes.cpp
@@ -189,7 +189,7 @@ cl_int NDRangeKernelAction::Execute(cl_command_queue queue, cl_uint numWaits,
 #pragma mark -------------------- Buffer Sub-Classes -------------------------
 
 cl_int BufferAction::Setup(cl_device_id device, cl_context context,
-                           cl_command_queue queue, bool allocate)
+                           cl_command_queue queue)
 {
     cl_int error;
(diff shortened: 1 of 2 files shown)


svenvh:
I went for the simplest solution here, to just remove the `allocate` parameter.  An alternative solution would be to actually make use of `allocate` (but we'd have to guess the intent of the original author).

EwanC (approved):
LGTM

Would be nice if these action classes used `override` on the appropriate virtual functions, but that's widening the scope of this change, so could be another follow-on PR if we wanted to do it.

svenvh:
> Would be nice if these action classes used override on the appropriate virtual functions, but that's widening the scope of this change, so could be another follow-on PR if we wanted to do it.

Yeah, there's a general lack of `override` in the CTS.  Adding it to this particular instance only would lead to new "inconsistent-missing-override" warnings, so I figured it would be best to leave it as a future PR.

bcalidas:
Qualcomm would like more time to review this change.

lakshmih (approved):
Looks good

Sreelakshmi Haridas Maruthur
Qualcomm

bashbaug:

PR #1587, opened 2022-11-29
api: Update semaphores to check for pending signal

lakshmih:
A semaphore must have a pending signal or be signaled before waiting on it.  Add a check for signal. Additionally, remove check for invalid semaphore usage, which is implementation defined behavior, and un-testable.

### DIFF
--- a/test_conformance/api/test_semaphores.cpp
+++ b/test_conformance/api/test_semaphores.cpp
@@ -72,6 +72,75 @@
 
 static const char* source = "__kernel void empty() {}";
 
+/* Infer that signal_event generates a pending signal.
+ * A command that is submitted and not held back by dependencies
+ * moves to READY. If signal_event has no dependencies, polling for submitted
+ * or higher is sufficient.
+ *
+ * If signal event is submitted and has dependencies, poll for the RUNNING
+ * state.
+ * */
+static int poll_for_pending_signal(cl_command_queue queue,
+                                   cl_event signal_event,
+                                   cl_uint num_dependencies,
+                                   cl_event* dependencies)
+{
+    int err = CL_SUCCESS;
+    bool has_dependencies = false;
+
+    for (cl_uint i = 0; i < num_dependencies; i++)
+    {
+        cl_int event_status = -1;
+        err = clGetEventInfo(dependencies[i], CL_EVENT_COMMAND_EXECUTION_STATUS,
+                             sizeof(cl_int), &event_status, nullptr);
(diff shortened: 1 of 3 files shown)


EwanC:
This comment isn't related to this change in particular, but semaphore testing in general in the CTS. Rather than adding lots of semaphore tests to `test_conformance/api`, should `cl_khr_semaphore` have its own directory in `test_conformance/extensions/`? Then all the semaphore test could be run as one test binary, rather than part of `test_api`. If that makes sense, we could add moving the tests as an action in https://github.com/KhronosGroup/OpenCL-CTS/issues/1588

bcalidas:
> A semaphore must have a pending signal or be signaled before waiting on it. Add a check for signal. Additionally, remove check for invalid semaphore usage, which is implementation defined behavior, and un-testable.

To add more detail to this comment - the test was doing things like enqueueing a WaitSemaphore  command with no dependencies and allowing it to transition to the ready state before the corresponding SignalSemaphore command was enqueued.

test_conformance/api/test_semaphores.cpp
```diff
@@ -72,6 +72,75 @@
 
 static const char* source = "__kernel void empty() {}";
 
+/* Infer that signal_event generates a pending signal.
+ * A command that is submitted and not held back by dependencies
+ * moves to READY. If signal_event has no dependencies, polling for submitted
+ * or higher is sufficient.
```
bashbaug, line 78: This could very well be right, but I'm not quite following this - why is it OK to only check for SUBMITTED if the event has no dependencies?

It might be useful to add a few more words why this is safe, or just to check for RUNNING unconditionally, for simplicity.

bashbaug (commented):
The changes themselves seem reasonable though I'm not finding in the semaphores spec why they are required.  Is there something we need to add or enhance in the spec?

Also, just to confirm: there is nothing saveable in the test_semaphores_invalid_command test, so the only option is to remove it completely?

bashbaug:
Needs one more week for broader review.  May also need corresponding spec updates.

bcalidas:
Let's keep this PR on hold pending updates to  binary semaphores  in the spec.

nikhiljnv:
@bcalidas 
As discussed in the teleconference on Jun 27th, I believe this is not needed after recent changes to spec and related CTS tests.
Can we close this one?

nikhiljnv:
Ping @bcalidas !!

PR #1591, opened 2022-12-01
[NFC] relationals: Tidy up get_order_string

svenvh:
Remove a misleading comment at the end.  Improve comments in general, and use `true`/`false` for `bool`s.

### DIFF
--- a/test_conformance/relationals/test_shuffles.cpp
+++ b/test_conformance/relationals/test_shuffles.cpp
@@ -201,14 +201,13 @@ const char *get_order_string( ShuffleOrder &order, size_t vecSize, cl_uint lengt
 
     size_t j, idx;
 
-    // Assume we don't have to use numbers
-    byNumber = 0;
-    // Check to see
+    // Assume we don't have to use numbered indices (.s0123...).
+    byNumber = false;
+    // Check if any index is beyond xyzw, which requires to use numbers.
     for( j = 0; j < lengthToUse; j++ )
     {
         if (order[j] > 3) {
-            // An index is > xyzw so we need to use numbers
-            byNumber = 1;
+            byNumber = true;
             break;
         }
     }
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
Thank you for fixing these!  This will aid understanding the next time we need to modify these tests.

test_conformance/relationals/test_shuffles.cpp
```diff
@@ -201,14 +201,13 @@ const char *get_order_string( ShuffleOrder &order, size_t vecSize, cl_uint lengt
 
     size_t j, idx;
 
-    // Assume we don't have to use numbers
-    byNumber = 0;
-    // Check to see
+    // Assume we don't have to use numbered indices (.s0123...).
+    byNumber = false;
```
EwanC, line 205: Can we remove the `byNumber` pass-by-value parameter of `get_order_string` if we always set the value to `false` at the start of the function? And just have `bool get_order_string = false;` here instead.

test_conformance/relationals/test_shuffles.cpp
```diff
@@ -201,14 +201,13 @@ const char *get_order_string( ShuffleOrder &order, size_t vecSize, cl_uint lengt
 
     size_t j, idx;
```
EwanC, line 202: Would be nice if we could move these declarations down into the usages

EwanC (approved):
LGTM, some minor suggestions it's fine to pass on if it's out of scope for this PR

svenvh:
> LGTM, some minor suggestions it's fine to pass on if it's out of scope for this PR

Thanks! There is indeed more to improve (I noticed the strange use of the pass-by-value parameter too), but I was trying to keep this PR mainly focused around the remark [here](https://github.com/KhronosGroup/OpenCL-CTS/pull/1567#discussion_r1031454226).

PR #1592, opened 2022-12-05
printf test cases for cl_khr_command_buffer extenstion

shajder:
Added two test cases according to point 1.6 of issue specification:

https://github.com/KhronosGroup/OpenCL-CTS/issues/1369


test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -0,0 +1,556 @@
...
+        // Query if device supports simultaneous use
+        cl_device_command_buffer_capabilities_khr capabilities;
+        cl_int error =
+            clGetDeviceInfo(device, CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR,
+                            sizeof(capabilities), &capabilities, NULL);
+        test_error(
+            error,
+            "Unable to query CL_COMMAND_BUFFER_CAPABILITY_KERNEL_PRINTF_KHR");
+
+        if ((capabilities & CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR)
```
EwanC, line 192: This is checking `CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR` rather than `CL_COMMAND_BUFFER_CAPABILITY_KERNEL_PRINTF_KHR`.

If `CL_COMMAND_BUFFER_CAPABILITY_KERNEL_PRINTF_KHR` isn't supported the printf tests should skip rather than fail.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -0,0 +1,556 @@
...
+
+    //--------------------------------------------------------------------------
+    int WaitForEvent(cl_event* event)
+    {
+        cl_int status = clWaitForEvents(1, event);
+        if (status != CL_SUCCESS)
+        {
+            log_error("clWaitForEvents failed");
+            return status;
+        }
```
EwanC, line 261: Think this could be `test_error(status, "clWaitForEvents failed")`

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -0,0 +1,556 @@
...
+        // redirect output stream to temporary file
+        file_descriptor = AcquireOutputStream(&error);
+        if (error != 0)
+        {
+            log_error("Error while redirection stdout to file");
+            return TEST_FAIL;
+        }
+
+        // enqueue command buffer with kernel containing printf command
+        cl_event wait_event;
```
EwanC, line 308: Using `clEventWrapper` rather than `cl_event` should release `wait_event` when it goes out of scope

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -0,0 +1,556 @@
...
+    {
+        cl_int error = CL_SUCCESS;
+
+        // record command buffer with primary queue
+        error = RecordCommandBuffer();
+        test_error(error, "RecordCommandBuffer failed");
+
+        if (simultaneous_use_support)
+        {
+            // enque simultaneous command-buffers with printf calls
```
EwanC, line 222: typo "enqueue", and in the `else`

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -0,0 +1,556 @@
...
+    clEventWrapper trigger_event = nullptr;
+
+    std::string temp_filename;
+    int file_descriptor;
+
+    // specifies max test length for printf pattern
+    const unsigned max_pattern_length = 6;
+    // specifies min test length for printf pattern
+    const unsigned min_pattern_length = 1;
+    // specifies number of command-buffer equeue iterations
```
EwanC, line 538: "enqueue"

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
shajder: Method removed due to switch from cl_event to clEventWrapper.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
EwanC: So although this check has now been updated to `CL_COMMAND_BUFFER_CAPABILITY_KERNEL_PRINTF_KHR` I still don't think it's quite right. The test should skip rather than fail if this capability isn't supported by a device, so think this check should be in `CommandBufferPrintfTest::Skip()`

shajder: hmm, then maybe I should add new flag "printf_use_support" and move this piece of code to parent structure, right ?

shajder: Flag added, condition moved to `CommandBufferPrintfTest::Skip()` method.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -13,157 +13,166 @@
...
+                            sizeof(capabilities), &capabilities, NULL);
+    test_error(error,
+               "Unable to query CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR");
+    simultaneous_use_support = simultaneous_use_requested
+        && (capabilities & CL_COMMAND_BUFFER_CAPABILITY_SIMULTANEOUS_USE_KHR)
+            != 0;
+    out_of_order_support =
+        (capabilities & CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR) != 0;
+
+    printf_use_support =
```
EwanC, line 142: I don't think `printf_use_support` needs to be in the base class, it's unlikely any of the derived tests other than these printf tests will need it. So would make `printf_use_support` a member of `CommandBufferPrintfTest` which is initialized in ``CommandBufferPrintfTest::Setup`

EwanC (approved):
A few more minor nitpicks, but think this is looking good 👍

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -0,0 +1,536 @@
...
+        }
+
+        // execute command buffers
+        error = clSetUserEventStatus(trigger_event, CL_COMPLETE);
+        test_error_release_stdout(error, "clSetUserEventStatus failed");
+
+        // flush streams
+        fflush(stdout);
+        error = clFlush(queue);
+        test_error_release_stdout(error, "clFlush failed");
```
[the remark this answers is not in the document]
shajder: done

EwanC (dismissed):

bashbaug (commented):
I haven't done a thorough review just yet but I did find one issue.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp
```diff
@@ -0,0 +1,531 @@
...
+        cl_int error = CL_SUCCESS;
+
+        const char* kernel_str =
+            R"(
+      __kernel void print(__global char* in, __global char* out, __global int* offset)
+      {
+          size_t id = get_global_id(0);
+          int ind = offset[0] + offset[1] * id;
+          for(int i=0; i<offset[1]; i++) out[ind+i] = in[i];
+          printf("%s", in);
```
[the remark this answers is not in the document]
shajder: Hello Ben

Thanks for your feedback, I appreciate it.

Please let me summarize the idea behind that kernel. The goal was to print 1-6 the same characters in a null terminated string. I am using `std::vector<cl_char>` which guarantee consistent buffer (which is not that obvious when it comes to `std::string`). Then, mentioned vector is copied into `cl_mem` object which is set as first argument of kernel object:

`__global char* in`

next, `in` is used as `printf` argument in kernel call.

Could you explain a bit more what correction is on your mind ? Thank you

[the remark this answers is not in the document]
EwanC: I think the issue is that `__global char* in` isn't a string literal, can see in the spec section on examples of invalid usage of `%s` we've fallen into the first case :

image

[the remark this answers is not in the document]
shajder: ops :o, good find. Can't find clear info in spec - will below correction be sufficient? Guess not.  

`__global const char* in`

In that case we will have to "bake" string literal into kernel source, right ? Other ideas ?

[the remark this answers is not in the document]
EwanC: Using `%c` as per your example seems like the easiest fix to the immediate problem of invalid Opencl-C to me.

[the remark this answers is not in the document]
shajder: Done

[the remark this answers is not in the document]
bashbaug: Thanks!  The OpenCL C program looks valid now and this test is now passing on the Intel implementation with the command buffer emulation layer :+1: .

Interestingly the test is still failing on a different implementation.  I haven't been able to figure out why but I don't think it's related to command buffers or the command buffer emulation layer, since the test also fails if I replace the call to `clEnqueueCommandBufferKHR` with a call to `clEnqueueNDRangeKernel`:

```c++
#if 0
        // enqueue command buffer with kernel containing printf command
        error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
                                          nullptr, &wait_event);
        test_error_release_stdout(error, "clEnqueueCommandBufferKHR failed");
#else
        error = clEnqueueNDRangeKernel(queue, kernel, 1, nullptr, &num_elements, nullptr, 0, nullptr, &wait_event);
        test_error_release_stdout(error, "clEnqueueNDRangeKernel failed");
#endif
```

I'm going to resolve this conversation since this looks reasonable to me now, but this might be something to keep an eye on...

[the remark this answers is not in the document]
EwanC: Interestingly, after the `%c` change I now see the `basic_printf` test fail on our implementation. This is because too few characters are printed on the 3rd & last iteration of `RunSingle`. The number of total characters printed by this point should still be less than [CL_DEVICE_PRINTF_BUFFER_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_PRINTF_BUFFER_SIZE), so I think this is a bug in our implementation. As we've currently implemented printf in command-buffers, every enqueue of a command-buffer will use up more of the same kernel printf buffer, rather than having the whole allocation each command-buffer enqueue.

However, since another implementation is also failing, and this issue wasn't there with `%s` I thought I'd point out that I could work around this issue by using a smaller global size in the kernel, e.g instead of `num_elements` (which is `16384`) I used `1024` work-items.

EwanC (dismissed):

lakshmih:
Still reviewing this.

Sreelakshmi Haridas Maruthur
Qualcomm

aharon-abramson (approved):

EwanC (dismissed):

test_common/harness/os_helpers.cpp
```diff
@@ -32,6 +32,14 @@
 #include <android/api-level.h>
 #endif
 
+#if !defined(_WIN32)
+#if defined(__APPLE__)
+#include <sys/sysctl.h>
+#endif
+#include <unistd.h>
+#endif
+
```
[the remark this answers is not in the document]
shajder: Thanks @lakshmih , could you please verify my correction ? Currently I don't have android environment to check that out.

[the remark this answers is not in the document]
shajder: Good catch, thanks. Additional remark: `test_printf` is not directly related to `pritnf` test for `cl_khr_command_buffer` but it was already modified in this PR (moving `get_temp_filename` procedure to `os_helpers.cpp`) so I think this is justified to apply correction in here.

test_conformance/printf/test_printf.cpp
```diff
@@ -946,9 +924,10 @@ int main(int argc, const char* argv[])
...
     }
 
-    if (getTempFileName() == -1)
+    strcpy(gFileName, get_temp_filename());
+    if (strlen(gFileName) == 0)
     {
-        log_error("getTempFileName failed\n");
+        log_error("get_temp_filename failed\n");
         return -1;
     }
```
bcalidas, line 932: would be good to release the temp file name and avoid the leak here as well.

shajder: That's correct. I think I didn't push previous commit.

bashbaug:
Discussed in the March 28th teleconference.  OK to merge when conflicts are resolved.

lakshmih:
Looks good, thanks!

EwanC (approved):

PR #1594, opened 2022-12-08
Added tests for 1.1 individual command entry-points.

pj87:


test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_fill.cpp
```diff
@@ -0,0 +1,164 @@
...
+        
+        error = clFinish(queue);
+
+        test_error(error, "clFinish failed");
+        
+        std::vector<cl_char> output_data(data_size);
+        error = clEnqueueReadBuffer(queue, buffer, CL_TRUE, 0,
+                                    data_size, output_data.data(), 0, nullptr,
+                                    nullptr);
+        test_error(error, "clEnqueueReadBuffer failed");
```
shajder, line 123: Normally `clFinish` should be called after `clEnqueueReadBuffer` but in this case it is not necessary at all due to blocking read operation (third parameter set to `CL_TRUE`)

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_fill.cpp
```diff
@@ -0,0 +1,164 @@
...
+        return CL_SUCCESS;
+    }
+
+    cl_int SetUp(int elements) override
+    {
+        cl_int error = BasicCommandBufferTest::SetUp(elements);
+        test_error(error, "BasicCommandBufferTest::SetUp failed");
+
+        buffer = clCreateBuffer(context, CL_MEM_READ_WRITE, data_size, nullptr, &error);
+        test_error(error, "Unable to create buffer");
```
shajder, line 139: No need for new buffer I think, you could use one of those created in `BasicCommandBufferTest::SetUp`

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_kernel.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+        cl_int Run() override
+        {
+            cl_int error = clCommandNDRangeKernelKHR(
+                command_buffer, nullptr, nullptr, kernel, 1, nullptr, &num_elements,
+                nullptr, 0, nullptr, nullptr, nullptr);
+            test_error(error, "clCommandNDRangeKernelKHR failed");
+
+            error = clCommandBarrierWithWaitListKHR(command_buffer, nullptr, 0, nullptr, nullptr, nullptr);
+            test_error(error, "clCommandBarrierWithWaitListKHR failed");
```
shajder, line 43: As far as I understand this function waits for sync-points or non-blocking operations to be completed, thus you should precede it with such calls which is missing at the moment

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_kernel.cpp
```diff
@@ -0,0 +1,98 @@
...
+            return CL_SUCCESS;
+        }
+
+        cl_int SetUp(int elements) override
+        {
+            cl_int error = BasicCommandBufferTest::SetUp(elements);
+            test_error(error, "BasicCommandBufferTest::SetUp failed");
+
+            user_event = clCreateUserEvent(context, &error);
+            test_error(error, "clCreateUserEvent failed");
```
shajder, line 81: It seems to me like this test does not require use of separate event

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
```diff
@@ -0,0 +1,459 @@
...
+        cl_int SetUp(int elements) override
+        {
+            cl_int error = BasicCommandBufferTest::SetUp(elements);
+            test_error(error, "BasicCommandBufferTest::SetUp failed");
+
+            in_mem = clCreateBuffer(context, CL_MEM_READ_WRITE, data_size, nullptr, &error);
+            test_error(error, "clCreateBuffer failed");
+
+            out_mem = clCreateBuffer(context, CL_MEM_READ_WRITE, data_size, nullptr, &error);
+            test_error(error, "clCreateBuffer failed");
```
shajder, line 167: Those buffers were already created in `BasicCommandBufferTest::SetUp`, I think there is no need to repeat that operation.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
```diff
@@ -0,0 +1,459 @@
...
+
+            return CL_SUCCESS;
+        }
+
+        const size_t data_size = 262144;
+        const char pattern = 0x14;
+        const cl_uint fill_color[4] = { pattern, pattern, pattern, pattern };
+        const cl_image_format formats = { CL_RGBA, CL_UNSIGNED_INT8 };
```
shajder, line 175: Are those necessary in `CopyBufferKHR` test ?

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
```diff
@@ -0,0 +1,459 @@
...
+
+    struct CopyBufferKHR : public BasicCommandBufferTest
+    {
+        using BasicCommandBufferTest::BasicCommandBufferTest;
+
+        cl_int Run() override
+        {
+            cl_int error = clCommandFillBufferKHR(
+                    command_buffer, nullptr, in_mem, &pattern, sizeof(cl_char), 0,
+                    data_size, 0, nullptr, nullptr, nullptr);
```
shajder, line 126: Using hard-coded `data_size` without involving `sizeof` and using the same name as parent method name is a bit misleading I think.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
```diff
@@ -0,0 +1,459 @@
...
+            error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
+                                            nullptr, nullptr);
+            test_error(error, "clEnqueueCommandBufferKHR failed");
+
+            error = clFinish(queue);
+
+            test_error(error, "clFinish failed");
+
+            std::vector<cl_char> output_data(data_size);
+            error = clEnqueueReadImage(queue, dst_image, CL_TRUE, origin, region, 0, 0, output_data.data(), 0, nullptr, nullptr);
```
shajder, line 81: The same comment as for `command_buffer_test_fill.cpp`: Normally `clFinish` should be called after `clEnqueueReadBuffer` but in this case it is not necessary at all due to blocking read operation (third parameter set to CL_TRUE)

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
```diff
@@ -0,0 +1,459 @@
...
+        cl_int SetUp(int elements) override
+        {
+            cl_int error = BasicCommandBufferTest::SetUp(elements);
+            test_error(error, "BasicCommandBufferTest::SetUp failed");
+
+            in_mem = clCreateBuffer(context, CL_MEM_READ_WRITE, data_size, nullptr, &error);
+            test_error(error, "clCreateBuffer failed");
+
+            out_mem = clCreateBuffer(context, CL_MEM_READ_WRITE, data_size, nullptr, &error);
+            test_error(error, "Unable to create buffer");
```
shajder, line 407: Due to consistency with other tests we could use buffers created in `BasicCommandBufferTest::SetUp` and `BasicCommandBufferTest::data_size()` method as an input for copy operations I think

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
```diff
@@ -0,0 +1,436 @@
...
+        dst_image = create_image_2d(context, CL_MEM_WRITE_ONLY, &formats,
+                                    img_width, img_height, 0, NULL, &error);
+        test_error(error, "create_image_2d failed");
+
+        return CL_SUCCESS;
+    }
+
+    bool Skip() override
+    {
+        cl_int error = BasicCommandBufferTest::Skip();
```
EwanC, line 101: Think `BasicCommandBufferTest::Skip()` returns a `bool`, so this can just be
```cpp
return imageSupport || BasicCommandBufferTest::Skip();
```

But I am seeing these image test skip correctly now on my implementation.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_fill.cpp
pj87: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_kernel.cpp
pj87: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
pj87: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
pj87: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
pj87: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
pj87: It was not necessary. Removed.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_fill.cpp
pj87: Done.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_kernel.cpp
pj87: Done.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
pj87: Done.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
pj87: Done.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_copy.cpp
```diff
@@ -0,0 +1,412 @@
...
+#include "harness/typeWrappers.h"
+#include "procs.h"
+
+#include <vector>
+
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
+// Command-queue fill tests which handles below cases:
```
EwanC, line 26: "Command-buffer copy tests .."

EwanC (commented):
I have a couple of superficial comments, and a suggestion about how the barrier test could be improved.

In future, we'd probably want wider testing of the usage of these copy & fill commands for things like offsets and full range of patterns. However, I didn't specify that in the test plan and it's low value in terms of finding spec issues since this API usage is taken from existing core spec entry-points. So think this coverage in this MR is fine as a first step.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -546,4 +546,4 @@ int test_out_of_order(cl_device_id device, cl_context context,
                       cl_command_queue queue, int num_elements)
 {
     return MakeAndRunTest<OutOfOrderTest>(device, context, queue, num_elements);
-}
+}
```
EwanC, line 549: The new files you've added in this PR end in a new line, so think we should keep this new line as-is

EwanC (approved):
LGTM, some very trivial comments

EwanC (approved):

aharon-abramson (approved):

PR #1597, opened 2022-12-09
relationals: Use stringstream in print_hex_mem_dump

svenvh:
print_hex_mem_dump was calling `sprintf` with source and destination buffers overlapping, which is explicitly forbidden.  Rewrite the function to use C++ stringstreams instead.

This commit only affects the error path and does not change the printed output.  Error logs before and after this commit should not differ.

### DIFF
--- a/test_conformance/relationals/test_shuffles.cpp
+++ b/test_conformance/relationals/test_shuffles.cpp
@@ -13,6 +13,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+
+#include <iomanip>
+
 #include "testBase.h"
 #include "harness/conversions.h"
 #include "harness/typeWrappers.h"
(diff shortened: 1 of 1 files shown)


test_conformance/relationals/test_shuffles.cpp
```diff
@@ -260,33 +263,48 @@ char * get_order_name( ExplicitType vecType, size_t inVecSize, size_t outVecSize
...
+                        const unsigned char *outDataPtr, size_t inVecSize,
+                        size_t outVecSize, size_t typeSize)
 {
-    char error [4096] = "";
-    strcat(error, "      Source: ");
-    for( unsigned int j = 0; j < inVecSize * typeSize; j++ )
+    auto print_hex_byte = [](unsigned char v) {
+        // Use a new stream to avoid manipulating state of outer stream.
+        std::ostringstream ss;
+        ss << std::setfill('0') << std::setw(2) << std::right << std::hex << +v;
```
StuartDBrady, line 275: `+v`?  Is that intentional?

test_conformance/relationals/test_shuffles.cpp
```diff
@@ -260,33 +263,48 @@ char * get_order_name( ExplicitType vecType, size_t inVecSize, size_t outVecSize
...
+void print_hex_mem_dump(const unsigned char *inDataPtr,
+                        const unsigned char *inDataPtr2,
+                        const unsigned char *expected,
+                        const unsigned char *outDataPtr, size_t inVecSize,
+                        size_t outVecSize, size_t typeSize)
 {
-    char error [4096] = "";
-    strcat(error, "      Source: ");
-    for( unsigned int j = 0; j < inVecSize * typeSize; j++ )
+    auto print_hex_byte = [](unsigned char v) {
```
StuartDBrady, line 272: Wondering if `hex_byte_to_string` might be a better name for this?  It doesn't _actually_ print anything itself, aside from using `ostringstream` (which you could arguably consider printing), but that's an internal detail of the lambda.

StuartDBrady (commented):
Mostly looks good, just a few minor questions.

test_conformance/relationals/test_shuffles.cpp
StuartDBrady: Does this promote its type, somehow, to prevent it from being appended to the string as an actual character?  (Apologies if that is so - it's not something I've come across - so I'm wondering if that's the generally accepted idiom for this?)

svenvh: > Does this promote its type, somehow, to prevent it from being appended to the string as an actual character?

Correct, the unary `+` operand is subject to implicit integer promotion as the rank of `unsigned char` is less than the rank of `int`.  I believe it's a fairly common idiom when dealing with output streams and `char` types (and in general, one of the very few practical uses of unary `+`: to force integer promotion).

test_conformance/relationals/test_shuffles.cpp
svenvh: Good point, I'll update the name.

test_conformance/relationals/test_shuffles.cpp
StuartDBrady: Thanks!  If that's the common idiom, then the change looks good to me.

(I would raise a bit of an eyebrow at iostream, here – in C, _char_ is a "byte" of unspecified signedness (i.e. usually 8 bits) and as I understood it, character constants are actually of type _int_ – so speaking as someone more used to printf, this seems highly questionable – although I suppose if you wanted an 8-bit integer and not a _character_, you'd use _signed char_ or _unsigned char_ (or _int8_t_ or _uint8_t_) – but then I suppose these get interpreted by iostream as characters too.  Oh well...)

EwanC (approved):

lakshmih:

PR #1602, opened 2022-12-16
Added tests for 3.1 event info.

pj87:


test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+        error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
+                                          nullptr, &event);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        error = clWaitForEvents(1, &event);
+        test_error(error, "clWaitForEvents failed");
+
+        error = clGetEventInfo(event, CL_EVENT_REFERENCE_COUNT, sizeof(count),
+                               &count, &size);
```
EwanC, line 190: should we be checking the returned value for `size`?

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,240 @@
...
+        test_error(error, "clGetEventInfo failed");
+
+        if (count != expected_count)
```
EwanC, line 193: The wording around spec reference counts means that I think it would still be legal for a implementation to return a value larger than 1 here, I'd just make this check `count > 0`

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    cl_int Run() override
+    {
+        cl_int error = clFinalizeCommandBufferKHR(command_buffer);
+        test_error(error, "clFinalizeCommandBufferKHR failed");
+
+        error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
+                                          nullptr, &event);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
```
EwanC, line 150: Could we query the execution status here as well, although  valid values will be all of
`status == CL_QUEUED || status == CL_SUBMITTED || status == CL_RUNNING || status == CL_COMPLETE`

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,240 @@
...
+        error = clWaitForEvents(1, &event);
+        test_error(error, "clWaitForEvents failed");
+
+        error = clGetEventInfo(event, CL_EVENT_CONTEXT, sizeof(ret_context),
+                               &ret_context, &size);
```
EwanC, line 121: Can check that `size == sizeof(context)` after this call too

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,240 @@
...
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        error = clWaitForEvents(1, &event);
+        test_error(error, "Unable to wait for event");
+
+        error = clGetEventInfo(event, CL_EVENT_COMMAND_QUEUE, sizeof(ret_queue),
+                               &ret_queue, &size);
+        test_error(error, "clGetEventInfo failed");
+
+        if (ret_queue != queue)
```
EwanC, line 90: Interesting it seems that the regular tests don't verify this, just the size of the queue - https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/events/test_events.cpp#L143-L144

And don't think we should be doing anything differently from them.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,256 @@
...
+        error = clGetEventInfo(event, CL_EVENT_COMMAND_QUEUE,
+                               sizeof(otherQueue), &otherQueue, &size);
+        test_error(error, "Unable to get event info!");
+
+        // We can not check if this is the right queue because this is an opaque
+        // object.
+        if (size != sizeof(queue) || otherQueue == NULL)
+        {
+            log_error("ERROR: Returned command queue size does not validate "
+                      "(expected %d, got %d)\n",
```
EwanC, line 92: Rather than `%d` and c-style casting, can use the `%zu` format specified to print `size_t` variables. Can update this and other log messages.

EwanC (commented):
Looking good, one more minor comment

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -432,7 +432,7 @@ struct OutOfOrderTest : public BasicCommandBufferTest
     OutOfOrderTest(cl_device_id device, cl_context context,
                    cl_command_queue queue)
         : BasicCommandBufferTest(device, context, queue),
-          out_of_order_queue(nullptr), out_of_order_command_buffer(this),
+          out_of_order_command_buffer(this), out_of_order_queue(nullptr),
```
EwanC, line 435: Don't think we should be changing this member initialization order. `out_of_order_queue` is defined before `out_of_order_command_buffer` on lines 504-505, so initializing `out_of_order_command_buffer` first I suspect would trigger a `-Wreorder` warning.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,254 @@
...
+                "ERROR: Incorrect status returned from clGetEventInfo (%d)\n",
+                status);
+
+            return TEST_FAIL;
+        }
+
+        return CL_SUCCESS;
+    }
+
+    clEventWrapper event;
```
EwanC, line 63: I think these variables can just be defined in `Run()`, rather than needing to be class member variables. Same for all class member variables in new tests.

test_conformance/extensions/cl_khr_command_buffer/main.cpp
```diff
@@ -23,7 +23,12 @@ test_definition test_list[] = { ADD_TEST(single_ndrange),
                                 ADD_TEST(out_of_order),
                                 ADD_TEST(queue_substitution),
                                 ADD_TEST(properties_queue_substitution),
-                                ADD_TEST(simultaneous_queue_substitution) };
+                                ADD_TEST(simultaneous_queue_substitution),
+                                ADD_TEST(command_type),
+                                ADD_TEST(command_queue),
+                                ADD_TEST(execution_status),
+                                ADD_TEST(context),
+                                ADD_TEST(reference_count) };
```
EwanC, line 31: Could we prefix these test names with `event_info_` to make their context clearer.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,254 @@
...
+        test_error(error, "Unable to get event info!");
+
+        // We can not check if this is the right queue because this is an opaque
+        // object.
+        if (size != sizeof(queue) || otherQueue == NULL)
+        {
+            log_error("ERROR: Returned command queue size does not validate "
+                      "(expected %zu, got %zu)\n",
+                      sizeof(queue), size);
+            return -1;
```
EwanC, line 92: There is some inconsistency between returning `-1` in these tests and `TEST_FAIL`,  would prefer to have `TEST_FAIL` consistently.

EwanC (dismissed):

test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_event_info.cpp
```diff
@@ -0,0 +1,254 @@
...
+////////////////////////////////////////////////////////////////////////////////
+// get event info tests which handles below cases:
+//
+// -command type
+// -command queue
+// -context
+// -execution status
+// -reference count
+
+struct CommandType : public BasicCommandBufferTest
```
aharon-abramson, line 31: BasicCommandBufferTest has a Skip() method, right?

pj87: Yes, it does have it:

```
struct BasicCommandBufferTest : CommandBufferTestBase
{

    BasicCommandBufferTest(cl_device_id device, cl_context context,
                           cl_command_queue queue);

    virtual bool Skip();
    virtual cl_int SetUp(int elements);
```

aharon-abramson (approved):

EwanC:
Closing this PR as it has been superseded by https://github.com/KhronosGroup/OpenCL-CTS/pull/1647 which is the same content for testing 3.1 but rebased cleanly

PR #1603, opened 2022-12-16
c11 atomic fence: relaxed requirements for an auxiliary atomic_store

FMarno:
I had a case where the test kernel wasn't building because atomic_store requires support for __opencl_c_atomic_order_seq_cst and the device didn't have it.
I believe this store is only used to record some data in the failure case, so it being relaxed is probably fine since further reads of the data come after barriers, like around here: https://github.com/KhronosGroup/OpenCL-CTS/blob/59a1198e5f2f04540736ea0ae03c88069b2a3543/test_conformance/c11_atomics/common.h#L1014-L1019

### DIFF
--- a/test_conformance/c11_atomics/test_atomics.cpp
+++ b/test_conformance/c11_atomics/test_atomics.cpp
@@ -2914,7 +2914,8 @@ public:
                 + "-1);\n"
                   "  if(hisAtomicValue != hisValue)\n"
                   "  { // fail\n"
-                  "    atomic_store(&destMemory[myId], myValue-1);\n";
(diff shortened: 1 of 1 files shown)


lakshmih:

lakshmih (approved):
Looks good.

Sreelakshmi Haridas Maruthur
Qualcomm

bashbaug (approved):
LGTM - thanks!

For future reference, a command line that exercises these changes is:

test_c11_atomics atomic_fence

bashbaug:
Merging as discussed in the January 31st teleconference.

PR #1605, opened 2022-12-20
math_brute_force: Remove unnecessary gotos

svenvh:
Simplify code by returning directly instead of using goto statements.

Although intended as an NFC commit, this changes the behaviour around
clFlush calls.  Before this commit, failure of the third clFlush call
would print "clFlush 3 failed" and return the clFlush error code.
This behaviour is inconsistent with the other clFlush calls in
math_brute_force, which are not fatal.  The lack of a `goto exit`
makes me suspect that this 3rd clFlush call was intended to be
non-fatal too.  As such, this commit makes all clFlush calls non-fatal
by returning `CL_SUCCESS` even when the third clFlush call fails.

### DIFF
--- a/test_conformance/math_brute_force/binary_i_double.cpp
+++ b/test_conformance/math_brute_force/binary_i_double.cpp
@@ -288,14 +288,14 @@ cl_int Test(cl_uint job_id, cl_uint thread_id, void *data)
                                       buffer_size, p, 0, NULL, NULL)))
     {
         vlog_error("Error: clEnqueueWriteBuffer failed! err: %d\n", error);
-        goto exit;
+        return error;
     }
 
     if ((error = clEnqueueWriteBuffer(tinfo->tQueue, tinfo->inBuf2, CL_FALSE, 0,
                                       buffer_size / 2, p2, 0, NULL, NULL)))
     {
         vlog_error("Error: clEnqueueWriteBuffer failed! err: %d\n", error);
-        goto exit;
+        return error;
     }
 
     for (auto j = gMinVectorSizeIndex; j < gMaxVectorSizeIndex; j++)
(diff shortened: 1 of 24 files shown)


test_conformance/math_brute_force/binary_i_double.cpp
```diff
@@ -464,7 +463,6 @@ cl_int Test(cl_uint job_id, cl_uint thread_id, void *data)
 
     if ((error = clFlush(tinfo->tQueue))) vlog("clFlush 3 failed\n");
```
EwanC, line 464: do we need to `return error;` in this if now? We'd now return CL_SUCCESS when the function returns even if this error was logged, where as before we'd return the error from the function (I think)

svenvh: Thanks, good catch!

svenvh: On a closer look, it looks like other `clFlush` calls are not fatal; see for example the lines that print "clFlush failed" and "clFlush 2 failed" in this file: the error variable is overwritten in subsequent API calls.  The lack of a `goto exit` here makes me suspect that it was intended for this 3rd `clFlush` call to be non-fatal too.

If you agree that returning `CL_SUCCESS` while "clFlush 3 failed" is acceptable, I'll update the commit message to mention this functional change (and drop the NFC tag).

EwanC: Interesting, thanks for looking into it. Seems like this behaviour affects other math_brute_force tests too and not just this file, but after this MR the behaviour will be consistent everywhere in the math_brute_force_tests of `clFlush` being non-fatal. I think the non-fatal behaviour odd, but that's a separate discussion from this useful clean-up PR.

Your proposal for updating the commit and progressing with this change sounds good to me.

svenvh: Thanks! I've updated the commit message and PR description.

EwanC (dismissed):

lakshmih:
Qualcomm needs more time to review this

lakshmih (approved):
Looks good.

bashbaug (dismissed):
LGTM

bashbaug:
Note: there are several conflicts that will need to be resoved before merging.

svenvh:
> Note: there are several conflicts that will need to be resoved before merging.

I have resolved these now (by cherry-picking the change onto latest `main` using the `-X ours` merge strategy option, followed by fixing up some newly appearing `goto exit` instances).

bashbaug (approved):
Merging as agreed in the February 7th teleconference.

PR #1610, opened 2022-12-27
gl: Fix array size expression

svenvh:
`sizes` is a pointer argument, so the expression does not compute the presumably intended number of elements in `sizes`.

Fixes a `-Wsizeof-pointer-div` warning.


EwanC (approved):
I've not tested this change, but makes sense to me

EwanC (approved):

PR #1611, opened 2022-12-29
clGetCommandBufferInfoKHR test cases for cl_khr_command_buffer extension

shajder:
Added 4 test cases according to point 1.4 of issue specification:

https://github.com/KhronosGroup/OpenCL-CTS/issues/1369


test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,334 @@
...
+    //--------------------------------------------------------------------------
+    cl_int Run() override
+    {
+        cl_int error = CL_SUCCESS;
+
+        switch (test_mode)
+        {
+            case CombufInfoTestMode::CITM_QUEUES:
+                error = RunQueuesInfoTest();
+                test_error(error, "RunCombufWaitForCombuf failed");
```
EwanC, line 69: Think these log messages need updated

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,334 @@
...
+        error = clGetCommandBufferInfoKHR(
+            command_buffer, CL_COMMAND_BUFFER_NUM_QUEUES_KHR, sizeof(cl_uint),
+            &num_queues, &ret_value_size);
+        test_error(error, "clGetCommandBufferInfoKHR failed");
+
+        test_expected_info(ret_value_size > sizeof(cl_int));
+
+        test_expected_info(num_queues != expect_queue_list.size());
+
+        std::vector<cl_command_queue> ql(num_queues);
```
EwanC, line 138: nitpick - `ql` doesn't mean anything to me ("queue list" ?), would prefer a more descriptive variable name

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,334 @@
...
+    {
+        cl_int error = CL_SUCCESS;
+
+        // record command buffer
+        error = RecordCommandBuffer();
+        test_error(error, "RecordCommandBuffer failed");
+
+        // collect initial reference count
+        cl_uint init_ref_count = 0;
+        size_t ret_value_size = 0;
```
EwanC, line 167: Don't think we ever check the returned `ret_value_size` in this test

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,334 @@
...
+    //--------------------------------------------------------------------------
+    cl_int RunStateInfoTest()
+    {
+        cl_int error = CL_SUCCESS;
+
+        // lambda to verify given state
+        auto verify_state = [&](const cl_command_buffer_state_khr &expected) {
+            cl_command_buffer_state_khr state =
+                CL_COMMAND_BUFFER_STATE_INVALID_KHR;
+            size_t ret_value_size = 0;
```
EwanC, line 221: we don't check this returned value

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,334 @@
...
+    cl_int RunPropArrayInfoTest()
+    {
+        cl_int error = CL_SUCCESS;
+
+        // record command buffer
+        error = RecordCommandBuffer();
+        test_error(error, "RecordCommandBuffer failed");
+
+        size_t ret_value_size = 0;
+        cl_command_buffer_properties_khr combuf_props[16];
```
EwanC, line 279: Would prefer to make this a vector and then resize it, rather than a C array.
Something like:
```
std::vector<cl_command_buffer_properties_khr> combuf_props;
clGetCommandBufferInfoKHR(
            command_buffer, CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR,
            0, NULL, &ret_value_size);
cl_uint num_props =  ret_value_size / sizeof(cl_command_buffer_properties_khr);
combuf_props.resize(num_props);
clGetCommandBufferInfoKHR(
            command_buffer, CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR,
            ret_value_size, combuf_props.data(), NULL );

**Review by EwanC — COMMENTED** [review]
Looks good and passes on Codeplay's implementation, some minor comments

File: test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp (hunk 1 above)
Comment (shajder) [reply test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp]: Done

File: test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp (hunk 3 above)
Comment (shajder) [reply test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp]: Done

File: test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp (hunk 4 above)
Comment (shajder) [reply test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp]: Done

File: test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp (hunk 5 above)
Comment (shajder) [reply test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp]: Done

File: test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp [hunk 6]
```diff
@@ -0,0 +1,339 @@
...
+            ret_value_size / sizeof(cl_command_buffer_properties_khr);
+        combuf_props.resize(num_ret_props);
+        error = clGetCommandBufferInfoKHR(
+            command_buffer, CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR,
+            num_ret_props * sizeof(cl_command_buffer_properties_khr),
+            combuf_props.data(), nullptr);
+        test_error_ret(error, "clGetCommandBufferInfoKHR failed", TEST_FAIL);
+
+        for (int i = 0; i < num_ret_props; i++)
+            if (combuf_props[i] == CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR)
```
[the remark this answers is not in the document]
shajder: understood, then we must memorize properties array for default command_buffer in `BasicCommandBufferTest` structure, right ?

We could also make a bald assumption that below code from base structure `BasicCommandBufferTest` will stay intact. In that case we could even skip iteration and simply check if there is only one property and it is equal `CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR`

```
    if (simultaneous_use_support)
    {
        cl_command_buffer_properties_khr properties[3] = {
            CL_COMMAND_BUFFER_FLAGS_KHR, CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR,
            0
        };
        command_buffer =
            clCreateCommandBufferKHR(1, &queue, properties, &error);
    }

```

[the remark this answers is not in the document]
EwanC: How things are currently setup in the command-buffer tests, if we want to test a command-buffer with more specific properties, then the derived test fixture should set that up themselves (e.g out-of-order command-buffer tests). And `BasicCommandBufferTest` will create these basic command-buffer for more generic tests. In the future it may be that we are smarter about setting properties in `BasicCommandBufferTest` so that some device specific flags are added to the passed properties based on query results. However, that's not how the command-buffer tests are currently setup, and should be a PR addressing that concern in isolation since it will likely affect other tests too. Therefore, for the scope of this PR I think it's fine to assume that the properties passed in `BasicCommandBufferTest`
will stay intact. 

So we could indeed skip iterating during validation as you describe. We should also check another case however (the `else` to the line i've commented on), that if simultaneous use  wasn't set, then the correct value is returned to stress this spec wording

> If the properties argument specified in clCreateCommandBufferKHR used to
create command_buffer was NULL, the implementation may return either a param_value_size_ret
of 0 (i.e. there is are no properties to be returned), or the implementation may return
a property value of 0 (where 0 is used to terminate the properties list).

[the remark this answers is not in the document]
shajder: > ... We should also check another case however (the else to the line i've commented on), that if simultaneous use wasn't set, then the correct value is returned to stress this spec wording

This changes the logic of test a bit. So far `Skip` method verified if `CombufInfoTestMode::CITM_PROP_ARRAY` test created command buffer with simultaneous use, otherwise test skipped. After correction this test accepts both scenarios (`NULL` and simultaneous use prop array).

EwanC (approved):

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,340 @@
...
+        test_error(error, "clGetCommandBufferInfoKHR failed");
+
+        test_expected_info(ret_value_size > expect_size);
+
+        // We can not check if this is the right queue because this is an opaque
+        // object, test against NULL.
+        for (int i = 0; i < queue_list.size(); i++)
+        {
+            test_expected_info(queue_list[i] == NULL);
+        }
```
bashbaug, line 142: I think we should be able to do a little better here, see: https://github.com/KhronosGroup/OpenCL-Docs/issues/797

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,340 @@
...
+            nullptr, 0, nullptr, nullptr, nullptr);
+        test_error(error, "clCommandNDRangeKernelKHR failed");
+
+        error = clFinalizeCommandBufferKHR(command_buffer);
+        test_error(error, "clFinalizeCommandBufferKHR failed");
+        return CL_SUCCESS;
+    }
+
+    //--------------------------------------------------------------------------
+#define test_expected_info(cond)                                               \
```
bashbaug, line 94: Any reason we can't use the existing `test_assert_error` instead of defining a new `test_expected_info`?

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
shajder: Done, confusion comes from two directions:

1) similar solution used in a different test (raise an issue?):
[/test_conformance/events/test_events.cpp](https://github.com/KhronosGroup/OpenCL-CTS/blob/296a7a574dc09e8b7c143e1de77ed5ae0aa8b63d/test_conformance/events/test_events.cpp#L143-L151)
```
    // We can not check if this is the right queue because this is an opaque
    // object.
    if (size != sizeof(queue))
    {
        log_error("ERROR: Returned command queue size does not validate "
                  "(expected %d, got %d)\n",
                  (int)sizeof(queue), (int)size);
        return -1;
    }
```
2) PoCL fails equality test at the moment

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,340 @@
...
+                                          &trigger_event, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        // verify pending state
+        error = verify_state(CL_COMMAND_BUFFER_STATE_PENDING_KHR);
+        test_error(error, "verify_state failed");
+
+        // execute command buffer
+        error = clSetUserEventStatus(trigger_event, CL_COMPLETE);
+        test_error(error, "clSetUserEventStatus failed");
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -0,0 +1,340 @@
...
+        error = clGetCommandBufferInfoKHR(
+            command_buffer, CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR,
+            num_ret_props * sizeof(cl_command_buffer_properties_khr),
+            combuf_props.data(), nullptr);
+        test_error_ret(error, "clGetCommandBufferInfoKHR failed", TEST_FAIL);
+
+        if (simultaneous_use_support)
+        {
+            if (combuf_props[0] == CL_COMMAND_BUFFER_FLAGS_KHR
+                && combuf_props[1] == CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR)
```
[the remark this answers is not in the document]
shajder: Done

EwanC (dismissed):

EwanC (approved):

bashbaug:

PR #1614, opened 2023-01-04
spirv: disable testing for OpFMod / OpFRem

### LINKED ISSUE #1548: spirv_new frem/fmod tests may require excessive precision
I am not sure if this is an error from my understanding of the Env spec or if the assumption made in the test is actually wrong, but what the test is doing is to assume that `OpFRem` is equal to the OpenCL C builtin `fmod` which I don't think is a valid assumption.

It should compare the `fmod` OpenCL ExtInstr against a OpenCL C kernel using `fmod` instead.

But maybe we have to assume that spir-v alu instructions have the same precision requirements as the OpenCL builtins? That would make layering OpenCL on top of Vulkan much harder though.

Thoughts on this?

karolherbst:
The OpenCL SPIR-V Environment Specification does not require the SPIR-V opcodes `OpFMod` and `OpFRem` to match any OpenCL C semantics, so implementations implementing those two instructions according to Vulkan and/or OpenGL semantics will fail those tests without actually violating any OpenCL specification.

Instead `OpExtInst fmod` should be checked against OpenCL C `fmod` as this is actually specified to match in precision by the SPIR-V Environment Specification and this is what the SPIRV-LLVM-Translator ends up using for fmod anyway.

This also allows implementations to relax their implementations of `OpFMod` and `OpFRem` to trade performance for precision, but also to allow for more consistent results between OpenCL and Vulkan using those SPIR-V instructions.

Closes #1548

Signed-off-by: Karol Herbst <kherbst@redhat.com>

### DIFF
--- a/test_conformance/spirv_new/test_op_fmath.cpp
+++ b/test_conformance/spirv_new/test_op_fmath.cpp
@@ -173,13 +173,14 @@ int test_fmath(cl_device_id deviceID,
                           lhs, rhs);                \
     }
 
-#define TEST_FMATH_MODE(TYPE, MODE)             \
-    TEST_FMATH_FUNC(TYPE, fadd, MODE)           \
-    TEST_FMATH_FUNC(TYPE, fsub, MODE)           \
-    TEST_FMATH_FUNC(TYPE, fmul, MODE)           \
-    TEST_FMATH_FUNC(TYPE, fdiv, MODE)           \
-    TEST_FMATH_FUNC(TYPE, frem, MODE)           \
-    TEST_FMATH_FUNC(TYPE, fmod, MODE)           \
+#define TEST_FMATH_MODE(TYPE, MODE)                                            \
+    TEST_FMATH_FUNC(TYPE, fadd, MODE)                                          \
+    TEST_FMATH_FUNC(TYPE, fsub, MODE)                                          \
+    TEST_FMATH_FUNC(TYPE, fmul, MODE)                                          \
+    TEST_FMATH_FUNC(TYPE, fdiv, MODE)                                          \
+// disable those tests until we figure out what the precision requierements are
+//    TEST_FMATH_FUNC(TYPE, frem, MODE)           \
+//    TEST_FMATH_FUNC(TYPE, fmod, MODE)           \
 
 #define TEST_FMATH_TYPE(TYPE)                   \
     TEST_FMATH_MODE(TYPE, regular)              \


StuartDBrady:
Thanks for looking into this!  However your change completely drops `OpFMod` and `OpFRem` coverage from the CTS which I don't think we should do. We should instead fix the reference computation or validation to cope with lower precision.  Keep in mind that `OpExtInst fmod` itself is already tested elsewhere in the `--compilation-mode spir-v` mode, which is why spirv_new _only_ tests `OpFMod` and `OpFRem` (and not `OpExtInst fmod`).

We should also extend the testing of `OpFMod` and `OpFRem` to test with negative numbers, but that can be done separately.  For `OpFMod`, this will require correcting the reference computation so that it correctly calculates a remainder with the same sign as the first operand (i.e. the dividend), as per the SPIR-V specification, in particular [§ 2.2.3. _Computation_](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_computation) (in § 2.2. _Terms_):

> _Remainder_: When dividing _a_ by _b_, a _remainder r_ is defined to be a value that satisfies _r_ + _q_ × _b_ = _a_ where _q_ is a whole number and |_r_| < |_b_|.

I have raised issue #1618 to address the missing coverage, but I do not mind whether we address the two issues together with a single commit, or with separate changes.

StuartDBrady (dismissed):
As above.

karolherbst:
> Thanks for looking into this! However your change completely drops `OpFMod` and `OpFRem` coverage from the CTS which I don't think we should do. We should instead fix the reference computation or validation to cope with lower precision. Keep in mind that `OpExtInst fmod` itself is already tested elsewhere in the `--compilation-mode spir-v` mode, which is why spirv_new _only_ tests `OpFMod` and `OpFRem` (and not `OpExtInst fmod`).
> 

The SPIR-V environment specification doesn't mention those opcodes and doesn't specify any precision requirements. So not sure how much of value it would be to test those. I think it makes sense to add proper tests for the other missing opcodes as well once we figure out if we just want to match vulkan here or do something else.

> We should also extend the testing of `OpFMod` and `OpFRem` to test with negative numbers, but that can be done separately. For `OpFMod`, this will require correcting the reference computation so that it correctly calculates a remainder with the same sign as the second operand (i.e. the divisor), as per the SPIR-V specification, in particular [§ 2.2.3. _Computation_](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_computation) (in § 2.2. _Terms_):
> 
> > _Remainder_: When dividing _a_ by _b_, a _remainder r_ is defined to be a value that satisfies _r_ + _q_ × _b_ = _a_ where _q_ is a whole number and |_r_| < |_b_|.
> 
> I have raised issue #1618 to address the missing coverage, but I do not mind whether we address the two issues together with a single commit, or with separate changes.

We can't address those anyway as we still have spec work to do on those.

StuartDBrady:
> The SPIR-V environment specification doesn't mention those opcodes and doesn't specify any precision requirements. So not sure how much of value it would be to test those. I think it makes sense to add proper tests for the other missing opcodes as well once we figure out if we just want to match vulkan here or do something else.

If we assume that there is no value in testing those operations, that would still leave the question of the value of testing _OpExtInst fmod_ twice.  Would it not make sense to remove all testing of _OpFMod_ and _OpFRem_ and raise an issue regarding the missing coverage?  Currently, the tests for _OpFMod_ and _OpFRem_ can pass even if the two are implemented identically – so IMO, our existing coverage gives false confidence.

> > I have raised issue #1618 to address the missing coverage, but I do not mind whether we address the two issues together with a single commit, or with separate changes.
> 
> We can't address those anyway as we still have spec work to do on those.

Do you mean SPIR-V spec work, or OpenCL SPIR-V environment spec work?  I agree that we need environment spec work to determine precision requirements, but the SPIR-V spec is unambiguous, IMO.  I would expect that a reference result could be calculated without knowing the full precision requirements even if we disable the test for the time being, with the intent of getting the code ready for the moment that the precision requirements are clarified.  This might not be so feasibly if we expect _OpFMod_ and _OpFRem_ to have high precision requirements, but it seems that is unlikely.  I'm not saying that we _should_ address this as a priority but I am confused as to why we apparently _cannot_ at least make progress.

karolherbst:
> > The SPIR-V environment specification doesn't mention those opcodes and doesn't specify any precision requirements. So not sure how much of value it would be to test those. I think it makes sense to add proper tests for the other missing opcodes as well once we figure out if we just want to match vulkan here or do something else.
> 
> If we assume that there is no value in testing those operations, that would still leave the question of the value of testing _OpExtInst fmod_ twice. Would it not make sense to remove all testing of _OpFMod_ and _OpFRem_ and raise an issue regarding the missing coverage? Currently, the tests for _OpFMod_ and _OpFRem_ can pass even if the two are implemented identically – so IMO, our existing coverage gives false confidence.
> 

yeah, that's probably a good alternative. I wasn't really aware of the `--compilation-mode spir-v` flag, so I wasn't looking into that yet.

I'd remove/disable the tests for now then given the spec doesn't specify it anyway.

> > > I have raised issue #1618 to address the missing coverage, but I do not mind whether we address the two issues together with a single commit, or with separate changes.
> > 
> > 
> > We can't address those anyway as we still have spec work to do on those.
> 
> Do you mean SPIR-V spec work, or OpenCL SPIR-V environment spec work? I agree that we need environment spec work to determine precision requirements, but the SPIR-V spec is unambiguous, IMO. I would expect that a reference result could be calculated without knowing the full precision requirements even if we disable the test for the time being, with the intent of getting the code ready for the moment that the precision requirements are clarified. This might not be so feasibly if we expect _OpFMod_ and _OpFRem_ to have high precision requirements, but it seems that is unlikely. I'm not saying that we _should_ address this as a priority but I am confused as to why we apparently _cannot_ at least make progress.

The OpenCL SPIR-V environment spec. The ULP values are not defined in the SPIR-V spec itself, but in the OpenCL and Vulkan specs. They are generally close, but not equal (e.g. OpenCL requires lower ULPs for FDiv). The OpenCL Env spec only defines it for 4 or 5 opcodes in total regardless.

Then I guess I'll simply comment out running the tests and leave a comment, so once we reached a conclusion (and I think people generally lean towards having them as close or identical to vulkan) we can re-enable it and add it for other opcodes as well.

StuartDBrady:
> I guess I'll simply comment out running the tests and leave a comment, so once we reached a conclusion (and I think people generally lean towards having them as close or identical to vulkan) we can re-enable it and add it for other opcodes as well.

That seems reasonable to me, although it might be good to have some confirmation from the working group.

Wondering out loud here: if we were to loosen the precision used in the _OpFRem_ and _OpFMod_ tests to that of the Vulkan environment spec, how much extra work would that be? Even if we haven't yet agreed on Vulkan's level of precision, I can't see a need for the ops to be allowed to be _less_ precise than in Vulkan, but I do not really understand the need for the _OpFRem_ and _OpFMod_ ops when using OpenCL.  If we did loosen the precision, we could perhaps add a comment for now making it clear that negative values are not tested (and maybe changing the OpFMod reference code to use the fmod function to remove the erroneous use of copysign).  If we don't have clear answer to these questions yet, I think that's okay.

I'm not going to insist on any of that, though – if you wish to simply disable the _OpFMod_/_OpFRem_ testing for now, I certainly won't object.

karolherbst:
> > I guess I'll simply comment out running the tests and leave a comment, so once we reached a conclusion (and I think people generally lean towards having them as close or identical to vulkan) we can re-enable it and add it for other opcodes as well.
> 
> That seems reasonable to me, although it might be good to have some confirmation from the working group.
> 
> Wondering out loud here: if we were to loosen the precision used in the _OpFRem_ and _OpFMod_ tests to that of the Vulkan environment spec, how much extra work would that be? Even if we haven't yet agreed on Vulkan's level of precision, I can't see a need for the ops to be allowed to be _less_ precise than in Vulkan, but I do not really understand the need for the _OpFRem_ and _OpFMod_ ops when using OpenCL. If we did loosen the precision, we could perhaps add a comment for now making it clear that negative values are not tested (and maybe changing the OpFMod reference code to use the fmod function to remove the erroneous use of copysign). If we don't have clear answer to these questions yet, I think that's okay.
> 
> I'm not going to insist on any of that, though – if you wish to simply disable the _OpFMod_/_OpFRem_ testing for now, I certainly won't object.

The main reason to match vulkan here is that it makes it way easier for layered implementations to handle those. Also, if one doesn't necessarily need the high precision the OpenCL C builtins provide, one could just use the SPIR-V ones.

Anyway, I'll try to bring this PR up on the next working group meeting.

bashbaug (approved):
LGTM as a temporary fix - matches the discussion in the working group calls.

bashbaug:
@StuartDBrady are you OK disabling the tests temporarily?

Should we keep #1548 open until we determine how much testing we can realistically do for these instructions?

PR #1615, opened 2023-01-04
test_harness: Update test status upon api failure

lakshmih:
In tests that use the harness, such as SVM svm_atomic_fetch_add, a failure in the clCreateContext call does not result in an overall fail status because of the way error codes are returned in the harness. This patch fixes it so that the conformance test correctly identifies and reports specific subtests that failed.

### DIFF
--- a/test_common/harness/testHarness.cpp
+++ b/test_common/harness/testHarness.cpp
@@ -819,6 +819,8 @@ test_status callSingleTestFunction(test_definition test,
         if (!context)
         {
             print_error(error, "Unable to create testing context");
+            gFailCount++;
+            gTestsFailed++;
             return TEST_FAIL;
         }
 
@@ -841,6 +843,8 @@ test_status callSingleTestFunction(test_definition test,
         {
             print_error(error, "Unable to create testing command queue");
             clReleaseContext(context);
+            gFailCount++;
+            gTestsFailed++;
             return TEST_FAIL;
         }
     }
@@ -888,6 +892,8 @@ test_status callSingleTestFunction(test_definition test,
         if (error)
         {
             log_error("clFinish failed: %s\n", IGetErrorString(error));
+            gFailCount++;
+            gTestsFailed++;
             status = TEST_FAIL;
         }
         clReleaseCommandQueue(queue);


MeyeChris (commented):
I can't help think that this function is a good example of why single returns lead to better code.

We probably don't want to refactor the whole function - we could but given what a nested mess it is we probably don't want to.

What we might want to do is nest it within the itself!

```
test_status callSingleTestFunction(test_definition test,
                                                       cl_device_id deviceToUse,
                                                       int forceNoContextCreation,
                                                       int numElementsToUse,
                                                       const cl_queue_properties queueProps) {
       switch (executeSingleTestFunction(test,
                                                              deviceToUse,
                                                              forceNoContextCreation,
                                                              numElementsToUse,
                                                              queueProps) {
               case TEST_SKIP:
                      // Do test skip stuff.
                      break;
               case TEST_FAIL:
                      gFailCount++;
                      gTestsFailed++;
                      break;
               case TEST_PASS:
                      gTestsPassed++;
                      break;
               case ...
                      ...
               default:
                      assert(false);
                      break;
}
```

The make the current function be executeSingleTestFunction
Then clean up the exit and error handling. The code we be more reliable, less repetitive, and easier to change.
                                   
Also is the commit not supposed to have a ticket number, and a description of the change. what and why.

bcalidas:
Would it be possible to merge the change as is to resolve a known issue and consider refactoring separately. The change itself is fairly simple and is consistent with the rest of the code in this file.
 Refactoring this code will be a complex challenge and will need to be scoped separately.

test_common/harness/testHarness.cpp
```diff
@@ -819,6 +819,8 @@ test_status callSingleTestFunction(test_definition test,
         if (!context)
         {
             print_error(error, "Unable to create testing context");
+            gFailCount++;
```
bashbaug, line 822: Elsewhere in this harness we only increment `gTestsFailed` when we return `TEST_FAIL`.  Is there some reason we increment both `gFailCount` and `gTestFailed` for these failures?

bashbaug (approved):
LGTM, just one small question.

bashbaug:

PR #1616, opened 2023-01-05
Image streams optimization

cdavis5e:
This change attempts to shorten the time it takes to run the image streams test, primarily through vectorization of the `read_image_pixel_float()` and `sample_image_pixel_float()` functions, where the most time is spent. With this, I was able to reduce the time the test takes from 45 minutes to 37 minutes.

Unfortunately, most of the time in the test is spent waiting for memory fetches, particularly in the 3D case, where the image does not fit neatly into the cache, especially with linear sampling, where pixels from two 2D slices must be sampled. Randomized offsets are likely a factor, because they throw off the hardware prefetcher. Software prefetching won't help, because that only moves around the long waits for memory; experimentally, this actually totally negates the performance improvement from vectorization.

One possible further optimization is, in the linear sampling case, to load two sampled pixels at once, which can be done for the 32 bpc formats with 256-bit vectors (e.g. Intel AVX). Unfortunately, I have little time to spare for this, and I don't know how many hours are left in the contract, but I imagine it isn't very much.

The vectorization changes currently only apply to x86 CPUs with at least SSE2, with the greatest performance gains seen with SSE4.1. Work is needed to support other CPUs' vector units--ARM (Advanced SIMD/NEON, SVE) is probably the most important of these. Another possibility that was mentioned on the call is arranging the code so that the compiler's autovectorization will kick in and do what I did here manually.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -28,6 +28,41 @@
 #if !defined(_WIN32)
 #include <cmath>
 #endif
+#ifdef __SSE2__
+#include <x86intrin.h>
+#if (defined(__GNUC__) && !defined(__clang__) && __GNUC__ < 12)                \
+    || (defined(__clang__) && !defined(__apple_build_version__)                \
+        && __clang_major__ < 8)
+// Add missing intrinsics that aren't in ancient compilers, but are used below
+#ifdef __clang__
+#define NODEBUG nodebug
+#else
+#define NODEBUG artificial
+#endif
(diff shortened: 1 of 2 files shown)


lakshmih:
We are reviewing this PR
Sreelakshmi Haridas Maruthur
Qualcomm

lakshmih:
Leads to minor improvement in execution time. We are okay with this change.

-
Sreelakshmi Haridas Maruthur
Qualcomm

bashbaug (approved):
Merging as discussed in the Februrary 7th teleconference.

Issue #142, opened 2019-04-02
cl_khr_fp16 is poorly tested


kpet:
Missing tests in most of the suites:
- [ ] bruteforce
- [ ] conversions
- [ ] relationals
- ... and probably others

EwanC:
Trying to figure out exactly what folders would be the most work to properly test this extension.
Any insights welcome.

## Bruteforce

Rather than reimplementing all the reference functions with a 32-bit single precision overload to verify half against, we could use [Ulp_Error_Half](https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_conformance/compatibility/test_common/harness/errorHelpers.c#L276) with the existing double references. Would that be admissible?

For functions which take a single parameter we can resonably test every possible 16-bit input. How many inputs should we test for functions which take multiple parameters? Not sure how this is currently done.
 
## Geometric

Looks a pain, seperate ~1000 line files for single and double precision, with a lot of duplication. C++ files rather than C, so scope for some refactoring but could be a fair bit of work.

## Relationals

Same as geometrics but not as large, seperate files for float and double which could be cleaned up into a single file for all floating point types.

Verification functions on host won't be able to use `half`, convert to float with `half2float` and do comparison there instead.

##  Conversions

Already a mass of function definitions enumerating all the possibilites that will need to be added to. Definetelty scope for refactoring this code to use templates.

This test already takes a long time, could be an issue if increase in combinations makes this worse

## Commonfns

Validation functions use type of same precision on, but host code won't be able to use `half`, upcast to float with `half2float` and do comparison there instead.

## Basic

* async_work_group_strided_copy
* async_work_group_copy
* vload/store
* astype

## Select

Glance through suggests not much to add and we can us `cl_ushort` & `cl_short` in reference functions.

## Printf

Doesn't look like too much work other than adding the format specifiers

## SPIR

Regenerate zips and update khr.csv?

## SPIRV New

Don't know much about these.

kpet:
Thanks for the initial analysis, that's great input.

I think we've reached the point where we need to take a step back and figure out how we want the code of these individual tests to be structured overall. We can't keep copy/pasting and adding special cases :). If there are specific refactorings you have in mind that you think would make it easier to add FP16 support, feel free to suggest them here, create individual issues or send PRs directly.

EwanC:
[...]

PR #1617, opened 2023-01-09
Added cl_half support for test_select

shajder:
According to issue description (select section):


test_conformance/select/util_select.cpp
```diff
@@ -242,6 +240,28 @@ void refselect_1u64u(void *dest, void *src1, void *src2, void *cmp, size_t count
         d[i] = m[i] ? y[i] : x[i];
 }
 
+void refselect_hhi(void *dest, void *src1, void *src2, void *cmp, size_t count)
+{
+    size_t i;
+    cl_short *d, *x, *y, *m;
+    d = (cl_short *)dest;
```
MeyeChris, line 247: This would be better written as 

`const cl_short *const d = (cl_short *)dest;`

1 less line and less opportunity for error, in the distant future.  Same for all, maybe to be like existing code I guess? Which I could live with.

test_conformance/select/test_select.h
```diff
@@ -28,18 +28,20 @@
...
-    kint = 5,
-    kfloat = 6,
-    kulong = 7,
-    klong = 8,
-    kdouble = 9,
-    kTypeCount  // always goes last
+    khalf = 4,
+    kuint = 5,
+    kint = 6,
+    kfloat = 7,
```
MeyeChris, line 40: Minor - but thhis change is big because of the values - however the standard says it will start at 0.

Why not remove the values - then we get more clarity when one is added or remove.

As I said minor.

test_conformance/select/util_select.cpp
```diff
@@ -724,6 +769,30 @@ size_t check_long(void *test, void *correct, size_t count, size_t vector_size) {
     return 0;
 }
 
+size_t check_half(void *test, void *correct, size_t count, size_t vector_size)
+{
+    const cl_ushort *t = (const cl_ushort *)test;
```
MeyeChris, line 774: `*const t;`

You never want t to change either, const is massively underused in general (Not general here - general in the world!)

But more importantly why are they not

'const cl_half *const t;'

So that you can do log_error?

But if so log_error should have the cast - because it is the 'broken' bit - you have pointers to cl_halfs!

MeyeChris (changes requested):

test_conformance/select/util_select.cpp
MeyeChris: My view others may differ is that if we make a change we make it as good (within reason) as possible. Sometimes the event in the future happens, and sometimes it doesn't.  So presume it doesn't and we always get code at its best.

shajder: Thanks Chris, let's clarify first one concern. Didn't you mean:

`cl_short *const d = (cl_short *)dest;`

with your additional qualifier `const` build of my project (under Ubuntu) fails.

EwanC: On the refactoring point, this is the kind of "modernization" I was thinking of from the OpenCL call discussion. Getting rid of old C89 conventions in the codebase that split variable definition & declaration, rather than using modern C++ features.

We suggested pushing separate PRs for these fp16 changes - A first one with the functional additions for fp16, and a second with the clean-up changes. However, what we could do is make them a single PR, but logically split the commits. So that the first commit is the functional addition, and a second commit is refactoring clean-ups, and later commits to address PR feedback. Then reviewers can look at the commits individually to understand the changes better, but the whole PR would be the best possible code.

However, I don't think we should start pushing clean-up changes to this PR in particular right now, until the in progress work by Holochip on the same test is merged (or even pushed), since we don't want conflicts and the work to be wasted.

EwanC (commented):
Tried this PR on Codeplay's implementation and can see the new tests pass (and skip correctly against an implementation without cl_khr_fp16), but need to review the PR in more detail.

EwanC:
> Tried this PR on Codeplay's implementation and can see the new tests pass (and skip correctly against an implementation without cl_khr_fp16), but need to review the PR in more detail.

To follow up, I've looked through this again, and happy that how it adds half support is consistent with the other tests.

How/when to do the refactoring to make it nicer is another question.

test_conformance/select/util_select.cpp
shajder: @MeyeChris I've went through all similar cases and added requested consts

EwanC (approved):
> @EwanC due to the fact my project comes to an end and I can't see any progress around https://github.com/KhronosGroup/OpenCL-CTS/pull/1665 I am blocked with I decided to move on.
>
> Except of corrections for above requests I've added optimization which seems to speed up the test ~18% at my machine. Moreover, I've replaced CL objects with specific wrappers.

Had a look through this again, and happy with where we've got to with this change.

svenvh (approved):

svenvh:
LGTM, but please remove the excess semicolon that Ewan pointed out.

test_conformance/select/test_select.cpp
```diff
@@ -66,6 +69,17 @@ static void printUsage( void );
 #define BUFFER_SIZE (1024*1024)
 #define KPAGESIZE 4096
 
+#define test_error_count(errCode, msg)                                         \
+    {                                                                          \
+        auto errCodeResult = errCode;                                          \
+        if (errCodeResult != CL_SUCCESS)                                       \
+        {                                                                      \
+            gFailCount++;                                                      \
+            ;                                                                  \
```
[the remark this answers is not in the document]
shajder: Done

svenvh (approved):

EwanC (dismissed):

test_conformance/select/test_select.cpp
```diff
@@ -396,111 +388,66 @@ static int doTest(cl_command_queue queue, cl_context context, Type stype, Type c
...
-        {
-            log_error("Error: Reading buffer from cmp to cmp_host failed\n");
-            ++s_test_fail;
-            goto exit;
-        }
+        initSrcBuffer(src1_host.data(), stype, d);
+        initSrcBuffer(src2_host.data(), stype, d);
+        initCmpBuffer(cmp_host.data(), cmptype, i * cmp_stride, block_elements);
+
+        clEventWrapper user_event = clCreateUserEvent(context, &err);
```
bashbaug, line 398: As far as I can tell this user event is never used.  Can it be removed?

bashbaug (dismissed):
LGTM, aside from one minor comment.

I didn't do a super detailed code review but I did verify that the additions pass on our implementation and I did not see any change in coverage (aside from the added half tests).  Thanks!

test_conformance/select/test_select.cpp
shajder: event removed

> I did not see any change in coverage ...

apparently you are right, my measurement must have been skewed, sorry for confusion

shajder:
Close/open to trigger CI

Issue #1579, opened 2022-11-24
cl_khr_extended_async_copies are size not type based.


ghost:
The latest version of the cl_khr_extended_async_copies extension uses element size rather the element type as its base. The means it can be called with arbitrary and in particular non power of 2 sizes, such as 3 or 13.

The CTS cl_khr_extended_async_copies tests need to be updated be size based rather than type based.

I will push a patch to do this in the next 24 hours as a suggestion.

MeyeChris:
Patch is here: https://gitlab.khronos.org/opencl/conformance-tests/-/merge_requests/486

Not sure is this is the correct location - I may push to github as well.

rjodinchr:
Fixed by PR #1619 (commit `19f4fc3f`), which updated `test_async_copy2D.cpp` and `test_async_copy3D.cpp` to test `cl_khr_extended_async_copies` based on arbitrary element byte sizes rather than primitive types, conforming to the updated extension specification.

PR #1619, opened 2023-01-10
Make extended_async_copy tests type agnostic

MeyeChris:
The latest version of the cl_khr_extended_async_copies extension uses element size rather the element type as its base. The means it can be called with arbitrary and in particular non power of 2 sizes, such as 3 or 13.

Update the test_async_copy2D and test_async_copy3D tests to make them element size based rather than type based.

As well as this run all tests that can fit into the memory of the target rather than presumed large elements cannot fit.

Make some addtional good practice changes in terms of const usage, declaring variables where they are use, and usage of iterators.

The test coverage increases from 1224 cases to 1332 cases for the test_async_copy2D and test_async_copy3D cases.

Ticket: #1579

Signed-off-by: Chris Gearing <chris.gearing@mobileye.com>

### DIFF
--- a/test_conformance/basic/test_async_copy2D.cpp
+++ b/test_conformance/basic/test_async_copy2D.cpp
@@ -27,17 +27,25 @@
 
 static const char *async_global_to_local_kernel2D = R"OpenCLC(
 #pragma OPENCL EXTENSION cl_khr_extended_async_copies : enable
-%s // optional pragma string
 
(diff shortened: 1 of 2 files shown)


lakshmih:
Still reviewing.

lakshmih:
Looks good.

lakshmih (approved):

lakshmih:
Can we get one more review and merge this or take the label off since it has been in review for two months now?

PR #1620, opened 2023-01-12
Subgroup test fixes

pjaaskel:
A couple of small subgroup-related fixes.

### DIFF
--- a/test_conformance/api/test_sub_group_dispatch.cpp
+++ b/test_conformance/api/test_sub_group_dispatch.cpp
@@ -188,8 +188,9 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
         }
     }
 
-    // test when input subgroup count exceeds max wg size
-    size_t large_sg_size = kernel_subgroup_count + 1;
+    // test when input subgroup count exceeds max wg size:
+    // there can be at most the local size of (1 WI) subgroups
+    size_t large_sg_size = max_local + 1;
     error = clGetKernelSubGroupInfo(kernel, deviceID, CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT, sizeof(size_t), &large_sg_size, sizeof(ret_ndrange1d), &ret_ndrange1d, &realSize);
         test_error(error, "clGetKernelSubGroupInfo failed for CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT");
     if (ret_ndrange1d != 0)


lakshmih (approved):

test_conformance/api/test_sub_group_dispatch.cpp
```diff
@@ -189,7 +189,7 @@ int test_sub_group_dispatch(cl_device_id deviceID, cl_context context, cl_comman
     }
 
     // test when input subgroup count exceeds max wg size
-    size_t large_sg_size = kernel_subgroup_count + 1;
+    size_t large_sg_size = max_local + 1;
     error = clGetKernelSubGroupInfo(kernel, deviceID, CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT, sizeof(size_t), &large_sg_size, sizeof(ret_ndrange1d), &ret_ndrange1d, &realSize);
```
bashbaug, line 193: I'm not sure that this is the right fix, though I do agree there is a problem here.

The value that is passed into `clGetKernelSubGroupInfo` for `CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT` should be a count of sub-groups, not a count of work-items.  This means that a) the comment is wrong, but also b) we shouldn't be passing in `max_local` (which is the queried value for `CL_DEVICE_MAX_WORK_GROUP_SIZE`) either.

Perhaps we should be querying for `CL_KERNEL_MAX_NUM_SUB_GROUPS` and passing in one greater than the queried value as this test value?

test_conformance/subgroups/subhelpers.h
```diff
@@ -1518,7 +1518,8 @@ template <typename Ty, typename Fns, size_t TSIZE = 0> struct test
 
         // Limit it a bit so we have muliple work groups
         // Ideally this will still be large enough to give us multiple
-        if (local > test_params.local_workgroup_size)
+        if (local > test_params.local_workgroup_size
+            && global % test_params.local_workgroup_size == 0)
```
bashbaug, line 1522: Can you say a little more what this change is doing and why it is required?  Thanks!

test_conformance/api/test_sub_group_dispatch.cpp
pjaaskel: To my understanding the test on purpose sets a SG count to be a larger number than the max work-items in the work group. Given the minimum SG size is 1 WI, it means that there can be maximum of maximum work-group size of SGs (of 1 WI of size). Thus, if we request a number of SGs that exceeds the local size, the query should fail as expected. Yes, your proposal would make it more explicit, but to my understanding this check is correct as well.

lakshmih (approved):

test_conformance/api/test_sub_group_dispatch.cpp
bashbaug: Thanks, this makes sense to me.

I'd find it easier to understand what this test is doing if the name of the variable were `large_sg_count` instead of `large_sg_size`, but this isn't something you've changed and I agree the test logic is correct now.

bashbaug (approved):

PR #1622, opened 2023-01-17
Added cl_half support for test_printf

shajder:
According to issue description (printf section):

### DIFF
--- a/test_common/harness/kernelHelpers.cpp
+++ b/test_common/harness/kernelHelpers.cpp
@@ -1526,7 +1526,39 @@ cl_device_fp_config get_default_rounding_mode(cl_device_id device)
     if (0 == (single & CL_FP_ROUND_TO_ZERO))
         test_error_ret(-1,
                        "FAILURE: device must support either "
-                       "CL_DEVICE_SINGLE_FP_CONFIG or CL_FP_ROUND_TO_NEAREST",
+                       "CL_FP_ROUND_TO_ZERO or CL_FP_ROUND_TO_NEAREST",
+                       0);
+
+    // Make sure we are an embedded device before allowing a pass
+    if ((error = clGetDeviceInfo(device, CL_DEVICE_PROFILE, sizeof(profileStr),
+                                 &profileStr, NULL)))
+        test_error_ret(error, "FAILURE: Unable to get CL_DEVICE_PROFILE", 0);
+
+    if (strcmp(profileStr, "EMBEDDED_PROFILE"))
+        test_error_ret(error,
+                       "FAILURE: non-EMBEDDED_PROFILE devices must support "
+                       "CL_FP_ROUND_TO_NEAREST",
+                       0);
+
+    return CL_FP_ROUND_TO_ZERO;
+}
+
+cl_device_fp_config get_default_rounding_mode_half(cl_device_id device)
+{
+    char profileStr[128] = "";
+    cl_device_fp_config config = 0;
+    int error = clGetDeviceInfo(device, CL_DEVICE_HALF_FP_CONFIG,
+                                sizeof(config), &config, NULL);
+    if (error)
+        test_error_ret(error, "Unable to get device CL_DEVICE_HALF_FP_CONFIG",
+                       0);
+
+    if (config & CL_FP_ROUND_TO_NEAREST) return CL_FP_ROUND_TO_NEAREST;
+
+    if (0 == (config & CL_FP_ROUND_TO_ZERO))
+        test_error_ret(-1,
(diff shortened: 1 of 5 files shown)


test_common/harness/kernelHelpers.cpp
```diff
@@ -1511,22 +1511,33 @@ size_t get_min_alignment(cl_context context)
...
+        test_error_ret(
+            -1,
+            "FAILURE: CL_DEVICE_DOUBLE_FP_CONFIG not supported by this routine",
+            0);
+
     char profileStr[128] = "";
     cl_device_fp_config single = 0;
-    int error = clGetDeviceInfo(device, CL_DEVICE_SINGLE_FP_CONFIG,
-                                sizeof(single), &single, NULL);
+    int error = clGetDeviceInfo(device, param, sizeof(single), &single, NULL);
```
EwanC, line 1525: I'm getting here an error here on a device which doesn't support cl_khr_fp16, but trying to query `CL_DEVICE_HALF_FP_CONFIG`.

From a call stack starting from `generateRef` on test_printf.cpp:1194 then calling `get_default_rounding_mode(device, CL_DEVICE_HALF_FP_CONFIG)`

test_conformance/printf/util_printf.cpp
```diff
@@ -593,23 +741,27 @@ std::vector<printDataGenParameters> printVectorGenParameters = {
 
     //(Minimum)Two-wide,two positions after decimal
 
-    {NULL,"(1.0f,2.0f,3.0f,4.0f)","%2.2","hlf","float","4"},
+    { NULL, "(1.0h,2.0h,3.0h,4.0h)", "%2.2", "hlf", "half", "4" },
```
EwanC, line 744: Could we put this new test in the list at index 5 so that it corresponds to the new `vector_5` test that's been added, rather than changing what the `vector_0` -> `vector_4` tests do.

This test also fails on Codeplay's half implementation (it's the only new test which does) but I need to look into if that's an implementation issue or test issue.

shajder: Done

test_conformance/printf/util_printf.cpp
```diff
@@ -593,24 +741,27 @@ std::vector<printDataGenParameters> printVectorGenParameters = {
...
 
-    //Alternative form,long argument
+    // Alternative form,long argument
 
-    {NULL,"(12345678,98765432)","%","ld","long","2"}
+    { NULL, "(12345678,98765432)", "%", "ld", "long", "2" },
 
+    //(Minimum)Two-wide,two positions after decimal
+
+    { NULL, "(1.0h,2.0h,3.0h,4.0h)", "%2.2", "hlf", "half", "4" }
```
EwanC, line 764: Think rather than `hlf` (where `hl` seems to be for float) we should use `hf` instead:

> hl This modifier can only be used with the vector specifier. Specifies that a following d, i, o, u, x, or X conversion specifier applies to a intn or uintn argument; that a following a, A, e, E, f, F, g, or G conversion specifier applies to a floatn argument.

> h Specifies that a following d, i, o, u, x, or X conversion specifier applies to a shortn or ushortn argument (the argument will not be promoted); that a following a, A, e, E, f, F, g, or G conversion specifier applies to a halfn [62] argument.

cc- @frasercrmck  who found this

frasercrmck: I think all vector literals should be explicitly cast to the vector type. For example here we should have `(half4)((1.0h,2.0h,3.0h,4.0h))`.

> A vector literal is written as a parenthesized vector type followed by a parenthesized comma delimited list of parameters.

EwanC: > I think all vector literals should be explicitly cast to the vector type. For example here we should have `(half4)((1.0h,2.0h,3.0h,4.0h))`.
> 

I think the cast is happening already, based on the [kernel code](https://github.com/KhronosGroup/OpenCL-CTS/blob/0bcb99088def37a710da0459d81bab6921d4464b/test_conformance/printf/test_printf.cpp#L255-L276) which will create a variable called `tmp` with the cast type (e.g `half4`), and then `tmp` is passed to printf

frasercrmck: Right you are! Sorry for the noise.

shajder: @EwanC I've applied `hf` correction. Test passes with Intel HDG driver.

EwanC (approved):

svenvh (approved):

bashbaug:

PR #1625, opened 2023-01-24
Fix semaphore tests

ahesham-arm:
Semaphore properties are null-terminated
lists. `CL_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR` is a null-terminated property. Each should be separately null-terminated.

Signed-off-by: Ahmed Hesham <ahmed.hesham@arm.com>


ahesham-arm:
The failing check is not related to the change. It was addressed in [this pull request](url). It now needs to be re-ran, which I do not have the permissions to do.

paulfradgley:
Fixed as part of PR #1675

ahesham-arm:
Closing the PR as it was fixed as part of https://github.com/KhronosGroup/OpenCL-CTS/pull/1675

PR #1628, opened 2023-01-27
[NFC] Fix BasicCommandBufferTest initialization order

svenvh:
The `command_buffer` member is initialized with `this`, and should thus be initialized after all other members have been initialized, to avoid any potential uninitialized data accesses by `command_buffer`'s constructor.  This is currently not a problem, but would become a problem when `command_buffer`'s constructor accesses the object through its parameter.

Fixes a `-Wreorder` warning.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
@@ -25,12 +25,12 @@ BasicCommandBufferTest::BasicCommandBufferTest(cl_device_id device,
                                                cl_context context,
                                                cl_command_queue queue)
     : CommandBufferTestBase(device), context(context), queue(nullptr),
-      num_elements(0), command_buffer(this), simultaneous_use_support(false),
+      num_elements(0), simultaneous_use_support(false),
       out_of_order_support(false),
       // try to use simultaneous path by default
       simultaneous_use_requested(true),
       // due to simultaneous cases extend buffer size
-      buffer_size_multiplier(1)
+      buffer_size_multiplier(1), command_buffer(this)
 
 {
     cl_int error = clRetainCommandQueue(queue);
--- a/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.h
+++ b/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.h
@@ -53,7 +53,6 @@ protected:
 
     cl_context context;
     clCommandQueueWrapper queue;
-    clCommandBufferWrapper command_buffer;
     clProgramWrapper program;
     clKernelWrapper kernel;
     clMemWrapper in_mem, out_mem, off_mem;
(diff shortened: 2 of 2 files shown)


svenvh:
cc: @shajder

aharon-abramson (approved):

bashbaug (commented):
This change is fine but it's rather confusing and inconsistent to have the command buffer wrapper be constructed from a `CommandBufferTestBase*` rather than a `cl_command_buffer_khr` like the other wrappers.  Are there any plans or ideas to fix this?

bashbaug:
Note, even with this initialization order fix we still have the potential for problems because the extension function pointers accessed via "this" aren't valid until `init_extension_functions` are called, which doesn't happen until `SetUp` is called.

svenvh:
> This change is fine but it's rather confusing and inconsistent to have the command buffer wrapper be constructed from a CommandBufferTestBase* rather than a cl_command_buffer_khr like the other wrappers. Are there any plans or ideas to fix this?

I haven't considered changing the overall class hierarchy, as I'm primarily focusing on solving compiler warnings for now.  Perhaps the original author(s) could look into this?

> Note, even with this initialization order fix we still have the potential for problems because the extension function pointers accessed via "this" aren't valid until init_extension_functions are called, which doesn't happen until SetUp is called.

At least the extension function pointers are initialized to `nullptr`, which is arguably better than potentially accessing uninitialized members (which this PR solves).  I see your point and wonder why the extension function pointers aren't initialized by the constructor, but I'd leave that out of this PR.

EwanC:
Think we should merge this PR as-is to avoid expanding its scope past fixing a warning, but to add to the discussion.

> This change is fine but it's rather confusing and inconsistent to have the command buffer wrapper be constructed from a CommandBufferTestBase* rather than a cl_command_buffer_khr like the other wrappers. Are there any plans or ideas to fix this?

I came up with this design originally due to the command-buffer wrapper needing the release/retain functions queried from the platform for function points, which I don't believe is the case with any other wrappers. I chose to get this function pointer information by constructing via a `CommandBufferTestBase`. If we don't do it that way, I think we'd need another way to pass into the wrapper either the platform to query for `clReleaseCommandBufferKHR`/`clRetainCommandBufferKHR` function pointers, or the function pointers themselves. I'm not aware of any plans to improve this approach, but we can create an Issue to revisit it if you think the wrapper has scope for improvement.

> Note, even with this initialization order fix we still have the potential for problems because the extension function pointers accessed via "this" aren't valid until init_extension_functions are called, which doesn't happen until SetUp is called.

Hadn't noticed this potential issue before, to resolve it I think we could move `init_extension_functions` from `BasicCommandBufferTest` Setup into the `BasicCommandBufferTest` constructor, and initialize the command_buffer member afterwards. I can look into that once this PR is merged.

bashbaug (approved):
I do think it would be worth looking at improvements to the wrapper because we're going to have a similar problem for semaphores (and any other objects that are retained or released by extension functions).  I agree though that this goes far beyond the warning fix, which is the intent of this PR.  Thanks!

PR #1632, opened 2023-02-07
Pass 'flag' variable through to clGetImageRequirementsInfoEXT

paulfradgley:
Almost all of the image_from_buffer cl_ext_image_requirements_info tests pass '0' as the 'flags' to clGetImageRequirementsInfoEXT.

Passing '0' as the flags is valid according to the spec, but since the tests are calling clGetImageRequirementsInfoEXT in a loop over a list of flags it doesn't seem intended that all the tests pass '0'.


ouakheli (approved):
The flag was not used in the query but is should be passed

ouakheli (approved):

bashbaug (commented):
Nice catch, this fix looks good.  I have a few follow-on questions though:

1. Is it OK that the `flag == 0` is never  tested now?  I don't see it in the `flagTypes` array.
2. Do we really want to be mucking with the `flag` in the inner loop iteration?

```c++
                flag = (flag == CL_MEM_KERNEL_READ_AND_WRITE)
                    ? CL_MEM_READ_WRITE
                    : flag;
```

As-written, I think we'll a) never test `clGetImageRequirementsInfoExt` with the flag value `CL_MEM_KERNEL_READ_AND_WRITE` (and I think this is valid?), and b) because we're modifying the flag valued used outside of the loop we'll only test `CL_MEM_KERNEL_READ_AND_WRITE` for the first image type because subsequent image types will instead have a flag value `CL_MEM_READ_WRITE`.

paulfradgley:
> 1. Is it OK that the `flag == 0` is never  tested now?  I don't see it in the `flagTypes` array.

We should test it, but maybe as a separate test case?

> 2. Do we really want to be mucking with the `flag` in the inner loop iteration?

I've fixed the inner loop modifying the flag variable for subsequent iterations.
I noticed that the flag override happens after the flag is passed to get_format_list though. Is that intended?

bashbaug (approved):
This is definitely an improvement and I suspect this reflects the intent of these tests.

It'd be good to have an approval from somebody who implements this extension though.

Do we want to file any issues for the gaps described above?

lakshmih (dismissed):

bashbaug:
Discussed in the March 7th teleconference.  Will merge as soon as CI is clean - suggested rebasing to see if this "fixes" the code formatting failure.

PR #1635, opened 2023-02-10
[NFC] Fix -Wparentheses warnings

svenvh:
Some warnings were recently introduced by b73c3149 ("Image streams optimization (#1616)", 2023-02-07).  Use parentheses to explicitly state precedence when mixing different bitwise operators, as people easily get confused about their precedence.


svenvh:
@cdavis5e I suppose these changes reflect what you intended with the original code?

cdavis5e:
> @cdavis5e I suppose these changes reflect what you intended with the original code?

Yes, that is correct.

bashbaug (approved):
LGTM, need to decide if we merge this or back out the changes (for now).

lakshmih (approved):

svenvh:
Won't need this any longer, because the breaking commit got reverted in #1638 .

PR #1645, opened 2023-02-18
Add external semaphore tests.

pj87:
Added basic test to check the functionality of cl_ext_external_semaphore_test extension. Tests rely on Vulkan API as the producer of the external semaphore functionality. The Vulkan wrapper has been copied from the Vulkan CTS tests and minor changes have been introduced. A separate copy of the Vulkan wrapper is for breaking any dependencies between different CTS modules and thus make the build system simpler.


nikhiljnv:
Consider moving common vulkan wrapper files to a more common place say OpenCL-CTS/test_common or OpenCL-CTS/test_conformance/extensions/common. These will be needed by cl_khr_external_memory and related extensions too.

nikhiljnv:
Also, if we are adding these tests under test_conformance/extensions, we should consider moving test_conformance/vulkan to test_conformance/extensions/vulkan as well.

test_conformance/extensions/cl_ext_external_semaphore/test_external_semaphore.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+    // Ensure all events are completed
+    test_assert_event_complete(signal_event);
+    test_assert_event_complete(wait_1_event);
+    test_assert_event_complete(wait_2_event);
+
+    return TEST_PASS;
+}
+
+int test_external_semaphores_in_order_queue(cl_device_id deviceID,
```
nikhiljnv, line 196: Can we reuse similar scenarios from cl_khr_semaphore by just replacing semaphore created within OpenCL by semaphore imported from vulkan?

nikhiljnv (changes requested):

test_conformance/extensions/cl_ext_external_semaphore/test_external_semaphore.cpp
pj87: > Consider moving common vulkan wrapper files to a more common place say OpenCL-CTS/test_common or OpenCL-CTS/test_conformance/extensions/common. These will be needed by cl_khr_external_memory and related extensions too.

I totally agree. Do we want to do it in a scope of current task or create a new one for this?

nikhiljnv:
> I totally agree. Do we want to do it in a scope of current task or create a new one for this?

We can do it via separate PR. Don't think we will need a separate issue, but can file one if that helps.

nikhiljnv:
@pj87 
Based on my review, it seems that there are three parts to this PR

1. The vulkan wrapper files are now moved to test_conformance/common/vulkan_wrapper and in a common library
2. test_vulkan is updated to use vulkan_wrapper sources from test_conformance/common/vulkan_wrapper and common lib built from it
3. Earlier changes to add external semaphore tests under test_conformance/extensions/cl_ext_external_semaphore/, but updated to use wrapper sources and lib from 1.

All good work. Some suggestions -

1. Can we move 1 and 2 into a separate PR and keep 3 that depends on this as a separate PR? May be cleaner and would help getting parts of it merged quicker. For 1 and 2, we need to just ensure build sanity.  3 is additional functionality/test_coverage which may need some validation from implementers. This will also allow moving test_conformance/vulkan to test_conformance/extensions/vulkan independent of 3. (We don't need to hold on the PR for this if we get approvals for 3 before we spend cycles on this refactor, we can get this PR in as is)
2. Can we rename cl_ext_external_semaphore dir to cl_khr_external_semaphore to be consistent with extension name  (similar to what we have for cl_khr_semaphore) ?
3. Given that vulkan_wrapper will be used by tests based on external semaphore/memory and similar other extensions and generally won't be required for core OpenCL spec coverage, it may be a good idea to move test_conformance/common/vulkan_wrapper  to test_conformance/extensions/common/vulkan_wrapper. (Again don't need to hold this PR for too long just for this, we can do this in a subsequent PR )

nikhiljnv (approved):
LGTM.

bashbaug:
Merging as discussed in the March 21st teleconference.

PR #1630, opened 2023-02-03
Enable test_vulkan by default in presubmit

nikhiljnv:
Enable test_vulkan by default in presubmit

test_vulkan is disabled by default as part of
PR #1530.
Given the ongoing development to test_vulkan,
it is important to have this enabled in CI builds
to avoid breaking test builds.
Enable the test build as part of presubmit script.

PR #1648, opened 2023-02-21
Add tests for external sharing not dependant on semaphores.

pj87:
Additional external sharing tests that use fences instead of semaphores.

### DIFF
--- a/test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
+++ b/test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
@@ -615,6 +615,16 @@ VulkanQueue::VulkanQueue(VkQueue vkQueue): m_vkQueue(vkQueue) {}
 
 VulkanQueue::~VulkanQueue() {}
 
+void VulkanQueue::submit(const VulkanCommandBuffer &commandBuffer, const VkFence &fence)
+{
+    VkSubmitInfo vkSubmitInfo = {};
+    vkSubmitInfo.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
+    vkSubmitInfo.pNext = NULL;
+    vkSubmitInfo.waitSemaphoreCount = (uint32_t)0;
+
+    vkQueueSubmit(m_vkQueue, 1, &vkSubmitInfo, fence);
+}
+
 void VulkanQueue::submit(const VulkanSemaphoreList &waitSemaphoreList,
                          const VulkanCommandBufferList &commandBufferList,
                          const VulkanSemaphoreList &signalSemaphoreList)
--- a/test_conformance/vulkan/CMakeLists.txt
+++ b/test_conformance/vulkan/CMakeLists.txt
@@ -23,6 +23,7 @@ include_directories (${CLConform_INCLUDE_DIR})
 set (${MODULE_NAME}_SOURCES
         main.cpp
         test_vulkan_interop_buffer.cpp
+        test_vulkan_interop_buffer_fence.cpp
         test_vulkan_interop_image.cpp
         test_vulkan_api_consistency.cpp
         test_vulkan_platform_device_info.cpp
--- a/test_conformance/vulkan/main.cpp
+++ b/test_conformance/vulkan/main.cpp
@@ -53,6 +53,8 @@ static void params_reset()
 
 extern int test_buffer_common(cl_device_id device_, cl_context context_,
                               cl_command_queue queue_, int numElements_);
+extern int test_buffer_common_fence(cl_device_id device_, cl_context context_,
+                                    cl_command_queue queue_, int numElements_);
 extern int test_image_common(cl_device_id device_, cl_context context_,
                              cl_command_queue queue_, int numElements_);
 
@@ -90,6 +92,44 @@ int test_buffer_multiImport_diffCtx(cl_device_id device_, cl_context context_,
              "IN DIFFERENT CONTEXT...... \n\n");
     return test_buffer_common(device_, context_, queue_, numElements_);
 }
+int test_buffer_single_queue_fence(cl_device_id device_, cl_context context_,
+                                   cl_command_queue queue_, int numElements_)
+{
+    params_reset();
+    log_info("RUNNING TEST WITH ONE QUEUE...... \n\n");
+    return test_buffer_common_fence(device_, context_, queue_, numElements_);
+}
+int test_buffer_multiple_queue_fence(cl_device_id device_, cl_context context_,
+                                     cl_command_queue queue_, int numElements_)
+{
+    params_reset();
+    numCQ = 2;
+    log_info("RUNNING TEST WITH TWO QUEUE...... \n\n");
+    return test_buffer_common_fence(device_, context_, queue_, numElements_);
+}
+int test_buffer_multiImport_sameCtx_fence(cl_device_id device_,
+                                          cl_context context_,
+                                          cl_command_queue queue_,
+                                          int numElements_)
+{
+    params_reset();
+    multiImport = true;
+    log_info("RUNNING TEST WITH MULTIPLE DEVICE MEMORY IMPORT "
+             "IN SAME CONTEXT...... \n\n");
+    return test_buffer_common_fence(device_, context_, queue_, numElements_);
+}
+int test_buffer_multiImport_diffCtx_fence(cl_device_id device_,
+                                          cl_context context_,
+                                          cl_command_queue queue_,
+                                          int numElements_)
+{
+    params_reset();
+    multiImport = true;
+    multiCtx = true;
+    log_info("RUNNING TEST WITH MULTIPLE DEVICE MEMORY IMPORT "
+             "IN DIFFERENT CONTEXT...... \n\n");
+    return test_buffer_common_fence(device_, context_, queue_, numElements_);
+}
 int test_image_single_queue(cl_device_id device_, cl_context context_,
                             cl_command_queue queue_, int numElements_)
 {
@@ -106,17 +146,24 @@ int test_image_multiple_queue(cl_device_id device_, cl_context context_,
     return test_image_common(device_, context_, queue_, numElements_);
 }
 
-test_definition test_list[] = { ADD_TEST(buffer_single_queue),
-                                ADD_TEST(buffer_multiple_queue),
-                                ADD_TEST(buffer_multiImport_sameCtx),
-                                ADD_TEST(buffer_multiImport_diffCtx),
-                                ADD_TEST(image_single_queue),
-                                ADD_TEST(image_multiple_queue),
-                                ADD_TEST(consistency_external_buffer),
-                                ADD_TEST(consistency_external_image),
-                                ADD_TEST(consistency_external_semaphore),
-                                ADD_TEST(platform_info),
-                                ADD_TEST(device_info) };
+test_definition test_list[] = {
+
+    ADD_TEST(buffer_single_queue),
+    ADD_TEST(buffer_multiple_queue),
+    ADD_TEST(buffer_multiImport_sameCtx),
+    ADD_TEST(buffer_multiImport_diffCtx),
+    ADD_TEST(buffer_single_queue_fence),
+    ADD_TEST(buffer_multiple_queue_fence),
+    ADD_TEST(buffer_multiImport_sameCtx_fence),
+    ADD_TEST(buffer_multiImport_diffCtx_fence),
+    ADD_TEST(image_single_queue),
+    ADD_TEST(image_multiple_queue),
+    ADD_TEST(consistency_external_buffer),
+    ADD_TEST(consistency_external_image),
+    ADD_TEST(consistency_external_semaphore),
+    ADD_TEST(platform_info),
+    ADD_TEST(device_info)
+};
 
 const int test_num = ARRAY_SIZE(test_list);
 
--- a/test_conformance/vulkan/test_vulkan_interop_buffer_fence.cpp
+++ b/test_conformance/vulkan/test_vulkan_interop_buffer_fence.cpp
@@ -79,10 +79,12 @@ __kernel void checkKernel(__global unsigned char *ptr, int size, int expVal, __g
         }                              \n\
 }";
 
-int run_test_with_two_queue(cl_context &context, cl_command_queue &cmd_queue1,
-                            cl_command_queue &cmd_queue2, cl_kernel *kernel,
-                            cl_kernel &verify_kernel, VulkanDevice &vkDevice,
-                            uint32_t numBuffers, uint32_t bufferSize)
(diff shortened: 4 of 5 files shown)


nikhiljnv:
Will need to merge https://github.com/KhronosGroup/OpenCL-CTS/pull/1630 before merging any changes to test_vulkan.

nikhiljnv:
Can we make VkFence + clFinish a synchronization option to existing tests rather than creating a separate test that uses fence?
That way we can extend existing coverage across both modes (synchronization via external semaphores vs synchronization via VkFence + clFinish)

nikhiljnv:
@pj87 
Looks like there will be some conflicts between this PR and part of #1645 due to refactoring of vulkan_wrapper moved into common place / lib.
We can go either ways, but do we have a preference which one we want to merge first? We will need to update other one based on the sequence.
Given that this is still in drafts, should we focus on #1645 first?

test_conformance/vulkan/vulkan_interop_common/vulkan_wrapper.cpp
```diff
@@ -615,6 +615,17 @@ VulkanQueue::VulkanQueue(VkQueue vkQueue): m_vkQueue(vkQueue) {}
 
 VulkanQueue::~VulkanQueue() {}
 
+void VulkanQueue::submit(const VulkanCommandBuffer &commandBuffer,
+                         const VkFence &fence)
+{
+    VkSubmitInfo vkSubmitInfo = {};
+    vkSubmitInfo.sType = VK_STRUCTURE_TYPE_SUBMIT_INFO;
+    vkSubmitInfo.pNext = NULL;
+    vkSubmitInfo.waitSemaphoreCount = (uint32_t)0;
```
nikhiljnv, line 624: Command buffer links are missing in vkSubmitInfo.
See below code at line number 642 and 643 for reference.
vkSubmitInfo.commandBufferCount = (uint32_t)commandBufferList.size();
    vkSubmitInfo.pCommandBuffers = commandBufferList();

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -229,14 +251,37 @@ int run_test_with_two_queue(cl_context &context, cl_command_queue &cmd_queue1,
 
                 if (iter == 0)
                 {
-                    vkQueue.submit(vkCommandBuffer, vkVk2CLSemaphore);
+                    if (use_fence)
+                    {
+                        vkQueue.submit(vkCommandBuffer, fence);
```
nikhiljnv, line 256: Will need similar changes to other files especially to test_vulkan_interop_image.cpp

nikhiljnv (changes requested):
Some early comments.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
pj87: So as I aready mentioned at the beginnig of the task, I don't know how it should be done for the image part. I tried changing it in this commit accordingly for bufffer and image but I ended up with deadlocks:
https://github.com/KhronosGroup/OpenCL-CTS/commit/9e83f008f5469af2e13ff4f1c7602351485d14ea#diff-2b3fe9aac0e9f06312676bdb7212bb775c8b1386f633d73db32aebb87b283f3f
Do you have any idea on how it should be implemented?

nikhiljnv: Will take a look. We can cover image changes later if there are more issues to address.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -229,14 +251,37 @@ int run_test_with_two_queue(cl_context &context, cl_command_queue &cmd_queue1,
...
+                    }
+                    else
+                    {
+                        vkQueue.submit(vkCommandBuffer, vkVk2CLSemaphore);
+                    }
+                }
+                else
+                {
+                    if (use_fence)
+                    {
```
nikhiljnv, line 266: Fence needs to be reset before reusing after previous Wait.

pj87: I've added it but now there are deadlocks there.

nikhiljnv: The sequence should be "Reset fence, use it in some submit operation where it gets signaled, wait on it" or  "use fence in some submit operation where it gets signaled, wait on it, Reset fence". Resetting just before wait but after signal would override effect of signalling at the end of completion of work submitted earlier and wait would not wait for the same.

pj87: After this change there is an issue with:

```
err = clEnqueueReadBuffer(cmd_queue1, error_1, CL_TRUE, 0,
                                              sizeof(uint8_t), error_3, 0, NULL,
                                              NULL);
```

it produces the following output:

```
log_error("&&&& vulkan_opencl_buffer test FAILED\n");
```

Yet I am not sure what the reason is.

bashbaug: Discussed May 9th.  @nikhiljnv will try to reproduce these failures.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -288,7 +323,14 @@ int run_test_with_two_queue(cl_context &context, cl_command_queue &cmd_queue1,
 
                 if (iter != (maxIter - 1))
                 {
-                    clCl2VkExternalSemaphore->signal(cmd_queue2);
+                    if (use_fence)
+                    {
+                        fence->wait();
```
nikhiljnv, line 328: Need clFinish here instead of fence wait to ensure OpenCL work is done before starting vulkan work  submission.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -564,7 +642,15 @@ int run_test_with_one_queue(cl_context &context, cl_command_queue &cmd_queue1,
                 }
                 if (iter != (maxIter - 1))
                 {
-                    clCl2VkExternalSemaphore->signal(cmd_queue1);
+                    if (use_fence)
+                    {
+                        fence->wait();
```
nikhiljnv, line 647: I think fence->wait can be removed and we only need clFinish here.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -876,7 +1001,14 @@ int run_test_with_multi_import_same_ctx(
                     }
                     if (iter != (maxIter - 1))
                     {
-                        clCl2VkExternalSemaphore->signal(cmd_queue1);
+                        if (use_fence)
+                        {
+                            fence->wait();
```
nikhiljnv, line 1006: Need clFinish here instead of fence wait to ensure OpenCL work is done before starting vulkan work submission.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -1288,7 +1491,14 @@ int run_test_with_multi_import_diff_ctx(
                     }
                     if (iter != (maxIter - 1))
                     {
-                        clCl2VkExternalSemaphore2->signal(cmd_queue2);
+                        if (use_fence)
+                        {
+                            fence->wait();
```
nikhiljnv, line 1496: Need clFinish here instead of fence wait to ensure OpenCL work is done before starting vulkan work submission.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -1237,22 +1409,53 @@ int run_test_with_multi_import_diff_ctx(
...
+                        else
+                        {
+                            vkQueue.submit(vkCl2VkSemaphore, vkCommandBuffer,
+                                           vkVk2CLSemaphore);
+                        }
+                    }
+
+                    if (use_fence)
+                    {
+                        vkQueue.submit(vkCommandBuffer, fence);
```
nikhiljnv, line 1453: Did you mean fence->wait here?

nikhiljnv (changes requested):

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
nikhiljnv: We tried repro of the issue and found that synchronizations are misplaced in some scenarios.
Have requested changes which should fix data mismatches seen here.

pj87:
After the changes requested by @nikhiljnv I am still seeing undeterministic messages:

`log_error("&&&& vulkan_opencl_buffer test FAILED\n");`

It mostly doesn't reproduce, but it seems that there still may occur from time to time.

nikhiljnv (changes requested):

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -1194,14 +1341,38 @@ int run_test_with_multi_import_diff_ctx(
...
+                        else
+                        {
+                            vkQueue.submit(vkCl2VkSemaphore, vkCommandBuffer,
+                                           vkVk2CLSemaphore);
+                        }
+                    }
+
+                    if (use_fence)
+                    {
+                        clFinish(cmd_queue1);
```
nikhiljnv, line 1370: Need fence->wait here.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -834,14 +933,39 @@ int run_test_with_multi_import_same_ctx(
...
+                        {
+                            vkQueue.submit(vkCl2VkSemaphore, vkCommandBuffer,
+                                           vkVk2CLSemaphore);
+                        }
                     }
-                    clVk2CLExternalSemaphore->wait(cmd_queue1);
+
+                    if (use_fence)
+                    {
+                        clFinish(cmd_queue1);
```
nikhiljnv, line 962: Need fence->wait here.

nikhiljnv (changes requested):

nikhiljnv (approved):
LGTM.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -117,6 +119,7 @@ int run_test_with_two_queue(cl_context &context, cl_command_queue &cmd_queue1,
         getSupportedVulkanExternalSemaphoreHandleTypeList()[0];
     VulkanSemaphore vkVk2CLSemaphore(vkDevice, vkExternalSemaphoreHandleType);
     VulkanSemaphore vkCl2VkSemaphore(vkDevice, vkExternalSemaphoreHandleType);
+    std::shared_ptr<VulkanFence> fence;
```
joshqti, line 122: Initialize to nullptr

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
joshqti: The reason for the else branch when iter > 0 was to wait on the semaphore dispatched at the end of the last frame.  We either need to wait on the last fence here, or wait on a fence at the end of the frame, and delete the if(iter==0) behavior.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -229,14 +239,39 @@ int run_test_with_two_queue(cl_context &context, cl_command_queue &cmd_queue1,
 
                 if (iter == 0)
```
joshqti, line 240: This code can be simplified.  We do not need to wait on a fence from the previous frame, so there is no need to check iter when using fence.  I suggest we always reset the fence at the start of the loop if use_fence.  Something like

`
if (use_fence)
{
    fence->reset();
    vkQueue.submit(vkCommandBuffer, fence);
}
else
{
    if (iter == 0)
    {
        vkQueue.submit(vkCommandBuffer, vkVk2CLSemaphore);
    }
    else
    {
        vkQueue.submit(vkCl2VkSemaphore, vkCommandBuffer,
                       vkVk2CLSemaphore);
    }
}
`

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -288,7 +323,14 @@ int run_test_with_two_queue(cl_context &context, cl_command_queue &cmd_queue1,
```
joshqti, line 323: We need to wait on OpenCL to finish the work.  I suggest we flush both queues to avoid deadlock, then either finish or wait_for events.  For example:

`
if(use_fence)
{
    clFlush(cmd_queue1);
    clFlush(cmd_queue2);
    err = clWaitForEvents(...);
    }
}
`

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -288,7 +323,14 @@ int run_test_with_two_queue(cl_context &context, cl_command_queue &cmd_queue1,
 
                 if (iter != (maxIter - 1))
```
joshqti, line 324: This iter check only needs to be done when !use_fence.  And if you make the change above, we already called finish or wait for events.  I suggest something like if (!use_fence && (iter != (maxIter - 1)))...

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -528,14 +582,38 @@ int run_test_with_one_queue(cl_context &context, cl_command_queue &cmd_queue1,
             {
                 if (iter == 0)
                 {
-                    vkQueue.submit(vkCommandBuffer, vkVk2CLSemaphore);
+                    if (use_fence)
```
joshqti, line 585: See previous comments for the subtest.  Similar changes apply

pj87:
@joshqti could you take a look if everything is OK now?

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -640,16 +612,14 @@ int run_test_with_one_queue(cl_context &context, cl_command_queue &cmd_queue1,
                                 " error\n");
                     goto CLEANUP;
                 }
-                if (iter != (maxIter - 1))
+                if (use_fence)
```
joshqti, line 615: Can you propagate fence improvements everywhere in the file where fences were added?

pj87: Done

bcalidas:
The latest commit looks good. This PR can be merged.

Thanks,
Balaji Calidas,
Qualcomm

nikhiljnv (approved):
LGTM.

PR #1638, opened 2023-02-11
Revert "Image streams optimization (#1616)"

bashbaug:
This reverts commit b73c3149ad7930048d00a0c9d2fab046e41d227c.

Fixes #1637.

Likely fix for #1636 - tagging @alan-baker.

PR #1652, opened 2023-02-27
Fix Intel compiler compilation issues on windows.

gwawiork:
Changes from PR #1616 caused compilation errors on Windows and Intel Compiler. This changes fix compilation issues.
- extend compiler flags to -QxCORE-AVX2
- looks like Intel Compiler should use intrin.h
- this type __m128i_u is unknown


svenvh:
I suppose we no longer need this after #1638 ?

gwawiork:
> I suppose we no longer need this after #1638 ?
Yes we do not need

PR #1654, opened 2023-02-28
Fix slice pitch for 1D array images on image_from_buffer_read_positive kernel_read_write images subtest.

AhmedAmraniAkdi:
The slice pitch for 1D array image corresponds to a 1D array slice, but the test was computing it as 2D resulting in an incorrect slice pitch after aligning to the slice pitch alignment.


lakshmih (approved):

bashbaug:

PR #1656, opened 2023-03-01
cmake: Fix `-Wno-` handling in add_cxx_flag_if_supported

svenvh:
The `-Wno-...` compiler flags only result in a diagnostic if another diagnostic is emitted.  When passing such flags to `add_cxx_flag_if_supported` (and `check_cxx_compiler_flag`), the test would always succeed.  For such cases, test the `-W...` flag instead of the `-Wno-...` flag.


lakshmih:
Still reviewing this, thanks.

bcalidas:
We are ok with this change.

MathiasMagnus (approved):

PR #1657, opened 2023-03-01
Disable vendor extension formats from image_from_buffer_fill_positive cl_ext_image_from_buffer test.

AhmedAmraniAkdi:
Disable vendor extension formats from image_from_buffer_fill_positive cl_ext_image_from_buffer test, as the verification step is incorrect and can cause seg fault, e.g. for YUV images.


lakshmih (approved):

bashbaug:
It would be good to have somebody from Arm review these changes - @kpet or @ouakheli?

paulfradgley:
The normal Images CTS tests do not test image formats added by vendor extensions.
These new image_from_buffer tests do.

We had consensus on the call to remove the testing of vendor extension image formats from the image_from_buffer tests, but ARM were not present for the discussion.
We're waiting to hear from them before merging this.

PR #1662, opened 2023-03-06
Add compiler test external memory/semaphore extensions

paulfradgley:
The OpenCL Headers have cl_khr_external_memory_dx, cl_khr_external_memory_dma_buf and cl_khr_external_semaphore_dx_fence as extension strings, but CTS Compiler fails if an implementation says that it supports them.


bashbaug (approved):
LGTM.

It bothers me a little that the external memory handle type extensions aren't in alphabetical order, but they were already out of order before this change. :smile:

bashbaug:
Merging as discussed in the March 7th teleconference.

PR #1664, opened 2023-03-07
add parallel test running to the framework.

gpx1000:
python parallel test updates.

### DIFF
--- a/test_conformance/run_conformance.py
+++ b/test_conformance/run_conformance.py
@@ -10,12 +10,16 @@
 
 from __future__ import print_function
 
+import multiprocessing
 import os
 import re
 import sys
 import subprocess
 import time
 import tempfile
+from itertools import repeat
+
+from distlib.compat import raw_input
 
 DEBUG = 0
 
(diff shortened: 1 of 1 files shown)


test_conformance/run_conformance.py
```diff
@@ -206,75 +210,90 @@ def run_test_checking_output(current_directory, test_dir, log_file):
...
+    for arg in sys.argv:
+        match = re.search("RUN_PARALLEL", arg)
+        if match:
+            run_in_parallel = True
+    if run_in_parallel:
+        pool = multiprocessing.Pool(4)
+        failures, previous_test, test_number = pool.starmap(run_specific_test, zip(tests, repeat([failures, previous_test, test_number])))
+    else:
+        for test_itr in tests:
+            failures, previous_test, test_number = run_specific_test(test_itr, [failures, previous_test, test_number])
```
lakshmih, line 296: From the code, to run in parallel, the word "RUN_PARALLEL" is expected in the arg list. But this causes errors right at the beginning because it is not ignored like other args in lines 337-340 (in changed file).
So the following should be added right after:

if re.search("RUN_PARALLEL", arg):
        continue 

Second,  after the above is fixed, it still can't run (at least in windows). Following errors are seen: "An attempt has been made to start a new process before the current process has finished its bootstrapping phase."
This can be fixed by wrapping the calling of the main execution function (run_tests) and maybe surrounding code in a "main" function and then calling it as below:

if __name__=="__main__":
    main() 

After this, the tests run to completion (for a small subset of tests (6)) but "ValueError: too many values to unpack" is seen at the end.

The test specific stdout is also all mixed up due to tests running in parallel.

The help message should also be updated to mention the new "RUN_PARALLEL" argument

Looks like change needs some re-work before it can be reviewed again.

lakshmih: Can we remove the focused review tag on this one if we are planning a different fix?

bashbaug:
Removing "focused review" for now - not quite ready yet.

PR #1665, opened 2023-03-07
rewrite test_select to run in a few seconds.

gpx1000:
Test select getting optimized for runtime.


lakshmih:
Need more time to review this.

test_conformance/select/test_select.cpp
```diff
@@ -368,84 +373,79 @@ static int doTest(cl_command_queue queue, cl_context context, Type stype, Type c
         cmp_stride = block_elements * step * (0xffffffffffffffffULL / 0x100000000ULL + 1);
 
     log_info("Testing...");
-    d = init_genrand( gRandomSeed );
-    uint64_t i;
-    for (i=0; i < blocks; i+=step)
+    d0 = init_genrand( gRandomSeed );
+    randBuf = clCreateBuffer( context, CL_MEM_READ_ONLY, BUFFER_SIZE, nullptr, &err );
+    if( err ) { log_error( "Error: could not allocate src1 buffer\n" );  ++s_test_fail; goto exit; }
+    initSrcBuffer(randBuf, stype, d0);
```
bcalidas, line 379: It appears that a memobject handle is being treated as a hosptr when passed to initSrcBuffer. This would be incorrect use of the spec.

test_conformance/select/test_select.cpp
```diff
@@ -368,84 +373,79 @@ static int doTest(cl_command_queue queue, cl_context context, Type stype, Type c
...
+    s1 = clEnqueueMapBuffer( queue, src1, CL_TRUE, CL_MAP_WRITE, 0, BUFFER_SIZE, 0, nullptr, nullptr, &err );
+    if( err ){ log_error( "Error: Could not map src1" ); goto exit; }
+    s2 = clEnqueueMapBuffer( queue, src2, CL_TRUE, CL_MAP_WRITE, 0, BUFFER_SIZE, 0, nullptr, nullptr, &err );
+    if( err ){ log_error( "Error: Could not map src2" ); goto exit; }
+    s3 = clEnqueueMapBuffer( queue, cmp, CL_TRUE, CL_MAP_WRITE, 0, BUFFER_SIZE, 0, nullptr, nullptr, &err );
+    if( err ){ log_error( "Error: Could not map cmp" ); goto exit; }
+    // Set up the input data to change for each block
+    clEnqueueCopyBuffer(queue, randBuf, src1, 0, 0, BUFFER_SIZE, 0, nullptr, nullptr);
+    // Set up the input data to change for each block
+    clEnqueueCopyBuffer(queue, randBuf, src2, 0, 0, BUFFER_SIZE, 0, nullptr, nullptr);
```
bcalidas, line 389: Is the intent here to initialize both input buffers to the same data. If so, this would not be good practice. The original intent of the code was to have 2  independently randomized inputs.

bcalidas:
There are major concerns with this PR. I would suggest a complete re-evaluation by the author.

gpx1000:
Sorry for that; I will naturally re-evaluate and re-submit when I have a better version.

bashbaug:
Removing "focused review" until the review issues are addressed - thanks!

lakshmih (dismissed):

bashbaug (dismissed):
Thanks, this doesn't quite reduce execution time to "a few seconds" but I'm definitely seeing a nice improvement and there doesn't seem to be any lost coverage.

I tested one of our devices with the command line `test_select select_int_int`.

For the old version, before these changes:

```sh
$ time ./test_conformance/select/test_select select_int_int
...
real	1m19.207s
user	1m9.578s
sys	0m11.316s

```

For the new version, with these changes:

```sh
real	0m33.572s
user	0m23.800s
sys	0m11.626s
```

Thanks!

Note, there is a merge conflict that needs to be resolved before merging.

gpx1000:
I'm reusing the PR from the first attempt at this.  The first attempt does complete in a few seconds but at the expense of losing the number of test cases.  I merged / resolved the merge conflict.

bashbaug:

PR #1669, opened 2023-03-09
added cl_half support for test_conversions

shajder:
According to issue description (conversions section):

Additional remarks:

- In the process of adding `cl_khr_fp16` support source code was unified and as result heavily refactored. 

- Some pieces of code (eg. `InitCL` fuction) stayed as it was previously for the sake of comparison purposes. For consistency I would suggest to move it to methods (eg. `ConversionsTest::SetUp`).

- Please take a closer look at method `DataInfoSpec<>::init` under condition `is_in_half()`. There is new code to initialize array of `cl_half` values for conversions. It was source of the problem due to some special values which brought undefined results. ATM this code avoids those problematic values.

- Capacity of test was decreased. I had some difficult time trying to figure out why this number is so extremely big. I will appreciate your feedback. I limited number of tests to 16 777 216 conversions (`uint64_t lastCase = 1000000ULL;`)

- At my machine intel driver do not implement some of kernel functions which I think should be available:

```
__kernel void test_convert_half_sat_uchar( __global uchar *src, __global half *dest )
{
   size_t i = get_global_id(0);
   dest[i] = convert_half_sat( src[i] );
}
Build not successful for device "Intel(R) UHD Graphics [0x9a60]", status: CL_BUILD_ERROR
Build log for device "Intel(R) UHD Graphics [0x9a60]" is: ------------
3632:5:14: error: implicit declaration of function 'convert_half_sat' is invalid in OpenCL
   dest[i] = convert_half_sat( src[i] );
```

thus I wasn't able to finalize all tests successfully. I am counting on some help here, thanks!


svenvh:
> In the process of adding cl_khr_fp16 support source code was unified and as result heavily refactored.

Have you considered doing this unification/refactoring in a new pull request on its own?  I would recommend doing so.  We should hopefully be able to merge such refactorings relatively sooner.  Then the changes for actually adding half support should be smaller, which benefits reviewers (a smaller patch is easier to review) and you (a smaller patch is easier to keep up to date).

shajder:
Thanks for your feedback!

> Have you considered doing this unification/refactoring in a new pull request on its own? I would recommend doing so. 

yes, I think I have specific [commit](https://github.com/KhronosGroup/OpenCL-CTS/pull/1669/commits/487102962b82e0fab53079dbbd6fd02c7aa2b94e) which is good candidate to separate work on unification and extension of test's functionality - so this is still an open door for us. On the other hand there were arguments to stick to single PR:

- it turns out that the unification code takes ~75%+ of whole task. Please correct me if I am wrong but my conclusion was that reviewing of unification PR will take approximately the same time as slightly bigger one introducing `cl_khr_fp16` extension support.

- vast majority of the code responsible for the logic of the test stayed intact. What really changed is a wrapper layer around that logic + couple of necessary conditions. Thus I don't expect any major issues to investigate. BTW I had no problems with successful launch of whole unified test.

If you think those are not good enough to stick to single PR then let me know and I will start refactoring work on splitting it.

svenvh:
> If you think those are not good enough to stick to single PR then let me know and I will start refactoring work on splitting it.

I can't speak for others so please don't take my opinion too strongly.  Maybe I'm the minority here who prefer multiple smaller commits/PRs that each do one thing instead of a large monolithic PR that does multiple things.

But to illustrate a point: this PR already seems to conflict with the main branch; and the larger the PR, the higher the chances of such conflicts occurring.  Hence I believe that doing the refactoring separately will simplify maintenance of this PR in the long term (as we will probably need to give vendors a fair amount of time to test the new fp16 coverage on their devices, whereas generic refactorings shouldn't need as much time).

bcalidas:
We are still reviewing this change.

test_conformance/conversions/basic_test_conversions.h
```diff
@@ -69,5 +82,475 @@ extern InitDataFunc gInitFunctions[ kTypeCount ];
...
+//--------------------------------------------------------------------------
+
+struct DataInitInfo
+{
+    ~DataInitInfo()
+    {
+        if (d)
+        {
+            for (unsigned i = 0; i < threads; i++) free_mtdata(d[i]);
+            free(d);
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -14,2243 +14,2373 @@
...
-    {
-        if( f >= -(double) LONG_MIN )
-            return LONG_MAX;
-    }
-    else
-    {
-        if( f >= LONG_MAX )
-            return LONG_MAX;
+template <typename InType, typename OutType, bool InFP, bool OutFP>
+DataInfoSpec<InType, OutType, InFP, OutFP>::DataInfoSpec(
```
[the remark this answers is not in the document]
shajder: Done, ATM I made a separate header to isolate `DataInitInfo` related structures. Do you think it makes sense to create one more separated header for helper structures like `IterOverTypes` ?

lakshmih:
We would like to double check results at our end, can we get some more time.

lakshmih:
We are having trouble testing and verifying this PR. Could this PR be split into two changes, one for the code refactoring and the other for enhancing fp16 coverage?

shajder:
> We are having trouble testing and verifying this PR. Could this PR be split into two changes, one for the code refactoring and the other for enhancing fp16 coverage?

@lakshmih here is [PR](https://github.com/KhronosGroup/OpenCL-CTS/pull/1719) with modernized conversions without `cl_khr_fp16` support

svenvh:
This will need a rebase now that #1719 has landed.

EwanC (changes requested):
I've rebased this PR, now the diff is a lot more manageable for review 😃 

I'm seeing fails on the half tests when running this patch however. When I try the commit before I merged in `main` I see the same fails, so I don't *think* i've done the rebase wrong, but it's a possibility (we can always revert the commit to rollback).

Marking this as "Request Changes" since I'd like to investigate the fails i'm seeing further.

bashbaug:
FYI, I'm seeing failures here too, looking into them now.

test_conformance/conversions/basic_test_conversions.h
```diff
@@ -232,6 +245,7 @@ struct TestType
             case kshort: return std::is_same<cl_short, T>::value;
             case kuint: return std::is_same<cl_uint, T>::value;
             case kint: return std::is_same<cl_int, T>::value;
+            case khalf: return std::is_same<cl_half, T>::value;
```
bashbaug, line 248: Note: since a `cl_half` and a `cl_ushort` are the same type this function doesn't quite behave as expected in all cases after this change.  It would be really nice if we could differentiate between these two types.

test_conformance/conversions/basic_test_conversions.h
```diff
@@ -268,8 +282,8 @@ struct IterOverTypes : public TestType
         if (!testType<OutType>(outType)) vlog_error("Unexpected data type!\n");
```
bashbaug, line 283: We need some extra checks now, to differentiate between a `cl_half` and a `cl_ushort`.  I'm not entirely sure if this is the right fix but this seems to be helping me to get farther.

```suggestion
        if (In != inType) return;
        if (Out != outType) return;

```

test_conformance/conversions/basic_test_conversions.h
```diff
@@ -341,7 +355,8 @@ struct IterOverSelectedTypes : public TestType
         {
             // run selected conversion
```
bashbaug, line 356: This needs some extra checks now too.  Something like:

```suggestion
            if (In != inType) return;
            if (Out != outType) return;

            // run selected conversion
```

test_conformance/conversions/conversions_data_info.h
```diff
@@ -608,18 +661,30 @@ void DataInfoSpec<InType, OutType>::conv_sat(OutType *out, InType *in)
...
+        const unsigned sclamp =
+            std::is_signed<OutType>::value ? 0xffff : 0x7fff;
+
+        InType *o = (InType *)pIn;
+
+        for (unsigned i = start; i < size; i++)
+        {
+            o[i] = static_cast<InType>(i % sclamp);
+        }
+    }
```
bashbaug, line 682: This whole section of code doesn't look right and doesn't follow the same pattern as for floats and doubles:

* Why the static cast to an `InType`?  This might be OK if it's trying to cast to a `cl_ushort` to effectively do a bitcast, but it's confusing at the very least.
* Why the modulo by `sclamp`?  The code for floats and doubles only has fixup code if we aren't saturating:

```c++
        if (kUnsaturated == sat)
        {
            InType *f = (InType *)pIn;
            for (i = 0; i < size; i++) f[i] = clamp(f[i]);
        }
```

test_conformance/conversions/conversions_data_info.h
```diff
@@ -608,18 +661,30 @@ void DataInfoSpec<InType, OutType>::conv_sat(OutType *out, InType *in)
...
+    else if (std::is_integral<InType>::value)
     {
         InType *o = (InType *)pIn;
         if (sizeof(InType) <= sizeof(cl_short))
-        { // char/uchar/ushort/short
+        { // char/uchar/ushort/short/half
```
bashbaug, line 687: Minor: I think this comment is incorrect - halfs are handled above.

```suggestion
        { // char/uchar/ushort/short
```

test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -282,15 +288,15 @@ static inline void Force64BitFPUPrecision(void)
...
+int CalcRefValsPat<InType, OutType, InFP, OutFP>::check_result(void *test,
+                                                               uint32_t count,
+                                                               int vectorSize)
 {
     const cl_uchar *a = (const cl_uchar *)gAllowZ;
 
-    if (std::is_integral<OutType>::value)
-    { // char/uchar/short/ushort/int/uint/long/ulong
+    if (std::is_integral<OutType>::value || is_half<OutType, OutFP>())
+    { // char/uchar/short/ushort/half/int/uint/long/ulong
```
bashbaug, line 299: I don't think this will work - the half case needs to allow NaNs to be binary different, whereas the integral types do not have this special case.

test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -445,11 +465,31 @@ cl_int ConversionsTest::Run()
...
+            get_default_rounding_mode(device, CL_DEVICE_HALF_FP_CONFIG);
+        if ((fpConfigHalf & CL_FP_ROUND_TO_NEAREST) != 0)
+        {
+            DataInitInfo::halfRoundingMode = CL_HALF_RTE;
+            ConversionsTest::defaultHalfRoundingMode = CL_HALF_RTE;
+        }
+        else if ((fpConfigHalf & CL_FP_ROUND_TO_ZERO) != 0)
+        {
+            DataInitInfo::halfRoundingMode = CL_HALF_RTZ;
+            ConversionsTest::defaultHalfRoundingMode = CL_HALF_RTZ;
```
bashbaug, line 480: Something seems funny here - shouldn't the default rounding mode be RTE, unconditionally?

test_conformance/conversions/conversions_data_info.h
```diff
@@ -125,8 +142,9 @@ struct DataInfoSpec : public DataInitBase
...
 };
 
-template <typename InType, typename OutType>
-DataInfoSpec<InType, OutType>::DataInfoSpec(const DataInitInfo &agg)
+template <typename InType, typename OutType, bool InFP, bool OutFP>
+DataInfoSpec<InType, OutType, InFP, OutFP>::DataInfoSpec(
+    const DataInitInfo &agg)
     : DataInitBase(agg), mdv(0)
 {
     if (std::is_same<cl_float, OutType>::value)
```
bashbaug, line 150: I think we need an if check here for the "half" OutType, to setup ranges properly for halfs.

test_conformance/conversions/conversions_data_info.h
```diff
@@ -159,9 +177,9 @@ DataInfoSpec<InType, OutType>::DataInfoSpec(const DataInitInfo &agg)
 
         InType eps = std::is_same<InType, cl_float>::value ? (InType) FLT_EPSILON : (InType) DBL_EPSILON;
         if (std::is_integral<OutType>::value)
-        { // to char/uchar/short/ushort/int/uint/long/ulong
+        { // to char/uchar/short/ushort/int/uint/long/ulong/half
             if (sizeof(OutType)<=sizeof(cl_short))
-            { // to char/uchar/short/ushort
+            { // to char/uchar/short/ushort/half
```
bashbaug, line 182: I'm a little skeptical this will work for half as an InType or as an OutType:

* For half an InType, note that `std::is_floating_point<InType>::value` will be false, so this will need to be handled specially.
* For half as an OutType, note that `std::is_integral<OutType>::value` will be true, so this will need to be handled specially, also.

bashbaug (changes requested):
I found and fixed a few problems and managed to get a little farther but this still isn't working for me.  Here are a few issues I found.

test_conformance/conversions/basic_test_conversions.h
shajder: Correct, please notice `isTypeFp` array which takes care of distinguishing `cl_half` and `cl_ushort` later. I agree it looks a bit hard coded but it may be difficult to figure out more elegant solution in this case.

test_conformance/conversions/conversions_data_info.h
shajder: Please notice my initial description for the PR above (_Please take a closer look at method DataInfoSpec<>::init under condition is_in_half()..._). The code you refer to is an result of searching a way to pass the test with `fp16` implementation I am working with (`11th Gen Intel(R) Core(TM) i7-11800H`). Approach of generating input values similar to `double`s and `float`s failed in couple of conversion cases. I will appreciate your expertise here. Should we test ALL possible values from cl_half (ushort) scope ?

bashbaug:
This PR needs a new owner, so for now I've removed "focused review" and added "help wanted".

test_conformance/conversions/basic_test_conversions.h
shajder: Neither of these conditions is ever met during test at my machine. In general those values shouldn't differ, the main difference is that `In`/`Out` are template arguments known during build time. 

To distinguish between `cl_half` and `cl_ushort` `isTypeFp` array was introduced and it is used on line below as an template argument.

test_conformance/conversions/basic_test_conversions.h
shajder: Comment above applies here as well.

shajder: Good find - that is correct, I will put additional checks in `testType` function template.

test_conformance/conversions/basic_test_conversions.cpp
shajder: @bashbaug good question, can we clarify that ? [Here](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_fp16-ieee754-compliance) spec states:

```
The required minimum half precision floating-point capability as implemented by this extension is:

CL_FP_ROUND_TO_ZERO, or CL_FP_ROUND_TO_NEAREST | CL_FP_INF_NAN.
```

Please correct me if I am mistaken, my understanding is that in some cases RTE may be unavailable. Should RTE rounding still be default in such situation ?

test_conformance/conversions/basic_test_conversions.h
shajder: Done (through additional condition in `testType` function template)

test_conformance/conversions/conversions_data_info.h
shajder: Done, I have adapted this piece of code to follow the same pattern as for floats and doubles.

The obstacle to do so earlier was unclear idea behind `specialValuesFloat`/`specialValuesDouble` (I have mentioned in the description of this PR). I just borrowed such values for `specialValuesHalf` from `bruteforce` test.

test_conformance/conversions/conversions_data_info.h
shajder: Done

test_conformance/conversions/basic_test_conversions.cpp
shajder: Done

test_conformance/conversions/conversions_data_info.h
shajder: Done, I have added additional if-check for both `cl_half` and `cl_ushort`

test_conformance/conversions/conversions_data_info.h
shajder: Done, separate checks for both `InType` and `OutType` added to handle `cl_half` impact on `clamp_ranges`

test_conformance/conversions/basic_test_conversions.h
bashbaug: This is fixed now, by also looking at the FP template parameter.  Resolving.

```c++
case khalf: return std::is_same<cl_half, T>::value && FP;
```

test_conformance/conversions/basic_test_conversions.h
bashbaug: This doesn't seem to be needed with the latest changes.  Resolving.

test_conformance/conversions/basic_test_conversions.cpp
bashbaug: Interesting - this does seem to be correct, although confusing and error prone.

From the OpenCL C spec: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_fp16-rounding-modes

> If CL_FP_ROUND_TO_NEAREST is supported, the default rounding mode for half-precision floating-point operations will be round to nearest even; otherwise the default rounding mode will be round to zero.

And also the OpenCL SPIR-V environment spec: https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_rounding_modes

> For half precision operations, devices may support either Round to nearest even or Round toward zero as the default rounding mode for half precision operations.

(For this last one we should probably say a bit more about how the default rounding mode is chosen.)

I'll just resolve this comment.  Thanks!

bashbaug (commented):
My comments seem to have been addressed - thanks!

I think it's a bit error-prone to need to check both the "type" and the "fp" bool to determine if the type is a "half", but I believe it's all correct now, so we can always tidy this up later.

The saturate issue mentioned above does need to be fixed, but I'm not seeing any failures with the fix.  For reference, I'm running:

```sh
test_conversions -w -1
```

This finishes in reasonable time.  If I think of it I'll kick off a more thorough test overnight.

bashbaug:
I spoke too soon, and I'm seeing a few failures in double to half conversions, for non-round-to-even rounding modes:

Example for: test_conversions half_rtn_double -w -1

```
Building convert_half_rtn( double ) test
Testing... .
Error for vector size 1 found at 0x00003f76:  *-0x1.8p+1 vs -0x1.804p+1
Input value: -0x1.8000000000001p+1  (convert_half_rtn( double ))
         *** convert_halfn_rtn( doublen ) FAILED ** 
conversions FAILED
```

I'm checking to see if this is an issue with our implementation or with the tests.

test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -445,11 +505,31 @@ cl_int ConversionsTest::Run()
...
+        }
+    }
+
     return CL_SUCCESS;
 }
 
-
-template <typename InType, typename OutType>
+template <typename InType, typename OutType, bool InFP, bool OutFP>
 void ConversionsTest::TestTypesConversion(const Type &inType,
```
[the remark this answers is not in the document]
shajder: That explains a lot. Corrected!

test_conformance/conversions/conversions_data_info.h
```diff
@@ -376,6 +455,8 @@ void DataInfoSpec<InType, OutType>::conv(OutType *out, InType *in)
     {
         if (std::is_same<cl_float, OutType>::value)
             *out = (OutType)*in;
+        else if (is_out_half())
+            *out = static_cast<OutType>(HFF(*in));
```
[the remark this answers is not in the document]
shajder: Thanks, good to know I can emulate `FP64`. I've corrected this case by using `cl_half_from_double`.

bashbaug (approved):
I think all of my issues are addressed - thanks!

svenvh (approved):

bashbaug:
I think we should merge this as-is and fix any bugs that appear.

@lakshmih can you run a "wimpy" test first to gain confidence that the changes are OK?  Maybe even a scalar-only "wimpy" test?  This finishes fairly promptly on our devices.

Example: test_conversions -w -1

bashbaug:
I just created an `fp16-staging` branch, as we discussed.  If this PR is retargeted at this branch I will merge it, also as we discussed.

It's been a few days now though.  @lakshmih has your testing completed?  If so, would we be OK to merge this to the main branch, instead?

shajder:
> I just created an `fp16-staging` branch, as we discussed. If this PR is retargeted at this branch I will merge it, also as we discussed.

@bashbaug I've created new [PR](https://github.com/KhronosGroup/OpenCL-CTS/pull/1864) to be merged with `fp16-staging` branch

bashbaug:
Removing "focused review", since we are reviewing and merging changes into the staging branch now.

shajder:
This PR was merged to main branch along with #1975, closing

PR #1675, opened 2023-03-15
Update cl_khr_semaphore test to conform with blocking semaphores

lakshmih:
Semaphore spec has been updated to reflect the fact that semaphores will be in the appropriate - pending signal or pending wait - state when returning from clEnqueueSignalSemaphore or clEnqueueWaitSemaphore commands: https://github.com/KhronosGroup/OpenCL-Docs/pull/882

Deleted the following tests to match the updated spec:

semaphores_order_1 - Test calls EnqueueWaitSemaphore before calling EnqueueSignalSemaphore and expects this wait to succeed. This behavior is not compatible with the recent spec updates to semaphores.

semaphores_order_2 & semaphores_order_3 - Calling clEnqueueSignalSemaphoresKHR with a dependency on a user event may cause the implementation to block until the user event is complete. This is unsafe usage of clEnqueueSignalSemaphoresKHR and may lead to deadlock.

semaphores_invalid_command - This test checks for specific behavior when waiting on a semaphore in an invalid state. According to the spec, this is undefined behavior, and therefore cannot be tested directly.


bashbaug:
We merged the spec changes previously so this one is ready for review.

nikhiljnv:
LGTM as most changes are about removing some of the test scenarios.

PR #1676, opened 2023-03-15
External memory updates

lakshmih:
Fix Vulkan spec violation issues:

- Program Vulkan descriptor set layouts as arrays where appropriate:

Vulkan shader code binds images and buffers as arrays:
layout(binding = 1) buffer Buffer
{
     uint8_t ptr[];
} bufferPtrList[MAX_BUFFERS];

bufferPtrList is an array of size MAX_BUFFERS at binding 1.

This requires a VkDescriptorSetLayoutBinding with binding=1, and descriptorCount=MAX_BUFFERS.

From the Khronos spec "descriptorCount is the number of descriptors contained in the binding, accessed in a shader as an array"

The test previously assumed bufferPtrList[i] maps to binding site 1 + i, with descriptorCount=1.  This is not valid, and not portable.

See https://registry.khronos.org/vulkan/specs/1.1-extensions/html/vkspec.html#descriptorsets-sets

- Change VULKAN_QUEUE_FLAG_MASK_ALL to VULKAN_QUEUE_FLAG_GRAPHICS | VULKAN_QUEUE_FLAG_COMPUTE. The VULKAN_QUEUE_TRANSFER_BIT is implied by either the graphics or compute bit, implementations are not required to report it.

- Implementations may require dedicated memory for images.  This can be queried with https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkMemoryDedicatedRequirements.html. This query must be done by default, and not enabled by a command line switch.  The conformance test should run on all platforms without extra options.  Suggest a follow up gerrit to add a debugging “disable dedicated memory” switch if manual control of the behavior is needed.

### DIFF
--- a/test_conformance/common/vulkan_wrapper/vulkan_api_list.hpp
+++ b/test_conformance/common/vulkan_wrapper/vulkan_api_list.hpp
@@ -75,7 +75,7 @@
     VK_FUNC_DECL(vkDestroyImageView)                                           \
     VK_FUNC_DECL(vkCreateImage)                                                \
     VK_FUNC_DECL(vkGetImageMemoryRequirements)                                 \
-    VK_FUNC_DECL(vkGetImageMemoryRequirements2KHR)                             \
+    VK_FUNC_DECL(vkGetImageMemoryRequirements2)                                \
     VK_FUNC_DECL(vkDestroyImage)                                               \
     VK_FUNC_DECL(vkDestroyBuffer)                                              \
     VK_FUNC_DECL(vkDestroyPipeline)                                            \
@@ -88,9 +88,9 @@
     VK_FUNC_DECL(vkDestroyDescriptorSetLayout)                                 \
     VK_FUNC_DECL(vkGetPhysicalDeviceQueueFamilyProperties)                     \
     VK_FUNC_DECL(vkGetPhysicalDeviceFeatures)                                  \
-    VK_FUNC_DECL(vkGetPhysicalDeviceProperties2KHR)                            \
+    VK_FUNC_DECL(vkGetPhysicalDeviceProperties2)                               \
     VK_FUNC_DECL(vkGetBufferMemoryRequirements)                                \
-    VK_FUNC_DECL(vkGetBufferMemoryRequirements2KHR)                            \
+    VK_FUNC_DECL(vkGetBufferMemoryRequirements2)                               \
     VK_FUNC_DECL(vkGetMemoryFdKHR)                                             \
     VK_FUNC_DECL(vkGetSemaphoreFdKHR)                                          \
     VK_FUNC_DECL(vkEnumeratePhysicalDeviceGroups)                              \
@@ -162,7 +162,7 @@
 #define vkDestroyImageView _vkDestroyImageView
 #define vkCreateImage _vkCreateImage
 #define vkGetImageMemoryRequirements _vkGetImageMemoryRequirements
-#define vkGetImageMemoryRequirements2KHR _vkGetImageMemoryRequirements2KHR
+#define vkGetImageMemoryRequirements2 _vkGetImageMemoryRequirements2
 #define vkDestroyImage _vkDestroyImage
 #define vkDestroyBuffer _vkDestroyBuffer
 #define vkDestroyPipeline _vkDestroyPipeline
@@ -176,9 +176,9 @@
 #define vkGetPhysicalDeviceQueueFamilyProperties                               \
     _vkGetPhysicalDeviceQueueFamilyProperties
 #define vkGetPhysicalDeviceFeatures _vkGetPhysicalDeviceFeatures
-#define vkGetPhysicalDeviceProperties2KHR _vkGetPhysicalDeviceProperties2KHR
+#define vkGetPhysicalDeviceProperties2 _vkGetPhysicalDeviceProperties2
 #define vkGetBufferMemoryRequirements _vkGetBufferMemoryRequirements
-#define vkGetBufferMemoryRequirements2KHR _vkGetBufferMemoryRequirements2KHR
+#define vkGetBufferMemoryRequirements2 _vkGetBufferMemoryRequirements2
 #define vkGetMemoryFdKHR _vkGetMemoryFdKHR
 #define vkGetSemaphoreFdKHR _vkGetSemaphoreFdKHR
 #define vkEnumeratePhysicalDeviceGroups _vkEnumeratePhysicalDeviceGroups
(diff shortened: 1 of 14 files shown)


bcalidas:
This PR adds a number of general fixes for OpenCL - Vulkan interop.

nikhiljnv:
@lakshmih / @bcalidas 
The PR description seems empty. A short summary of changes would help.
Also, there is mix of different set of changes in this PR.
1. General improvements - Formatting fixes, memory leaks 
2. Additional functionality - Multiple semaphore type support, SYNC_FD support
3. Removal/refactoring of existing functionality - Changes to dedicated memory allocation usage, Delete multi-context/import test

Can we move them into separate PRs so that it is easier to review and start merging the easier ones?
Category 3 may require some discussion and possibly spec clarifications (especially, for [OpenCL-Docs#861](https://github.com/KhronosGroup/OpenCL-Docs/issues/866)

nikhiljnv:
Implementations may require dedicated memory for images --> May be we will need to discuss this as part of https://github.com/KhronosGroup/OpenCL-Docs/issues/866. Are there cases where non-dedicated memory can not be safely imported as image or a particular image setting requires dedicated memory?

The conformance test should run on all platforms without extra options. --> Sure.

Suggest a follow up gerrit to add a debugging “disable dedicated memory” switch if manual control of the behavior is needed. --> Any reason we can not reuse existing knob which is getting removed with this PR with the semantics suggested above i.e run for both non-dedicated vs dedicated cases, but have the command line option to restrict the usage just to dedicated or non-dedicated imports.

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
```diff
@@ -1886,6 +1992,61 @@ VulkanDeviceMemory::VulkanDeviceMemory(
...
+        VK_STRUCTURE_TYPE_MEMORY_DEDICATED_ALLOCATE_INFO;
+    vkMemoryDedicatedAllocateInfo.pNext =
+        externalMemoryHandleType ? &vkExportMemoryAllocateInfoKHR : NULL;
+    vkMemoryDedicatedAllocateInfo.image = VK_NULL_HANDLE;
+    vkMemoryDedicatedAllocateInfo.buffer = buffer;
+
+    VkMemoryAllocateInfo vkMemoryAllocateInfo = {};
+    vkMemoryAllocateInfo.sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+    vkMemoryAllocateInfo.pNext =
+        m_isDedicated ? &vkMemoryDedicatedAllocateInfo : nullptr;
```
nikhiljnv, line 2041: We still need to pass vkExportMemoryAllocateInfoKHR even if the memory is not dedicated, but looks like we are skipping all of AllocateInfo in case of non-dedicated memory. Is that a bug or intentional?

test_conformance/common/vulkan_wrapper/vulkan_api_list.hpp
```diff
@@ -89,6 +90,7 @@
     VK_FUNC_DECL(vkGetPhysicalDeviceFeatures)                                  \
     VK_FUNC_DECL(vkGetPhysicalDeviceProperties2KHR)                            \
     VK_FUNC_DECL(vkGetBufferMemoryRequirements)                                \
+    VK_FUNC_DECL(vkGetBufferMemoryRequirements2KHR)                            \
```
nikhiljnv, line 93: Any reason we can not use core version of these APIs (vkGetBufferMemoryRequirements2) and need to use KHR extension version? Are we assuming any particular minimum Vk version ?
May be we can check for VkVersion supported and then choose between core vs KHR version of these APIs.
Also, for KHR version usage, we will need to enable the required extension during VkDevice creation.

nikhiljnv (changes requested):

test_conformance/common/vulkan_wrapper/vulkan_api_list.hpp
joshqti: The only reason I used KHR was because the automated testing failed to link on Ubuntu 20.02.  I suspect the packaged Vulkan version is older than on the other platforms, however I have no way to debug github's build server.

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
joshqti: Nice catch, I need to reverse the pChain, this is a bug.

bcalidas:
@nikhiljnv 

In response to your earlier comment  "Implementations may require dedicated memory for images --> May be we will need to discuss this as part of https://github.com/KhronosGroup/OpenCL-Docs/issues/866. Are there cases where non-dedicated memory can not be safely imported as image or a particular image setting requires dedicated memory?"

The motivation for our code change comes purely from Vulkan. Our Vulkan implementation requires that all exportable allocations be dedicated. This is allowed by the Vulkan spec.

Regarding the reference to https://github.com/KhronosGroup/OpenCL-Docs/issues/866 , it's tricky to articulate this issue as an OpenCL requirement since this restriction comes entirely from Vulkan. We do need to broadly discuss importing dedicated vs. non dedicated allocations and what general message we can give to OpenCL developers.

In the context of this conformance test which uses Vulkan, I'm hoping that this  will be a simple decision since it relates to behavior that is entirely contained within Vulkan.

Thanks,
Balaji

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
joshqti: Thanks for noticing, that was a bug.  I need to reverse the pChain, otherwise vkExportMemoryAllocateInfoKHR  will not get passed in if the allocation is not dedicated.

test_conformance/common/vulkan_wrapper/vulkan_api_list.hpp
nikhiljnv: As far as I can see, https://github.com/KhronosGroup/OpenCL-CTS/blob/main/presubmit.sh uses Vulkan Loader and Headers from GitHub TOT. So, not sure why the automatic build may have failed while using the newer APIs.
Do you have link to the failed logs?

joshqti: Unfortunately I do not have the logs, I updated with a force push.

joshqti:
Addressed all review comments as best as possible, pushed a new commit.

bashbaug:
Discussed in the August 29th teleconference.  Should be clean now that the header changes have been merged.  Need to fix the merge conflict.

lakshmih:
Fixed merge conflict, ready to merge.

nikhiljnv (approved):
LGTM.

PR #1678, opened 2023-03-16
fixing test events

bananAshkar:
fixing the test of events

### DIFF
--- a/test_conformance/events/test_event_dependencies.cpp
+++ b/test_conformance/events/test_event_dependencies.cpp
@@ -52,7 +52,7 @@ int test_event_enqueue_wait_for_events_run_test(
 {
     cl_int error = CL_SUCCESS;
     size_t threads[3] = { TEST_SIZE, 0, 0 };
-    int i, loop_count, event_count, expected_value, failed;
+    int i, loop_count, expected_value, failed;
     int expected_if_only_queue[2];
     int max_count = TEST_SIZE;
 
@@ -65,8 +65,6 @@ int test_event_enqueue_wait_for_events_run_test(
     clMemWrapper data;
     clProgramWrapper program;
     clKernelWrapper kernel1[TEST_COUNT], kernel2[TEST_COUNT];
-    clEventWrapper event[TEST_COUNT * 4 + 2]; // If we usemarkers we get 2 more
-                                              // events per iteration
 
     if (test_enqueue_wait_for_events)
         log_info("\tTesting with clEnqueueBarrierWithWaitList as barrier "
@@ -249,11 +247,13 @@ int test_event_enqueue_wait_for_events_run_test(
     else
         log_info("Queues chosen alternatily for each kernel execution.\n");
 
-    event_count = 0;
+    clEventWrapper pre_loop_event;
+    clEventWrapper last_loop_event;
+
     for (i = 0; i < (int)TEST_SIZE; i++) values[i] = 1;
     error = clEnqueueWriteBuffer(queues[0], data, CL_FALSE, 0,
                                  TEST_SIZE * sizeof(cl_int), values, 0, NULL,
-                                 &event[event_count]);
+                                 &pre_loop_event);
     test_error(error, "clEnqueueWriteBuffer 2 failed");
     expected_value = 1;
     expected_if_only_queue[0] = 1;
@@ -263,7 +263,7 @@ int test_event_enqueue_wait_for_events_run_test(
     if (test_enqueue_wait_for_events)
     {
         error = clEnqueueBarrierWithWaitList(queues[queue_to_use], 1,
-                                             &event[event_count], NULL);
+                                             &pre_loop_event, NULL);
         test_error(error, "Unable to queue wait for events");
     }
     else if (test_barrier)
@@ -275,6 +275,13 @@ int test_event_enqueue_wait_for_events_run_test(
 
     for (loop_count = 0; loop_count < TEST_COUNT; loop_count++)
     {
+        int event_count = 0;
+        clEventWrapper first_dependency =
+            (loop_count == 0) ? pre_loop_event : last_loop_event;
+        clEventWrapper
+            event[5]; // A maximum of 5 events are created in the loop
+        event[event_count] = first_dependency;
+
         // Execute kernel 1
         event_count++;
         if (use_waitlist | use_marker)
@@ -424,6 +431,7 @@ int test_event_enqueue_wait_for_events_run_test(
                                                  NULL);
             test_error(error, "Unable to queue barrier");
         }
+        last_loop_event = event[event_count];
     }
 
     // Now finish up everything
@@ -435,7 +443,7 @@ int test_event_enqueue_wait_for_events_run_test(
 
     error = clEnqueueReadBuffer(queues[0], data, CL_TRUE, 0,
                                 TEST_SIZE * sizeof(cl_int), values, 1,
-                                &event[event_count], NULL);
+                                &last_loop_event, NULL);
 
     test_error(error, "clEnqueueReadBuffer failed");
 


EwanC (commented):
Could you elaborate on what issue this patch is fixing that requires manually resetting the events set each loop iteration?

I'd expect the resources for `clEventWrapper`s to get released automatically without the reset when the `event` variable goes [out of scope](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_common/harness/typeWrappers.h#L106) at the end of the function.

bananAshkar:
> Could you elaborate on what issue this patch is fixing that requires manually resetting the events set each loop iteration?
> 
> I'd expect the resources for `clEventWrapper`s to get released automatically without the reset when the `event` variable goes [out of scope](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_common/harness/typeWrappers.h#L106) at the end of the function.

The function produces  events and automatically releases them at the end of its execution. However, if the function allocates an event, uses it, and then finishes with it, there is no need to retain the event until the end of the function. In other words, releasing the event once it is no longer needed can improve memory management and overall efficiency.

EwanC:
> The function produces events and automatically releases them at the end of its execution. However, if the function allocates an event, uses it, and then finishes with it, there is no need to retain the event until the end of the function. In other words, releasing the event once it is no longer needed can improve memory management and overall efficiency. @EwanC

Thanks for clarifying, it makes sense that the change is for performance rather than correctness. I feel like defining some of these event wrappers inside the loop scope would be a better solution however. Then the events get released at the end of the
loop iteration without having to manually count back and reset, and don't need to worry about guarding some resets with `use_marker` as you can destroy a null wrapper.

bananAshkar:
> > thank you for your comment , but in line 445 (after the loop ends) , we have the function clEnqueueReadBuffer , which uses event[event_count] which is the last event , so we cant define the cleventwrapper inside the loop . what do you think?
> 
> We can still define events outside the loop, but for the dependencies of commands inside the loop, we can put them in the loop scope. Something like this psuedocode
> 
> ```c++
> clEventWrapper pre_loop_event;
> 
> // pre_loop_event set by last commands to be enqueued prior to the loop 
> 
> clEventWrapper last_loop_event;
> for (loop_count = 0; loop_count < TEST_COUNT; loop_count++) {
>   clEventWrapper first_dependency = (loop_count == 0) ? pre_loop_event : last_loop_event;
>   clEventWrapper events[N]; // used for commands within loops, and freed at end of loop iteration (apart from event copied to last_loop_event)
> 
>   ...  
> 
>   last_loop_event = // event returned by final command in loop
> }
> 
> // clEnqueueReadBuffer on line 445 waits on last_loop_event;
> ```

Expressing gratitude and appreciation for reviewing my code , thank you

test_conformance/events/test_event_dependencies.cpp
```diff
@@ -435,7 +441,7 @@ int test_event_enqueue_wait_for_events_run_test(
 
     error = clEnqueueReadBuffer(queues[0], data, CL_TRUE, 0,
                                 TEST_SIZE * sizeof(cl_int), values, 1,
-                                &event[event_count], NULL);
+                                &last_loop_even, NULL);
```
EwanC, line 444: `last_loop_event`

test_conformance/events/test_event_dependencies.cpp
```diff
@@ -275,6 +276,10 @@ int test_event_enqueue_wait_for_events_run_test(
 
     for (loop_count = 0; loop_count < TEST_COUNT; loop_count++)
     {
+        clEventWrapper first_dependency = (loop_count == 0) ? pre_loop_event : last_loop_event;
+        clEventWrapper event[TEST_COUNT*4+2]; // If we usemarkers we get 2 more events per iteration
```
EwanC, line 280: Thanks for making these changes, this is looking like a nice improvement.

I think we can do better still, and make this array of size 4 (or however many events we actually need in the loop) rather than `[TEST_COUNT*4+2]`  and move the `event_count` variable into the loop to increment through those 4 events.

bananAshkar: > Thanks for making these changes, this is looking like a nice improvement.
> 
> I think we can do better still, and make this array of size 4 (or however many events we actually need in the loop) rather than `[TEST_COUNT*4+2]` and move the `event_count` variable into the loop to increment through those 4 events.

thank you !!

test_conformance/events/test_event_dependencies.cpp
```diff
@@ -435,9 +443,10 @@ int test_event_enqueue_wait_for_events_run_test(
 
     error = clEnqueueReadBuffer(queues[0], data, CL_TRUE, 0,
                                 TEST_SIZE * sizeof(cl_int), values, 1,
-                                &event[event_count], NULL);
+                                &last_loop_event, NULL);
 
     test_error(error, "clEnqueueReadBuffer failed");
+    last_loop_event.reset();
```
EwanC, line 449: Shouldn't need this manual reset, will happen when `last_loop_event` goes out of scope at the end of the function

test_conformance/events/test_event_dependencies.cpp
```diff
@@ -275,6 +277,10 @@ int test_event_enqueue_wait_for_events_run_test(
 
     for (loop_count = 0; loop_count < TEST_COUNT; loop_count++)
     {
```
EwanC, line 279: think we can move the definition of `int event_count` from being at the function scope to being defined as `int event_count = 0;` at the start of this loop. So we also wouldn't need `event_count = 0;` at the end of the loop.

test_conformance/events/test_event_dependencies.cpp
bananAshkar: thank you , done

test_conformance/events/test_event_dependencies.cpp
bananAshkar: done

test_conformance/events/test_event_dependencies.cpp
```diff
@@ -275,6 +276,11 @@ int test_event_enqueue_wait_for_events_run_test(
...
+        int event_count = 0;
+        clEventWrapper first_dependency = (loop_count == 0) ? pre_loop_event : last_loop_event;
```
EwanC, line 280: This needs a [clang-format to pass CI](https://github.com/KhronosGroup/OpenCL-CTS/actions/runs/4512804381/jobs/7978464519?pr=1678)

```diff
diff --git a/test_conformance/events/test_event_dependencies.cpp b/test_conformance/events/test_event_dependencies.cpp
index cd0aa67..e94c8cc 100644
--- a/test_conformance/events/test_event_dependencies.cpp
+++ b/test_conformance/events/test_event_dependencies.cpp
@@ -277,8 +277,10 @@ int test_event_enqueue_wait_for_events_run_test(
     for (loop_count = 0; loop_count < TEST_COUNT; loop_count++)
     {
         int event_count = 0;
-        clEventWrapper first_dependency = (loop_count == 0) ? pre_loop_event : last_loop_event;
-        clEventWrapper event[SIZE]; // If we usemarkers we get 2 more events per iteration
+        clEventWrapper first_dependency =
+            (loop_count == 0) ? pre_loop_event : last_loop_event;
+        clEventWrapper
+            event[SIZE]; // If we usemarkers we get 2 more events per iteration
         event[event_count] = first_dependency;
 
         // Execute kernel 1
```

test_conformance/events/test_event_dependencies.cpp
```diff
@@ -275,6 +276,11 @@ int test_event_enqueue_wait_for_events_run_test(
...
+        clEventWrapper event[SIZE]; // If we usemarkers we get 2 more events per iteration
```
EwanC, line 281: Don't think we need a new `SIZE` macro since it's only ever used here. Would just make this line:

```cpp
// A maximum of 5 events are created in the loop
clEventWrapper event[5];
```

EwanC (commented):
Looking good to me now, just some minor syntactical changes needed to pass CI

test_conformance/events/test_event_dependencies.cpp
bananAshkar: thank you done!

test_conformance/events/test_event_dependencies.cpp
bananAshkar: done , thank you !!

EwanC (approved):

lakshmih:
Reviewing

lakshmih (approved):
Looks good

PR #1634, opened 2023-02-09
math_brute_force: Drop BuildKernelInfo2

svenvh:
Replace occurrences of `BuildKernelInfo2` with `BuildKernelInfo`. This aligns the kernel creation code of the non-threaded tests to the kernel creation code of the threaded tests, giving opportunities for further code deduplication.

Signed-off-by: Sven van Haastregt <sven.vanhaastregt@arm.com>

PR #1433, opened 2022-06-07
Enable fp16 in math bruteforce

gwawiork:
This is almost the duplicate of https://github.com/KhronosGroup/OpenCL-CTS/pull/529 after rebase.

PR #1681, opened 2023-03-23
added cl_half support for test_bruteforce

shajder:
According to issue description (bruteforce section):

Additional remark: This PR is based on @gwawiork work from PR [#1433](https://github.com/KhronosGroup/OpenCL-CTS/pull/1433). Due to long approval process it was necessary to rebase and adapt to [#1634](https://github.com/KhronosGroup/OpenCL-CTS/pull/1634) modernization. Due to reassignment to **`mobica-backlog`** it was more convenient to introduce new PR (which was discussed with Grzegorz).


test_conformance/math_brute_force/CMakeLists.txt
```diff
@@ -15,12 +17,16 @@ set(${MODULE_NAME}_SOURCES
...
+    i_unary_half.cpp
     macro_binary_double.cpp
     macro_binary_float.cpp
+    macro_binary_half.cpp
     macro_unary_double.cpp
     macro_unary_float.cpp
+    macro_unary_half.cpp
     mad_double.cpp
     mad_float.cpp
+    mad_half.cpp
```
EwanC, line 29: we've added `mad` but we've not added a `ternary_half.cpp` file to test `fma()`.

Similarly we need a `unary_two_resuts_half` to test `sincos`, and `unary_two_results_i_float.cpp` to test `lgamma_r` and `frexp`.

Are half `fma()`, `lgamma_r`, `frexp` being testing in some other way, or are they just missing?

lakshmih:
Testing / reviewing.

lakshmih:
We would like to double check results at our end, can we get some more time.

test_conformance/math_brute_force/CMakeLists.txt
shajder: I've added missing source files, couple of remarks:

1) Whole test should be unified due to huge duplication of code. I think current progress is not a good starting point for modernization so my proposition is to create new issue.
2) I had to modify the logic adapted from ternary_float in order to pass the test with cl_half. Source of the confusion is Ulp_Half_Error. I've added [additional conditions](https://github.com/KhronosGroup/OpenCL-CTS/blob/04033bf8bd958c9c053e753363c4588f7dc75ccf/test_conformance/math_brute_force/ternary_half.cpp#L293-L294) to work this problem around but I am not sure if that is correct approach.

test_conformance/math_brute_force/CMakeLists.txt
```diff
@@ -15,28 +17,37 @@ set(${MODULE_NAME}_SOURCES
...
     reference_math.h
     sleep.cpp
     sleep.h
     ternary_double.cpp
     ternary_float.cpp
+    ternary_half.cpp
     test_functions.h
     unary_double.cpp
     unary_float.cpp
+    unary_half.cpp
```
lakshmih, line 41: Could you verify that the kernels that are added that use the half type have the pragma for enabling cl_khr_fp16:
#pragma OPENCL EXTENSION cl_khr_fp16 : enable"

shajder: I just run all the `bruteforce` tests related to `cl_khr_fp16` and yup - I can confirm all kernels contain necessary `pragma`, roughly header of each kernel looks very similar:

```
#define RETTYPE half2
#define RETTYPE_SCALAR half
#define TYPE1 half2
#define TYPE1_SCALAR half
#define UNDEF1 NAN

__kernel void math_kernel2(__global RETTYPE* ....
```

lakshmih: This is the kernel that's getting generated for nan:

    Build options:  -cl-std=CL3.0
     Original source is: ------------
     #define RETTYPE half16
     #define RETTYPE_SCALAR half
     #define TYPE1 ushort16
     #define TYPE1_SCALAR ushort
     #define UNDEF1 0x5678
     
     __kernel void math_kernel16(__global RETTYPE* out,
                               __global TYPE1* in1)
     {
         size_t i = get_global_id(0);
         out[i] = nan(in1[i]);
     }

Can you specifically check unary_u_half.cpp?

svenvh: Good catch, this is actually a pre-existing issue in `common.cpp`.  I've opened #1705 to fix this for fp64.  That fix will then have to be extended for fp16.

shajder: Just to comprehend the problem - why I can't see this ? Is it specifically related to ARM architecture build ?

shajder: Answering to myself - for some reason missing `pragma` did not cause neither kernel built error nor test fail with my Intel device.

test_conformance/math_brute_force/macro_unary_half.cpp
```diff
@@ -0,0 +1,418 @@
...
+        // there was overflow
+        test_info.jobCount = 1;
+    }
+    else
+    {
+        test_info.jobCount = (cl_uint)((1ULL << 32) / test_info.step);
+    }
+
+    test_info.f = f;
+    test_info.ftz = f->ftz || gForceFTZ;
```
lakshmih, line 110: This test isn't handling half denorms correctly. There might be other cases, still checking.

shajder: > This test isn't handling half denorms correctly. There might be other cases, still checking.

Done in this case

test_conformance/math_brute_force/CMakeLists.txt
svenvh: @shajder #1705 has been merged now, so if you rebase this PR on top of latest `main` then the missing pragma issue should be fixed.

svenvh: @shajder yes you will need to add `ParameterType::Short`/`UShort` to the switch (which is now inside a loop so it's one level deeper compared to your previous patch).  Then you also need to update the case for `ParameterType::Half` to reflect how `Double` is handled now.

shajder: Correction merged, `nan` kernel now looks correct:

  _2:             nan fp16     #define RETTYPE half_

```
#define RETTYPE_SCALAR half
#define TYPE1 ushort3
#define TYPE1_SCALAR ushort
#define UNDEF1 0x5678

__kernel void math_kernel3(__global RETTYPE_SCALAR* out,
                          __global TYPE1_SCALAR* in1)
{
    size_t i = get_global_id(0);
...
```

test_conformance/math_brute_force/reference_math.cpp
```diff
@@ -4699,6 +4699,40 @@ double reference_nextafter(double xx, double yy)
...
+{
+    cl_ushort u;
+    cl_half h;
+    u = x | 0x7e00U;
+    memcpy(&h, &u, sizeof(cl_half));
+    return h;
+}
+
+float reference_nextafterh(float xx, float yy)
+{
```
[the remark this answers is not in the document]
shajder: @lakshmih understood. Spec also states:

[1.](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#edge-case-behavior-in-flush-to-zero-mode:~:text=Any%20conforming%20result%20for%20non%2Dflush%2Dto%2Dzero%20mode) Any conforming result for non-flush-to-zero mode

so ... correct subnormal result is still possible even if `CL_FP_DENORM` is not supported. I think we should check both subnormal and "flushed" result.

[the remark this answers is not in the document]
shajder: @lakshmih I will appreciate if you could repeat test for your `nextafter` implementation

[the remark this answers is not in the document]
lakshmih: @shajder  nextafter is now passing with your latest commit. Thanks.

test_conformance/math_brute_force/ternary_half.cpp
```diff
@@ -0,0 +1,774 @@
...
+            {
+                vlog_error("ReadArray failed %d\n", error);
+                return error;
+            }
+        }
+
+        if (gSkipCorrectnessTesting) break;
+
+        // Verify data
+        uint32_t *t = (uint32_t *)gOut_Ref;
```
lakshmih, line 265: Needs this change without which this code causes an OOB access
    cl_half *t = (cl_half *)gOut_Ref;

test_conformance/math_brute_force/ternary_half.cpp
```diff
@@ -0,0 +1,774 @@
...
+        for (size_t j = 0; j < half_buffer_size; j++)
+        {
+            for (auto k = gMinVectorSizeIndex; k < gMaxVectorSizeIndex; k++)
+            {
+                uint32_t *q = (uint32_t *)(gOut[k]);
```
lakshmih, line 270: Same thing here.
cl_half *q = (cl_half *)(gOut[k]);

lakshmih (changes requested):
Requesting changes.

test_conformance/math_brute_force/ternary_half.cpp
shajder: Thanks, it is corrected.

test_conformance/math_brute_force/ternary_half.cpp
shajder: Done

lakshmih:
Fix for FMA verified. Looks good.

bashbaug:
I haven't looked through the code, but as a high level comment: Are these fp16 tests perhaps testing too much?

Consider a unary function like `acos`.  We could test every possible input to this function and we would only need to call the function 2^16 = 64K times, yet I am seeing the kernel invoked 128K times with a global work size of 32K, which means we're calling this function 2^32 = 4B times instead.  Or, in other words, we seem to be doing about 64K times as much testing as we need to.

```
                                       Function Name,  Calls,     Time (ns), Time (%),  Average (ns),      Min (ns),      Max (ns)
...
 math_kernel SIMD32 GWS[ 32768 ] LWS[ NULL ], 131072,     815348497,   10.88%,          6220,          5104,         30468
math_kernel16 SIMD32 GWS[ 2048 ] LWS[ NULL ], 131072,    1940212796,   25.88%,         14802,         12031,         65833
math_kernel2 SIMD32 GWS[ 16384 ] LWS[ NULL ], 131072,     703839750,    9.39%,          5369,          4427,         23802
math_kernel3 SIMD32 GWS[ 10923 ] LWS[ NULL ], 131072,    1566422556,   20.90%,         11950,          9843,         47395
 math_kernel4 SIMD32 GWS[ 8192 ] LWS[ NULL ], 131072,     824479277,   11.00%,          6290,          5156,         30937
 math_kernel8 SIMD32 GWS[ 4096 ] LWS[ NULL ], 131072,     966104570,   12.89%,          7370,          6093,         35052
```

The binary (e.g. `pow`) and ternary (e.g. `fma`) functions may need to run a few more cases or do some amount of random sampling, but unary functions are the common case, so I still think it should be possible to reduce testing time significantly.

bashbaug:
> Quick experiment (for `half` test) with below modification gave promising result: [...]

This seems reasonable to me and appears to produce the expected amount of testing:

```
                                      Function Name,  Calls,     Time (ns), Time (%),  Average (ns),      Min (ns),      Max (ns)
        math_kernel SIMD32 GWS[ 32768 ] LWS[ NULL ],      2,         12134,   10.70%,          6067,          5937,          6197
       math_kernel16 SIMD32 GWS[ 2048 ] LWS[ NULL ],      2,         26197,   23.11%,         13098,         13020,         13177
       math_kernel2 SIMD32 GWS[ 16384 ] LWS[ NULL ],      2,         10676,    9.42%,          5338,          5260,          5416
       math_kernel3 SIMD32 GWS[ 10923 ] LWS[ NULL ],      2,         25832,   22.79%,         12916,         12395,         13437
        math_kernel4 SIMD32 GWS[ 8192 ] LWS[ NULL ],      2,         12551,   11.07%,          6275,          6041,          6510
        math_kernel8 SIMD32 GWS[ 4096 ] LWS[ NULL ],      2,         15155,   13.37%,          7577,          7239,          7916
```

The one thing I might check is whether the input buffers have the expected data.  At first glance it seems like we're mostly testing every value, but I'm not sure if there are some duplicates (and hence some missing values, too).

shajder:
> The one thing I might check is whether the input buffers have the expected data.

@bashbaug I just went through bruteforce procedures with similar pattern for jobs count calculation. 

As it comes to `macro_unary_half` and `unary_half` I can confirm every value for `half` type is used in input buffer - I've applied mentioned corrections for those cases.

As it comes to `macro_binary_half`, `binary_half` and `binary_i_half`, input buffers are initialized with random numbers which is aligned with `float` and `double` equivalents, thus I am leaving those cases as they are.

lakshmih (approved):

test_conformance/math_brute_force/macro_unary_half.cpp
```diff
@@ -0,0 +1,421 @@
...
+        if ((error = clReleaseEvent(e[j])))
+        {
+            vlog_error("Error: clReleaseEvent failed! err: %d\n", error);
+            return error;
+        }
+
+        // Fill the result buffer with garbage, so that old results don't carry
+        // over
+        uint16_t pattern = 0xdead;
+        memset_pattern4(out[j], &pattern, buffer_size);
```
svenvh, line 250: This triggers an AddressSanitizer: stack-buffer-overflow.

test_conformance/math_brute_force/binary_operator_float.cpp
svenvh, line 1: There is no `binary_operator_half.cpp` being added. Is that intentional?

test_conformance/math_brute_force/unary_u_half.cpp
```diff
@@ -0,0 +1,238 @@
...
+        {
+            vlog_error("\n*** Error %d in clEnqueueWriteBuffer ***\n", error);
+            return error;
+        }
+
+        // write garbage into output arrays
+        for (auto j = gMinVectorSizeIndex; j < gMaxVectorSizeIndex; j++)
+        {
+            uint16_t pattern = 0xdead;
+            memset_pattern4(gOut[j], &pattern, bufferSize);
```
svenvh, line 94: AddressSanitizer is reporting an overflow here.

test_conformance/math_brute_force/binary_half.cpp
```diff
@@ -0,0 +1,791 @@
...
+        if ((error = clReleaseEvent(e[j])))
+        {
+            vlog_error("Error: clReleaseEvent failed! err: %d\n", error);
+            return error;
+        }
+
+        // Fill the result buffer with garbage, so that old results don't carry
+        // over
+        uint16_t pattern = 0xdead;
+        memset_pattern4(out[j], &pattern, buffer_size);
```
svenvh, line 356: AddressSanitizer is reporting an overflow here.

svenvh (changes requested):

test_conformance/math_brute_force/binary_half.cpp
svenvh: I haven't checked all other `memset_pattern4` calls, but please check them as they might have the same issue.

test_conformance/math_brute_force/unary_half.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+    // Thread-specific kernels for each vector size:
+    // k[vector_size][thread_id]
+    KernelMatrix k;
+};
+
+}
+
+////////////////////////////////////////////////////////////////////////////////
+static cl_int TestHalf(cl_uint job_id, cl_uint thread_id, void *p);
```
svenvh, line 86: Why not provide the definition of `TestHalf` here right away?  Then you won't need the prototype, and you can move it into the the anonymous namespace that ends a few lines above (so you don't need `static`).  This would also make it easier to diff against `unary_float.cpp`.

test_conformance/math_brute_force/binary_half.cpp
```diff
@@ -0,0 +1,791 @@
...
+
+                if (fabsf(err) > tinfo->maxError)
+                {
+                    tinfo->maxError = fabsf(err);
+                    tinfo->maxErrorValue = s[j];
+                    tinfo->maxErrorValue2 = s2[j];
+                }
+                if (fail)
+                {
+                    vlog_error("\nERROR: %s%s: %f ulp error at {%a (0x%0.4x), "
```
svenvh, line 734: This introduces more Wformat errors:
```
error: '0' flag ignored with precision and ‘%x’ gnu_printf format
```

From the `printf` man page:
> If a precision is given with a numeric conversion (d, i, o, u, x, and X), the 0 flag is ignored.

This affects other `_half.cpp` files too.

bcalidas:
Latest commit looks good.

svenvh:
There are still open comments to be addressed before merging.

lakshmih:
Can we remove focused review from this until we find someone to pick it up?

bashbaug:
This PR needs a new owner, so for now I've removed "focused review" and added "help wanted".

test_conformance/math_brute_force/binary_operator_float.cpp
shajder: @svenvh  There is also missing `binary_two_results_i_half.cpp` to handle `remquo` `half` version. No idea what was initial intention, I am continuing that work after @gwawiork . I will add missing tests now.

test_conformance/math_brute_force/binary_i_half.cpp
```diff
@@ -0,0 +1,537 @@
...
+#include "function_list.h"
+#include "test_functions.h"
+#include "utility.h"
+
+#include <climits>
+#include <cstring>
+
+namespace {
+
+////////////////////////////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/math_brute_force/macro_unary_half.cpp
shajder: Done

test_conformance/math_brute_force/binary_operator_float.cpp
shajder: Added both `binary_operator_half.cpp` and `binary_two_results_i_half.cpp`

shajder: Added both `binary_operator_half.cpp` and `binary_two_results_i_half.cpp`

test_conformance/math_brute_force/unary_u_half.cpp
shajder: Done

test_conformance/math_brute_force/binary_half.cpp
shajder: Done, all other calls from `bruteforce` tested with Address Sanitizer.

test_conformance/math_brute_force/binary_half.cpp
```diff
@@ -0,0 +1,791 @@
...
+// A table of more difficult cases to get right
+static const cl_half specialValuesHalf[] = {
+    0xffff,
+    0x0000,
+    0x0001,
+    0x7c00 /*INFINITY*/,
+    0xfc00 /*-INFINITY*/,
+    0x8000 /*-0*/,
+    0x7bff /*HALF_MAX*/,
+    0x0400 /*HALF_MIN*/
```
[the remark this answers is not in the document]
shajder: Done, list expanded.

test_conformance/math_brute_force/mad_half.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+int TestFunc_mad_Half(const Func *f, MTdata d, bool relaxedMode)
+{
+    int error;
+    Programs programs;
+    KernelMatrix kernels;
+    const unsigned thread_id = 0; // Test is currently not multithreaded.
+    float maxError = 0.0f;
+    //    int ftz = f->ftz || gForceFTZ;
```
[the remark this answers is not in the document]
shajder: Yup, done

test_conformance/math_brute_force/mad_half.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+#include "common.h"
+#include "function_list.h"
+#include "test_functions.h"
+#include "utility.h"
+
+#include <cstring>
+
+////////////////////////////////////////////////////////////////////////////////
+cl_int BuildKernel_HalfFn(cl_uint job_id, cl_uint thread_id UNUSED, void *p)
```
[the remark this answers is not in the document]
shajder: Corrected and applied to other source files with similar problem.

test_conformance/math_brute_force/unary_half.cpp
shajder: Sounds reasonable, applied to all half source files with similar problem.

test_conformance/math_brute_force/binary_half.cpp
shajder: Done

shajder:
> This PR needs a new owner, so for now I've removed "focused review" and added "help wanted".

@bashbaug all pending remarks were addressed, maybe we could remove **`help wanted`** label to avoid confusion of potential helper ?

svenvh:
@shajder thanks for the update!  I will take a more detailed look asap.

test_conformance/math_brute_force/binary_half.cpp
```diff
@@ -0,0 +1,765 @@
...
+        {
+            vlog_error("Error: Unable to create sub-buffer of gInBuffer for "
+                       "region {%zd, %zd}\n",
+                       region.origin, region.size);
+            return error;
+        }
+        test_info.tinfo[i].inBuf2 =
+            clCreateSubBuffer(gInBuffer2, CL_MEM_READ_ONLY,
+                              CL_BUFFER_CREATE_TYPE_REGION, &region, &error);
+        if (error || NULL == test_info.tinfo[i].inBuf)
```
svenvh, line 686:
```suggestion
        if (error || NULL == test_info.tinfo[i].inBuf2)
```

shajder: good find, thanks - correction applied here and for similar cases.

test_conformance/math_brute_force/binary_half.cpp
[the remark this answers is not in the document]
svenvh: Suggesting a few more:
```
0xfbff, /* -HALF_MAX */
0x8400, /* -HALF_MIN */
0x4248, /* M_PI_H */
0xc248, /* -M_PI_H */
0xbbff, /* Largest negative fraction */
```

test_conformance/math_brute_force/i_unary_half.cpp
```diff
@@ -0,0 +1,215 @@
...
+                           error, j);
+                return error;
+            }
+        }
+
+        // Run the kernels
+        for (auto j = gMinVectorSizeIndex; j < gMaxVectorSizeIndex; j++)
+        {
+            size_t vectorSize = sizeValues[j] * sizeof(cl_int);
+            size_t localCount = (bufferSize + vectorSize - 1) / vectorSize;
```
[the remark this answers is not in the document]
shajder: Actually I think this calculation may be correct. In order to prevent access violation due to overflowing of the index of output buffer (which is the same size as input half buffer) we need to use size of bigger type to calculate work size of the kernel. Am I missing something here ?

test_conformance/math_brute_force/mad_half.cpp
```diff
@@ -0,0 +1,193 @@
...
+    float maxErrorVal = 0.0f;
+    float maxErrorVal2 = 0.0f;
+    float maxErrorVal3 = 0.0f;
+    size_t bufferSize = BUFFER_SIZE;
+
+    logFunctionInfo(f->name, sizeof(cl_half), relaxedMode);
+    uint64_t step = bufferSize / sizeof(cl_half);
+    if (gWimpyMode)
+    {
+        step = (1ULL << 32) * gWimpyReductionFactor / (512);
```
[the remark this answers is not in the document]
shajder: Yeah, that's a good question :) . No idea but I can spot the same formula in `utility.h` as well. Do you have suggestion for other value ?

test_conformance/math_brute_force/i_unary_half.cpp
```diff
@@ -0,0 +1,215 @@
...
+int TestFunc_Int_Half(const Func *f, MTdata d, bool relaxedMode)
+{
+    int error;
+    Programs programs;
+    KernelMatrix kernels;
+    const unsigned thread_id = 0; // Test is currently not multithreaded.
+    int ftz = f->ftz || 0 == (gHalfCapabilities & CL_FP_DENORM) || gForceFTZ;
+    size_t bufferSize = BUFFER_SIZE;
+    uint64_t step = getTestStep(sizeof(cl_half), BUFFER_SIZE);
+    uint64_t bufferElements = bufferSize / sizeof(cl_int);
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/math_brute_force/unary_u_half.cpp
```diff
@@ -0,0 +1,239 @@
...
+            {
+                LogBuildError(programs[j]);
+                return error;
+            }
+
+            if ((error = clEnqueueNDRangeKernel(gQueue, kernels[j][thread_id],
+                                                1, NULL, &localCount, NULL, 0,
+                                                NULL, NULL)))
+            {
+                vlog_error("FAILURE -- could not execute kernel\n");
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/math_brute_force/macro_binary_half.cpp
```diff
@@ -0,0 +1,520 @@
...
+    0x0400, /*HALF_MIN*/
+    0x03ff, /* Largest denormal */
+    0x3c00, /* 1 */
+    0xbc00, /* -1 */
+    0x3555, /*nearest value to 1/3*/
+    0x3bff, /*largest number less than one*/
+    0xc000, /* -2 */
+};
+
+size_t specialValuesHalfCount = ARRAY_SIZE(specialValuesHalf);
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/math_brute_force/i_unary_half.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+    // Init the kernels
+    {
+        BuildKernelInfo build_info = { 1, kernels, programs, f->nameInCode };
+        if ((error = ThreadPool_Do(BuildKernel_HalfFn,
+                                   gMaxVectorSizeIndex - gMinVectorSizeIndex,
+                                   &build_info)))
+            return error;
+    }
+    s.resize(bufferElements);
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/math_brute_force/ternary_half.cpp
```diff
@@ -0,0 +1,776 @@
...
+    const unsigned thread_id = 0; // Test is currently not multithreaded.
+    KernelMatrix kernels;
+    float maxError = 0.0f;
+    int ftz = f->ftz || gForceFTZ || 0 == (CL_FP_DENORM & gHalfCapabilities);
+    float maxErrorVal = 0.0f;
+    float maxErrorVal2 = 0.0f;
+    float maxErrorVal3 = 0.0f;
+    uint64_t step = getTestStep(sizeof(cl_half), BUFFER_SIZE);
+
+    constexpr size_t half_buffer_size = BUFFER_SIZE / sizeof(cl_half);
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/math_brute_force/ternary_half.cpp
```diff
@@ -0,0 +1,776 @@
...
+        {
+            vlog_error("\n*** Error %d in clEnqueueWriteBuffer3 ***\n", error);
+            return error;
+        }
+
+        // Write garbage into output arrays
+        for (auto j = gMinVectorSizeIndex; j < gMaxVectorSizeIndex; j++)
+        {
+            uint32_t pattern = 0xffffdead;
+            if (gHostFill)
```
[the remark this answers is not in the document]
shajder: Done, I have added missing `gHostFill` support for consistency.

test_conformance/math_brute_force/macro_binary_half.cpp
```diff
@@ -0,0 +1,520 @@
...
+    };
+    return BuildKernels(info, job_id, generator);
+}
+
+struct ThreadInfo
+{
+    clMemWrapper inBuf; // input buffer for the thread
+    clMemWrapper inBuf2; // input buffer for the thread
+    clMemWrapper outBuf[VECTOR_SIZE_COUNT]; // output buffers for the thread
+    MTdata d;
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/math_brute_force/mad_half.cpp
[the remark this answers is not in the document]
svenvh: Ah, I think you'll probably want to use `getTestStep` from `utility.h` here?  `mad_float.cpp` and `mad_double.cpp` are already using that helper function.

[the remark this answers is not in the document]
shajder: Indeed, done

test_conformance/math_brute_force/unary_two_results_half.cpp
```diff
@@ -0,0 +1,459 @@
...
+    float maxError1 = 0.0f;
+    int ftz = f->ftz || gForceFTZ || 0 == (CL_FP_DENORM & gHalfCapabilities);
+    float maxErrorVal0 = 0.0f;
+    float maxErrorVal1 = 0.0f;
+    uint64_t step = getTestStep(sizeof(cl_half), BUFFER_SIZE);
+
+    constexpr size_t half_buffer_size = BUFFER_SIZE / sizeof(cl_half);
+
+    cl_uchar overflow[half_buffer_size];
+    int isFract = 0 == strcmp("fract", f->nameInCode);
```
svenvh, line 57: I'm seeing a verification failure for fract with a very small input value:
```
ERROR: fract2: {-0.874878, 0.000000} ulp error at -0x1.004p-14: *{0x1p+0, -0x1p+0} vs. {0x1.ffcp-1, -0x1p+0}
```

I believe the reference calculation of the return value (`0x1p+0`) is wrong, as the OpenCL C spec s7.5.1 says:
> fract(x, iptr) shall not return a value greater than or equal to 1.0,

shajder: Indeed, input `-0x1.004p-14` (`-6.109476e-05`, hex `0x8401`) returns `0.999938905` reference float which is converted into `1.0` half float due to default `RTE` rounding method. 

Still, according to `cl_fp16_khr` extension spec `fract` should return correctly rounded result which is `1.0`, right ?

If I will put additional condition to replace reference value `1.0` with largest number less than one (`0x3bff`) my intel implementation does not pass the test (kernel `fract` returns `1.0`).

svenvh: > Still, according to cl_fp16_khr extension spec fract should return correctly rounded result which is 1.0, right ?

It doesn't say explicitly when the rounding should happen, but doing the rounding after the `fmin` step violates s7.5.1.  Hence I would argue that the intermediate result (i.e. `x - floor(x)`) should be rounded, so that the `fmin` step only sees correctly rounded fp16 values.

Intuitively I would expect `fract` to never return `1.0`, as its name suggests it returns the fractional part of the input. The purpose of the `fmin` operator is to enforce that, by capping the return value to `0x1.ffcp-1` (the largest fp16 value less than `1.0`).

shajder: >If I will put additional condition to replace reference value 1.0 with largest number less than one (0x3bff) my intel implementation does not pass the test (kernel `fract` returns 1.0).

@bashbaug could you be so kind to comment on that ? Thanks

test_conformance/math_brute_force/i_unary_half.cpp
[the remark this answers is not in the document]
svenvh: After some digging I believe the calculation may very well be correct, but the inputs to the calculation are probably too large.

For example, for the scalar case in non-wimpy mode: `bufferSize = 1024 * 1024 * 2`, which is `524288` `int`s, so we're enqueuing the kernel with a global work size of `524288`.  That is 8 times more than the 65536 possible fp16 input values.  So every possible fp16 input value is repeated 8 times, which makes the test unnecessarily slow (particularly noticeable on embedded platforms).

I guess this problem stems from the original (float/double) structure of the code, which doesn't consider the case when all possible input values can easily fit in a single buffer.

test_conformance/math_brute_force/unary_two_results_half.cpp
bashbaug: I think this is an issue with our `fract` implementation.  I just filed https://github.com/intel/intel-graphics-compiler/issues/307.  Thanks!

test_conformance/math_brute_force/i_unary_half.cpp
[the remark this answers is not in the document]
shajder: @svenvh I just applied corrections for `bufferSize` in `i_unary_half.cpp`. Two remarks here:

-buffer size computation schema is common for most of `cl_khr_fp16` `bruteforce` tests, please let me know if this correction works for you so I could propagate it across remaining tests
-with this correction previous approach for `gWimpyMode` make no longer sense so I removed section to initialize buffer for wimpy mode, if you have other suggestions I will appreciate it

test_conformance/math_brute_force/unary_two_results_half.cpp
shajder: Thanks, correction applied.

test_conformance/math_brute_force/i_unary_half.cpp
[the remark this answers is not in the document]
svenvh: > I just applied corrections for bufferSize in i_unary_half.cpp.

You only touched `bufferElements`, which doesn't affect the global work size that gets passed to `clEnqueueNDRangeKernel` as far as I can see.  So with your latest change the test still enqueues 8 times the amount of work needed, but only validates 1/8 of that work (i.e. it now validates the "correct" amount).

test_conformance/math_brute_force/unary_two_results_half.cpp
svenvh: Thanks, latest correction fixes the `fract` issue.

test_conformance/math_brute_force/i_unary_half.cpp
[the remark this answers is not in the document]
shajder: Good catch, thanks. Uninitialized buffer passed the test which is odd. Size related variables were corrected.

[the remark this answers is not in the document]
svenvh: Thanks; with your latest correction I see the global work size is as expected. And the separation of buffer size in/out variables looks good to me.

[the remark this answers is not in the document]
shajder: I found similar computation schema in `unary_two_results_half.cpp`, `unary_two_results_i_half.cpp` and `unary_u_half.cpp`. Relevant corrections applied.

svenvh (approved):
This looks good to me now; thanks!

bashbaug (commented):
I did a quick run with these changes on our implementation with:

`test_bruteforce -f -r -w -1`

Aside from the known failure with `fract` I'm also seeing a failure with `divide_cr`:

```
 83:       divide_cr fp16     
ERROR: divide_cr: -0.500000 ulp error at {-0x1.394p-12 (0x8ce5), -0x1.cp+5 (0xd300)}
Expected: 0x1.68p-18  (half 0x005a) 
Actual: 0x1.64p-18 (half 0x0059) at index: 22220
ERROR: ThreadPool_Do: TestHalf failed
! (CL_DEVICE_NOT_FOUND from /home/bashbaug/git/OpenCL-CTS/test_conformance/math_brute_force/binary_operator_half.cpp:667)
divide_cr FAILED
```

I haven't investigated further but I wanted to raise this ASAP.

bashbaug:
`fma` is failing also, if I take off `-w`:

```
 25:             fma fp16     ....
ERROR: fma: -0.500000 ulp error at {0x1.eacp+7, 0x1.3f4p+4, 0x1.c04p+14} ({0x5bab, 0x4cfd, 0x7701}): *0x1.068p+15 vs. 0x1.064p+15
fma FAILED
```

svenvh:
> It looks like the tests are checking for a correctly rounded result in this case, but it's not clear this is required. The build option specifically says (**emphasis** mine):
> 
> "-cl-fp32-correctly-rounded-divide-sqrt The -cl-fp32-correctly-rounded-divide-sqrt build option to [clBuildProgram](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clBuildProgram) or [clCompileProgram](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCompileProgram) allows an application to specify that **single precision floating-point** divide (x/y and 1/x) and sqrt used in the program source are correctly rounded. If this build option is not specified, the minimum numerical accuracy of single precision floating-point divide and sqrt are as defined in the OpenCL C or OpenCL SPIR-V Environment specifications."
> 
> I think the safest thing to do would be to disable the `divide_cr` test for fp16 while we figure out what to do with it.

Correctly rounded results for fp16-divide and fp16-sqrt are not dictated by `CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT` indeed.  Instead, "Table 11. ULP Values for Half Precision Floating-Point Arithmetic Operations" of the extension spec states that fp16-divide and fp16-sqrt need to be correctly rounded for Full Profile, and <= 1 ULP for Embedded Profile.

That made me realize that the additions to `function_list.cpp` only add the Full Profile ULP values.  For quite a few functions (e.g. `cos`, `exp`, ...), the Embedded Profile ULP value is different, but this is not taken into account in this patch.

lakshmih:
We are seeing failures in multiple tests with this patch and haven't fully resolved them to bugs in the test or in our implementation.

test_conformance/math_brute_force/binary_operator_half.cpp
```diff
@@ -0,0 +1,680 @@
...
+                    }
+
+                    // retry per section 6.5.3.3
+                    if (IsHalfSubnormal(p[j]))
+                    {
+                        double correct2, correct3;
+                        float err2, err3;
+
+                        correct2 = HTF(func.f_ff(0.0, s2[j]));
+                        correct3 = HTF(func.f_ff(-0.0, s2[j]));
```
lakshmih, line 378: The half_to_float is unnecessary here since func.f_ff() returns a double. Similar changes are needed elsewhere in the code as well. 

correct2 = func.f_ff(0.0, s2[j]);
 correct3 = func.f_ff(-0.0, s2[j]);

bashbaug:
We're still debugging a few unexpected failures on our side, also.

Can we merge these changes but with half testing disabled by default, able to be enabled via command line arguments?  Then, when debug is complete and any issues are fixed, we can re-enable half testing by default?

svenvh:
> Can we merge these changes but with half testing disabled by default, able to be enabled via command line arguments? Then, when debug is complete and any issues are fixed, we can re-enable half testing by default?

That sounds good to me; as this is a very large patch, it will be hard to get everything right the first time without the patch going stale.

bashbaug:
I just created an `fp16-staging` branch, as we discussed.  If this PR is retargeted at this branch I will merge it, also as we discussed.

shajder:
> I just created an `fp16-staging` branch, as we discussed. If this PR is retargeted at this branch I will merge it, also as we discussed.

@bashbaug I've created new [PR](https://github.com/KhronosGroup/OpenCL-CTS/pull/1863) to be merged with `fp16-staging` branch

Nuullll:
Got 2 fails with `test_bruteforce -f -r -w -1` on our device. Investigating:
```
fract...
 57:           fract fp64     ................Wimp pass	{    0.00,     0.00} @ {0x0p+0, 0x0p+0}
 58:           fract fp16     
ERROR: fract: {0.125000, 0.000000} ulp error at -0x1p-14: *{0x1.ffcp-1, -0x1p+0} vs. {0x1p+0, -0x1p+0}
fract FAILED

nextafter...
117:       nextafter fp64     ................Wimp pass	    0.00 @ {0x0p+0, 0x0p+0}
118:       nextafter fp16     
ERROR: nextafter: -nan ulp error at {0x0p+0 (0x0000), 0x1.3ccp-2 (0x34f3)}
Expected: 0x1p-24  (half 0x0001) 
Actual: -nan (half 0xffff) at index: 54

ERROR: nextafter: -nan ulp error at {0x0p+0 (0x0000), 0x1.578p-6 (0x255e)}
Expected: 0x1p-24  (half 0x0001) 
Actual: -nan (half 0xffff) at index: 4670
```

hvdijk:
I'm seeing
```
ERROR: fma: -0.500000 ulp error at {0x1.eacp+7, 0x1.3f4p+4, 0x1.c04p+14} ({0x5bab, 0x4cfd, 0x7701}): *0x1.068p+15 vs. 0x1.064p+15
fma FAILED
```
on our implementation. `0x1.eacp+7 * 0x1.3f4p+4 + 0x1.c04p+14` is exactly `0x1.065fffp+15`. Rounding to half-precision directly should give `0x1.064p+15`. What I suspect one version is doing is rounding to single-precision first to give `0x1.066p+15`, and then rounding that to half-precision to give `0x1.068p+15`, which is the wrong result. But it never actually says which of the two values is the value the CTS considers correct, which would tell us whether it is the CTS that is doing the wrong calculation, or our implementation. Is the `*` meant to indicate the correct version? "expected: <...>, actual: <...>" would be clearer, in my opinion.

bashbaug:
Removing "focused review", since we are reviewing and merging changes into the staging branch now.

hvdijk:
> I'm seeing
> 
> ```
> ERROR: fma: -0.500000 ulp error at {0x1.eacp+7, 0x1.3f4p+4, 0x1.c04p+14} ({0x5bab, 0x4cfd, 0x7701}): *0x1.068p+15 vs. 0x1.064p+15
> fma FAILED
> ```
> 
> on our implementation. `0x1.eacp+7 * 0x1.3f4p+4 + 0x1.c04p+14` is exactly `0x1.065fffp+15`. Rounding to half-precision directly should give `0x1.064p+15`. What I suspect one version is doing is rounding to single-precision first to give `0x1.066p+15`, and then rounding that to half-precision to give `0x1.068p+15`, which is the wrong result. But it never actually says which of the two values is the value the CTS considers correct, which would tell us whether it is the CTS that is doing the wrong calculation, or our implementation. Is the `*` meant to indicate the correct version? "expected: <...>, actual: <...>" would be clearer, in my opinion. Edit: Have confirmed that the CTS is the one giving the incorrect result.

I was asked to open a PR for this against `fp16-staging` for this; done as #1882.

Nuullll:
> Got 2 fails with `test_bruteforce -f -r -w -1` on our device. Investigating:
> 
> ```
> fract...
>  57:           fract fp64     ................Wimp pass	{    0.00,     0.00} @ {0x0p+0, 0x0p+0}
>  58:           fract fp16     
> ERROR: fract: {0.125000, 0.000000} ulp error at -0x1p-14: *{0x1.ffcp-1, -0x1p+0} vs. {0x1p+0, -0x1p+0}
> fract FAILED
> 
> nextafter...
> 117:       nextafter fp64     ................Wimp pass	    0.00 @ {0x0p+0, 0x0p+0}
> 118:       nextafter fp16     
> ERROR: nextafter: -nan ulp error at {0x0p+0 (0x0000), 0x1.3ccp-2 (0x34f3)}
> Expected: 0x1p-24  (half 0x0001) 
> Actual: -nan (half 0xffff) at index: 54
> 
> ERROR: nextafter: -nan ulp error at {0x0p+0 (0x0000), 0x1.578p-6 (0x255e)}
> Expected: 0x1p-24  (half 0x0001) 
> Actual: -nan (half 0xffff) at index: 4670
> ```

The `fract` and `nextafter` errors were caused by our implementation bugs. After fixing our bugs, `test_bruteforce -f -r -w` is passing on our device.

cycheng:
When is this PR expected to be merged?

shajder:
This PR was merged to main branch along with [this](https://github.com/KhronosGroup/OpenCL-CTS/pull/1975)

Issue #1481, opened 2022-08-31
Test cl_khr_command_buffer_mutable_dispatch


EwanC:
Test plan for [cl_khr_command_buffer_mutable_dispatch](https://github.com/KhronosGroup/OpenCL-Docs/pull/819) breaking down the areas which need to be covered.

## 1 Device Info Query

Verify the following queries can be performed successfully and return
defined values:

- [x] `CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR`

EDIT - Done in https://github.com/KhronosGroup/OpenCL-CTS/pull/1651

### 2.1 Mutable-dispatch

For each property in `cl_mutable_dispatch_properties_khr` reported by device query.

* Create a command-buffer with `CL_COMMAND_BUFFER_MUTABLE_KHR`
* Record `clCommandNDRangeKernelKHR` returning a `cl_mutable_command_khr` object
* Enqueue command-buffer and verify
* Set a new value for the mutable value, and enqueue again, verify result
* Enqueue command-buffer again, without updating mutable-dispatch, and verify that new values have persisted.

- [x] `CL_MUTABLE_DISPATCH_GLOBAL_OFFSET_KHR` https://github.com/KhronosGroup/OpenCL-CTS/pull/1743
- [x] `CL_MUTABLE_DISPATCH_GLOBAL_SIZE_KHR` https://github.com/KhronosGroup/OpenCL-CTS/pull/1744
- [x] `CL_MUTABLE_DISPATCH_LOCAL_SIZE_KHR` https://github.com/KhronosGroup/OpenCL-CTS/pull/1745
- [x] `CL_MUTABLE_DISPATCH_ARGUMENTS_KHR` for normal arguments and SVM. Checking variants of:
  * __global arguments https://github.com/KhronosGroup/OpenCL-CTS/pull/1749
  * __local arguments https://github.com/KhronosGroup/OpenCL-CTS/pull/1749
  * POD (plain-old-data) arguments https://github.com/KhronosGroup/OpenCL-CTS/pull/1749
  * NULL arguments https://github.com/KhronosGroup/OpenCL-CTS/pull/1749
  * image arguments https://github.com/KhronosGroup/OpenCL-CTS/pull/1750
- [x] `CL_MUTABLE_DISPATCH_EXEC_INFO_KHR` Part of SVM test in https://github.com/KhronosGroup/OpenCL-CTS/pull/1749
[...]

PR #1683, opened 2023-03-28
Add tests for cl_khr_command_buffer_mutable_dispatch.

pj87:
Add following tests for cl_khr_command_buffer_mutable_dispatch:
- Mutable-dispatch
- Simultaneous use


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_interleave.cpp
```diff
@@ -0,0 +1,818 @@
...
+        error = clEnqueueCommandBufferKHR(0, nullptr, mutable_command_buffer, 0,
+                                          nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        // cl_ndrange_kernel_command_properties_khr test_props[] = { 0, 0, 0 };
+        size_t size;
+
+        cl_mutable_dispatch_exec_info_khr exec_info;
+
+        error = clGetMutableCommandInfoKHR(
```
pj87, line 596: It returns `CL_INVALID_VALUE` and I don't know how to solve it.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_interleave.cpp
```diff
@@ -0,0 +1,818 @@
...
+        test_error(error, "clUpdateMutableCommandsKHR failed");
+
+        error = clEnqueueCommandBufferKHR(0, nullptr, mutable_command_buffer, 0,
+                                          nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        cl_ndrange_kernel_command_properties_khr test_props[] = { 0, 0, 0 };
+        size_t size;
+
+        error = clGetMutableCommandInfoKHR(
```
pj87, line 435: Should there be any specific kernel configurations for:

`CL_MUTABLE_DISPATCH_ARGUMENTS_KHR` for normal arguments and SVM. Checking variants of:

- __global arguments
- __local arguments
- POD (plain-old-data) arguments
- NULL arguments
- image arguments

?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_interleave.cpp
```diff
@@ -0,0 +1,818 @@
...
+    cl_mutable_command_khr command = nullptr;
+
+    clUpdateMutableCommandsKHR_fn clUpdateMutableCommandsKHR = nullptr;
+    clCommandQueueWrapper out_of_order_queue;
+    clCommandBufferWrapper out_of_order_command_buffer;
+};
+
+struct CrossQueueSimultaneousUse : public BasicMutableCommandBufferTest
+{
+    using BasicMutableCommandBufferTest::BasicMutableCommandBufferTest;
```
pj87, line 754: I don't know where to start with this particular test case:

**2.1.3 cross-queue simultaneous-use**
 cross-queue simultaneous-use
If a device supports simultaneous-use

Test updating mutable-dispatch objects of a command-buffer while a previous instance of the command-buffer is in flight to a different queue. See 1.3.3 from the https://github.com/KhronosGroup/OpenCL-CTS/issues/1369, but changing one of the command-buffer enqueues to the original queue.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_interleave.cpp
```diff
@@ -0,0 +1,818 @@
...
+    {
+        cl_int *pA;
+        cl_int *pB;
+        cl_int *pC;
+    } BufPtrs;
+};
+
+struct MutableUpdate : public BasicMutableCommandBufferTest
+{
+    using BasicMutableCommandBufferTest::BasicMutableCommandBufferTest;
```
pj87, line 635: I don't know where to start with this particular test case:

**2.1.1 Multiple updates**
 Test using `clUpdateMutableCommandsKHR()` in a way which updates more than one command-handle in the same call.
 Test calling `clUpdateMutableCommandsKHR()` several times on the same command-handle and make sure that the most recent value persists.

EwanC: > Test using clUpdateMutableCommandsKHR() in a way which updates more than one command-handle in the same call.

The command-buffer will need more than one command, so we have more than one handle. Then for each of these handles there is a  `cl_mutable_dispatch_config_khr` entry in the `mutable_dispatch_list` member of `cl_mutable_base_config_khr`.

> Test calling clUpdateMutableCommandsKHR() several times on the same command-handle and make sure that the most recent value persists.

Call `clUpdateMutableCommandsKHR` with some new value in `cl_mutable_dispatch_config_khr`, then change the value in the same field and immediately call `clUpdateMutableCommandsKHR`  again. Then enqueue the command-buffer and verify that the most recent value set is the one which took effect,

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_interleave.cpp
```diff
@@ -0,0 +1,808 @@
...
+        };
+
+
+        cl_mutable_dispatch_config_khr dispatch_config{
+            CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR,
+            nullptr,
+            command,
+            0 /* num_args */,
+            0 /* num_svm_arg */,
+            0 /* num_exec_infos */,
```
EwanC, line 587: this should be getting set to 1 since you're passing 1 element to`exec_info_list`

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_interleave.cpp
```diff
@@ -0,0 +1,808 @@
...
+
+        error =
+            clUpdateMutableCommandsKHR(mutable_command_buffer, &mutable_config);
+        test_error(error, "clUpdateMutableCommandsKHR failed");
+
+        error = clEnqueueCommandBufferKHR(0, nullptr, mutable_command_buffer, 0,
+                                          nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        error = clGetMutableCommandInfoKHR(
```
EwanC, line 121: For all these tests we need to verify that there has also been a successful update to device execution. So although checking the `clGetMutableCommandInfoKHR` for a correctly updated value is good, we should also have kernels with outputs that depend on these updated fields, so we can verify from the kernel output they updated corrected.
For work offset this could be something like:

```c
void kernel (__global unsigned *out) {
  out[get_global_id(0)] = get_global_offset(0);
}
```  

This potentially means each test having it's own command-buffer, with a kernel command to test the specific behaviour we're trying to verify.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+
+    cl_int Run() override
+    {
+        const char *sample_const_arg_kernel[] = {
+            "__kernel void sample_test(__constant int *src, int "
+            "dst )\n"
+            "{\n"
+            "    size_t  tid = get_global_id(0);\n"
+            "\n"
+            "    dst = src[tid];\n"
```
EwanC, line 403: can we make `dst` a buffer instead and `src` a scalar `int` so that we can verify the results.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+
+        streams[1] = clCreateBuffer(context, CL_MEM_READ_WRITE, sizeToAllocate,
+                                    NULL, &error);
+        test_error(error, "Creating test array failed");
+
+        clMemWrapper newBuffer;
+        newBuffer = clCreateBuffer(context, CL_MEM_READ_WRITE, sizeToAllocate,
+                                   NULL, &error);
+
+        cl_mutable_dispatch_arg_khr arg_1{ 1, sizeof(streams[1]) * 2, nullptr };
```
EwanC, line 477: Should be `sizeof(int)` I think and pass a new int value, such as `intarg`  earlier.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+        error = clEnqueueCommandBufferKHR(0, nullptr, mutable_command_buffer, 0,
+                                          nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        streams[1] = clCreateBuffer(context, CL_MEM_READ_WRITE, sizeToAllocate,
+                                    NULL, &error);
+        test_error(error, "Creating test array failed");
+
+        clMemWrapper newBuffer;
+        newBuffer = clCreateBuffer(context, CL_MEM_READ_WRITE, sizeToAllocate,
```
EwanC, line 474: don't need a new buffer

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+
+    clUpdateMutableCommandsKHR_fn clUpdateMutableCommandsKHR = nullptr;
+    clCommandBufferWrapper mutable_command_buffer;
+    cl_mutable_command_khr command = nullptr;
+
+    const cl_ulong max_size = 16;
+    cl_ulong currentSize;
+};
+
+struct MutableDispatchImageArguments : public BasicMutableCommandBufferTest
```
EwanC, line 693: I can't see where images are tested in this test

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+        properties[1] = CL_COMMAND_BUFFER_MUTABLE_KHR;
+
+        mutable_command_buffer =
+            clCreateCommandBufferKHR(1, &queue, properties, &error);
+        test_error(error, "clCreateCommandBufferKHR failed");
+
+        return 0;
+    }
+
+    cl_int Run() override
```
EwanC, line 893: Image tests need to check if [CL_DEVICE_IMAGE_SUPPORT](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_IMAGE_SUPPORT) is supported before running, and skip if not.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+
+        clProgramWrapper program = clCreateProgramWithSource(
+            context, 1, &set_kernel_exec_info_svm_ptrs_kernel, nullptr, &error);
+        test_error(error, "Unable to create program");
+
+        error = clBuildProgram(program, 1, &device, nullptr, nullptr, nullptr);
+        test_error(error, "Unable to build program");
+
+        kernel = clCreateKernel(program, "set_kernel_exec_info_test", &error);
+        test_error(error, "Unable to create kernel");
```
EwanC, line 1287: Can use `create_single_kernel_helper`

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+        return CL_SUCCESS;
+    }
+
+    cl_mutable_command_khr command = nullptr;
+
+    clUpdateMutableCommandsKHR_fn clUpdateMutableCommandsKHR = nullptr;
+    clCommandQueueWrapper queue_1, queue_2;
+    clCommandBufferWrapper mutable_command_buffer;
+
+    typedef struct
```
EwanC, line 1376: should pack this on host and device to try ensur it's the same size

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+        };
+        cl_mutable_base_config_khr mutable_config{
+            CL_STRUCTURE_TYPE_MUTABLE_BASE_CONFIG_KHR, nullptr, 1,
+            &dispatch_config
+        };
+        error =
+            clUpdateMutableCommandsKHR(mutable_command_buffer, &mutable_config);
+        test_error(error, "clUpdateMutableCommandsKHR failed");
+
+        clSVMFree(context, pBuf);
```
EwanC, line 1365: need to free pA, pB, pC

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+
+        error = clFinalizeCommandBufferKHR(mutable_command_buffer);
+        test_error(error, "clFinalizeCommandBufferKHR failed");
+
+        error = clEnqueueCommandBufferKHR(0, nullptr, mutable_command_buffer, 0,
+                                          nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        cl_mutable_dispatch_exec_info_khr exec_info_list{
+            CL_KERNEL_EXEC_INFO_SVM_PTRS, sizeof(BufPtrs), pBuf
```
EwanC, line 1339: should change `pBuf` of something new to update too.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_kernel.cpp
```diff
@@ -0,0 +1,1453 @@
...
+    {
+        BasicMutableCommandBufferTest::SetUp(elements);
+
+        cl_platform_id platform;
+        cl_int error =
+            clGetDeviceInfo(device, CL_DEVICE_PLATFORM, sizeof(cl_platform_id),
+                            &platform, nullptr);
+        test_error(error, "clGetDeviceInfo for CL_DEVICE_PLATFORM failed");
+
+        GET_EXTENSION_ADDRESS(clUpdateMutableCommandsKHR);
```
EwanC, line 47: Rather than duplicating this code in every test, can we put it in a new base class that every mutable test can inherit from

EwanC:
Could we split this PR up please into smaller
single purpose patches. We're trying to cover a lot right now
and it's hard to review. 

An example of how we could start breaking this is:

* PR for global offset `CL_MUTABLE_DISPATCH_GLOBAL_OFFSET_KHR`
* PR for global size `CL_MUTABLE_DISPATCH_GLOBAL_SIZE_KHR`.
* PR for local size. `CL_MUTABLE_DISPATCH_LOCAL_SIZE_KHR`
* PR for changing __global arguments. `CL_MUTABLE_DISPATCH_ARGUMENTS_KHR`
* PR for changing image arguments `CL_MUTABLE_DISPATCH_ARGUMENTS_KHR`
* PR for changing exec info SVM arguments CL_MUTABLE_DISPATCH_EXEC_INFO_KHR/

But if you have a better idea of a split go for that, or alternatively I could try to put more detail in the #1481 test plan to achieve this.

pj87:
> But if you have a better idea of a split go for that, or alternatively I could try to put more detail in the #1481 test plan to achieve this.

I can split it, but please review at least the common parts of the tests which will be used by all of the tests. After I split it into several PRs, it will be very diffucult to sychronise all of the changes in the common parts. I am talking especially about
`BasicMutableCommandBufferTest` and `ComplexMutableCommandBufferTest` structs.

PR #1686, opened 2023-03-29
cmake: Enable -Wall for non-release builds

svenvh:
Enable a larger set of compiler warnings globally for gcc/clang-like compilers.  This should help catch issues early and result in better code quality.  These warnings are primarily aimed at OpenCL-CTS developers, so only enable them for non-release builds.

Ensure `-Wall` and `-Werror` are set before any `-Wno-...` options, as the `-Wall`/`-Werror` flags undo `-Wno-...` for some compilers.

The code base does not compile cleanly with `-Wall` yet, so disable some warnings.  Do this globally for now, to keep this patch small and localized.

Silence `-Wunknown-pragmas` (instead of only making them non-fatal), to avoid verbose warning logs.


lakshmih:
Reviewing

EwanC (approved):

lakshmih (approved):

PR #1687, opened 2023-03-29
Add basic support to the harness for parallel test execution

kpet:
This change introduces a new command-line parameter to enable parallel execution by a specified number of worker threads. When parallel execution is requested, tests are distributed across the worker threads. This behaviour is disabled by default.

This does not currently work for all suites as some of them are using global variables to configure tests. For the suites that do not use global state, this change reduced the execution time by up to 5x on an 8-core machine.


lakshmih:
Reviewing.

lakshmih:
Is there a list of tests that this could affect?

kpet:
>Is there a list of suites that this could affect?

It affects all the suites that use `runTestHarness`, `runTestHarnessWithCheck`, or `parseAndCallCommandLineTests`. I think that's actually all of them.

lakshmih (approved):

Issue #1679, opened 2023-03-17
Undefined behaviour in test_absdiff


alan-baker:
The CTS for integer_ops/abs_diff executes undefined behaviour in some cases. A recent integration started triggering different verification results internally.

Here is the verification code for scalar int.
```c++
static int verify_absdiff_int( const void *p, const void *q, const void *r, size_t n, const char *sizeName, size_t vecSize )
{
    const cl_int *inA = (const cl_int *)p;
    const cl_int *inB = (const cl_int *)q;
    const cl_uint *outptr = (const cl_uint *)r;
    size_t i;
    for( i = 0; i < n; i++ )
    {
        cl_uint r = inA[i] - inB[i];
        if( inB[i] > inA[i] )
            r = inB[i] - inA[i];
        if( r != outptr[i] )
        {
            log_info( "%ld) Failure for absdiff( (int%s) 0x%8.8x, (int%s) 0x%8.8x) = *0x%8.8x vs 0x%8.8x\n", i, sizeName, inA[i], sizeName, inB[i], r, outptr[i] );
            return -1;
        }
    }
    return 0;
}
```

The test case we caught was:
> 0) Failure for absdiff( (int) 0x8c7f0aac, (int) 0x1902f8c8) = *0x737c11e4 vs 0x8c83ee1c

In this case `inB[i] - inA[i]` overflows.

An alternative implementation that passes is:
```c++
template<typename ST, typename UT>
UT absdiff(ST a, ST b) {
  const bool aneg = a < 0;
  const bool bneg = b < 0;
  if (aneg == bneg) {
    // Subtraction won't overflow.
    const ST diff = a-b;
    return diff < 0 ? -diff : diff;
  } else {
    const UT higher = aneg ? b : a;
    const ST lower = aneg ? a : b;
    // Can't trust negation of the most negative value.
    const UT neg_lower = (lower == std::numeric_limits<ST>::min()) ? static_cast<UT>(lower) : static_cast<UT>(-lower);
    const UT diff = higher + neg_lower;
    return diff;
  }
}
```

PR #1689, opened 2023-03-31
Avoid integer overflows when computing reference in test_absdiff

jmmartinez:
After compiling OpenCL CTS with -fsanitize=undefined I stumbled upon some signed integer overflows.

test_absdiff.cpp:103:24: runtime error: signed integer overflow:
  419625160 - -1937831252 cannot be represented in type 'int'
test_absdiff.cpp:101:28: runtime error: signed integer overflow:
  1277901399 - -1294103363 cannot be represented in type 'int'
test_absdiff.cpp:140:24: runtime error: signed integer overflow:
  8945130301981949496 - -2872789959208163723 cannot be represented in type
  'long int'
test_absdiff.cpp:138:29: runtime error: signed integer overflow:
  5488544718097069860 - -5558131619970145206 cannot be represented in type
  'long int'

The result from these operations was used to compare against the kernel result.

This patch replaces the abs_diff manual computations by a function that handles the signed integer overflow case.

### DIFF
--- a/test_conformance/integer_ops/test_absdiff.cpp
+++ b/test_conformance/integer_ops/test_absdiff.cpp
@@ -22,6 +22,17 @@
 
 #include "procs.h"
 
+template <class Integer>
+static typename std::make_unsigned<Integer>::type abs_diff(Integer a, Integer b)
+{
+    using Unsigned = typename std::make_unsigned<Integer>::type;
+    Unsigned ua = a;
+    Unsigned ub = b;
+    Unsigned diff = ua - ub;
+    if (a < b) diff = -diff;
+    return diff;
+}
+
 static int verify_absdiff_char( const void *p, const void *q, const void *r, size_t n, const char *sizeName, size_t vecSize )
 {
     const cl_char *inA = (const cl_char *)p;
@@ -30,9 +41,7 @@ static int verify_absdiff_char( const void *p, const void *q, const void *r, siz
     size_t i;
     for( i = 0; i < n; i++ )
     {
-        cl_uchar r = inA[i] - inB[i];
-        if( inB[i] > inA[i] )
-            r = inB[i] - inA[i];
+        cl_uchar r = abs_diff(inA[i], inB[i]);
         if( r != outptr[i] )
         { log_info( "%ld) Failure for absdiff( (char%s) 0x%2.2x, (char%s) 0x%2.2x) = *0x%2.2x vs 0x%2.2x\n", i, sizeName, inA[i], sizeName, inB[i], r, outptr[i] ); return -1; }
     }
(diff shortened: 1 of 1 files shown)


kpet:
This relates https://github.com/KhronosGroup/OpenCL-CTS/issues/1679 and probably fixes it.

jmmartinez:
@EwanC , I've noticed you are quite active in the project so I wanted to ask: do you know a reviewer that could be assigned to this PR? Or how to make this PR move forward?

Otherwise if the PR is not relevant I think we can close it.

EwanC:
> @EwanC , I've noticed you are quite active in the project so I wanted to ask: do you know a reviewer that could be assigned to this PR? Or how to make this PR move forward?
> 
> Otherwise if the PR is not relevant I think we can close it.

Hi @jmmartinez - thanks for your contribution, I think it is relevant. I'll find time to review this and will also attach a label that means this PR should get prioritized by the OpenCL Working Group for review.

jmmartinez:
Thank you for the quick response! I'll wait for the reviews then.

EwanC (approved):
I tried this out and verified the UBsan warning disappeared and the test still passed on a couple of different implementations. I'm not 100% sure about the difference between this solution and the one proposed in #1679. However, I tried both with a couple values and they produced the same result, so have a preference for this more concise version.

```cpp
/*
$ ./main
a -1937831252
b 419625160
wrong 1937510884
abs_diff_1(a,b) 2357456412
abs_diff_2(a,b) 2357456412

min_int -2147483648
abs_diff_1(min_int, b) 2567108808
abs_diff_2(min_int, b) 2567108808

*/

#include <iostream>

template <class Integer>
typename std::make_unsigned<Integer>::type abs_diff_1(Integer a, Integer b)
{
    using Unsigned = typename std::make_unsigned<Integer>::type;
    Unsigned ua = a;
    Unsigned ub = b;
  
    Unsigned diff = ua - ub;
    if (a < b) diff = -diff;
    return diff;
}

template <class ST>
typename std::make_unsigned<ST>::type 
abs_diff_2(ST a, ST b) {
  using UT = typename std::make_unsigned<ST>::type;
  const bool aneg = a < 0;
  const bool bneg = b < 0;
  if (aneg == bneg) {
    // Subtraction won't overflow.
    const ST diff = a-b;
    return diff < 0 ? -diff : diff;
  } else {
    const UT higher = aneg ? b : a;
    const ST lower = aneg ? a : b;
    // Can't trust negation of the most negative value.
    const UT neg_lower = (lower == std::numeric_limits<ST>::min()) ? static_cast<UT>(lower) : static_cast<UT>(-lower);
    const UT diff = higher + neg_lower;
    return diff;
  }
}

int main() {
 int a = 0x8c7f0aac;
 int b = 0x1902f8c8;
 unsigned bad = a - b;
 std::cout << "a " << a << "\n";
 std::cout << "b " << b << "\n";
 std::cout << "wrong " << bad << "\n";

 unsigned v1 = abs_diff_1(a, b);
 unsigned v2 = abs_diff_2(a, b);

 std::cout << "abs_diff_1(a,b) " << v1 << "\n";
 std::cout << "abs_diff_2(a,b) " << v2 << "\n";

 int min_int = std::numeric_limits<int>::min();
 v1 = abs_diff_1(min_int, b);
 v2 = abs_diff_2(min_int, b);

 std::cout << "abs_diff_1(min_int, b) " << v1 << "\n";
 std::cout << "abs_diff_2(min_int, b) " << v2 << "\n";
 return 0;
}
```

lakshmih:
We are still reviewing and testing this. Thanks.

lakshmih (approved):
Looks good.

svenvh (approved):

PR #1695, opened 2023-04-04
Added cl_khr_fp16 extension support for test_commonfns

shajder:
This PR is direct extension of modernization PR [#1694](https://github.com/KhronosGroup/OpenCL-CTS/pull/1694) . To avoid conflicts I would suggest to merge modernization PR first. Additional remarks:
-in degrees/radians test cases I had to add condition to verify if result and reference are both infinite
-I've changed ranges of half generation to stick to reasonable values - that could be possible candidate for corrections


test_conformance/commonfns/test_unary_fn.cpp
```diff
@@ -0,0 +1,427 @@
...
+
+    error = test_unary_fn<float>(device, context, queue, num_elems,
+                                 fnName.c_str(), verify_sign<float>);
+    test_error(error, "SignTest::Run<float> failed");
+
+    if (is_extension_available(device, "cl_khr_fp64"))
+    {
+        error = test_unary_fn<double>(device, context, queue, num_elems,
+                                      fnName.c_str(), verify_sign<double>);
+        test_error(error, "SignTest::Run<cl_half> failed");
```
EwanC, line 398:
```suggestion
        test_error(error, "SignTest::Run<cl_double> failed");
```

test_conformance/commonfns/test_unary_fn.cpp
```diff
@@ -0,0 +1,427 @@
...
+        else
+        {
+            log_info("%s %s%d test passed\n", fnName.c_str(), tname.c_str(),
+                     ((g_arrVecSizes[i])));
+        }
+
+        if (err) break;
+    }
+
+    if (err) return err;
```
EwanC, line 324: don't need this line, returning err unconditionally on the next line

test_conformance/commonfns/test_unary_fn.cpp
```diff
@@ -0,0 +1,427 @@
...
+                  const std::string &fnName, VerifyFuncUnary<T> verifyFn)
+{
+    clMemWrapper streams[2];
+    std::vector<T> input_ptr, output_ptr;
+
+    std::vector<clProgramWrapper> programs;
+    std::vector<clKernelWrapper> kernels;
+
+    int err, i;
+    MTdata d;
```
EwanC, line 194: Since this is brand new file, preference for using `MTDataHolder`

test_conformance/commonfns/test_base.h
```diff
@@ -0,0 +1,282 @@
...
+{
+    int sformat = std::snprintf(nullptr, 0, format.c_str(), args...) + 1;
+    if (sformat <= 0)
+        throw std::runtime_error("string_format: string processing error.");
+    auto format_size = static_cast<size_t>(sformat);
+    std::unique_ptr<char[]> buffer(new char[format_size]);
+    std::snprintf(buffer.get(), format_size, format.c_str(), args...);
+    return std::string(buffer.get(), buffer.get() + format_size - 1);
+}
+
```
EwanC, line 194: I feel like i've seen this function, and maybe some others here in other half PRs. If we're going to duplicate the code it might be worth adding the functino somewhere in `<root>/test_common/harness/`  and reusing it.

test_conformance/commonfns/test_base.h
```diff
@@ -0,0 +1,282 @@
...
+    {
+        return Ulp_Error(val, r);
+    }
+    else if (std::is_same<T, double>::value)
+    {
+        return Ulp_Error_Double(val, r);
+    }
+    else
+    {
+        log_error("GeometricsFPTest::UlpError: unsupported data type\n");
```
EwanC, line 212: ~~`GeometricsFPTest`~~

test_conformance/commonfns/main.cpp
```diff
@@ -32,32 +37,57 @@ static void initVecSizes() {
...
+    return TEST_PASS;
+}
+
+//--------------------------------------------------------------------------
 int main(int argc, const char *argv[])
 {
     initVecSizes();
-    return runTestHarness(argc, argv, test_num, test_list, false, 0);
+
+    if (BaseFunctionTest::type2name.empty())
```
EwanC, line 82: Could this be an assert? Assuming this condition would almost always hold since we're at the start of `main()`

test_conformance/commonfns/test_unary_fn.cpp
shajder: Done

test_conformance/commonfns/main.cpp
shajder: Correct, condition redundant

test_conformance/commonfns/test_base.h
shajder: Done

test_conformance/commonfns/test_base.h
shajder: @EwanC I moved it to `test_common/stringHelpers.h`

test_conformance/commonfns/test_mix.cpp
```diff
@@ -142,18 +161,33 @@ int test_mix_fn(cl_device_id device, cl_context context, cl_command_queue queue,
         test_error(err, "clCreateBuffer failed");
     }
 
-    for (i = 0; i < num_elements; i++)
+    std::string pragma_str;
+    if (std::is_same<T, double>::value)
```
EwanC, line 165: We're already setting this sting a couple of lines later, should this be checking for half and setting the half pragma?

test_conformance/commonfns/test_unary_fn.cpp
```diff
@@ -207,33 +225,38 @@ int test_unary_fn(cl_device_id device, cl_context context,
                 get_random_double(-100000.0 * M_PI, 100000.0 * M_PI, d);
         }
     }
+    else if (std::is_same<T, half>::value)
+    {
+        pragma_str = "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n";
+        for (int j = 0; j < num_elements; j++)
+        {
+            input_ptr[j] = conv_to_half(get_random_float(
+                (float)(-65503.f * M_PI), (float)(65503.f * M_PI), d));
```
EwanC, line 234: for the other tests we've just copied the random float range from the float and double tests (even when it's a hex values like `0x20000000` I don't really understand). Why `-65503.f` here?

test_conformance/commonfns/test_smoothstep.cpp
```diff
@@ -95,32 +101,45 @@ int verify_smoothstep(const T *const edge0, const T *const edge1,
...
+                        return -1;
+                    }
                 }
+                else
+                    max_delta = std::max(max_delta, delta);
             }
         }
     }
+
+    if (std::is_same<T, half>::value)
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/commonfns/test_mix.cpp
shajder: Done

test_conformance/commonfns/test_mix.cpp
```diff
@@ -120,7 +139,7 @@ int test_mix_fn(cl_device_id device, cl_context context, cl_command_queue queue,
     std::vector<clKernelWrapper> kernels;
 
     int err, i;
-    MTdataHolder d = MTdataHolder(gRandomSeed);
+    MTdata d;
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/commonfns/test_unary_fn.cpp
EwanC: `10000.0 * M_PI` is more in line with the multiplier used in the float/double tests, so happy enough with that. Though I can't say I know why `+/- 100000.0 * M_PI` is used, I don't have any better suggestions for moving forward here without reverse engineering how all the random input limits are set.

test_conformance/commonfns/test_mix.cpp
shajder: Corrected the same way as `smoothstep` - due to the fact that accuracy of `mix` for `cl_khr_fp16` is implementation defined this test should only report maximum error without testing maximum error threshold

test_conformance/commonfns/test_mix.cpp
```diff
@@ -90,25 +109,48 @@ int verify_mix(const T *const inptrX, const T *const inptrY,
...
+                if (!std::is_same<T, half>::value)
                 {
-                    log_error("{%d, element %d}) verification error: mix(%a, "
-                              "%a, %a) = *%a vs. %a\n",
-                              ii, j, inptrX[vi], inptrY[vi], inptrA[i], r,
-                              outptr[vi]);
-                    return -1;
+                    if (delta > MAX_ERR)
+                    {
+                        if (std::is_same<T, half>::value)
```
EwanC, line 120: Don't think we need this `if` condition any more given we check `if (!std::is_same<T, half>::value)` a few lines above

shajder: of course, thanks!

EwanC (approved):

test_conformance/commonfns/test_step.cpp
```diff
@@ -140,6 +147,16 @@ int test_step_fn(cl_device_id device, cl_context context,
             input_ptr[1][i] = get_random_double(-0x40000000, 0x40000000, d);
         }
     }
+    else if (std::is_same<T, half>::value)
+    {
+        const float fval = 0x40000000;
```
svenvh, line 152: I think we need to decrease this value, as the generated data results in infinities most of the time.

svenvh (changes requested):

test_conformance/commonfns/test_clamp.cpp
```diff
@@ -169,6 +199,17 @@ int test_clamp_fn(cl_device_id device, cl_context context,
             input_ptr[2][j] = get_random_double(input_ptr[1][j], 0x20000000, d);
         }
     }
+    else if (std::is_same<T, half>::value)
+    {
+        const float fval = 0x200000;
```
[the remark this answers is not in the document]
svenvh: Thanks; the input values look more useful now.  Using `CL_HALF_MAX` here sounds reasonable to me.

It would indeed be nice to test infinity cases too, but obtaining them from random generation feels a bit unreliable. We could add some special cases in the future perhaps (but that's separate work).  For now, this looks good enough to me.

svenvh (approved):

lakshmih (dismissed):
Looks good except for the merge conflict.

lakshmih (approved):

PR #1696, opened 2023-04-04
[NFC] Remove unused depth_lod variables

svenvh:
Remove the remaining unused variables in the test harness.

Even though this commit does not change behavior, I'm not sure if the lack of a use of these variables highlights an underlying issue?


lakshmih:
Reviewing

lakshmih (approved):
Looks good.

PR #1705, opened 2023-04-21
math_brute_force: consider all types for extension pragmas

svenvh:
When generating the kernel code, consider the return type(s) and the types of all parameters, instead of only the first parameter type. This fixes a missing extension pragma for certain cases (such as `nan`).

### DIFF
--- a/test_conformance/math_brute_force/common.cpp
+++ b/test_conformance/math_brute_force/common.cpp
@@ -67,22 +67,28 @@ void EmitDefineUndef(std::ostringstream &kernel, const char *name,
     kernel << "#define " << name << " " << GetUndefValue(type) << '\n';
 }
 
-void EmitEnableExtension(std::ostringstream &kernel, ParameterType type)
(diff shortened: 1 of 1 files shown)


test_conformance/math_brute_force/common.cpp
```diff
@@ -67,22 +67,28 @@ void EmitDefineUndef(std::ostringstream &kernel, const char *name,
...
+        switch (type)
+        {
+            case ParameterType::Double: needsFp64 = true; break;
+
+            case ParameterType::Float:
+            case ParameterType::Int:
+            case ParameterType::UInt:
+            case ParameterType::Long:
+            case ParameterType::ULong:
+                // No extension required.
```
shajder, line 86: I've noticed this test does not verify neither `gHasLong` nor `cles_khr_int64` extension ? Is it unnecessary in here ?

EwanC (approved):

PR #1709, opened 2023-04-25
atomics: Remove unused variable in atomic_add_index_bin

svenvh:
Remove a variable that was set when an OpenCL API call failed, but never read again.  Instead, return immediately upon failure.

### DIFF
--- a/test_conformance/atomics/test_indexed_cases.cpp
+++ b/test_conformance/atomics/test_indexed_cases.cpp
@@ -201,7 +201,6 @@ int add_index_bin_test(size_t *global_threads, cl_command_queue queue,
     int number_of_bins = number_of_items / divisor;
     int max_counts_per_bin = divisor * 2;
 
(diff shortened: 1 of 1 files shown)


EwanC (approved):

lakshmih:
From the test-spec, it looks like all subtests are run in the same process.

In this case, there might be memory leakage since we're returning before the mallocs are freed. (The CL buffers aren't released either, even if their allocations were successful which is not even checked for.)
This would only matter if this modified function (add_index_bin_test) fails and one/more of the other failing tests failed because of a leakage from this function all in the same process.

svenvh:
This test (and many other tests in fact) leak on the error paths indeed, and I guess my change has made that slightly worse (although continuing to run the test when CL calls have failed isn't great either).  To solve that, we should be using cl*Wrapper objects and `unique_ptr`s; I'll see if I can put up a patch for this test.

Issue #1054, opened 2020-11-10
CTS select test takes extremely long execution time on mobile GPU


yzhang-verisilicon:
OpenCL CTS test program test_select.cpp has several hot spots that need to be optimized or changed. In the function

`static int doTest(cl_command_queue queue, cl_context context, Type stype, Type cmptype, cl_device_id device)`

The following code segments cause the extremely long running time:

1. `cl_ulong blocks = type_size[stype] * 0x100000000ULL / BUFFER_SIZE;`

2. `initCmpBuffer(s3, cmptype, i * cmp_stride, block_elements);`

3. `(*sfunc)(ref, s1, s2, s3, block_elements);`

4. `(*sfunc)(sref, s1, s2, s3, block_elements);`

5. `if ((*checkResults[stype])(d, vecsize == 0 ? sref : ref, block_elements, element_count[vecsize])!=0){`

yzhang-verisilicon:
Some more data for CTS running time on our test platforms.

These two:
    test_conversions    60+ hours
    test_bruteforce      28 hours
are the most critical CTS tests that need to be improved to reduce the running time. 

The following seven tests take 2-3 hours (each) to complete. Nice to improve but they are not critical.
    Test_thread_dimensions full*
    Test_select
    Test_integer_ops
    Test_half
    Test_image_streams CL_FILTER_NEAREST
    Test_image_streams CL_FILTER_NEAREST use_pitches
    Test_image_streams CL_FILTER_NEAREST max_images

Jeremy-Kemp:
I would echo the same sentiments about conversions and bruteforce, both take up the majority of our CTS runtime. 

Select, integer ops and half are also tests that take between 7-12 hours each.

yzhang-verisilicon:
We have tested OpenCL-CTS-2021-02-01-00 release tag on one of VSI GC7000 platforms. 

With the following changes in "opencl_conformance_tests_full.csv". The OpenCL 3.0 CTS can be completed in about 24 hours on a production SoC platform.

diff opencl_conformance_tests_full.csv opencl_conformance_tests_full_VSI.csv
2c2
< # OpenCL Conformance Test Suite (full version)
---
> # OpenCL Conformance Test Suite (VSI version)
27c27
< Thread Dimensions,thread_dimensions/test_thread_dimensions full*
---
> Thread Dimensions,thread_dimensions/test_thread_dimensions quick*
77c77
< Conversions,conversions/test_conversions
---
[...]

PR #1712, opened 2023-04-26
select: using clEnqueueReadBuffer rather than clEnqueueMapBuffer

dreamibor:
In this PR, I'd propose to replace clEnqueueMapBuffer with clEnqueueReadBuffer while reading data from buffers.

In my opinion:
- clEnqueueMapBuffer should be used to read/write small data randomly.
- clEnqueueReadBuffer should be used to read the whole data from the buffer **frequently**.

Vendors implement the clEnqueueMapBuffer API differently. If the buffer is located on the device, and clEnqueueMapBuffer is returning a CPU virtual address of it, the reading speed vary. For example, PCIe writing is asynchronous while PCIe reading is synchronous. This results a gap between the speed of reading and writing to the memory/pointer returned by clEnqueueMapBuffer.

Loading the buffer back to CPU using clEnqueueReadBuffer ensures the frequent data reading always happen on CPU, and it will increase the speed of creating reference result and comparing/checking result. For example, on my test machine, the runtime of test_select program **reduced from 82 hours to 45 minutes**. This is also relevant to the ticket here: #1054 .

Also, the clEnqueueMapBuffer function used for creating reference result is actaully created for "writing" (CL_MAP_WRITE), and later we used it for "reading" as well.
```
void *s3 = clEnqueueMapBuffer( queue, cmp, CL_TRUE, CL_MAP_WRITE, 0, BUFFER_SIZE, 0, NULL, NULL, &err );
if( err ){ log_error( "Error: Could not map cmp" ); goto exit; }
// Setup the input data to change for each block
initCmpBuffer(s3, cmptype, i * cmp_stride, block_elements);

// Create the reference result
Select sfunc = (cmptype == ctype[stype][0]) ? vrefSelects[stype][0] : vrefSelects[stype][1];
(*sfunc)(ref, s1, s2, s3, block_elements);
```
I've tested the patch with Nvidia RTX 3090 (Driver Version: 530.30.02, CUDA version 12.1) as well, the correctness is okay. It won't reduce the runtime of their test_select, but I'd guess they have an optimized implementation for clEnqueueMapBuffer.

### DIFF
--- a/test_conformance/select/test_select.cpp
+++ b/test_conformance/select/test_select.cpp
@@ -363,14 +363,30 @@ static int doTest(cl_command_queue queue, cl_context context, Type stype, Type c
     dest = clCreateBuffer( context, CL_MEM_WRITE_ONLY, BUFFER_SIZE, NULL, &err );
     if( err ) { log_error( "Error: could not allocate dest buffer\n" );  ++s_test_fail; goto exit; }
 
-    src1_host = malloc( BUFFER_SIZE );
-    if( NULL == src1_host ){ log_error("Error: could not allocate src1_host buffer\n" ); goto exit; }
-    src2_host = malloc( BUFFER_SIZE );
-    if( NULL == src2_host ){ log_error("Error: could not allocate src2_host buffer\n" ); goto exit; }
-    cmp_host = malloc( BUFFER_SIZE );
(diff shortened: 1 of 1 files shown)


bashbaug:
Merging as discussed in the May 30th teleconference.

PR #1713, opened 2023-04-27
Added cl_khr_fp16 extension support for test_explicit_s2v from basic

shajder:

Additional remark (1): original test skips conversions between different types, in other words supported conversions were eg. `char scalar`->`char vector`, `int scalar`->`int vector`, `float scalar`->`float vector` and so on. This PR extend the scope of original test to `integral scalar types`->`integral vector types`, `floating scalar types`->`floating vector types`

Additional remark (2): test fails with my Intel device and kernel with `half` conversions: 
_**ERROR: Explicit cast of scalar float to vector half2 FAILED; skipping other half vector tests**_
I will appreciate additional attention when it comes to testing `cl_khr_fp16` with this test


test_conformance/basic/test_explicit_s2v.cpp
```diff
@@ -170,11 +139,13 @@ int test_explicit_s2v_function(cl_device_id deviceID, cl_context context, cl_com
...
      do a memcpy instead of relying on the actual type of data */
-    error = clEnqueueReadBuffer( queue, streams[1], CL_TRUE, 0, destStride * count, outData, 0, NULL, NULL );
+    error =
+        clEnqueueReadBuffer(queue, streams[1], CL_TRUE, 0, destStride * count,
+                            outData.data(), 0, NULL, NULL);
     test_error( error, "Unable to read output values!" );
 
     inPtr = (unsigned char *)inputData;
-    outPtr = (unsigned char *)outData;
+    outPtr = (unsigned char *)outData.data();
```
EwanC, line 148: With this change I'm seeing fails when converting char to other integer types, I suspect it's because of differences in the signedness of char between platforms, as I see these fails on Arm builds but not x86 builds. 

`char` has implementation defined sign in C, and is generally signed on x86 and unsigned on Arm. However, in OpenCL-C `char` is always signed. If I modify the verification function `convert_explicit_value` to use `signed char` to match the device these fails dissapear (i'm also seeing fails on float types but need to investigate further):

```diff
diff --git a/test_common/harness/conversions.cpp b/test_common/harness/conversions.cpp
index d52a2ac..03c3718 100644
--- a/test_common/harness/conversions.cpp
+++ b/test_common/harness/conversions.cpp
@@ -386,6 +386,7 @@ static Long sLowerLimits[kNumExplicitTypes] = {
     break;
 
 typedef unsigned char uchar;
+typedef signed char schar;
 typedef unsigned short ushort;
 typedef unsigned int uint;
 typedef unsigned long ulong;
@@ -395,7 +396,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                             ExplicitType outType)
 {
     bool *boolPtr;
-    char *charPtr;
+    schar *scharPtr;
     uchar *ucharPtr;
     short *shortPtr;
     ushort *ushortPtr;
@@ -449,29 +450,29 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
             break;
 
         case kChar:
-            charPtr = (char *)inRaw;
+            scharPtr = (schar *)inRaw;
             switch (outType)
             {
-                BOOL_CASE(char)
+                BOOL_CASE(schar)
 
                 case kChar:
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(char, kUChar, uchar, saturate)
-                    SIMPLE_CAST_CASE(char, kUnsignedChar, uchar)
-                    SIMPLE_CAST_CASE(char, kShort, short)
-                    SIMPLE_CAST_CASE(char, kUShort, ushort)
-                    SIMPLE_CAST_CASE(char, kUnsignedShort, ushort)
-                    SIMPLE_CAST_CASE(char, kInt, int)
-                    SIMPLE_CAST_CASE(char, kUInt, uint)
-                    SIMPLE_CAST_CASE(char, kUnsignedInt, uint)
-                    SIMPLE_CAST_CASE(char, kLong, Long)
-                    SIMPLE_CAST_CASE(char, kULong, ULong)
-                    SIMPLE_CAST_CASE(char, kUnsignedLong, ULong)
-
-                    TO_FLOAT_CASE(char)
-                    TO_DOUBLE_CASE(char)
+                    DOWN_CAST_CASE(schar, kUChar, uchar, saturate)
+                    SIMPLE_CAST_CASE(schar, kUnsignedChar, uchar)
+                    SIMPLE_CAST_CASE(schar, kShort, short)
+                    SIMPLE_CAST_CASE(schar, kUShort, ushort)
+                    SIMPLE_CAST_CASE(schar, kUnsignedShort, ushort)
+                    SIMPLE_CAST_CASE(schar, kInt, int)
+                    SIMPLE_CAST_CASE(schar, kUInt, uint)
+                    SIMPLE_CAST_CASE(schar, kUnsignedInt, uint)
+                    SIMPLE_CAST_CASE(schar, kLong, Long)
+                    SIMPLE_CAST_CASE(schar, kULong, ULong)
+                    SIMPLE_CAST_CASE(schar, kUnsignedLong, ULong)
+
+                    TO_FLOAT_CASE(schar)
+                    TO_DOUBLE_CASE(schar)
 
                 default:
                     log_error("ERROR: Invalid type given to "
@@ -491,7 +492,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(uchar, kChar, char, saturate)
+                    DOWN_CAST_CASE(uchar, kChar, schar, saturate)
                     SIMPLE_CAST_CASE(uchar, kShort, short)
                     SIMPLE_CAST_CASE(uchar, kUShort, ushort)
                     SIMPLE_CAST_CASE(uchar, kUnsignedShort, ushort)
@@ -523,7 +524,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(uchar, kChar, char, saturate)
+                    DOWN_CAST_CASE(uchar, kChar, schar, saturate)
                     SIMPLE_CAST_CASE(uchar, kShort, short)
                     SIMPLE_CAST_CASE(uchar, kUShort, ushort)
                     SIMPLE_CAST_CASE(uchar, kUnsignedShort, ushort)
@@ -554,7 +555,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(short, kChar, char, saturate)
+                    DOWN_CAST_CASE(short, kChar, schar, saturate)
                     DOWN_CAST_CASE(short, kUChar, uchar, saturate)
                     DOWN_CAST_CASE(short, kUnsignedChar, uchar, saturate)
                     DOWN_CAST_CASE(short, kUShort, ushort, saturate)
@@ -587,7 +588,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(ushort, kChar, char, saturate)
+                    DOWN_CAST_CASE(ushort, kChar, schar, saturate)
                     DOWN_CAST_CASE(ushort, kUChar, uchar, saturate)
                     DOWN_CAST_CASE(ushort, kUnsignedChar, uchar, saturate)
                     DOWN_CAST_CASE(ushort, kShort, short, saturate)
@@ -619,7 +620,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(ushort, kChar, char, saturate)
+                    DOWN_CAST_CASE(ushort, kChar, schar, saturate)
                     DOWN_CAST_CASE(ushort, kUChar, uchar, saturate)
                     DOWN_CAST_CASE(ushort, kUnsignedChar, uchar, saturate)
                     DOWN_CAST_CASE(ushort, kShort, short, saturate)
@@ -650,7 +651,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(int, kChar, char, saturate)
+                    DOWN_CAST_CASE(int, kChar, schar, saturate)
                     DOWN_CAST_CASE(int, kUChar, uchar, saturate)
                     DOWN_CAST_CASE(int, kUnsignedChar, uchar, saturate)
                     DOWN_CAST_CASE(int, kShort, short, saturate)
@@ -683,7 +684,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(uint, kChar, char, saturate)
+                    DOWN_CAST_CASE(uint, kChar, schar, saturate)
                     DOWN_CAST_CASE(uint, kUChar, uchar, saturate)
                     DOWN_CAST_CASE(uint, kUnsignedChar, uchar, saturate)
                     DOWN_CAST_CASE(uint, kShort, short, saturate)
@@ -715,7 +716,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(uint, kChar, char, saturate)
+                    DOWN_CAST_CASE(uint, kChar, schar, saturate)
                     DOWN_CAST_CASE(uint, kUChar, uchar, saturate)
                     DOWN_CAST_CASE(uint, kUnsignedChar, uchar, saturate)
                     DOWN_CAST_CASE(uint, kShort, short, saturate)
@@ -746,7 +747,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    DOWN_CAST_CASE(Long, kChar, char, saturate)
+                    DOWN_CAST_CASE(Long, kChar, schar, saturate)
                     DOWN_CAST_CASE(Long, kUChar, uchar, saturate)
                     DOWN_CAST_CASE(Long, kUnsignedChar, uchar, saturate)
                     DOWN_CAST_CASE(Long, kShort, short, saturate)
@@ -779,7 +780,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    U_DOWN_CAST_CASE(ULong, kChar, char, saturate)
+                    U_DOWN_CAST_CASE(ULong, kChar, schar, saturate)
                     U_DOWN_CAST_CASE(ULong, kUChar, uchar, saturate)
                     U_DOWN_CAST_CASE(ULong, kUnsignedChar, uchar, saturate)
                     U_DOWN_CAST_CASE(ULong, kShort, short, saturate)
@@ -811,7 +812,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
                     memcpy(outRaw, inRaw, get_explicit_type_size(inType));
                     break;
 
-                    U_DOWN_CAST_CASE(ULong, kChar, char, saturate)
+                    U_DOWN_CAST_CASE(ULong, kChar, schar, saturate)
                     U_DOWN_CAST_CASE(ULong, kUChar, uchar, saturate)
                     U_DOWN_CAST_CASE(ULong, kUnsignedChar, uchar, saturate)
                     U_DOWN_CAST_CASE(ULong, kShort, short, saturate)
@@ -838,7 +839,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
             {
                 BOOL_CASE(float)
 
-                FLOAT_ROUND_CASE(kChar, char, roundType, saturate)
+                FLOAT_ROUND_CASE(kChar, schar, roundType, saturate)
                 FLOAT_ROUND_CASE(kUChar, uchar, roundType, saturate)
                 FLOAT_ROUND_CASE(kUnsignedChar, uchar, roundType, saturate)
                 FLOAT_ROUND_CASE(kShort, short, roundType, saturate)
@@ -870,7 +871,7 @@ void convert_explicit_value(void *inRaw, void *outRaw, ExplicitType inType,
             {
                 BOOL_CASE(double)
 
-                DOUBLE_ROUND_CASE(kChar, char, roundType, saturate)
+                DOUBLE_ROUND_CASE(kChar, schar, roundType, saturate)
                 DOUBLE_ROUND_CASE(kUChar, uchar, roundType, saturate)
                 DOUBLE_ROUND_CASE(kUnsignedChar, uchar, roundType, saturate)
                 DOUBLE_ROUND_CASE(kShort, short, roundType, saturate)
```

shajder: @EwanC  I've applied your corrections, it looks necessary to me.

EwanC:
Regarding
```
Additional remark (2): test fails with my Intel device and kernel with half conversions:
ERROR: Explicit cast of scalar float to vector half2 FAILED; skipping other half vector tests
I will appreciate additional attention when it comes to testing cl_khr_fp16 with this test
```

I'm also seeing this error locally:
```sh
ERROR: Output value 0:0 does not validate for size 2:2!
       Input:   0x4dbe2551
       Actual:  0x7c007c00 0x7c007c00 0x7c007c00 0x7c007c00
ERROR: Explicit cast of scalar float to vector half2 FAILED; skipping other half vector tests
```

`0x4dbe2551` is `398764576.0f` which is outside the range of half, which is why the half2 is `(0x7c00, 0x7c00)`, i.e INF values. So think we might need to check for overflow during the verification.

shajder:
@EwanC I just added missing part of `convert_explicit_value` to support `kHalf` `ExplicitType`. Test passes now but I will appreciate reviewing it more carefully - I am not quite sure it is 100% correct

EwanC (approved):
Passes for me now 🙂  `HALF_ROUND_CASE` has a lot of code duplication with `FLOAT_ROUND_CASE` & `DOUBLE_ROUND_CASE` which would be nice to improve upon but i'm okay with the change without it.

lakshmih:
Reviewing

lakshmih (changes requested):
This test is failing (on below kernel) with an error that cl_khr_fp16 is not enabled. Can we add the pragma for kernels that use half?

   __kernel void test_conversion(__global float *sourceValues, __global half2 *destValues )
     {
         int  tid = get_global_id(0);
         float  src = sourceValues[tid];

         destValues[tid] = (half2)src;

     }

test_conformance/basic/test_explicit_s2v.cpp
```diff
@@ -20,131 +20,100 @@
...
 #include <sys/stat.h>
-
+#include <vector>
 
 #include "procs.h"
 #include "harness/conversions.h"
 #include "harness/typeWrappers.h"
 
+// clang-format off
+
```
lakshmih, line 30: Needs "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n" here.

lakshmih (changes requested):

test_conformance/basic/test_explicit_s2v.cpp
shajder: Corrected - good find, thanks!

bcalidas:
We are seeing failures with the latest commit. Debugging at our end. Need some more time.
Thanks,
Balaji

test_conformance/basic/test_explicit_s2v.cpp
```diff
@@ -170,11 +141,13 @@ int test_explicit_s2v_function(cl_device_id deviceID, cl_context context, cl_com
...
      do a memcpy instead of relying on the actual type of data */
-    error = clEnqueueReadBuffer( queue, streams[1], CL_TRUE, 0, destStride * count, outData, 0, NULL, NULL );
+    error =
+        clEnqueueReadBuffer(queue, streams[1], CL_TRUE, 0, destStride * count,
+                            outData.data(), 0, NULL, NULL);
     test_error( error, "Unable to read output values!" );
 
     inPtr = (unsigned char *)inputData;
-    outPtr = (unsigned char *)outData;
+    outPtr = (unsigned char *)outData.data();
```
lakshmih, line 150: For half-float conversions, we need to add a NAN check in this code. i.e. the if the input is NAN the output needs to be NAN but the fraction part isn't required to be the same.

lakshmih (approved):

lakshmih:
Added a patch that does the NAN check.

test_conformance/basic/test_explicit_s2v.cpp
```diff
@@ -186,6 +179,14 @@ int test_explicit_s2v_function(cl_device_id deviceID, cl_context context, cl_com
         {
             if( memcmp( convertedData, outPtr + destTypeSize * s, destTypeSize ) != 0 )
             {
+                if ((srcType == kHalf) && (destType == kFloat)
+                    && IsHalfNaN(*reinterpret_cast<cl_half *>(inPtr))
+                    && IsFloatNaN(*reinterpret_cast<cl_float *>(
+                        outPtr + destTypeSize * s)))
```
bashbaug, line 185: This seems OK but it's a little odd to me that this is only needed for the half -> float case.  Couldn't this special-case be needed for any floating-point -> floating-point case, and especially the float -> double case?

test_conformance/basic/test_explicit_s2v.cpp
```diff
@@ -20,131 +20,122 @@
...
-    DECLARE_EMPTY_SET
+    DECLARE_S2V_IDENT_KERNELS_SET(half),
+    DECLARE_S2V_IDENT_KERNELS_SET(double)
 };
 
-int test_explicit_s2v_function(cl_device_id deviceID, cl_context context, cl_command_queue queue, const char *programSrc,
-                               ExplicitType srcType, unsigned int count, ExplicitType destType, unsigned int vecSize, void *inputData )
+// clang-format on
+
+int IsFloatNaN(double x)
```
bashbaug, line 101: This seems a little odd and error prone since the name of the function is `IsFloatNaN`, but it's passed a `double`, then it immediately casts and assigns it to a `cl_float`.

Could we just use `std::isnan` instead - [link](https://en.cppreference.com/w/cpp/numeric/math/isnan)?

bashbaug (commented):
Seems OK but I have a few questions, mostly around NaN handling.

The additional test coverage does increase the runtime of this sub-test, but it still runs in ~30 seconds on my system so probably not worth worrying about (yet).

bashbaug:
Discussed in the October 3rd teleconference.  Will wait a week to address the comments above.

test_conformance/basic/test_explicit_s2v.cpp
```diff
@@ -13,6 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+#include <cmath>
+#define isnan std::isnan
```
bashbaug, line 17: I'd much rather just spell out `std::isnan` in the places where it is used, but if we're going to keep something like this, I'd suggest using a more C++-like solution, something like:

```c++
using std::isnan;
```

bashbaug (dismissed):

test_conformance/basic/test_explicit_s2v.cpp
```diff
@@ -13,138 +13,122 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
+#include <cmath>
+using std::isnan;
+#define isnan isnan
```
bashbaug, line 18: We can remove the define completely now.

```suggestion
using std::isnan;
```

bashbaug (approved):
LGTM with one small nice-to-have - thanks!

bashbaug (approved):

PR #1717, opened 2023-05-08
Enable clCopyImage suite to run in parallel

kpet:


svenvh (approved):
I've tried digging into the git history/ies to see why it was written like this originally, but that didn't give any insights.

PR #1718, opened 2023-05-08
Added cl_khr_fp16 extension support for test_fpmath from basic

shajder:

Additional remark: previous test handle only `float`/`float2`/`float4`/`float8` types. This PR extends scope of test to `float`/`half`/`double` of sizes 1/2/4/8/16


EwanC (dismissed):

test_conformance/basic/test_fpmath.cpp
```diff
@@ -0,0 +1,381 @@
...
+    if (std::is_same<cl_half, T>::value)
+        return HTF(val);
+    else
+        return val;
+}
+
+bool isHalfNan(cl_half v)
+{
+    // Extract FP16 exponent and mantissa
+    uint16_t h_exp = (((cl_half)v) >> (CL_HALF_MANT_DIG - 1)) & 0x1F;
```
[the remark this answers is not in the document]
shajder: Done

EwanC (approved):

test_conformance/basic/test_fpmath.cpp
```diff
@@ -0,0 +1,381 @@
...
+                                     MAKE_HEX_DOUBLE(0x1.0p63, 0x1LL, 63),
+                                     seed);
+        };
+        for (auto &v : input)
+            std::generate(v.begin(), v.end(), random_generator);
+    }
+    else
+    {
+        auto random_generator = [&seed]() {
+            return get_random_float(HFF(-MAKE_HEX_FLOAT(0x1.0p15f, 0x1, 15)),
```
svenvh, line 142: I don't think this produces a good set of test values.  From a quick inspection of the generated input values for half, the vast majority of them seem to have the sign bit set for example.  Perhaps we need a dedicated `get_random_half` helper function?

svenvh (changes requested):

test_conformance/basic/test_fpmath.cpp
```diff
@@ -0,0 +1,381 @@
...
+static char extension[128] = { 0 };
+
+// clang-format off
+static const char* fp_kernel_code[] = {
+extension,
+"__kernel void test_fp(__global TYPE *srcA, __global TYPE *srcB, __global TYPE *dst)\n"
+"{\n"
+"    int tid = get_global_id( 0 );\n",
+"    dst[tid] = srcA[tid] OP srcB[tid];\n",
+"}\n"};
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/basic/test_fpmath.cpp
```diff
@@ -0,0 +1,381 @@
...
+        std::ostringstream sstr;
+        if (std::is_same<T, double>::value)
+            sstr << "#pragma OPENCL EXTENSION cl_khr_fp64 : enable\n";
+
+        if (std::is_same<T, cl_half>::value)
+            sstr << "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n";
+
+        snprintf(extension, sizeof(extension), "%s", sstr.str().c_str());
+
+        for (unsigned i = 0; i < sizeof(vecSizes) / sizeof(size_t); i++)
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/basic/test_fpmath.cpp
```diff
@@ -0,0 +1,381 @@
...
+            vecNameStr << test.type_str;
+            if (test.vec_size != 1) vecNameStr << test.vec_size;
+
+            clMemWrapper streams[3];
+            clProgramWrapper program;
+            clKernelWrapper kernel;
+
+            size_t length = sizeof(T) * num_elements * test.vec_size;
+
+            int isRTZ = 0;
```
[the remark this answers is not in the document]
shajder: Done

test_conformance/basic/test_fpmath.cpp
shajder: Good find, corrected. 

> Perhaps we need a dedicated get_random_half helper function?

probably, this is only an impression but it seems like most of test cases uses customized input values

svenvh:
> @svenvh any idea what could be wrong with macOS build (missing vulkan symbols) ?

Surely it's not caused by your commit, as it suddenly fails to build the Vulkan-Loader which happens before building the CTS.  I suspect this might be related to https://github.com/KhronosGroup/Vulkan-Loader/pull/1212 ; perhaps @charles-lunarg can provide any insights?

test_conformance/basic/test_fpmath.cpp
svenvh: It's a bit better now, but I still have some concerns on whether the generated test data for fp16 is meaningful enough:

1. Most randomly generated values are > 1024, so most of the test values don't have any interesting fractional part.
2. For multiplications, the randomly generated input operands are that large that the resulting product is almost always infinity.  Perhaps it makes sense to fine-tune the fp16 generation to generate smaller numbers more often?
3. Behavior near 0 isn't covered, but the float/double cases also seem to have this problem, so we could ignore that for now.

svenvh: Actually point 1 applies to float/double too indeed, so like point 3 we can leave that as a future improvement.  But point 2 seems specific to fp16: because of the very limited range of fp16, it's much easier to hit +/- infinity.  I just checked, for float the multiplication test doesn't produce infinities.

charles-lunarg:
Yes, I recently updated the macOS build to support assembly code. Unfortunately, it seems I did not test the 'turn off assembly code support' branch. I will create an issue to track that, but in the meantime this repo should be able to make use of the assembly code.

The change is simple and I'm willing to do it - unless someone else wants to take charge.
Change `USE_GAS=OFF` to `USE_GAS=ON`.
https://github.com/KhronosGroup/OpenCL-CTS/blob/63a8cb6b9d3345feec7621c7e0b0a4ca21cf545a/presubmit.sh#L80

Apologies for breaking your build.

svenvh:
> The change is simple and I'm willing to do it - unless someone else wants to take charge. Change `USE_GAS=OFF` to `USE_GAS=ON`. (or just remove the line entirely, since ON is the default)

Thanks for the guidance @charles-lunarg! I've opened #1756 and that seems to fix the CI build.

test_conformance/basic/test_fpmath.cpp
shajder: > I just checked, for float the multiplication test doesn't produce infinities.

of course, that's probably because float range (`+/- 0x1.0p31f`) leaves enough space for multiplication meanwhile half range is at the edge of precision. I think we should decrease it to `+/- 0x1.0p8f`

svenvh (dismissed):

svenvh:
Closing/reopening to retrigger CI.

EwanC (approved):

bashbaug:
Merging as discussed in the June 20th teleconference.

PR #1725, opened 2023-05-12
math_brute_force: remove gotos in macro_unary_float

svenvh:
Simplify code by returning directly instead of using goto statements.

Although intended as an NFC commit, this changes the behaviour when verification fails: the output buffer is no longer unmapped.  Although not ideal, this aligns the behaviour of `macro_unary_float` to the other math_brute_force tests.


bashbaug (approved):
Agree this is not ideal, but the verification failures shouldn't occur during typical execution of the tests, and it's nice to have the same behavior across all bruteforce tests.

PR #1728, opened 2023-05-15
Added cl_khr_fp16 extension support for test_vector_creation from basic

shajder:

Additional remark: previous test seems to skip cases of 16-size vectors creation. I am not quite sure if that was intentional, with this specific correction the test takes significantly longer but it passes with nvidia and intel implementation on my machine.


test_conformance/basic/test_vector_creation.cpp
```diff
@@ -171,236 +182,312 @@ int create_kernel(ExplicitType type, int output_size, char *program, int *number
...
-int test_vector_creation(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
+int test_vector_creation(cl_device_id deviceID, cl_context context,
+                         cl_command_queue queue, int num_elements)
 {
-    ExplicitType vecType[] = { kChar, kUChar, kShort, kUShort, kInt, kUInt, kLong, kULong, kFloat, kDouble };
-    unsigned int vecSizes[] = { 1, 2, 3, 4, 8, 16};
+    std::vector<ExplicitType> vecType = { kChar,  kUChar, kShort, kUShort,
+                                          kInt,   kUInt,  kLong,  kULong,
+                                          kFloat, kHalf,  kDouble };
+    std::vector<unsigned int> vecSizes = { 2, 3, 4, 8, 16 };
```
svenvh, line 246: These can be `const`.

shajder: Done

test_conformance/basic/test_vector_creation.cpp
```diff
@@ -171,236 +182,312 @@ int create_kernel(ExplicitType type, int output_size, char *program, int *number
...
+        {
+            size_t global[] = { 1, 1, 1 };
             int number_generated = -1;
             int previous_number_generated = 0;
 
-            log_info("Testing %s%s...\n", get_explicit_type_name(vecType[type_index]), size_names[size_index]);
-            while (number_generated != 0) {
+            log_info("Testing %s%s...\n",
+                     get_explicit_type_name(vecType[type_index]),
+                     size_names[size_index]);
```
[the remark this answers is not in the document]
shajder: Correct, `vecSizes` should be in sync with `size_names` since it uses the same index. Good find!

shajder:
> Also, this seems to trigger `ERROR: Invalid type given to convert_explicit_value!!` messages.

Done

svenvh (approved):

EwanC:
I'm seeing a pretty dramatic increase in runtime from the 16-wide vector additions. Admittedly, i'm testing it on an emulator, but it could be representative of embedded platforms

Before this PR running all the tests takes ~73 seconds
With this PR running char just tests takes more than 20 minutes (that's when I killed it). From the log:
```quote
vector_creation...
Testing char.
Testing char2...
Testing char3...
Testing char4...
Testing char8...
Testing char16...
Code size greater than 32kB; splitting test into multiple kernels.
        Executing vector permutations 0 to 238...
        Executing vector permutations 239 to 479...
        Executing vector permutations 480 to 721...
        Executing vector permutations 722 to 963...
        Executing vector permutations 964 to 1203...
        Executing vector permutations 1204 to 1455...
        Executing vector permutations 1456 to 1698...
        Executing vector permutations 1699 to 1938...
        Executing vector permutations 1939 to 2173...
        Executing vector permutations 2174 to 2423...
        Executing vector permutations 2424 to 2672...
        Executing vector permutations 2673 to 2939...
        Executing vector permutations 2940 to 3175...
        Executing vector permutations 3176 to 3412...
        Executing vector permutations 3413 to 3650...
        Executing vector permutations 3651 to 3889...
        Executing vector permutations 3890 to 4126...
        Executing vector permutations 4127 to 4373...
        Executing vector permutations 4374 to 4625...
        Executing vector permutations 4626 to 4877...
        Executing vector permutations 4878 to 5125...
        Executing vector permutations 5126 to 5388...
        Executing vector permutations 5389 to 5659...
        Executing vector permutations 5660 to 5898...
        Executing vector permutations 5899 to 6136...
        Executing vector permutations 6137 to 6373...
        Executing vector permutations 6374 to 6612...
        Executing vector permutations 6613 to 6848...
        Executing vector permutations 6849 to 7095...
        Executing vector permutations 7096 to 7337...
        Executing vector permutations 7338 to 7572...
        Executing vector permutations 7573 to 7804...
        Executing vector permutations 7805 to 8048...
        Executing vector permutations 8049 to 8292...
        Executing vector permutations 8293 to 8549...
        Executing vector permutations 8550 to 8804...
        Executing vector permutations 8805 to 9054...
        Executing vector permutations 9055 to 9303...
        Executing vector permutations 9304 to 9556...
        Executing vector permutations 9557 to 9805...
        Executing vector permutations 9806 to 10067...
        Executing vector permutations 10068 to 10334...
        Executing vector permutations 10335 to 10597...
        Executing vector permutations 10598 to 10864...
        Executing vector permutations 10865 to 11124...
        Executing vector permutations 11125 to 11363...
        Executing vector permutations 11364 to 11598...
        Executing vector permutations 11599 to 11840...
        Executing vector permutations 11841 to 12073...
        Executing vector permutations 12074 to 12316...
        Executing vector permutations 12317 to 12565...
        Executing vector permutations 12566 to 12797...
        Executing vector permutations 12798 to 13031...
        Executing vector permutations 13032 to 13274...
        Executing vector permutations 13275 to 13518...
        Executing vector permutations 13519 to 13771...
        Executing vector permutations 13772 to 14020...
        Executing vector permutations 14021 to 14253...
        Executing vector permutations 14254 to 14485...
        Executing vector permutations 14486 to 14726...
        Executing vector permutations 14727 to 14955...
        Executing vector permutations 14956 to 15193...
        Executing vector permutations 15194 to 15443...
        Executing vector permutations 15444 to 15687...
        Executing vector permutations 15688 to 15934...
        Executing vector permutations 15935 to 16185...
        Executing vector permutations 16186 to 16444...
        Executing vector permutations 16445 to 16708...
        Executing vector permutations 16709 to 16956...
        Executing vector permutations 16957 to 17205...
        Executing vector permutations 17206 to 17463...
        Executing vector permutations 17464 to 17707...
        Executing vector permutations 17708 to 17963...
        Executing vector permutations 17964 to 18226...
        Executing vector permutations 18227 to 18470...
        Executing vector permutations 18471 to 18715...
        Executing vector permutations 18716 to 18972...
        Executing vector permutations 18973 to 19229...
        Executing vector permutations 19230 to 19499...
        Executing vector permutations 19500 to 19769...
        Executing vector permutations 19770 to 20034...
        Executing vector permutations 20035 to 20299...
        Executing vector permutations 20300 to 20564...
        Executing vector permutations 20565 to 20828...
<I killed it here>
```

shajder:
> I'm seeing a pretty dramatic increase in runtime from the 16-wide vector additions. Admittedly, i'm testing it on an emulator, but it could be representative of embedded platforms
> 
> Before this PR running all the tests takes ~73 seconds With this PR running char just tests takes more than 20 minutes (that's when I killed it). From the log:
> 
> ```
> vector_creation...
> Testing char.
> Testing char2...
> Testing char3...
> Testing char4...
> ```

@EwanC I just went through execution times. Indeed for `intel` device I have ~8 minutes result. On the other hand for `nvidia` the test finishes in 7 seconds. Please correct me if I am wrong, with approximately the same number of kernel enqueues (`nvidia` has no `cl_khr_fp16` but it equalizes number of tests with `cl_khr_fp64`) my conclusion is the bottleneck must be located at the kernels driver end. Thus, I am not sure is there much to optimize at the test level, maybe we could randomly limit number of 16-cases variations ?

EwanC:
>@EwanC I just went through execution times. Indeed for intel device I have ~8 minutes result. On the other hand for nvidia the test finishes in 7 seconds. Please correct me if I am wrong, with approximately the same number of kernel enqueues (nvidia has no cl_khr_fp16 but it equalizes number of tests with cl_khr_fp64) my conclusion is the bottleneck must be located at the kernels driver end. Thus, I am not sure is there much to optimize at the test level, maybe we could randomly limit number of 16-cases variations ?

Looking more closely at what happens each loop iteration, we are [compiling a new kernel](https://github.com/KhronosGroup/OpenCL-CTS/blob/19f4fc3f3d69bea530182164dd061c7dfda166b0/test_conformance/basic/test_vector_creation.cpp#L297). That is probably what's hitting me, as the host code is emulated as well the device. So it might actually be bottlenecked on host CPU rather than kernel device driver.

test_conformance/basic/test_vector_creation.cpp
```diff
@@ -67,102 +60,123 @@ int create_kernel(ExplicitType type, int output_size, char *program, int *number
...
+            "%s%s *result) {\n",
+            extension, get_explicit_type_name(type),
+            get_explicit_type_name(type),
+            (number_of_sizes == 3) ? "" : size_names[number_of_sizes - 1]);
     total_program_length += (int)strlen(program);
 
-    char storePrefix[ 128 ], storeSuffix[ 128 ];
+    char storePrefix[128], storeSuffix[128];
 
     // Start out trying sizes 1,1,1,1,1...
```
EwanC, line 84: Can we update this comment to describe what we're doing with `inital_no_sizes` to omit the already tested vload vector widths

EwanC (dismissed):
The most recent change fixes the performance issues I was seeing, thanks.

test_conformance/basic/test_vector_creation.cpp
shajder: Done

EwanC (approved):

svenvh:
Aren't we losing coverage with the latest changes?

For example, for `char4` we used to test:

```
__kernel void test_vector_creation(__global char *src, __global char4 *result) {
  result[0] = (char4)(src[0],src[1],src[2],src[3]);
  result[1] = (char4)(src[0],src[1],vload2(0,src+2));
  result[2] = (char4)(src[0],vload2(0,src+1),src[3]);
  result[3] = (char4)(src[0],vload3(0,src+1));
  result[4] = (char4)(vload2(0,src+0),src[2],src[3]);
  result[5] = (char4)(vload2(0,src+0),vload2(0,src+2));
  result[6] = (char4)(vload3(0,src+0),src[3]);
  result[7] = (char4)(vload4(0,src+0));
}
```

but now we only seem to test

```
__kernel void test_vector_creation(__global char *src, __global char4 *result) {
  result[0] = (char4)(vload2(0,src+0),vload2(0,src+2));
  result[1] = (char4)(vload3(0,src+0),src[3]);
  result[2] = (char4)(vload4(0,src+0));
}
```

shajder:
> Aren't we losing coverage with the latest changes?

That is true. Due to performance issues reported above limiter was introduced. As far as I understand full coverage is not acceptable. Do you have some other balance on your mind ?

svenvh:
> Due to performance issues reported above limiter was introduced. As far as I understand full coverage is not acceptable. Do you have some other balance on your mind ?

Perhaps we should split out the vector size-16 additions from this PR?  Then we can discuss in the new PR how to handle the performance issues.  For example, I don't think we should reduce coverage for the non-problematic cases (i.e., vector sizes < 16); they seem to be fast enough?  For vector sizes of 16 only we can skip some cases.  Either way, I think we should do so in a separate PR to ensure visibility of the additional vector size-16 coverage.

shajder:
> For vector sizes of 16 only we can skip some cases.

@EwanC what would you say about that ? I could leave rest of the test as previously and limit only 16-sized vector case. I just measured timings and this should be still acceptable I guess. Could you verify performance of such case:

`static const int initial_no_sizes[] = { 0, 0, 0, 0, 0, 0, 2 };`

Thanks!

lakshmih (approved):

EwanC:
> > For vector sizes of 16 only we can skip some cases.
> 
> @EwanC what would you say about that ? I could leave rest of the test as previously and limit only 16-sized vector case. I just measured timings and this should be still acceptable I guess. Could you verify performance of such case:
> 
> `static const int initial_no_sizes[] = { 0, 0, 0, 0, 0, 0, 2 };`
> 
> Thanks!

I tested this out, and it's still an order of magnitude increase in the testing time, up to 30mins for the all the tests from 2 minutes. However, doing testing on this emulated platform is slow for us regardless, so this isn't a breaking issue now, and I think is a good compromise for this PR. If it does become an issue for us, we can revisit it ourselves with our own patch proposal.

shajder:
> > For vector sizes of 16 only we can skip some cases.
> 
> @EwanC what would you say about that ? I could leave rest of the test as previously and limit only 16-sized vector case. I just measured timings and this should be still acceptable I guess. Could you verify performance of such case:
> 
> `static const int initial_no_sizes[] = { 0, 0, 0, 0, 0, 0, 2 };`

@svenvh original capacity is restored, only 16-sizes vectors were limited

svenvh (approved):
> @svenvh original capacity is restored, only 16-sizes vectors were limited

Thanks, LGTM.

test_conformance/basic/test_vector_creation.cpp
```diff
@@ -171,236 +187,319 @@ int create_kernel(ExplicitType type, int output_size, char *program, int *number
...
+int test_vector_creation(cl_device_id deviceID, cl_context context,
+                         cl_command_queue queue, int num_elements)
 {
-    ExplicitType vecType[] = { kChar, kUChar, kShort, kUShort, kInt, kUInt, kLong, kULong, kFloat, kDouble };
-    unsigned int vecSizes[] = { 1, 2, 3, 4, 8, 16};
+    const std::vector<ExplicitType> vecType = { kChar,  kUChar, kShort, kUShort,
+                                                kInt,   kUInt,  kLong,  kULong,
+                                                kFloat, kHalf,  kDouble };
+    // should be in sync with global array size_names
+    const std::vector<unsigned int> vecSizes = { 1, 2, 3, 4, 8, 16 };
```
bashbaug, line 252: This is so minor it probably isn't worth changing, but I was initially confused because this test now outputs e.g.

```sh
Testing half.
Testing half2...
Testing half3...
Testing half4...
Testing half8...
```

Then, since this array of `vecSizes` includes `1`, so it's easy to incorrectly assume that scalars are also tested.  The loop that iterates over this array properly starts at index `1` though, so the `vecSize == 1` case is properly skipped, and `Testing half.` is only generic output that "half" vectors are being tested.

I was originally thinking that the value `1` could simply be removed from this array, but the `size_index` is also used to index into some other arrays, so additional changes will be required.  Probably not worth it.

The one change I _might_ consider is to put a bogus value here instead, so it is more clear that it is unused, but even that might be more trouble than it's worth.

bashbaug (approved):
The testing time does increase with this change, but not dramatically so.

PR #1729, opened 2023-05-16
Added cl_khr_fp16 extension support for test_vector_swizzle from basic

shajder:

### DIFF
--- a/test_conformance/basic/test_vector_swizzle.cpp
+++ b/test_conformance/basic/test_vector_swizzle.cpp
@@ -22,6 +22,8 @@
 #include "procs.h"
 #include "harness/testHarness.h"
 
+std::string pragma_extension;
+
 template <int N> struct TestInfo
 {
 };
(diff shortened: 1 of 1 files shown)


EwanC (approved):

test_conformance/basic/test_vector_swizzle.cpp
```diff
@@ -22,6 +22,8 @@
 #include "procs.h"
 #include "harness/testHarness.h"
 
+std::string pragma_extension;
```
svenvh, line 25: Wouldn't it be better to make this a local variable inside `test_vectype`?  You should then be able to use an `if constexpr` to set the extension pragma based on the template argument type.  That way we don't need any unnecessary global state.

shajder: To be honest it crossed my mind but I deduced it is not worth the effort. The thing is that `constexpr` condition will confuse `cl_ushort` and `cl_half`, I would have to pass an additional template argument to inform `test_vectype` we are dealing with FP type which eventually makes the code less readable.

EwanC (approved):

lakshmih:
Reviewing, thanks.

PR #1732, opened 2023-05-17
atomics: fix memory leaks on error paths

svenvh:
Before this change, `add_index_bin_test` would not always release `cl_mem` resources or `malloc`ed memory when encountering an error.  Fix by using `clMemWrapper` and `std::unique_ptr` to automatically release resources.


EwanC (approved):
LGTM, always good to see an update to use smart pointers and wrappers 😃

bashbaug:

Issue #1424, opened 2022-05-04
missing test coverage for cl_khr_expect_assume


bashbaug:
see: https://github.com/KhronosGroup/OpenCL-Docs/issues/723

We need to add CTS coverage for the `cl_khr_expect_assume` extension.

bashbaug:
I wrote up a rough test plan for this extension in case it is helpful.  This is written from a SPIR-V extension point of view, so applicable to both OpenCL and Vulkan, but some parts are not applicable for OpenCL (e.g. there is only one "shader pipeline stage" for OpenCL):

## SPV_KHR_expect_assume Test Plan

The test plan includes new features added in the SPV_KHR_expect_assume extension.

Verify:
* OpExtension "SPV_KHR_expect_assume" is present in the SPIR-V Module
* OpCapability "ExpectAssumeKHR" is present in the SPIR-V Module.

Test all shader pipeline stages.

* Test OpAssumeTrueKHR instruction:
  * Create SPIR-V Module with OpAssumeTrueKHR on a condition known to be true.
  * Verify that the SPIR-V Module executes correctly.

* Test OpExpectKHR instruction for integer types and Value == ExpectedValue:
  * For all supported integer types (8-bit, 16-bit, 32-bit, 64-bit):
    * For all supported vector sizes:
      * Create a SPIR-V Module with OpExpectKHR and a Value matching ExpectedValue.
      * Verify that the SPIR-V Module executes correctly.

* Test OpExpectKHR instruction for Boolean types and Value == ExpectedValue:
  * For all supported vector sizes:
    * Create a SPIR-V Module with OpExpectKHR and a Value matching ExpectedValue.
    * Verify that the SPIR-V Module executes correctly.

* Test OpExpectKHR instruction for integer types and Value != ExpectedValue:
  * For all supported integer types (8-bit, 16-bit, 32-bit, 64-bit):
    * For all supported vector sizes:
      * Create a SPIR-V Module with OpExpectKHR and a Value different from ExpectedValue.
      * Verify that the SPIR-V Module executes correctly.

* Test OpExpectKHR instruction for Boolean types and Value != ExpectedValue:
  * For all supported vector sizes:
    * Create a SPIR-V Module with OpExpectKHR and a Value different from ExpectedValue.
    * Verify that the SPIR-V Module executes correctly.

bashbaug:
Linking https://github.com/KhronosGroup/OpenCL-CTS/pull/1888.

[...]

PR #1735, opened 2023-05-19
add  cl_khr_expect_assume to test_compiler_defines_for_extensions

norablackcat:
https://github.com/KhronosGroup/OpenCL-CTS/issues/1424#issuecomment-1554680436

### DIFF
--- a/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
+++ b/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
@@ -76,6 +76,7 @@ const char *known_extensions[] = {
     "cl_khr_device_uuid",
     "cl_khr_pci_bus_info",
     "cl_khr_suggested_local_work_size",
+    "cl_khr_expect_assume",
     "cl_khr_spirv_linkonce_odr",
     "cl_khr_semaphore",
     "cl_khr_external_semaphore",


test_conformance/compiler/test_compiler_defines_for_extensions.cpp
```diff
@@ -63,6 +63,7 @@ const char *known_extensions[] = {
     "cl_khr_dx9_media_sharing",
     "cl_khr_egl_event",
     "cl_khr_egl_image",
+    "cl_khr_expect_assume",
```
svenvh, line 66: As this isn't an API-only extension, I believe this should go right before `cl_khr_subgroup_rotate`. See also the comments on lines 23 and 56.

Please also update `first_API_extension` accordingly.

svenvh (changes requested):

svenvh (dismissed):
LGTM, thanks!

lakshmih (approved):

bashbaug (approved):
Thank you!

bashbaug (commented):
Actually, I'm sorry, I did have one question/comment: because this extension does not extend OpenCL C, is it really correct to consider it to be a non-API-only extension?

As-written, this test will fail if an implementation reports support for `cl_khr_expect_assume` but the proper support is NOT "defined in the OpenLC C kernel environment", which does not quite seem correct.

norablackcat:
I actually had it as an api only extension at first because the extension spec doesn't define any feature macros for it. It does seem a bit silly to me to define cl_khr_expect_assume macro that doesn't do anything because spirv does the work. at least in the rusticl implementation.

karolherbst:
There are a handful of SPIR-V only extensions now and I suspect we might want to figure how to treat those.

"API only" here basically means, that the online compiler of the OpenCL runtime doesn't expose any feature macro and I think that term is therefore misleading.

So if an extension doesn't define anything on the OpenCL C level, we should just treat them as "API only". Some extensions might define a feature macro in additions to new OpenCL C features and we'll have to check those explicitly anyway.

karolherbst:
Side note: I don't think we should allow SPIR-V only extensions under `khr`, because they just fragment how compilers actually implement those extensions in their OpenCL C frontend and we might end up with implementations doing it differently. We should rather just define this alongside the SPIR-V side so it's all consistent across various OpenCL C to SPIR-V compilers, which in almost every case will be clang anyway)

bashbaug:
> If I had a say in it I would say it's api only.

Yeah, I think I agree with this too, though I'm starting to put "api only" in quotes.

I think what really happened here is that we used to only have two categories of extensions: "API-only" extensions, which only affected OpenCL runtime APIs, and "non-API-only" extensions, which affected the (OpenCL C) compiler.  The first category did not need any macros, because they only affected runtime APIs, but the second did.  This is how the two categories of extensions are used in this test.

We've since added a third category of extensions that affect the compiler, but does not affect the OpenCL C compilers, and hence does not need or include a macro.  These aren't really "API-only" extensions, but this categorization works for the purposes of this test.

> Side note: I don't think we should allow SPIR-V only extensions under khr, because they just fragment how compilers actually implement those extensions in their OpenCL C frontend and we might end up with implementations doing it differently. We should rather just define this alongside the SPIR-V side so it's all consistent across various OpenCL C to SPIR-V compilers, which in almost every case will be clang anyway

Can you say a little more about how this might work?  I'd love to have a way to streamline SPIR-V-only extensions but I haven't thought of a clever way to do this just yet.

Note that some of these SPIR-V-only extensions may not have any representation in OpenCL C (`cl_khr_spirv_linkonce_odr` is a good example), but they're still useful so kernels written in other source languages may executed via an OpenCL runtime.

We should probably move this conversation to a separate issue so it doesn't disappear when this PR gets merged.

karolherbst:
@bashbaug https://github.com/KhronosGroup/OpenCL-Docs/issues/941

karolherbst:
btw, we already treat `cl_khr_spirv_no_integer_wrap_decoration` as API only, which is also spirv only in its current form.

bashbaug (changes requested):
Just so this doesn't get merged by accident: we should be considering this an "API-only" extension because it does not formally affect OpenCL C, even though it is really primarily a SPIR-V extension.  See discussion above.

svenvh (approved):
@bashbaug the latest version treats this as an API-only extension again.  I suppose we can merge it now or do you have any further concerns?

bashbaug (approved):

bashbaug:
All good from my end, thanks for the nudge!

bashbaug:
This has two approvals now, plus @lakshmih's prior approval.  Merging.

PR #1736, opened 2023-05-21
Bug fix in test_async_copy_fence - wrong events usage

BZZiv:


EwanC (approved):
I don't have an implementation of `cl_khr_async_work_group_copy_fence` to test this on, but the patch makes sense to me.

bashbaug:
It might be best for somebody from Mobileye to review this PR - @chrisgearing perhaps?

This change seems reasonable to me, although the subsequent call to `wait_group_events` is waiting on two events, not just one...

```c
wait_group_events( 2, &events );
```

... which makes me wonder if I'm missing something about how this test is supposed to work.

BZZiv:
> It might be best for somebody from Mobileye to review this PR - @chrisgearing perhaps?
> 
> This change seems reasonable to me, although the subsequent call to `wait_group_events` is waiting on two events, not just one...
> 
> ```c
> wait_group_events( 2, &events );
> ```
> 
> ... which makes me wonder if I'm missing something about how this test is supposed to work.

You're right! Updated number of events to wait on.

lakshmih (approved):

bashbaug (approved):

bashbaug:
Merging as discussed in the March 12th teleconference.

PR #1740, opened 2023-05-24
Add check for CL_COMMAND_BUFFER_STATE_EXECUTABLE_KHR after completion.

pj87:
Added check for state CL_COMMAND_BUFFER_STATE_EXECUTABLE_KHR in state info tests.


test_conformance/extensions/cl_khr_command_buffer/command_buffer_get_command_buffer_info.cpp
```diff
@@ -255,6 +255,10 @@ struct CommandBufferGetCommandBufferInfo : public BasicCommandBufferTest
 
         test_error(signal_error, "clSetUserEventStatus failed");
 
+        // verify executable state
+        error = verify_state(CL_COMMAND_BUFFER_STATE_EXECUTABLE_KHR);
```
EwanC, line 259: It's not guaranteed that a command-buffer will have finished executing by this point, think you need a `clFinish(queue)`  first or return an event from the enqueue on line 244 and wait on it here.

EwanC (approved):

bashbaug (approved):
LGTM!

PR #1742, opened 2023-05-26
Added cl_khr_fp16 extension support for test_int2fp from basic

shajder:


test_conformance/basic/test_int2fp.cpp
```diff
@@ -0,0 +1,327 @@
...
+cl_half get_random_half(MTdata d)
+{
+    const unsigned m_size = 0x3fe;
+    const unsigned e_size = 0x1d;
+    const unsigned s_size = 0x1;
+    unsigned ind = genrand_int32(d) % (s_size * e_size * m_size);
+    return static_cast<cl_half>((((ind / (e_size * m_size)) << 15)
+                                 | (((ind / m_size) % e_size + 1) << 10)
+                                 | (ind % m_size + 1)));
+}
```
EwanC, line 54: I'm not quite following this code, first of all why it's needed it'd expect zero mantissa numbers to be some of the easier values to convert as there's no fractional component.

Secondly, once it's documented why we need this function itself needs more comments. There's a lot of expressions in here, and i'd prefer them to be in their own statements with comments for this code to be maintainable.

test_conformance/basic/test_int2fp.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+    template <typename Tx, typename Ty> static bool equal_value(Tx a, Ty b)
+    {
+        return a == (Tx)b;
+    }
+
+    static bool equal_value_from_half(cl_short a, cl_half b)
+    {
+        bool ret = a == (cl_short)HTF(b);
+        if (!ret) vlog("ds");
```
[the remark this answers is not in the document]
shajder: ops :), done

test_conformance/basic/test_int2fp.cpp
shajder: This code has been replaced with generator based on `get_random_float` in specific range:

```
-MAKE_HEX_FLOAT(0x1.ffcp+14, 1.9990234375f, 14),
 MAKE_HEX_FLOAT(0x1.ffcp+14, 1.9990234375f, 14)
```

Previous procedure turns out to be a temporary solution for test failing on my local device at the border values.

test_conformance/basic/test_int2fp.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+        std::tuple<cl_float, cl_half, cl_double> it;
+        for_each_elem(it);
+    }
+
+    template <typename T> void generate_random_inputs(std::vector<T> &v)
+    {
+        RandomSeed seed(gRandomSeed);
+
+        if (sizeof(T) == sizeof(cl_half))
```
EwanC, line 66: Could we have a comment here about the bound. e.g
"Bound generated half values to 0x1.ffcp+14(32752.0) which is the largest cl_half value smaller than the max value of cl_short, 32767."

EwanC (dismissed):
LGTM

test_conformance/basic/test_int2fp.cpp
```diff
@@ -0,0 +1,316 @@
...
+        {
+            if (int2fp)
+            {
+                auto random_generator = [&seed]() {
+                    return (short)get_random_float(
```
EwanC, line 71: preference to use `cl_short` for cast rather than `short`

test_conformance/basic/test_int2fp.cpp
shajder: Done

test_conformance/basic/test_int2fp.cpp
shajder: Done

EwanC (approved):

svenvh (approved):

bashbaug:
Merging as discussed in the June 20th teleconference.

PR #1743, opened 2023-05-26
Add global offset tests for cl_khr_command_buffer_mutable_dispatch.

pj87:


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
```diff
@@ -0,0 +1,136 @@
...
+        test_error(error, "clUpdateMutableCommandsKHR failed");
+
+        error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
+                                          nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        error = clGetMutableCommandInfoKHR(
+            command, CL_MUTABLE_DISPATCH_GLOBAL_WORK_OFFSET_KHR,
+            sizeof(test_global_work_offset), &test_global_work_offset, &size);
+        test_error(error, "clGetMutableCommandInfoKHR failed");
```
EwanC, line 111: We should check that's some observable output from the kernel as a result of the change to global work offset, not just that `clGetMutableCommandInfoKHR` has been updated. This means calling `get_global_offset()` inside of the kernel, writing that to a buffer, and reading the buffer after the command-buffer enqueue has finished.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
```diff
@@ -0,0 +1,185 @@
...
+            else if (i > global_offset && global_offset != resultData[i])
+            {
+                log_error("Data failed to verify: global_offset != "
+                          "resultData[%d]=%d\n",
+                          i, resultData[i]);
+                return TEST_FAIL;
+            }
+        return CL_SUCCESS;
+
+        return CL_SUCCESS;
```
EwanC, line 168: Duplicated returns here

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    bool Skip() override
+    {
+        cl_mutable_dispatch_fields_khr mutable_capabilities;
+
+        bool mutable_support =
+            !clGetDeviceInfo(
+                device, CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR,
+                sizeof(mutable_capabilities), &mutable_capabilities, nullptr)
+            && mutable_capabilities != CL_MUTABLE_DISPATCH_GLOBAL_OFFSET_KHR;
```
EwanC, line 62: This test isn't skipping correctly for me because we're using `!=` rather than `&` when checking for the bit. This fixes it, same change might affect other PRs.
```diff
obal_offset.cpp b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
index 992e721..b70eca0 100644
--- a/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
@@ -59,7 +59,7 @@ struct MutableDispatchGlobalOffset : InfoMutableCommandBufferTest
             !clGetDeviceInfo(
                 device, CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR,
                 sizeof(mutable_capabilities), &mutable_capabilities, nullptr)
-            && mutable_capabilities != CL_MUTABLE_DISPATCH_GLOBAL_OFFSET_KHR;
+            && (mutable_capabilities & CL_MUTABLE_DISPATCH_GLOBAL_OFFSET_KHR);

         return !mutable_support || InfoMutableCommandBufferTest::Skip();
     }
```

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
```diff
@@ -0,0 +1,187 @@
...
+        test_error(error, "clGetMutableCommandInfoKHR failed");
+
+        if (info_global_offset != update_global_offset)
+        {
+            log_error("ERROR: Wrong size returned from "
+                      "clGetMutableCommandInfoKHR.");
+            return TEST_FAIL;
+        }
+
+        size_t num_elements = sizeToAllocate / sizeof(cl_int);
```
EwanC, line 145: I'd move this next to where `sizeToAllocate` is defined as a constant class member variable.

EwanC (commented):
I have a couple more minor comments, but I ran this with the command-buffer emulation layer and it passes. So once these comments are resolved I'm happy with it.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+#include <CL/cl.h>
+#include <CL/cl_ext.h>
+
+////////////////////////////////////////////////////////////////////////////////
+// mutable dispatch tests which handle following cases:
+//
+// CL_MUTABLE_DISPATCH_GLOBAL_WORK_OFFSET_KHR
+
+struct MutableDispatchGlobalOffset : InfoMutableCommandBufferTest
```
bashbaug, line 38: It's a little odd to me that this test is deriving from `InfoMutableCommandBufferTest`.  Why `Info`?

Do we even need `InfoMutableCommandBuffertest`, or would it make more sense to remove it and just get the extension function pointer for `clGetMutableCommandInfoKHR` as part of `BasicMutableCommandBufferTest`?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
```diff
@@ -0,0 +1,185 @@
...
+                                cl_command_queue queue)
+        : InfoMutableCommandBufferTest(device, context, queue)
+    {}
+
+    virtual cl_int SetUp(int elements) override
+    {
+        InfoMutableCommandBufferTest::SetUp(elements);
+
+        return CL_SUCCESS;
+    }
```
bashbaug, line 52: Since this is only calling into the base class, can we just remove it?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_basic.h
```diff
@@ -84,24 +95,52 @@ struct BasicMutableCommandBufferTest : BasicCommandBufferTest
...
-        return TEST_FAIL;                                                      \
+        GET_EXTENSION_ADDRESS(clUpdateMutableCommandsKHR);
+
+        return CL_SUCCESS;
     }
+
+    clUpdateMutableCommandsKHR_fn clUpdateMutableCommandsKHR = nullptr;
+
+    const char* kernelString = "__kernel void empty() {}";
+    const size_t global_work_size = 4 * sizeof(cl_int);
```
bashbaug, line 106: These are confusing units - the global work size is the number of work-items, which is unrelated to any sort of size in bytes and hence should not be computed using `sizeof`.  Should this just be `16` instead?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
```diff
@@ -0,0 +1,185 @@
...
+                          "resultData[%d]=%d\n",
+                          i, resultData[i]);
+                return TEST_FAIL;
+            }
+        return CL_SUCCESS;
+    }
+
+    size_t info_global_offset = 0;
+    const size_t update_global_offset = 3;
+    const size_t sizeToAllocate = 76;
```
bashbaug, line 173: One more - I think it would be best to compute this from other parameters vs. hard-coding it.  For example, could this instead be:

```suggestion
    const size_t sizeToAllocate = (global_work_size + update_global_offset) * sizeof(cl_int);
```

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
pj87: The naming convention was agreed with @EwanC earlier. It's a common part of all of the `mutable dispatch` tests.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_basic.h
pj87: This will mean update in all PRs, maybe we could postpone this particular change until other things are done?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
EwanC: I had envisaged that just the `mutable_command_info.cpp` tests would inherit from `InfoMutableCommandBufferTest` and the behaviour tests wouldn't be using `clGetMutableCommandInfoKHR` as part of their verification, and so could just inherit from `BasicMutableCommandBufferTest`. But the fact that we do check `clGetMutableCommandInfoKHR` as part of the test verification makes them better tests, so I'm not proposing we change it, and was an oversight on my part.

> Do we even need InfoMutableCommandBuffertest, or would it make more
sense to remove it and just get the extension function pointer for
clGetMutableCommandInfoKHR as part of BasicMutableCommandBufferTest?

I think Ben's suggestion of rolling `clGetMutableCommandInfoKHR` into `BasicMutableCommandBufferTest`
makes sense if we're going to use it in the majority of circumstance, would also make the diff in this PR smaller. Apologies for the confusion, and the overhead of having to make this change in the other PRs

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_offset.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+        for (size_t i = 0; i < num_elements; i++)
+            if (i < update_global_offset && 0 != resultData[i])
+            {
+                log_error("Data failed to verify: update_global_offset != "
+                          "resultData[%d]=%d\n",
+                          i, resultData[i]);
+                return TEST_FAIL;
+            }
+            else if (i > update_global_offset
```
EwanC, line 153: Think we need to cover the case where `i == update_global_offset ` too
```suggestion
            else if (i >= update_global_offset
```

EwanC (approved):

PR #1744, opened 2023-05-26
Add global size tests for cl_khr_command_buffer_mutable_dispatch.

pj87:


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
```diff
@@ -0,0 +1,135 @@
...
+        test_error(error, "clUpdateMutableCommandsKHR failed");
+
+        error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
+                                          nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        error = clGetMutableCommandInfoKHR(
+            command, CL_MUTABLE_DISPATCH_GLOBAL_WORK_SIZE_KHR,
+            sizeof(test_global_work_size), &test_global_work_size, &size);
+        test_error(error, "clGetMutableCommandInfoKHR failed");
```
EwanC, line 111: We should check that's some observable output from the kernel as a result of the change to global work offset, not just that `clGetMutableCommandInfoKHR` has been updated. For example, getting every work-item to call `get_global_size()` inside of the kernel and writing it to a buffer, then reading the buffer after the command-buffer enqueue has finished and check it matches what we expect.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
```diff
@@ -0,0 +1,183 @@
...
+            nullptr /* global_work_offset */,
+            &global_size /* global_work_size */,
+            nullptr /* local_work_size */
+        };
+        cl_mutable_base_config_khr mutable_config{
+            CL_STRUCTURE_TYPE_MUTABLE_BASE_CONFIG_KHR, nullptr, 1,
+            &dispatch_config
+        };
+
+        error = clUpdateMutableCommandsKHR(command_buffer, &mutable_config);
```
EwanC, line 122: Same comment as https://github.com/KhronosGroup/OpenCL-CTS/pull/1749#discussion_r1209329308 about waiting on execution to complete before calling update. Should also make this change in other PRs too.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
```diff
@@ -0,0 +1,183 @@
...
+                return TEST_FAIL;
+            }
+
+        return CL_SUCCESS;
+    }
+
+    size_t test_global_work_size = 0;
+    const size_t global_size = 3;
+    const size_t sizeToAllocate = 64;
+    const size_t default_global_size = 16;
```
EwanC, line 172: We're using variables 4 very similar named variables in this test: `global_work_size`, `global_size`, `default_global_size`, and `test_global_work_size`.

Could we make them more identifiable: e.g `info_global_size` for the `clGetMutableCommandInfoKHR` query and `update_global_size` for the value we're updating too. I'm not sure on the difference between `default_global_size` and `global_work_size`

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+        std::vector<cl_int> resultData;
+        resultData.resize(num_elements);
+
+        error = clEnqueueReadBuffer(queue, stream, CL_TRUE, 0, sizeToAllocate,
+                                    resultData.data(), 0, nullptr, nullptr);
+        test_error(error, "clEnqueueReadBuffer failed");
+
+        for (size_t i = 0; i < num_elements; i++)
+            if (i > update_global_size && global_work_size != resultData[i])
```
bashbaug, line 152: We check for `i > update_global_size` here, and we check for `i < update_global_size` below, but this means we never check for `i == update_global_size`.  Should one of these conditions be `>=` or `<=` so we test ever element?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
```diff
@@ -0,0 +1,184 @@
...
+                return TEST_FAIL;
+            }
+
+        return CL_SUCCESS;
+    }
+
+    size_t info_global_size = 0;
+    const size_t update_global_size = 3;
+    const size_t sizeToAllocate = 64;
+    const size_t num_elements = sizeToAllocate / sizeof(cl_int);
```
bashbaug, line 174: Nitpick: I think these variables are only used inside of `Run` - do they need to be class member variables?

bashbaug (commented):
Please also see my comments on #1743, many of which apply here.

(Some of these comments may apply to #1743 also!)

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
pj87: I decided to put it there for more tidy code, since they are mostly `const`s.

EwanC (dismissed):
Needs rebased to remove conflicts, but happy with test content.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include "procs.h"
+#include "testHarness.h"
+#include "imageHelpers.h"
+#include <vector>
+#include <iostream>
+#include <random>
+#include <cstring>
+#include <algorithm>
+#include <memory>
+#include "mutable_command_basic.h"
```
EwanC, line 28: same comment as other PRs about not needed all these includes

EwanC (approved):

EwanC (dismissed):

EwanC (approved):

PR #1745, opened 2023-05-26
Add local size tests for cl_khr_command_buffer_mutable_dispatch.

pj87:


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_local_size.cpp
```diff
@@ -0,0 +1,134 @@
...
+        test_error(error, "clUpdateMutableCommandsKHR failed");
+
+        error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
+                                          nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        error = clGetMutableCommandInfoKHR(
+            command, CL_MUTABLE_DISPATCH_LOCAL_WORK_SIZE_KHR,
+            sizeof(test_local_work_size), &test_local_work_size, &size);
+        test_error(error, "clGetMutableCommandInfoKHR failed");
```
EwanC, line 111: We should check that's some observable output from the kernel as a result of the change to local work size, not just that `clGetMutableCommandInfoKHR` has been updated. For example, getting every work-item to call get_local_size() inside of the kernel and writing it to a buffer, then reading the buffer after the command-buffer enqueue has finished and check it matches what we expect.

bashbaug (commented):
This mostly looks good to me, though many of my comments from #1743 and #1744 also apply here.

My biggest question - which may be as much of a spec question as a test question - is that strictly speaking modifying a command buffer is not supposed to introduce more (or fewer, I thought?) work-groups.  See the error condition in the spec:

> CL_INVALID_OPERATION if values of local_work_size and/or global_work_size result in an increase to the number of work-groups in the ND-range.

This test modifies a command buffer by changing the local work size from `nullptr` to the fixed size `8`.  I suspect that this will (probably?) count as increasing the number of work-groups in the ND-range?

If we really do need to ensure there are no additional work-groups in the ND-range is there any way to avoid a) sending a non-`nullptr` local work size in both cases, and b) modifying both the global work size *and* the local work size?

Full disclosure: I've always thought this was a pretty big restriction and I'd love if it could be removed!

EwanC:
> This mostly looks good to me, though many of my comments from #1743 and #1744 also apply here.
> 
> My biggest question - which may be as much of a spec question as a test question - is that strictly speaking modifying a command buffer is not supposed to introduce more (or fewer, I thought?) work-groups. See the error condition in the spec:
> 
> > CL_INVALID_OPERATION if values of local_work_size and/or global_work_size result in an increase to the number of work-groups in the ND-range.
> 
> This test modifies a command buffer by changing the local work size from `nullptr` to the fixed size `8`. I suspect that this will (probably?) count as increasing the number of work-groups in the ND-range?
> 
> If we really do need to ensure there are no additional work-groups in the ND-range is there any way to avoid a) sending a non-`nullptr` local work size in both cases, and b) modifying both the global work size _and_ the local work size?
> 
> Full disclosure: I've always thought this was a pretty big restriction and I'd love if it could be removed!

For now, I think we do need to modify both the global work size and the local work size proportionally to match spec behaviour, which I see has just been pushed to this PR

We could create a OpenCL-Docs issue to revisit this constraint before the spec moves from provisional to final ? Could make changing the number of workgroups optional (the intersection of optionality of being able to change local-size, global-size, and/or num workgroups might become a bit confusing, but I don't think it's a blocker).

EwanC (approved):
Needs rebased, but happy with the test content.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_local_size.cpp
```diff
@@ -0,0 +1,180 @@
...
+#include "procs.h"
+#include "testHarness.h"
+#include "imageHelpers.h"
+#include <vector>
+#include <iostream>
+#include <random>
+#include <cstring>
+#include <algorithm>
+#include <memory>
+#include "mutable_command_basic.h"
```
EwanC, line 28: Don't think we need all these headers, locally I can remove the following and still build correctly:
* `#include "imageHelpers.h"`
* `#include <iostream>`
* `#include <random>`
* `#include <cstring>`
* `#include <algorithm>`
* `#include <memory>`

EwanC (dismissed):

EwanC:
> 



> > This mostly looks good to me, though many of my comments from #1743 and #1744 also apply here.
> > My biggest question - which may be as much of a spec question as a test question - is that strictly speaking modifying a command buffer is not supposed to introduce more (or fewer, I thought?) work-groups. See the error condition in the spec:
> > > CL_INVALID_OPERATION if values of local_work_size and/or global_work_size result in an increase to the number of work-groups in the ND-range.
> > 
> > 
> > This test modifies a command buffer by changing the local work size from `nullptr` to the fixed size `8`. I suspect that this will (probably?) count as increasing the number of work-groups in the ND-range?
> > If we really do need to ensure there are no additional work-groups in the ND-range is there any way to avoid a) sending a non-`nullptr` local work size in both cases, and b) modifying both the global work size _and_ the local work size?
> > Full disclosure: I've always thought this was a pretty big restriction and I'd love if it could be removed!
> 
> For now, I think we do need to modify both the global work size and the local work size proportionally to match spec behaviour, which I see has just been pushed to this PR
> 
> We could create a OpenCL-Docs issue to revisit this constraint before the spec moves from provisional to final ? Could make changing the number of workgroups optional (the intersection of optionality of being able to change local-size, global-size, and/or num workgroups might become a bit confusing, but I don't think it's a blocker).

As mentioned on the 27/06/23 WG call I've created an issue for this https://github.com/KhronosGroup/OpenCL-Docs/issues/946

EwanC (approved):

PR #1748, opened 2023-05-29
Add dispatch arguments tests for cl_khr_command_buffer_mutable_dispatch.

pj87:


pj87:
> It's not clear to me what the intention of this test is compared to PR #1749 which is more comprehensive argument testing

The intention here was to query for `CL_MUTABLE_DISPATCH_ARGUMENTS_KHR` but from what I understand, there is no such query. Please confirm that this is the case so I will close this issue.

EwanC:
> > It's not clear to me what the intention of this test is compared to PR #1749 which is more comprehensive argument testing
> 
> The intention here was to query for `CL_MUTABLE_DISPATCH_ARGUMENTS_KHR` but from what I understand, there is no such query. Please confirm that this is the case so I will close this issue.


`CL_MUTABLE_DISPATCH_ARGUMENTS_KHR`  is a bitfield bit that can be set as part of the `cl_mutable_dispatch_fields_khr` type. It can be either be returned by the query for device capabilites, or set as a bit on the properties to `clCommandNDRangeKernelKHR`. However, this would all naturally be covered in #1749 as a necessary part of argument testing, so indeed we don't this separate PR and it can be closed.

PR #1711, opened 2023-04-26
Added cl_khr_fp16 extenstion support for test_async_strided_copy from basic

shajder:
According to issue description (basic section):

https://github.com/KhronosGroup/OpenCL-CTS/issues/142

PR #1754, opened 2023-06-01
fix async strided test outputing error during verify

rjodinchr:
This bug was introduced by https://github.com/KhronosGroup/OpenCL-CTS/pull/1711

Ref google/clspv#1127


svenvh (approved):
LGTM, thanks!

cc: @shajder

Issue #1582, opened 2022-11-25
[Ubuntu 22.10/GCC 12.2.0] error: argument range exceeds maximum object size


jansol:
Building from tag v2022-10-27-00 with CMake's ExternalProject_Add

```
Consolidate compiler generated dependencies of target test_mem_host_flags
[ 69%] Building CXX object test_conformance/mem_host_flags/CMakeFiles/test_mem_host_flags.dir/mem_host_buffer.cpp.o
In file included from /code/pocl/build/examples/conformance/src/conformance/test_conformance/mem_host_flags/checker.h:25,
                 from /code/pocl/build/examples/conformance/src/conformance/test_conformance/mem_host_flags/checker_mem_host_read_only.hpp:19,
                 from /code/pocl/build/examples/conformance/src/conformance/test_conformance/mem_host_flags/mem_host_buffer.cpp:25:
In member function ‘void C_host_memory_block<T>::Init(int, T&) [with T = signed char]’,
    inlined from ‘cl_int cBuffer_check_mem_host_write_only<T>::verify_RW_Buffer_rect() [with T = signed char]’ at /code/pocl/build/examples/conformance/src/conformance/test_conformance/mem_host_flags/checker_mem_host_write_only.hpp:194:18:
/code/pocl/build/examples/conformance/src/conformance/test_conformance/mem_host_flags/C_host_memory_block.h:80:13: error: argument 1 range [18446744071562067968, 18446744073709551615] exceeds maximum object size 9223372036854775807 [-Werror=alloc-size-larger-than=]
   80 |     pData = new T[num_elem];
      |             ^~~~~~~~~~~~~~~
In file included from /usr/include/c++/12/bits/new_allocator.h:34,
                 from /usr/include/x86_64-linux-gnu/c++/12/bits/c++allocator.h:33,
                 from /usr/include/c++/12/bits/allocator.h:46,
                 from /usr/include/c++/12/vector:61,
                 from /code/pocl/build/examples/conformance/src/conformance/test_common/harness/imageHelpers.h:26,
                 from /code/pocl/build/examples/conformance/src/conformance/test_conformance/mem_host_flags/testBase.h:32,
                 from /code/pocl/build/examples/conformance/src/conformance/test_conformance/mem_host_flags/procs.h:19,
                 from /code/pocl/build/examples/conformance/src/conformance/test_conformance/mem_host_flags/mem_host_buffer.cpp:23:
[...]

PR #1755, opened 2023-06-01
mem_host_flags: use size_t for element count

svenvh:
More recent GCC versions (e.g. 12.2, 13.1) report that the array size argument to `new[]` in the `Init` methods exceeds the maximum object size, seemingly related to the negative range of the widened `int`.

Use an unsigned type to avoid the warning and propagate the signedness change to other uses of the `num_elements` member.

Fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/1582


EwanC (approved):
Haven't tried this on a gcc version which produces the error, but using an unsigned type here makes sense regardless.

PR #1756, opened 2023-06-06
[CI] Build Vulkan loader with USE_GAS=ON

svenvh:
https://github.com/KhronosGroup/Vulkan-Loader/pull/1212 broke builds that set `USE_GAS=OFF`.

It was originally added to fix aarch64 builds by @nikhiljnv 
>    Disable USE_GAS for Vulkan Loader build to fix aarch64 build.

but it seems we no longer need it.


EwanC (approved):
LGTM, CI on this PR shows that this fixes the macos-latest build fail appearing in other PRS

PR #1758, opened 2023-06-07
Command buffer wait_for_sec_queue_event subtest, call clFinish in the correct order.

AhmedAmraniAkdi:
queue has a command that depends on a command that resides in queue_sec, calling clFinish(queue) before clFinish(queue_sec) causes the test to hang as the queue_sec command never got a chance to finish.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
@@ -603,6 +603,9 @@ struct CommandBufferEventSync : public BasicCommandBufferTest
                                     event_ptrs[1], nullptr);
         test_error(error, "clEnqueueReadBuffer failed");
 
+        error = clFlush(queue);
+        test_error(error, "clFlush failed");
+
         error = clFinish(queue_sec);
         test_error(error, "clFinish failed");
 
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
@@ -603,10 +603,10 @@ struct CommandBufferEventSync : public BasicCommandBufferTest
                                     event_ptrs[1], nullptr);
         test_error(error, "clEnqueueReadBuffer failed");
 
-        error = clFinish(queue);
+        error = clFinish(queue_sec);
         test_error(error, "clFinish failed");
 
-        error = clFinish(queue_sec);
+        error = clFinish(queue);
         test_error(error, "clFinish failed");
 
         // verify the result - result buffer must contain initial pattern


svenvh:
> It says build macos-latest failed, but I am unsure the reason is my change.

That should be fixed by #1756

EwanC (dismissed):
LGTM

test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
```diff
@@ -603,10 +603,10 @@ struct CommandBufferEventSync : public BasicCommandBufferTest
                                     event_ptrs[1], nullptr);
         test_error(error, "clEnqueueReadBuffer failed");
 
-        error = clFinish(queue);
+        error = clFinish(queue_sec);
         test_error(error, "clFinish failed");
 
-        error = clFinish(queue_sec);
+        error = clFinish(queue);
         test_error(error, "clFinish failed");
```
EwanC, line 610: Looking at this again, could we actually just set blocking to `CL_TRUE`  on the `clEnqueueReadBuffer` call above and then remove these two `clFinish` calls

AhmedAmraniAkdi: In that case we would have the same problem again, setting `clEnqueueReadBuffer` to be a blocking command is the same as having clFinish(queue).

EwanC: ok cool, I feel like I'd expect these sorts of synchronization operations to transitively wait on the dependencies in other queues to complete, but that's behaviour from the core spec, and if it's ambiguous we shouldn't start mandating it in order to pass testing with this extension.

bashbaug (commented):
This change isn't bad but I do wonder if it is defeating the purpose of the test.  Would it be better to at least call `clFlush(queue)` first, before calling `clFinish(queue_sec)`?  So we'd have:

```c
        error = clFlush(queue);
        test_error(error, "clFlush failed");

        error = clFinish(queue_sec);
        test_error(error, "clFinish failed");

        error = clFinish(queue);
        test_error(error, "clFinish failed");
```

AhmedAmraniAkdi:
Yes, that works too as clFlush is not blocking and it also goes more "in line" with what the test was doing. Should I modify the commit with your suggestion then?

lakshmih:
Looks good.

EwanC:
> Yes, that works too as clFlush is not blocking and it also goes more "in line" with what the test was doing. Should I modify the commit with your suggestion then?

I'd also prefer the patch was updated to this suggestion

AhmedAmraniAkdi:
Updated PR.

EwanC (approved):

bashbaug (approved):
This change works for me - thanks!

Do we need any further discussion or spec clarifications to document whether or why the previous code was incorrect?

bashbaug:
Merging as discussed in the September 5th teleconference.

PR #1761, opened 2023-06-12
compiler: fix memory leak from unnecessary strdup

svenvh:
The result of the `strdup` was never freed.  The string duplication isn't necessary, so remove it.


bashbaug (approved):
No reason to wait to merge this one.

PR #1767, opened 2023-06-16
[NFC] device_execution: use raw string literals for block kernels

svenvh:
Modernize by using raw string literals, which makes the kernel sources easier to read/extract.


EwanC (approved):
I've not run this change, but looks like a good improvement.

PR #1770, opened 2023-06-19
Added cl_khr_fp16 extension support for test_decorate from spirv_new

shajder:
According to issue request https://github.com/KhronosGroup/OpenCL-CTS/issues/142


shajder:
Close/Open to trigger CI

test_conformance/spirv_new/test_decorate.cpp
```diff
@@ -389,41 +461,62 @@ template<typename Ti, typename To>
...
+        typedef cl_##Ti clTi;                                                  \
+        typedef cl_##To clTo;                                                  \
+        const int num = 1 << 16;                                               \
+        std::vector<clTi> in(num);                                             \
+        std::vector<clTo> out(num);                                            \
+        RandomSeed seed(gRandomSeed);                                          \
+                                                                               \
+        for (int i = 0; i < num; i++)                                          \
+        {                                                                      \
+            in[i] = num * genrand<clTi>(seed) - num / 2;                       \
```
coldav, line 509: Is there a possibility with the above code we could overflow half float with the rounding but the float result  we calculate would not overflow, so the results don't really match? Do we need to constrain the inputs more?

bashbaug (approved):
I think these changes look OK and they can be merged in this state.  I've run them on a few of our devices and the new tests are passing.

Recording a few nice-to-have fixes, none of which should hold up merging this PR, more things we could consider doing in the future:

- Implementation-wise think a lot of the checks for `std::is_same<Ti, cl_half>::value` could be removed with some clever template speciailizations and the code would likely be easier to read and understand as well.
- Testing-wise there are only 64K halfs, so we could test all of them exhaustively vs. relying on random sampling.

bashbaug:
@coldav would you like to give this one more review before we merge these changes?

test_conformance/spirv_new/test_decorate.cpp
coldav: is it possible to comment on this points before I approve?

bashbaug: I took a look at this yesterday and it's actually a little worse than I expected: `genrand<cl_half>(seed)` returns a `short`, not a floating-point value between zero and one, so almost all of these values overflow and we really only test one value.  This needs to be fixed.

bashbaug (changes requested):
See comment above - we need to fix the input data so it is generated properly.

bashbaug (dismissed):
I think I've fixed these now - please take a look.  My review approval isn't sufficient anymore since I've made some of the changes.

A few things I didn't do:

* We could be a bit more clever with `std::enable_if` to remove the runtime checks for `std::is_same`, but I've left the runtime checks for now.
* I've kept the random sampling vs. trying to test every half exhaustively.

shajder:
> * We could be a bit more clever with `std::enable_if` to remove the runtime checks for `std::is_same`, but I've left the runtime checks for now.

@bashbaug it looks to me that `std::is_same` will be unrolled in compile time without conditional code (at least with modern compilers and some basic optimizations enabled). Am I missing something [here](https://godbolt.org/z/necE74nhr) ?

bashbaug:
> it looks to me that `std::is_same` will be unrolled in compile time without conditional code (at least with modern compilers and some basic optimizations enabled).

Ah, good point!  Yes, the "runtime" checks should be just fine then.

Nuullll:
Verified with Intel OpenCL CPU implementation. New tests are passing!

test_conformance/spirv_new/test_decorate.cpp
```diff
@@ -177,20 +175,79 @@ TEST_SPIRV_FUNC(decorate_cpacked)
...
+
+template <typename Ti, typename Tl, typename To>
+static inline Ti generate_saturated_rhs_input(RandomSeed &seed)
+{
+    constexpr auto hiVal = std::numeric_limits<To>::max();
+
+    Tl val = genrand<Tl>(seed) % hiVal;
+    if (std::is_same<cl_half, Ti>::value)
+    {
+        if (val > 0 && val * 20 < hiVal)
```
coldav, line 201: could you add some comments to explain the * 20 please?

coldav:
The test changes in general look good to me, but we see some failures locally, so I'd like to have a few days to check if it relates to the test.

test_conformance/spirv_new/test_decorate.cpp
```diff
@@ -370,60 +446,77 @@ int test_fp_rounding(cl_device_id deviceID,
...
+    return std::floor(to_double(in));
+}
+
+template <typename Ti, typename To>
+static inline Ti generate_fprounding_input(RandomSeed &seed)
 {
-    To out = std::floor(in);
-    return out;
+    constexpr auto minVal = std::numeric_limits<To>::min() / 2;
+    constexpr auto maxVal = std::numeric_limits<To>::max() / 2;
```
hvdijk, line 486: This results in values not representable in `cl_float`, which triggers a warning (promoted to error) when building with Clang, and potential silent unwanted behavior (inputs outside of `[minVal, maxVal]`) when building with GCC.
```
/home/harald/OpenCL-CTS/test_conformance/spirv_new/test_decorate.cpp:494:42: error: implicit conversion from 'const int' to 'float' changes value from 1073741823 to 1073741824 [-Werror,-Wimplicit-const-int-float-conversion]
    return genrandReal_range<Ti>(minVal, maxVal, seed);
           ~~~~~~~~~~~~~~~~~             ^~~~~~
/home/harald/OpenCL-CTS/test_conformance/spirv_new/test_decorate.cpp:521:1: note: in instantiation of function template specialization 'generate_fprounding_input<float, int>' requested here
TEST_SPIRV_FP_ROUNDING_DECORATE(rte, round_to_even, float, int);
^
/home/harald/OpenCL-CTS/test_conformance/spirv_new/test_decorate.cpp:509:21: note: expanded from macro 'TEST_SPIRV_FP_ROUNDING_DECORATE'
            in[i] = generate_fprounding_input<clTi, clTo>(seed);               \
                    ^
```

test_conformance/spirv_new/types.hpp
```diff
@@ -89,6 +89,13 @@ T genrandReal(RandomSeed &seed)
     return genrand_real1(seed);
 }
 
+// Longer-term this could be refactored out and replace random_float():
+template <typename T> T genrandReal_range(T low, T high, RandomSeed &seed)
+{
+    T t = genrand_real1(seed);
+    return (1.0 - t) * low + t * high;
```
hvdijk, line 96: Floating point math is really difficult. If this had been `t * low + (1.0 - t) * high`, I know this could easily have resulted in random values outside of `[low, high]`. For `(1.0 - t) * low + t * high`, as long as `low < high`, I have not seen any values for which that can happen, but if this truly cannot happen, can you add a comment explaining why?

But I am actually not sure a uniform distribution is what we want here. Suppose we have a range of `[1, 4]`. There are as many floating point values in `[1, 2)` as there are in `[2, 4)`. This means individual problematic cases in `[1, 2)` only have half the probability of being tested of those in `[2, 4)`. At the same time, if there is a whole range of values that is problematic, you do want a uniform distribution. So I do not know what is best here.

coldav:
> The test changes in general look good to me, but we see some failures locally, so I'd like to have a few days to check if it relates to the test.

I think we are happy with it if you want to merge apart from the clang compile issue mentioned by Harald.

test_conformance/spirv_new/test_decorate.cpp
bashbaug: I've reproduced this issue.  Working on a fix.

bashbaug: I think the right thing to do here may be to simply silence the warning by inserting explicit casts.  Here's why:

This function is generating input data for tests that round from a floating-point value to an integer value.  The `minVal` and the `maxVal` are used to determine the range of input data that is tested, and is determined by the integer type we are rounding to.

Since we're dividing the minimum value and the maximum value by two we don't necessarily test the entire range of integer data, but we do test a good portion of it.  This also means, though, if we happen to test a slightly smaller range or a slightly larger range because the conversion from the integer minimum or maximum value to the floating-point minimum or maximum value is inexact, that's fine, and nothing to worry about.

I'll make a suggestion in a different review comment for review.

test_conformance/spirv_new/test_decorate.cpp
```diff
@@ -370,60 +446,77 @@ int test_fp_rounding(cl_device_id deviceID,
...
+
+    if (std::is_same<cl_half, Ti>::value)
+    {
+        cl_float f = genrandReal_range<cl_float>(minVal, maxVal, seed);
+        return cl_half_from_float(f, CL_HALF_RTE);
+    }
+
+    return genrandReal_range<Ti>(minVal, maxVal, seed);
```
bashbaug, line 494: Proposed fix for Clang implicit conversion warning:

```suggestion
    if (std::is_same<cl_half, Ti>::value)
    {
        constexpr auto minVal =
            static_cast<cl_float>(std::numeric_limits<To>::min() / 2);
        constexpr auto maxVal =
            static_cast<cl_float>(std::numeric_limits<To>::max() / 2);
        cl_float f = genrandReal_range<cl_float>(minVal, maxVal, seed);
        return cl_half_from_float(f, CL_HALF_RTE);
    }

    constexpr auto minVal = static_cast<Ti>(std::numeric_limits<To>::min() / 2);
    constexpr auto maxVal = static_cast<Ti>(std::numeric_limits<To>::max() / 2);
    return genrandReal_range<Ti>(minVal, maxVal, seed);
```

bashbaug (commented):
Here is the proposed fix for the Clang warning - see discussion above.

test_conformance/spirv_new/test_decorate.cpp
hvdijk: " This also means ... nothing to worry about." Thanks for the explanation, that seems like it makes sense.

bashbaug: Great, I've committed my suggestion and I think this is good-to-go now, just needs a review approval.

bashbaug:
@coldav @hvdijk if either of you can approve this PR we can probably merge it (finally) in tomorrow's teleconference.  Thanks!

coldav (approved):

coldav:
@hvdijk is away this week, but I've approved it.

alycm (approved):
Doing the GitHub approval on @coldav's behalf.

bashbaug:
Merging as discussed in the February 13th teleconference.

PR #1771, opened 2023-06-19
test_common: fix -Wformat warnings

svenvh:
In preparation of re-enabling -Wformat globally, fix format string warnings in test_common.

Printing a `size_t` requires the `%zu` specifier.

### DIFF
--- a/test_common/gl/helpers.cpp
+++ b/test_common/gl/helpers.cpp
@@ -1047,8 +1047,10 @@ void * CreateGLTexture2DMultisample( size_t width, size_t height, size_t samples
   GLint max_samples = get_gl_max_samples(target, internalFormat);
   check_gl_error()
 
-  if (max_samples < (GLint)samples)
-    log_error("GL error: requested samples (%d) exceeds renderer max samples (%d)\n", samples, max_samples);
+      if (max_samples < (GLint)samples)
+          log_error("GL error: requested samples (%zu) exceeds renderer max "
+                    "samples (%d)\n",
+                    samples, max_samples);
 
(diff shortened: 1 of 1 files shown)


test_common/gl/helpers.cpp
```diff
@@ -1047,8 +1047,10 @@ void * CreateGLTexture2DMultisample( size_t width, size_t height, size_t samples
   GLint max_samples = get_gl_max_samples(target, internalFormat);
   check_gl_error()
 
-  if (max_samples < (GLint)samples)
-    log_error("GL error: requested samples (%d) exceeds renderer max samples (%d)\n", samples, max_samples);
+      if (max_samples < (GLint)samples)
```
EwanC, line 1050: why is this `if` indented now? Is it because the lack of a semi-colon on `check_gl_error` is confusing clang-format?

svenvh: Yes, that seems to be the cause of confusion indeed.  Adding a semicolon or capitalizing the macro name fixes the indentation.  I think it'd be best to capitalize the macro, which I'd prefer to do in a followup PR (perhaps together with a full reformat of this file?).

EwanC: follow-up PR is fine

EwanC (approved):

test_common/gl/helpers.cpp
svenvh: Followup: https://github.com/KhronosGroup/OpenCL-CTS/pull/1813

PR #1776, opened 2023-06-24
Replace OBSOLETE_FORAMT with OBSOLETE_FORMAT

### LINKED ISSUE #1769: Misspell of OBSOLETE_FORMAT as OBSOLETE_FORAMT in imageHelpers.cpp
In imageHelpers.cpp line 101 and a few other places OBSOLETE_FORMAT is misspelled as OBSOLETE_FORAMT. Line 305 has it spelled correctly though. This misspelling also appears in the image tests, specifically images/common.cpp line 22.

niranjanjoshi121:
In imageHelpers.cpp and few other places in image tests, OBSOLETE_FORMAT is misspelled as OBSOLETE_FORAMT. Fix misspelling by replcaing it with OBSOLETE_FORMAT.

Fixes #1769


svenvh (dismissed):

kpet (changes requested):
If we change the ifdef this will break the code for whomever might have been using this. We don't have definitions for those formats in the standard headers so this can only build with custom headers. I don't think this is code we should have at all in the CTS, certainly not without knowing who's doing what with it. I suggest we just remove any and all code guarded by `OBSOLETE_FORAMT` instead.

bashbaug (approved):

kpet (approved):

PR #1778, opened 2023-06-30
subgroups: fix Wsign-compare warnings

svenvh:
The subgroup and workgroup sizes reported by clGetKernelSubGroupInfo and clGetKernelWorkGroupInfo are of type `size_t`.  Avoid changing the values to an `int` type as they are propagated through the tests and then compared against `size_t` again.


bashbaug (approved):

PR #1779, opened 2023-07-04
Fix more -Wsign-compare warnings

svenvh:
Fix `-Wsign-compare` warnings in multiple tests.  Most of these are trivial `for`-loop iterator type changes.

### DIFF
--- a/test_conformance/basic/test_fpmath.cpp
+++ b/test_conformance/basic/test_fpmath.cpp
@@ -94,7 +94,7 @@ int verify_fp(std::vector<T> (&input)[2], std::vector<T> &output,
 {
     auto &inA = input[0];
     auto &inB = input[1];
-    for (int i = 0; i < output.size(); i++)
+    for (size_t i = 0; i < output.size(); i++)
     {
         bool nan_test = false;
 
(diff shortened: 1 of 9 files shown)


test_conformance/basic/test_fpmath.cpp
```diff
@@ -106,7 +106,7 @@ int verify_fp(std::vector<T> (&input)[2], std::vector<T> &output,
         if (r != output[i] && nan_test)
         {
             log_error("FP math test for type: %s, vec size: %zu, failed at "
-                      "index %d, %a '%c' %a, expected %a, get %a\n",
+                      "index %u, %a '%c' %a, expected %a, get %a\n",
```
EwanC, line 109: suggestion that we could use `%zu` rather than `%u` for `size_t`

EwanC (approved):
LGTM

test_conformance/basic/test_fpmath.cpp
svenvh: Good catch, thanks!  We really need to get `Wformat` fixed and enabled. :-)

EwanC (approved):

PR #1780, opened 2023-07-06
New commands for profiling test

gwawiork:
* clEnqueueMarkerWithWaitList
* clEnqueueBarrierWithWaitList
The test verifies if:
- sync commands have proper timestamps
- ndrange commands before and after sync have proper timestamps (marker, barrier works correctly)

### DIFF
--- a/test_conformance/profiling/enqueueSyncFunctions.cpp
+++ b/test_conformance/profiling/enqueueSyncFunctions.cpp
@@ -1,0 +1,272 @@
+//
+// Copyright (c) 2025-2026 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "harness/compat.h"
+
+#include <stdio.h>
+#include <string.h>
+#include <time.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+
+#include "procs.h"
+#include "harness/testHarness.h"
+
+
+static const char *test_kernel = R"CLC(
+__kernel void test(__global int* dst) {
+    size_t id = get_global_linear_id();
+    size_t loop_end = 1<<12UL;
+    for (size_t i = 0; i <= loop_end; i++) {
+        if(i%2 == 0) {
+            dst[id] = 0;
+        } else {
+            dst[id] += 1;
+        }
+    }
+}
+)CLC";
+
+// Helper that executes one variant (in-order or out-of-order)
+static int run_enqueue_variant(cl_device_id device, cl_context context,
+                               cl_command_queue queue, bool is_out_of_order,
+                               int (*fn)(cl_command_queue, cl_uint,
+                                         const cl_event *, cl_event *))
+{
+    cl_int error = CL_SUCCESS;
+    int test_status = TEST_PASS;
+
+    const size_t global_work_size = 256;
+    const size_t allocSize = global_work_size * sizeof(cl_int);
+
+    clMemWrapper buffer1 =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, nullptr, &error);
+    test_error(error, "Unable to create buffer1");
+    clMemWrapper buffer2 =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, nullptr, &error);
+    test_error(error, "Unable to create buffer2");
+    clMemWrapper buffer3 =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, nullptr, &error);
+    test_error(error, "Unable to create buffer3");
+    clMemWrapper buffer4 =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, nullptr, &error);
+    test_error(error, "Unable to create buffer4");
+    clMemWrapper buffer5 =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, nullptr, &error);
+    test_error(error, "Unable to create buffer5");
+    clMemWrapper buffer6 =
+        clCreateBuffer(context, CL_MEM_READ_WRITE, allocSize, nullptr, &error);
+    test_error(error, "Unable to create buffer6");
+
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+    error = create_single_kernel_helper(context, &program, &kernel, 1,
+                                        &test_kernel, "test");
+    test_error(error, "Unable to build program / create kernel 'test'");
+
+    clEventWrapper events_set1[3] = { nullptr, nullptr, nullptr };
+    clEventWrapper events_set2[3] = { nullptr, nullptr, nullptr };
+    clEventWrapper sync_event1, sync_event2;
+
+    // First set (3 launches of the same kernel with different buffers)
+    error = clSetKernelArg(kernel, 0, sizeof(buffer1), &buffer1);
+    test_error(error, "Set arg buffer1 failed");
+    error = clEnqueueNDRangeKernel(queue, kernel, 1, nullptr, &global_work_size,
+                                   nullptr, 0, nullptr, &events_set1[0]);
+
+    error |= clSetKernelArg(kernel, 0, sizeof(buffer2), &buffer2);
+    test_error(error, "Set arg buffer2 failed");
+    error |=
+        clEnqueueNDRangeKernel(queue, kernel, 1, nullptr, &global_work_size,
+                               nullptr, 0, nullptr, &events_set1[1]);
+
+    error |= clSetKernelArg(kernel, 0, sizeof(buffer3), &buffer3);
+    test_error(error, "Set arg buffer3 failed");
+    error |=
+        clEnqueueNDRangeKernel(queue, kernel, 1, nullptr, &global_work_size,
+                               nullptr, 0, nullptr, &events_set1[2]);
+    test_error(error, "Unable to enqueue first kernel set");
+
+    error = fn(queue, 3, &events_set1[0], &sync_event1);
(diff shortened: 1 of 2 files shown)


test_conformance/profiling/enqueueSyncFunctions.cpp
```diff
@@ -0,0 +1,287 @@
...
+
+    error = create_single_kernel_helper(context, &program, &kernel2, 1,
+                                        &test_kernel, "test2");
+    test_error(error, "Unable to create test kernel");
+
+    error = clSetKernelArg(kernel2, 0, sizeof(buffer2), &buffer2);
+    test_error(error, "Unable to set argument for test kernel");
+
+    error = create_single_kernel_helper(context, &program, &kernel3, 1,
+                                        &test_kernel, "test1");
```
lakshmih, line 149: Should this be test3?

test_conformance/profiling/enqueueSyncFunctions.cpp
```diff
@@ -0,0 +1,287 @@
...
+    error |= clEnqueueNDRangeKernel(queue_with_props, kernel3, 1, NULL,
+                                    global_work_size, NULL, 0, NULL,
+                                    &events_list_set2[2]);
+    test_error(error, "Unable to enqueue kernels in set 2");
+
+    error =
+        fn(queue_with_props, 3, &events_list_set2[0], &eventEnqueueMarkerSet2);
+    test_error(error, "Unable to enqueue sync command");
+
+    error = clWaitForEvents(1, &eventEnqueueMarkerSet2);
```
lakshmih, line 189: Should clWaitForEvents be switched to eventEnqueueMarkerSet1 since that's what's being checked in the following lines, the command queue is out of order, and set 1 doesn't wait for set 2 to complete? Or maybe clFinish would be more readable?

gwawiork:
@lakshmih Thanks for review. Unfortunately I see some errors on our internal driver version. Until these issues will not be solved I do not plan any updates in this PR. I will let you know when it will be ready for review.

lakshmih:
> @lakshmih Thanks for review. Unfortunately I see some errors on our internal driver version. Until these issues will not be solved I do not plan any updates in this PR. I will let you know when it will be ready for review.

Thank you. Could you review the focused review label and add it back when ready?

gwawiork:
> > @lakshmih Thanks for review. Unfortunately I see some errors on our internal driver version. Until these issues will not be solved I do not plan any updates in this PR. I will let you know when it will be ready for review.
> 
> Thank you. Could you review the focused review label and add it back when ready?

I think I do not have permission to modify labels.

gwawiork:
This PR is refactored and fixed. Ready for review 
@lakshmih FYI
I think https://github.com/KhronosGroup/OpenCL-CTS/pull/2374 is not needed

test_conformance/profiling/main.cpp
bashbaug, line 1: Confirming: There are no changes to this file other than formatting?

test_conformance/profiling/procs.h
bashbaug, line 1: Confirming: There are no changes to this file other than formatting, too?

test_conformance/profiling/enqueueSyncFunctions.cpp
```diff
@@ -0,0 +1,261 @@
...
+                                    cl_event *event))
+{
+    cl_int error = CL_SUCCESS;
+    int test_status = TEST_PASS;
+
+    const cl_command_queue_properties props_out_of_order =
+        CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE | CL_QUEUE_PROFILING_ENABLE;
+    clCommandQueueWrapper queue_with_props =
+        clCreateCommandQueue(context, device, props_out_of_order, &error);
+    test_error(error, "Unable to create profiling command queue");
```
bashbaug: Does this need to be an out-of-order queue?  If so, we need checks to ensure that the test is properly skipped if the test device does not support out-of-order queues.

Should we run this twice - once with in-order queues, and then again with out-of-order queues, assuming the device supports out-of-order queues?

test_conformance/profiling/enqueueSyncFunctions.cpp
```diff
@@ -0,0 +1,261 @@
...
+    size_t loop_end = 1<<12UL;
+    for (size_t i = 0; i <= loop_end; i++) {
+        if(i%1 == 0) {
+            dst[id] = 0;
+        } else {
+            dst[id] += 1;
+        }
+    }
+}
+__kernel void test2(__global int* dst) {
```
bashbaug: Minor: Do we actually need three different kernels here, or since the content of `test2` is the same as `test1` (and `test3`), could we instantiate `test1` three times instead?

test_conformance/profiling/enqueueSyncFunctions.cpp
```diff
@@ -0,0 +1,261 @@
...
+            dst[id] = 0;
+        } else {
+            dst[id] += 1;
+        }
+    }
+}
+)CLC";
+
+int test_enqueue_function(cl_device_id device, cl_context context,
+                          cl_command_queue queue, int num_elements,
```
bashbaug: Since the passed-in queue is unused and intended to be unused, I'd recommend removing it, to avoid accidentally using it.

```suggestion
                          int num_elements,
```

If we do this, we could name the test queue `queue` instead of `queue_with_props`, also.

test_conformance/profiling/procs.h
gwawiork: yes

test_conformance/profiling/main.cpp
gwawiork: Yes, just date in Copyright header

test_conformance/profiling/enqueueSyncFunctions.cpp
gwawiork: fixed

test_conformance/profiling/enqueueSyncFunctions.cpp
gwawiork: fixed

test_conformance/profiling/enqueueSyncFunctions.cpp
gwawiork: fixed

test_conformance/profiling/main.cpp
bashbaug: Not going to lose sleep over this one, but should the file be removed from the PR then?

test_conformance/profiling/enqueueSyncFunctions.cpp
bashbaug: This is fine the way you currently have it, but if you still want the test to operate on three kernel handles still, something like this is what I was thinking of:

```c++
    clProgramWrapper program;
    clKernelWrapper kernel;
    error = create_single_kernel_helper(context, &program, &kernel, 1,
                                        &test_kernel, "test");
    // create two more kernel handles (error checking removed for brevity)
    clKernelWrapper kernel2 = clCreateKernel(program, "test", &error);
    clKernelWrapper kernel3 = clCreateKernel(program, "test", &error);
```

No action required if the test is doing what you want it to do, so I'll just resolve this thread.

test_conformance/profiling/enqueueSyncFunctions.cpp
```diff
@@ -0,0 +1,286 @@
...
+                     "guaranteed).\n");
+        else
+            log_info(
+                "Marker on in-order: ordering implied; no explicit check.\n");
+    }
+
+    // Sync event 1 should start only after all set1 kernels finished (only
+    // meaningful on OOO)
+    if (is_out_of_order)
+    {
```
bashbaug: I'd consider removing this check.  The ordering should be trivially satisfied due to the in-order queue semantics, but I think it's still woth checking, and the checks aren't very expensive.

test_conformance/profiling/enqueueSyncFunctions.cpp
```diff
@@ -0,0 +1,286 @@
...
+        ? TEST_FAIL
+        : TEST_PASS;
+}
+
+REGISTER_TEST(enqueue_marker)
+{
+    int (*foo)(cl_command_queue command_queue, cl_uint num_events_in_wait_list,
+               const cl_event *event_wait_list, cl_event *event);
+    foo = clEnqueueMarkerWithWaitList;
+    return test_enqueue_function(device, context, foo);
```
bashbaug: Is there a need for `foo`, or can this be simplified?

```suggestion
    return test_enqueue_function(device, context, clEnqueueMarkerWithWaitList);
```

test_conformance/profiling/enqueueSyncFunctions.cpp
gwawiork: added waiting on sync_event1 beside of waiting on sync_event2

test_conformance/profiling/enqueueSyncFunctions.cpp
gwawiork: now there is one kernel in the code run multiple times

test_conformance/profiling/enqueueSyncFunctions.cpp
gwawiork: done

test_conformance/profiling/enqueueSyncFunctions.cpp
gwawiork: Ok I is removed

bashbaug: Sorry, in hindsight my comment may have been ambiguous.  I was suggesting that we should remove the check for out-of-order queues and perform the tests against the end of set 1 and the start of set 2 even for in-order queues.  This ordering is due to the in-order queue semantics and unrelated to the marker, but since we're testing this case, we may as well test it completely.

Specifically, could we do something like this?

```c++
    // Ordering:
    //  - barrier: second set must start only after ALL kernels of first set end
    //  - marker: on OOO queue ordering is not guaranteed (skip); on in-order
    //  queue ordering is implied
    if (is_barrier || !is_out_of_order)  // test both barriers and in-order queues
    {
        for (int i = 0; i < 3; i++)
        {
            /* existing checks against start_set2 and end_set1 here */
        }
    }
    else
    {
        log_info("Marker on OOO: skipping cross-set ordering (not "
                 "guaranteed).\n");
    }
```

Note also, strictly speaking I believe there is a data race in the out-of-order queue and marker case, since the kernels in set1 are operating on the same memory objects as the kernels in set2 without any ordering dependency between them.

gwawiork: Restored removed code for as above example. I've added more buffers to avoid data race.

Issue #1781, opened 2023-07-06
empty command buffer


bashbaug:
See internal issue 334.

It is currently not an error to create, finalize, and execute an empty command buffer.  We should have a test for this case.

bashbaug:
This was fixed by #1782.  Closing.

PR #1782, opened 2023-07-07
[Command-buffer] Test finalizing empty command-buffer

EwanC:
This patch adds two test cases related to command-buffer finalization:

1) That it is an error to finalize and already finalized command-buffer. See https://github.com/KhronosGroup/OpenCL-Docs/pull/817

2) That it is not an error to create, finalize, and execute an empty command-buffer. Closes issue #1781

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_finalize.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_finalize.cpp
@@ -1,0 +1,85 @@
+//
+// Copyright (c) 2023 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


test_conformance/extensions/cl_khr_command_buffer/command_buffer_finalize.cpp
```diff
@@ -0,0 +1,85 @@
...
+{
+    return MakeAndRunTest<FinalizeInvalid>(device, context, queue,
+                                           num_elements);
+}
+
+int test_finalize_empty(cl_device_id device, cl_context context,
+                        cl_command_queue queue, int num_elements)
+{
+    return MakeAndRunTest<FinalizeEmpty>(device, context, queue, num_elements);
+}
```
bcalidas, line 85: Does the spec explicitly allow for empty command buffers.

EwanC: There's no error specified for empty command-buffers, which is intentional based on the discussion in internal issue 334 about empty command-buffers.

We could add some text to the spec to reinforce that empty command-buffers are permitted, if that would be helpful.

bashbaug (approved):
I verified that these tests pass with the command buffer emulation layer.

bcalidas:
On further reflection, empty command buffers might be helpful to have. We are ok with this PR.

PR #1785, opened 2023-07-12
fix a sporadic failure in the mutable_command_dimensions test

bashbaug:
This change fixes a sporadic failure in the mutable_command_dimensions test.

The previous test passed an uninitialized `size_t` variable as the return location for the `CL_MUTABLE_DISPATCH_DIMENSIONS_KHR` query, though this query only returns a `cl_uint`.  This meant that the test sometimes passes, especially for 32-bit builds, where `sizeof(size_t) == sizeof(cl_uint)`, but will frequently fail for larger values of `size_t` if the upper bits of the uninitialized variable are non-zero.  Changing to a `cl_uint` instead fixes the sporadic failures.

I also fixed a few ambiguous or incorrect error messages.

With these changes I'm seeing a 100% pass rate for the mutable dispatch tests with the command buffer emulation layer.


EwanC (approved):
LGTM

bashbaug:
It'd be really nice to get this merged, since I trip over this sporadic failure anytime I try command buffer testing.

Adding "focused review" - thanks!

bashbaug:

PR #1790, opened 2023-07-27
Fix minimum image size for cl_ext_image_requirements_info

JohnKesapidesARM:
Some new extension formats require the width to be a multiple of 4 or 2 (see cl_ext_image_raw10_raw12).
Change has not functional impact as the affected image is only used for generic queries.


lakshmih (approved):

bashbaug (approved):
Thanks - I think this could affect a couple of our image formats too (see [cl_intel_packed_yuv](https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_packed_yuv.html)).

bashbaug:

PR #1791, opened 2023-07-27
Fix CopyImage verification for 1D/2D images

JohnKesapidesARM:
The verification uses a common function with nested loops to verify the result of the copy operation. The upper loop limits thirdDim and SecondDim should be set according to the image type under test. Previously for 1D/2D they were set from dstImageInfo->depth and dstImageInfo->height. The issue is that the depth and height are set to 0 when unused. This caused the verification loop to be skipped.


lakshmih (approved):

bashbaug:

PR #1799, opened 2023-08-14
make atomic tests in prog var testing optional

karolherbst:
This is helpful for testing progvars on implementations not implementing all the optional atomic features.


test_conformance/basic/test_progvar.cpp
```diff
@@ -547,12 +549,28 @@ static cl_int print_build_log(cl_program program, cl_uint num_devices,
...
     l_has_double = is_extension_available(device, "cl_khr_fp64");
     l_has_cles_int64 = is_extension_available(device, "cles_khr_int64");
 
+    if (get_device_cl_version(device) >= Version(3, 0)) {
+        OpenCLCFeatures features;
+        status = get_device_cl_c_features(device, features);
+        if (status)
+        {
+            log_error("Couldn't query OpenCL C features for the device!\n");
+            return TEST_FAIL;
```
bashbaug, line 566: Nitpick: I think this is mixing CL error codes (e.g. CL_SUCCESS) with test statuses (TEST_FAIL).  Could this just return the CL error code instead?

```suggestion
            return status;
```

Alternatively this could just `assert(status == CL_SUCCESS)`, which isn't great but is already used below.

test_conformance/basic/test_progvar.cpp
```diff
@@ -547,12 +549,28 @@ static cl_int print_build_log(cl_program program, cl_uint num_devices,
...
+        }
+
+        l_has_atomics =
+           features.supports__opencl_c_atomic_order_seq_cst
+           && features.supports__opencl_c_atomic_scope_device;
```
bashbaug, line 571: This may be correct, but this will essentially mean that the atomic tests wont run on devices supporting OpenCL 1.2-level atomics.  Can this check be weakened at all?

karolherbst: this would require rewriting the OpenCL code as it's using the optional `atomic_load` and `atomic_store` functions.

It could be done, just need to figure out if the kernels needs to be rewritten as we can't do device wide atomic operations without optional features.

Maybe the code is even fine with the scope being reduced to `memory_scope_work_group`. Would have to dig deeper into how it all works.

bashbaug: Ah, gotcha.  I did a quick test and it does indeed appear that the kernels are using the `atomic_load` and `atomic_store` overloads with no additional arguments, so they do currently require both the `seq_cst` and `scope_device` features.  I do think these could be weakened to use a narrower scope or memory order, but that's beyond the scope of this PR.

Worth mentioning also, there is still pretty good testing coverage even without atomics - I count 141 total kernels with atomics, and 123 total kernels without atomics.

I'll just resolve this comment.  Thanks!

bashbaug (dismissed):

bashbaug:
Discussed in the October 3rd teleconference Waiting one more week for additional review.

It would be nice to fix the code formatting check also.

lakshmih (approved):

bashbaug (approved):

bashbaug:

PR #1801, opened 2023-08-26
Add NULL CL_MEM_HOST_PTR check for clGetMemObjectInfo

### LINKED ISSUE #1752: Value returned for `CL_MEM_HOST_PTR` not validated when `CL_MEM_USE_HOST_PTR` not present in flags
The spec requires implementations return `NULL` for `CL_MEM_HOST_PTR` when the flags passed at memory object creation time do not contain `CL_MEM_USE_HOST_PTR` but it seems the CTS is not checking this.

niranjanjoshi121:
The spec requires implementations return NULL for CL_MEM_HOST_PTR when the flags passed at memory object creation time do not contain CL_MEM_USE_HOST_PTR

CTS was not checking this. Add the same check.

Fixes #1752

### DIFF
--- a/test_conformance/api/test_mem_object_info.cpp
+++ b/test_conformance/api/test_mem_object_info.cpp
@@ -217,6 +217,9 @@ int test_get_buffer_info( cl_device_id deviceID, cl_context context, cl_command_
             // Create a buffer object to test against.
             bufferObject = clCreateBuffer( context, bufferFlags[ i ], addressAlign * 4, NULL, &error );
             test_error( error, "Unable to create buffer to test with" );
+            void *ptr;
+            TEST_MEM_OBJECT_PARAM(bufferObject, CL_MEM_HOST_PTR, ptr, NULL,
+                                  "host pointer", "%p", void *)
         }
 
         // Perform buffer object queries.


lakshmih (approved):

test_conformance/api/test_mem_object_info.cpp
```diff
@@ -217,6 +217,9 @@ int test_get_buffer_info( cl_device_id deviceID, cl_context context, cl_command_
             // Create a buffer object to test against.
             bufferObject = clCreateBuffer( context, bufferFlags[ i ], addressAlign * 4, NULL, &error );
             test_error( error, "Unable to create buffer to test with" );
+            void *ptr;
+            TEST_MEM_OBJECT_PARAM(bufferObject, CL_MEM_HOST_PTR, ptr, NULL,
+                                  "host pointer", "%p", void *)
```
bashbaug, line 222: This is fine, but as-written it is only testing two scenarios: the USE_HOST_PTR case (tested above, ~line 177), and this case when no other flags are passed.  Would it be better to refactor this part of the test slightly, so this query gets tested in more cases as part of "Perform buffer object queries" below, vs. being performed here when the buffer is created?

bashbaug:
Discussed in the October 3rd teleconference.  Will wait one more week.

bashbaug:
Merging as discussed in the October 10th teleconference.

PR #1802, opened 2023-08-26
Add negative test for test_svm size check

### LINKED ISSUE #1701: Add a negative test creating a buffer from an SVM pointer with too large a size
See https://github.com/KhronosGroup/OpenCL-Docs/issues/875 and https://github.com/KhronosGroup/OpenCL-Docs/pull/879

niranjanjoshi121:
Isuue #875 clarified that CL_INVALID_BUFFER_SIZE should be returned if clCreateBuffer is passed a pointer returned by clSVMAlloc as its host_ptr and the size of the buffer exceeds the size of the SVM allocation.

Add a new negative test to ensure CL_INVALID_BUFFER_SIZE is returned when the size of buffer exceeds the size of the SVM allocation.

Fixes #1701

### DIFF
--- a/test_conformance/SVM/test_allocate_shared_buffer_negative.cpp
+++ b/test_conformance/SVM/test_allocate_shared_buffer_negative.cpp
@@ -1,0 +1,102 @@
+//
+// Copyright (c) 2017 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "common.h"
+
+const cl_mem_flags svm_flag_set[] = {
+    CL_MEM_READ_WRITE,
+    CL_MEM_WRITE_ONLY,
+    CL_MEM_READ_ONLY,
+    CL_MEM_READ_WRITE | CL_MEM_SVM_FINE_GRAIN_BUFFER,
+    CL_MEM_WRITE_ONLY | CL_MEM_SVM_FINE_GRAIN_BUFFER,
+    CL_MEM_READ_ONLY | CL_MEM_SVM_FINE_GRAIN_BUFFER,
+    CL_MEM_READ_WRITE | CL_MEM_SVM_FINE_GRAIN_BUFFER | CL_MEM_SVM_ATOMICS,
+    CL_MEM_WRITE_ONLY | CL_MEM_SVM_FINE_GRAIN_BUFFER | CL_MEM_SVM_ATOMICS,
+    CL_MEM_READ_ONLY | CL_MEM_SVM_FINE_GRAIN_BUFFER | CL_MEM_SVM_ATOMICS,
+    0
+};
+const char* svm_flag_set_names[] = {
+    "CL_MEM_READ_WRITE",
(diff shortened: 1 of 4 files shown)


niranjanjoshi121:
With this change the test fails on some OpenCL implementations.

lakshmih (approved):

test_conformance/SVM/test_allocate_shared_buffer_negative.cpp
```diff
@@ -0,0 +1,102 @@
+//
+// Copyright (c) 2017 The Khronos Group Inc.
```
bashbaug, line 2: Nitpick:

```suggestion
// Copyright (c) 2024 The Khronos Group Inc.
```

test_conformance/SVM/test_allocate_shared_buffer_negative.cpp
```diff
@@ -0,0 +1,102 @@
...
+    "0"
+};
+
+
+int test_svm_allocate_shared_buffer_negative(cl_device_id deviceID,
+                                             cl_context context2,
+                                             cl_command_queue queue,
+                                             int num_elements)
+{
+    clContextWrapper context = NULL;
```
bashbaug, line 49: Not a must-fix since this follows the same pattern used by other SVM tests, but the multiple contexts (and multiple queues) confused me initially.  I would consider renaming the function arguments to something like `context_from_harnes` or `queue_from_harness` so it's more obvious that they should not be used, or perhaps to use the single-device context (and queue?) as-is for the tests where we do not need a multi-device context (are there any of these?).

test_conformance/SVM/test_allocate_shared_buffer_negative.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+    cl_device_svm_capabilities caps;
+    err = clGetDeviceInfo(deviceID, CL_DEVICE_SVM_CAPABILITIES,
+                          sizeof(cl_device_svm_capabilities), &caps, NULL);
+    test_error(err, "clGetDeviceInfo failed for CL_DEVICE_SVM_CAPABILITIES");
+
+    // under construction...
+    err = create_cl_objects(deviceID, NULL, &context, &program, &queues[0],
+                            &num_devices, CL_DEVICE_SVM_COARSE_GRAIN_BUFFER);
+    if (err) return -1;
```
bashbaug, line 63:
```suggestion
    if (err) return err;
```

Note: `create_cl_objects` has the line:

```
return 1; // 1 indicates do not execute, but counts as passing.
```

So, I think it could return a non-zero value and the test would still "pass"?

I'm pretty sure this error is spread throughout the SVM tests...

test_conformance/SVM/test_allocate_shared_buffer_negative.cpp
```diff
@@ -0,0 +1,102 @@
...
+                                               2 * size, pBufData1, &err);
+            test_failure_error(err, CL_INVALID_BUFFER_SIZE,
+                               "clCreateBuffer did not return expected error"
+                               "CL_INVALID_BUFFER_SIZE");
+        }
+
+        clSVMFree(context, pBufData1);
+    }
+
+    return 0;
```
bashbaug, line 101: Nitpick:
```suggestion
    return TEST_PASS;
```

bashbaug (commented):
Have you considered adding this test to the existing file "test_allocate_shared_buffers.cpp"?

To be clear, I am NOT suggesting that this test be merged into the existing `test_svm_allocate_shared_buffer()` function, but rather that the new `test_svm_allocate_shared_buffer_negative()` function be added to the existing file.  This would allow reuse of the `flag_set` and `flag_set_names` arrays, and if a change is needed in one of the tests it would likely be applicable to the other test as well.

bashbaug (approved):
This one has been around for a while.  I'm fine merging it as-is, and addressing the review comments in a subsequent PR.

bashbaug:

Issue #1794, opened 2023-08-01
math_bruteforce fp64 precision issues on aarch64


nikhiljnv:
math_bruteforce seems to be failing on aarch64 for fp64 tests due to difference in CPU reference calculation.
The issues seems to be arising due to difference in clang and gcc behavior when handling ffp-contract default case.
While clang assumes ffp-contract=off by default and doesn't generate any fma instructions, but does so with ffp-contract=fast,
gcc does the opposite where it assumes ffp-contract=fast by default and generates fma instructions, but doesn't do so when ffp-contract=off set explicitly.
This difference in behavior leads to different CPU reference value calculation across x86_64 and aarch64 as well as differences between CPU reference and CPU device computation on aarch64. 

https://godbolt.org/z/bW6M1f47W
https://godbolt.org/z/P9Mqx8Yr9

Apparently, this doesn't cause any issues for fp32 tests which is why 
A simple fix could be to enforce ffp-contract=off for CTS build on aarch64.

PR #1803, opened 2023-08-27
Add ffp-contract=off Compilation flag for CTS build.

nikhiljnv:
GCC defaults to using ffp-contract=fast even when fast math is disabled in the
case of GNU C
This creates precision issues when comparing the results with that of x86_64.
GNU options reference:
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#

This fix is to disable floating-point expression contractions with
flag ffp-contract=off for math_brute_force tests

Fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/1794


nikhiljnv:
Closing in favor of #1805

PR #1805, opened 2023-09-03
 Add ffp-contract=off Compilation flag for CTS build.

### LINKED ISSUE #1794: math_bruteforce fp64 precision issues on aarch64

nikhiljnv:
GCC defaults to using ffp-contract=fast even when fast math is disabled in the
case of GNU C
This creates precision issues when comparing the results with that of x86_64.
GNU options reference:
https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#

This fix is to disable floating-point expression contractions with
 flag ffp-contract=off for math_brute_force tests

    Fixes #1794


lakshmih:
Can this be rebased into the single commit that is needed? We will still review and test meanwhile.

nikhiljnv:
> Can this be rebased into the single commit that is needed? We will still review and test meanwhile. Thanks.

I guess we can just squeeze the commit history into one single commit while merging. I can try to have a clean commit history, but shouldn't affect the actual change.

lakshmih:
> > Can this be rebased into the single commit that is needed? We will still review and test meanwhile. Thanks.
> 
> I guess we can just squeeze the commit history into one single commit while merging. I can try to have a clean commit history, but shouldn't affect the actual change.

That sounds reasonable. The change itself looks fine. Thanks.

lakshmih (dismissed):

bashbaug:
Discussed in the October 3rd teleconference.  Will wait another week to address the comments above.

test_conformance/math_brute_force/CMakeLists.txt
```diff
@@ -44,5 +44,9 @@ set(${MODULE_NAME}_SOURCES
 # math_brute_force compiles cleanly with -Wall (except for a few remaining
 # warnings), but other tests not (yet); so enable -Wall locally.
 set_gnulike_module_compile_flags("-Wall -Wno-strict-aliasing -Wno-unknown-pragmas")
+if(${CLConform_TARGET_ARCH} STREQUAL "ARM64")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ffp-contract=off")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ffp-contract=off")
+endif()
```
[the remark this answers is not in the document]
nikhiljnv: Updated PR.

test_conformance/math_brute_force/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ set(${MODULE_NAME}_SOURCES
 
 # math_brute_force compiles cleanly with -Wall (except for a few remaining
 # warnings), but other tests not (yet); so enable -Wall locally.
-set_gnulike_module_compile_flags("-Wall -Wno-strict-aliasing -Wno-unknown-pragmas")
+set_gnulike_module_compile_flags("-Wall -Wno-strict-aliasing -Wno-unknown-pragmas -ffp-contract=off")
```
bashbaug, line 46: Was this intended to be added in two places?  I think the line below should be sufficient on its own.

```suggestion
set_gnulike_module_compile_flags("-Wall -Wno-strict-aliasing -Wno-unknown-pragmas")
```

test_conformance/math_brute_force/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ set(${MODULE_NAME}_SOURCES
...
+
+add_cxx_flag_if_supported("-ffp-contract=off")
```
bashbaug, line 48: Minor, but if you're fixing the issue above, I'd consider fixing this too: the other uses of add_cxx_flag_if_supported do not use quotes:

```suggestion
add_cxx_flag_if_supported(-ffp-contract=off)
```

bashbaug (commented):
Thanks, just a few things to check.

bashbaug:
Closing - unneeded after #1824.

PR #1806, opened 2023-09-04
Test IMAGE1D_BUFFER in more scenario

rjodinchr:
add `IMAGE1D_BUFFER` test case:
- cl_copy_images:
    - `IMAGE1D_BUFFER` to `IMAGE1D_BUFFER`
    - `IMAGE1D_BUFFER` to `IMAGE1D`
    - `IMAGE1D` to `IMAGE1D_BUFFER`
- cl_get_info
- cl_fill_image
- cl_read_write_image
- kernel_image_methods

Most of the tests are strongly inspired from the `IMAGE1D` tests already present.
Most significant changes are:
- Using the correct limit to choose the size (using `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE`)
- Creating the buffer needed for `IMAGE1D_BUFFER`
- Add `IMAGE1D_BUFFER` in many switch/if statements in common code to support the new tests.

I have made those tests because working on implementing 1dbuffer supports in clvk shows that lots of cases were not covered by the OpenCL-CTS to truly validate the implementation.
Ref kpet/clvk#403


test_conformance/images/clFillImage/test_loops.cpp
```diff
@@ -62,6 +67,12 @@ int test_image_type( cl_device_id device, cl_context context, cl_command_queue q
             break;
         default: log_error("Unhandled method\n"); return -1;
     }
+    else if (testMethod == k1DBuffer)
```
lakshmih, line 70: This is failing compilation since there's no 'if' to match the 'else if'
Also, from reviewing the code it looks like k1DBuffer would hit the default and error out at line 68?

test_conformance/images/clCopyImage/test_loops.cpp
```diff
@@ -88,6 +100,21 @@ int test_image_type( cl_device_id device, cl_context context, cl_command_queue q
             imageType = CL_MEM_OBJECT_IMAGE3D;
             break;
     }
+    else if (testMethod == k1DBuffer)
```
lakshmih, line 103: The switch case above doesn't have a default so this is failing compilation with "enumeration values 'k1DBuffer', 'k1DTo1DBuffer', and 'k1DBufferTo1D' not handled in switch "
And here again, not seeing an if to match the else if cascade so might be better to move everything inside the switch?

test_conformance/images/clFillImage/test_loops.cpp
bashbaug: I'm seeing the same build errors.

rjodinchr: Sorry about that, it was a rebase error on my side.

test_conformance/images/clCopyImage/test_loops.cpp
rjodinchr: Sorry about that, it was a rebase error on my side.

rjodinchr:
About the code formatting issue. I will leave the PR as it is formatted to ease the review. Once we agree on it, I'll fix those errors.

lakshmih:
Need some more time to review this.

lakshmih:
Still reviewing/testing.

test_conformance/images/clGetInfo/test_1D_buffer.cpp
```diff
@@ -0,0 +1,174 @@
...
+    image_descriptor imageInfo = { 0 };
+    RandomSeed seed(gRandomSeed);
+    size_t pixelSize;
+
+    cl_mem_flags all_host_ptr_flags[5] = { flags, CL_MEM_ALLOC_HOST_PTR | flags,
+                                           CL_MEM_COPY_HOST_PTR | flags,
+                                           CL_MEM_ALLOC_HOST_PTR
+                                               | CL_MEM_COPY_HOST_PTR | flags,
+                                           CL_MEM_USE_HOST_PTR | flags };
+
```
lakshmih, line 39: As per the spec section on creating images (USE_|ALLOC_|COPY_)_HOST_PTR flags are not valid for use when creating IMAGE1D_BUFFER objects so they might have to be handled differently in these tests. 

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_creating_image_objects

rjodinchr: I have updated the test accordingly

test_common/harness/imageHelpers.cpp
```diff
@@ -796,7 +796,8 @@ void get_max_sizes(
 
     (*numberOfSizes) = 0;
 
-    if (image_type == CL_MEM_OBJECT_IMAGE1D)
+    if (image_type == CL_MEM_OBJECT_IMAGE1D
+        || image_type == CL_MEM_OBJECT_IMAGE1D_BUFFER)
```
lakshmih, line 800: It looks like ln:806 isn't taking into account pixel size and is assigning the entire maxWidth value as the maxSize for the image 1D buffer.

test_conformance/images/clReadWriteImage/test_read_1D_buffer.cpp
```diff
@@ -0,0 +1,327 @@
...
+    }
+
+    BufferOwningPtr<char> resultValues(malloc(fullImageSize));
+    size_t imgValMipLevelOffset = 0;
+
+    for (size_t lod = 0; (gTestMipmaps && lod < imageInfo->num_mip_levels)
+         || (!gTestMipmaps && lod < 1);
+         lod++)
+    {
+        origin[1] = lod;
```
lakshmih, line 103: This should be 0 for 1D image or 1D image buffer objects.  

" If image is a 1D image or 1D image buffer object, origin[1] and origin[2] must be 0."

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueReadImage

rjodinchr: I don't know much about `gTestMipmaps` and `lod`. This is coming from `test_read_1D.cpp`, which is breaking the specification as you pointed out. Should I just get rid of the for loop and replace all `lod` with `0`?

lakshmih: From section 29 of the extension spec, 1D image buffers don't support mipmaps. So we should remove this testing entirely for this object type:
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_mipmap_image

"Mip-mapped images cannot be created for CL_MEM_OBJECT_IMAGE1D_BUFFER images, depth images or multi-sampled (i.e. msaa) images."

test_common/harness/imageHelpers.cpp
```diff
@@ -796,7 +796,8 @@ void get_max_sizes(
...
     {
 
         size_t M = maximum_sizes[0];
```
lakshmih, line 803: M should be adjusted to be within the max_pixels calculated allow to account for max allocation size.
For other object types, max_pixels factors into the math for other dimensions (after the max is locked) but there's no adjustment for the 1D cases.

rjodinchr: Done

test_conformance/images/clReadWriteImage/test_read_1D_buffer.cpp
rjodinchr: Done

test_common/harness/imageHelpers.cpp
```diff
@@ -801,6 +801,9 @@ void get_max_sizes(
...
+        size_t A = max_pixels;
+
+        M = static_cast<size_t>(fmin(A / M, M));
```
lakshmih, line 806: This can cause M to be 0 because of the division.

test_conformance/images/clReadWriteImage/test_read_1D_buffer.cpp
```diff
@@ -0,0 +1,256 @@
...
+    {
+        log_error("ERROR: Unable to create 1D image buffer of size %d (%s)",
+                  (int)imageInfo->width, IGetErrorString(error));
+        return -1;
+    }
+
+    if (gDebugTrace) log_info(" - Writing image...\n");
+
+    size_t origin[3] = { 0, 0, 0 };
+    size_t region[3] = { 0, 1, 1 };
```
lakshmih, line 59: Values in region cannot be 0 as per spec:

"region defines the (width, height, depth) in pixels of the 1D, 2D or 3D rectangle, the (width, height) in pixels of the 2D rectangle and the number of images of a 2D image array or the (width) in pixels of the 1D rectangle and the number of images of a 1D image array. If image is a 2D image object, region[2] must be 1. If image is a 1D image or 1D image buffer object, region[1] and region[2] must be 1. If image is a 1D image array object, region[2] must be 1. The values in region cannot be 0."

lakshmih:
We are still reviewing and testing the latest PS

test_conformance/images/clReadWriteImage/test_read_1D_buffer.cpp
rjodinchr: Good catch, thanks.
This is coming from the refactoring when removing mipmap code from this file.

test_common/harness/imageHelpers.cpp
rjodinchr: As `x0` can be 0 for `IMAGE1D_ARRAY` and `IMAGE2D`, and `x0` and `x1` can be 0 for `IMAGE2D_ARRAY` and `IMAGE3D`.
I'll add a `fmax` with `1` for this case that I'm modifying to make sure we never end up with `0`.

test_conformance/images/clCopyImage/test_copy_1D_buffer.cpp
```diff
@@ -0,0 +1,516 @@
...
+    }
+
+    imageInfo.format = format;
+    imageInfo.height = imageInfo.depth = imageInfo.arraySize =
+        imageInfo.slicePitch = 0;
+    imageInfo.type = CL_MEM_OBJECT_IMAGE1D_BUFFER;
+    pixelSize = get_pixel_size(imageInfo.format);
+
+    int error = clGetDeviceInfo(device, CL_DEVICE_IMAGE2D_MAX_WIDTH,
+                                sizeof(maxWidth), &maxWidth, NULL);
```
bashbaug, line 242: Can you check if this is correct?  It's not clear to me if this value is used to create a 1D image, or a 1D image from a buffer, or both.  If it's used to create a 1D image from a buffer then should we be querying [CL_DEVICE_IMAGE_MAX_BUFFER_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_IMAGE_MAX_BUFFER_SIZE) instead?  Or, if it's both, should we be querying both and using the smallest value?

bashbaug (approved):
LGTM, I've run all of these tests on an Intel GPU and they're passing, and I took a quick look through the code and everything looks in order.

There is a LOT of duplication among these tests, which makes them easier to review, but we might want to refactor to reduce the code duplication at some point (not in this PR!).

test_conformance/images/clCopyImage/test_copy_1D_buffer.cpp
rjodinchr: It is used to create 1D images. We could query both to get the smallest, but it would mean we expect to have an implementation with 1D image from buffer smaller than standard 1D image. Is that really possible?

bashbaug:
Discussed in the February 13th teleconference.  Waiting one more week for final reviews.

nikhiljnv:
We are seeing some issues on NVIDIA implementation with the changes proposed in this PR.
Will update in a couple of days once we dig into this more.

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -70,6 +74,26 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
+                if (err != CL_SUCCESS)
+                {
+                    log_error("ERROR: Could not create buffer for 1D buffer "
+                              "image. %ld bytes\n",
+                              imageInfo->rowPitch);
+                    return NULL;
+                }
+                imageDesc.buffer = buffer;
+            }
+            break;
```
lakshmih, line 96: This code doesn't seem to be handling the usePitches case for 1D image buffer correctly. The spec requires that the buffer be created with use_host_ptr and a host_ptr but the image object inherit those from the buffer, meaning the host_ptr in that case should be NULL. 
<The CL_MEM_USE_HOST_PTR, CL_MEM_ALLOC_HOST_PTR and CL_MEM_COPY_HOST_PTR values cannot be specified in flags but are inherited from the corresponding memory access qualifiers associated with mem_object>

lakshmih (changes requested):

rjodinchr:
> > We are seeing some issues on NVIDIA implementation with the changes proposed in this PR. Will update in a couple of days once we dig into this more.
> 
> We have fixed the issue on our side. One small change needed here to add image_1d_buffer case.
> 
> https://github.com/KhronosGroup/OpenCL-CTS/blob/e0a31a03fc8f816d59fd8b3051ac6a61d3fa50c6/test_conformance/images/clCopyImage/test_copy_generic.cpp#L445

I don't understand, I believe this PR already has the `CL_MEM_OBJECT_IMAGE1D_BUFFER` case in the switch you mentioned. But the link you gave is not the one from this PR, thus that's why it is missing.

nikhiljnv:
> > > We are seeing some issues on NVIDIA implementation with the changes proposed in this PR. Will update in a couple of days once we dig into this more.
> > 
> > 
> > We have fixed the issue on our side. One small change needed here to add image_1d_buffer case.
> > https://github.com/KhronosGroup/OpenCL-CTS/blob/e0a31a03fc8f816d59fd8b3051ac6a61d3fa50c6/test_conformance/images/clCopyImage/test_copy_generic.cpp#L445
> 
> I don't understand, I believe this PR already has the `CL_MEM_OBJECT_IMAGE1D_BUFFER` case in the switch you mentioned. But the link you gave is not the one from this PR, thus that's why it is missing.

I think you are referring to https://github.com/KhronosGroup/OpenCL-CTS/blob/9c6c16609b079f89bd045acae559bd0006c72748/test_conformance/images/clCopyImage/test_copy_generic.cpp#L458 from this PR, whereas the one I was referring to is here in this PR https://github.com/KhronosGroup/OpenCL-CTS/blob/9c6c16609b079f89bd045acae559bd0006c72748/test_conformance/images/clCopyImage/test_copy_generic.cpp#L597

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -78,12 +78,16 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
-            if (gEnablePitch) host_ptr = malloc(imageInfo->rowPitch);
             {
                 cl_int err;
-                cl_mem buffer =
-                    clCreateBuffer(context, CL_MEM_READ_WRITE,
-                                   imageInfo->rowPitch, host_ptr, &err);
+                cl_mem_flags buffer_flags = CL_MEM_READ_WRITE;
+                if (gEnablePitch)
+                {
+                    host_ptr = malloc(imageInfo->rowPitch);
```
lakshmih, line 86: As per section 5.3.1.2. Image Descriptor this host_ptr needs to be aligned with [CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT]

If the buffer object specified by mem_object was created with [CL_MEM_USE_HOST_PTR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_USE_HOST_PTR), the host_ptr specified to [clCreateBuffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateBuffer) or [clCreateBufferWithProperties](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateBufferWithProperties) must be aligned to the maximum of the [CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT) value for all devices in the context associated with the buffer specified by mem_object that support images

lakshmih (changes requested):

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -70,6 +74,70 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
+                        clGetCommandQueueInfo(queue, CL_QUEUE_DEVICE,
+                                              sizeof(device), &device, nullptr);
+                    if (err != CL_SUCCESS)
+                    {
+                        log_error(
+                            "Error: Could not get CL_QUEUE_DEVICE from queue");
+                        return NULL;
+                    }
+                    char major_version;
+                    err = clGetDeviceInfo(device, CL_DEVICE_VERSION,
```
lakshmih, line 97: This needs to be CL_DEVICE_NUMERIC_VERSION and the check on line 106 should change to:  'if (CL_VERSION_MAJOR(version) == 1)'

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -101,6 +181,21 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
             free(host_ptr);
```
lakshmih, line 181: Should the free change to align_free to match the malloc?

lakshmih (commented):
Thank you for addressing comments so far, just a couple more questions.

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -83,27 +103,7 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
-                                          sizeof(major_version), &major_version,
-                                          nullptr);
-                    if (err != CL_SUCCESS)
-                    {
-                        log_error("Error: Could not get CL_DEVICE_VERSION from "
-                                  "device");
-                        return NULL;
-                    }
-                    if (major_version == '1')
+                    if (CL_VERSION_MAJOR(version) == 1)
```
lakshmih, line 106: - If major version is not 1 and image type is other than IMAGE1D_BUFFER this code will call align_free instead of free

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -42,6 +42,26 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
+            return NULL;
+        }
+        err = clGetDeviceInfo(device, CL_DEVICE_NUMERIC_VERSION,
+                              sizeof(version), &version, nullptr);
+        if (err != CL_SUCCESS)
+        {
+            log_error("Error: Could not get CL_DEVICE_VERSION from "
+                      "device");
+            return NULL;
+        }
```
lakshmih, line 62: minor: Error log shows CL_DEVICE_VERSION instead of CL_DEVICE_NUMERIC_VERSION.

lakshmih (commented):
Thanks for the changes, just some more minor comments.

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -79,26 +161,63 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
+                            nullptr, error);
+    }
 
     if (gEnablePitch)
     {
-        if ( *error == CL_SUCCESS )
+        struct pitch_buffer_data *data = (struct pitch_buffer_data *)malloc(
+            sizeof(struct pitch_buffer_data));
+        data->buf = host_ptr;
+        data->is_aligned = CL_VERSION_MAJOR(version) == 1
```
lakshmih, line 186: We believe this needs to be:

-        data->is_aligned = CL_VERSION_MAJOR(version) == 1
-            && imageInfo->type == CL_MEM_OBJECT_IMAGE1D_BUFFER;
+        data->is_aligned =  !((CL_VERSION_MAJOR(version) == 1)
+            || (imageInfo->type != CL_MEM_OBJECT_IMAGE1D_BUFFER));

lakshmih (commented):
Hopefully the last one to fix and we should be good. Thank you for the incremental changes!

test_conformance/images/clFillImage/test_fill_generic.cpp
rjodinchr:
```
data->is_aligned = !((CL_VERSION_MAJOR(version) == 1)
|| (imageInfo->type != CL_MEM_OBJECT_IMAGE1D_BUFFER));
```
does not match 
> Actual: if version != 1 and type is 1Dbuffer then buffer is aligned

I think `version != 1 && type is 1DBuffer` is correct. I'll push that modification.

lakshmih (approved):

bashbaug:

test_conformance/images/kernel_image_methods/test_1D_buffer.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    // Generate some data to test against
+    BufferOwningPtr<char> imageValues;
+    generate_random_image_data(imageInfo, imageValues, d);
+
+    // Construct testing source
+    if (gDebugTrace)
+        log_info(" - Creating 1D image %d ...\n", (int)imageInfo->width);
+
+    buffer = clCreateBuffer(
```
ahesham-arm, line 70: The size passed to `clCreateBuffer` and the one in the call to `log_error` are different. The correct value that should be passed to both is `imageInfo->rowPitch`.

rjodinchr: Shouldn't `rowPitch` always equal `width * pixel_size`?

But I agree, it would be better to remove the use of `imageInfo->width * get_pixel_size(imageInfo->format)` in favour of `imageInfo->rowPitch`.

ahesham-arm: It should, but before the call to this function there is:
`imageInfo.rowPitch += extraWidth;`

rjodinchr: Then the caller might also need to be fixed either way, right?

ahesham-arm: Sorry, I didn't dig into what the test is trying to do but my general comment is that whatever we pass in as the image row pitch for creating the image, should be the same value we use for creating the buffer (size of the buffer is required to be >= row pitch in this case), and the same value that we log.

test_conformance/images/kernel_image_methods/test_1D_buffer.cpp
```diff
@@ -0,0 +1,282 @@
...
+        for (int i = 0; i < NUM_IMAGE_ITERATIONS; i++)
+        {
+            cl_ulong size;
+            // Loop until we get a size that a) will fit in the max alloc size
+            // and b) that an allocation of that image, the result array, plus
+            // offset arrays, will fit in the global ram space
+            do
+            {
+                imageInfo.width =
+                    (size_t)random_log_in_range(16, (int)maxWidth / 32, seed);
```
qichaogu, line 255: Integer overflow here: `maxWidth` is truncated from `size_t` to `int`.
It will cause problems:
`(int)maxWidth` may be a negative integer. As a result, the function `random_log_in_range` will have a second parameter which is negative so it will yield a negative result integer. Then this negative result integer will be converted to `size_t` and `imageInfo.width` will be a very large integer. Finally, the test will fail when trying to allocating memory of a very big size calculated from  `imageInfo.width`.

@rjodinchr Could you please fix the issue?

rjodinchr: `random_log_in_range` is used more than 200 times in the CTS, with this same pattern many times. It feels like a proper fix would involve changing that function and making sure everything is still working as intended. I'm sorry to say I will  not have time for that in the short term.

PR #1808, opened 2023-09-07
math_brute_force: don't set/restore FTZ mode twice

svenvh:
The suite's `main()` function already disables the FTZ mode prior to invoking `runTestHarnessWithCheck` and restores the FP state afterwards, so tests don't have to do so themselves.


lakshmih (approved):

bashbaug (approved):

bashbaug:

PR #1811, opened 2023-09-14
basic: fix more unused-but-set variables

lakshmih:
Removed the num_printed variable and replaced asserts on status with error-checking


svenvh (approved):

EwanC (approved):

lakshmih:
@bashbaug  can this be merged since it has two approvals. It is breaking release builds right now for us.

bashbaug (commented):
I think this change is generally OK, but just to be sure I understand: the issue is variables that are assigned to but that are only used in asserts, which triggers an unused variable warning, which causes a build failure due to warnings-as-errors?

It's a little unfortunate to lose otherwise useful asserts, but using asserts like this isn't best practice in the CTS tests, so I mostly just want to understand exactly what's happening.

I also have one specific question, see below.  Thanks!

bashbaug (approved):
Thanks!

I'm fine merging this as-is especially if unblocks you, but I was curious how the "Determine the max global id" code was used because it doesn't seem to be determining the maximum global ID, and it turns out it's dead code and `l_max_global_id0` is never used. If we don't want to remove it in this PR, could we remove it in a subsequent PR, or at least file an issue so we don't forget to clean this up in the future?

svenvh (approved):

lakshmih:
Ben, can we merge this, I'll push a follow up change.

PR #1812, opened 2023-09-18
Update test_compile.cpp

aharon-abramson:
Prevent the error:
"variable 'temp' is uninitialized when used here"

Since one of the generated kernels starts with:

__kernel void sample_test(__global float *src, __global int *dst) {
    float temp;
    int  tid = get_global_id(0);
    dst[tid] = dst[tid] + temp;

### DIFF
--- a/test_conformance/compiler/test_compile.cpp
+++ b/test_conformance/compiler/test_compile.cpp
@@ -28,7 +28,7 @@
 const char *sample_kernel_start =
     "__kernel void sample_test(__global float *src, __global int *dst)\n"
     "{\n"
-    "    float temp = 0.0;\n"
+    "    float temp = 0.0f;\n"
     "    int  tid = get_global_id(0);\n";
 
 const char *sample_kernel_end = "}\n";
--- a/test_conformance/compiler/test_compile.cpp
+++ b/test_conformance/compiler/test_compile.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
LGTM - I'm guessing the first commit has the only real change and the second is just formatting updates?

test_conformance/compiler/test_compile.cpp
```diff
@@ -16,148 +16,147 @@
...
+#define MAX_LOG_SIZE_IN_PROGRAM 2048
 
 const char *sample_kernel_start =
-"__kernel void sample_test(__global float *src, __global int *dst)\n"
-"{\n"
-"    float temp;\n"
-"    int  tid = get_global_id(0);\n";
+    "__kernel void sample_test(__global float *src, __global int *dst)\n"
+    "{\n"
+    "    float temp = 0.0;\n"
```
bashbaug, line 31: Actually one small nitpick - this should probably be "0.0f" to avoid issues on devices that do not support fp64.

bashbaug (commented):
LGTM - I'm guessing the first commit has the only real change and the second is just formatting updates?

test_conformance/compiler/test_compile.cpp
aharon-abramson: Sure, of course.

bashbaug (approved):

bashbaug:
Waiting one more week for additional review.

lakshmih (approved):
Looks fine, not sure we need to put the kernel code snippet in the commit message.

aharon-abramson:
Thank you, Lakshmi. @bashbaug, can you please merge this?

bashbaug:
Merging as discussed in the October 10th teleconference.

PR #1815, opened 2023-09-20
gl: fix sometimes-uninitialized warning

svenvh:
Bail out when hitting the default case, so that we don't attempt to access the uninitialized `error` variable.


EwanC (approved):

bashbaug (approved):
No reason to wait to merge this.

PR #1816, opened 2023-09-21
c11_atomics: align signedness of host types

svenvh:
Some `HOST_` types were declared with signedness different from the `HOST_ATOMIC_` counterparts, leading to sign-compare warnings when comparing between types.  Fix by aligning the signedness.


bashbaug (approved):
LGTM.

It would be nice to find a way to make this more robust in the future, but that would be a much larger change.

bashbaug:
Merging as discussed on the November 14th teleconference.

PR #1817, opened 2023-09-21
select: fix -Wformat warnings

svenvh:
Fix the following issues in the select tests:
 * Printing of a `size_t` requires the `%z` specifier.
 * Printing of `cl_long`/`cl_ulong` is now done using the `PRI*64` macros to ensure portability across 32 and 64-bit builds.


bashbaug (approved):
Thanks, LGTM!

EwanC (approved):

PR #1822, opened 2023-09-28
Semaphore types bug fixes revised

joshqti:
This PR fixes a number of issues, all of which we need to fix to fully pass.

- Detect semaphore handle types that are supported by both OpenCL and Vulkan implementations, using API queries. Test all of the detected handle types.
- Update the clExternalSemaphore class to handle signaling/waiting for SYNC_FD semaphore handle type.

- Delete obsolete semaphore test for double signal/wait. As per the current version of the spec, this behavior is undefined and cannot be tested.
- Delete inapplicable clCreateImageWithProperties test cases not related to the semaphore or external memory extensions.
- Fix error handling.  The multi-import tests, and possibly others did not fail when the test result was incorrect
- Delete dead code in the multi-import external memory tests related to "offset".  clExternalMemory had an unused parameter called "offset", deleted all test code directly related to that parameter.  Offset is no longer part of the spec.

Joshua Kelly, Qualcomm


test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -1688,108 +1545,99 @@ int test_buffer_common(cl_device_id device_, cl_context context_,
...
         {
-            print_error(errNum,
-                        "Error in clGetDeviceInfo for getting device_extension "
-                        "size....\n");
-            goto CLEANUP;
+            log_info(
+                "Device %u does not support cl_khr_external_memory_opaque_fd, "
+                "which is required for this test. Skipping\n",
+                device_no);
+            continue;
```
nikhiljnv, line 1583: What if cl_khr_external_memory_opaque_fd is not present, but some other handle type extension is present (e.g. cl_khr_external_memory_win32)? Consider either removing the check or adding checks for other handle types too.

test_conformance/vulkan/test_vulkan_interop_image.cpp
```diff
@@ -1360,61 +1365,73 @@ int test_image_common(cl_device_id device_, cl_context context_,
...
+        if (!is_extension_available(devices[device_no],
+                                    "cl_khr_external_memory_opaque_fd"))
         {
-            print_error(err, "clGetDeviceInfo failed with error");
-            goto CLEANUP;
+            log_info(
+                "Device %u does not support cl_khr_external_memory_opaque_fd, "
+                "which is required for this test. Skipping\n",
+                device_no);
+            continue;
```
nikhiljnv, line 1380: What if cl_khr_external_memory_opaque_fd is not present, but some other handle type extension is present (e.g. cl_khr_external_memory_win32)? Consider either removing the check or adding checks for other handle types too.

nikhiljnv (changes requested):
Some early comments.

test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
```diff
@@ -236,30 +237,119 @@ getSupportedVulkanExternalMemoryHandleTypeList()
...
+    // allow NT macro on Windows 8 and later
 #if _WIN32
     if (IsWindows8OrGreater())
     {
-        externalSemaphoreHandleTypeList.push_back(
-            VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_WIN32_NT);
+        all_known_handle_types.push_back(
+            { "VK_KHR_external_semaphore_win32",
+              VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_WIN32_BIT_KHR,
+              VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_WIN32_NT });
```
nikhiljnv, line 272: Is there any reason why we push opaque_fd, sync_fd and win32 KMT unconditionally, but need to push win32 NT conditionally? If we are anyway going to check against supported handle types later and this is merely a reference for all known handle types, don't see a need to check for Windows version condition here.

test_conformance/vulkan/test_vulkan_interop_image.cpp
```diff
@@ -1299,38 +1313,29 @@ int test_image_common(cl_device_id device_, cl_context context_,
     cl_kernel kernel_unsigned[num_kernels] = { NULL, NULL, NULL, NULL };
```
nikhiljnv, line 1313: Since you are at it, can you also initialize program array above to NULL for the sake of consistency and completeness?

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
```diff
@@ -104,48 +114,60 @@ int test_external_semaphores_queries(cl_device_id deviceID, cl_context context,
...
 
-    // Confirm that querying CL_SEMAPHORE_CONTEXT_KHR returns the right context
-    SEMAPHORE_PARAM_TEST(CL_SEMAPHORE_CONTEXT_KHR, cl_context, context);
+        SEMAPHORE_PARAM_TEST(CL_DEVICE_HANDLE_LIST_KHR, cl_device_id, deviceID);
 
-    // Confirm that querying CL_SEMAPHORE_REFERENCE_COUNT_KHR returns the right
-    // value
-    SEMAPHORE_PARAM_TEST(CL_SEMAPHORE_REFERENCE_COUNT_KHR, cl_uint, 1);
+        SEMAPHORE_PARAM_TEST(
+            CL_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR, cl_uint,
```
nikhiljnv, line 145: A couple of things -
1. If we are creating semaphore by importing it from Vulkan, export handle types list can be NULL/absent. So, this place doesn't seem about right for me to check for exportable handle types query.
2. The return type for CL_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR query is cl_external_semaphore_handle_type_khr[], not cl_uint.

nikhiljnv (changes requested):
Some more comments.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
joshqti: Agreed.  This isn't new code for my change.  Lets raise an issue to ensure we properly test this query, I would rather not add a new subtest in this change.

test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
joshqti: The original structure of the code used conditional compilation, I tried to maintain the original structure, as I do not understand or have the ability to test windows handles.  For example, we still have this code in the test from the original PR:
`
const std::vector<VulkanExternalMemoryHandleType>
getSupportedVulkanExternalMemoryHandleTypeList()
{
    std::vector<VulkanExternalMemoryHandleType> externalMemoryHandleTypeList;

#if _WIN32
    if (IsWindows8OrGreater())
    {
        externalMemoryHandleTypeList.push_back(
            VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_NT);
    }
    externalMemoryHandleTypeList.push_back(
        VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_KMT);
#else
    externalMemoryHandleTypeList.push_back(
        VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD);
#endif

    return externalMemoryHandleTypeList;
}
`

I was unclear on how to query Vulkan for the appropriate handle type.  For example,  the NT enum is a bitwise or of Vulkan flags:

`
VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_NT_KMT =
        VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT_KHR
        | VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_KMT_BIT_KHR
`

I am not sure how to interpret this, in the context of querying support from Vulkan for semaphore or memory handles.

Can you advise?

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
joshqti: Done

test_conformance/vulkan/test_vulkan_interop_image.cpp
joshqti: Done

test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
nikhiljnv: Since the test was iterating over all handle types pushed here, it made sense to push them conditionally here earlier so that we don't accidentally test for unsupported handle type based on OS version. Now that we anyway cross check this against supported handle types and use this just as reference of all known handle types, I think we can clean this up and push the types unconditionally.

joshqti:
Pushed patch to address comments

nikhiljnv (approved):
LGTM.

nikhiljnv:
Merging as agreed in Memory subgroup call on November 28th, 2023.

PR #1827, opened 2023-10-17
conversions: Build fix for x86 toolchain

lakshmih:
We haven't tested this since we don't support the data type but the build fix is needed to build for x86 Windows.


bashbaug (approved):

Issue #1818, opened 2023-09-22
Vulkan Interop tests have hardcoded workgroup dimensions


ahesham-arm:
`run_test_with_one_queue` and `run_test_with_two_queue` under the Vulkan interop tests have hardcoded workgroup dimensions that are passed to `vkCmdDispatch`. The hardcoded values are 512, 1, 1. The test does not query the device to check for the maximum supported workgroup dimensions first, which causes it to fail on older devices that support a maximum size < 512.

nikhiljnv:
@ahesham-arm I think we can close this as we have already merged the required changes as part of #1828 ?

PR #1828, opened 2023-10-18
Reduce the Vulkan interop buffer kernel workgroup size to 256

ahesham-arm:
Using `local_size_x = 512` in the shader used in the Vulkan interop tests fails on older hardware because this exceeds the maximum barrier size limit.

Changed to 256, which is still a multiple of 32 and 64, and is below the hardware limitations. The tests are not performance tests anyways, so the performance penalty, if any, is acceptable.

This fixes issue #1818


nikhiljnv (approved):
LGTM.

bashbaug (approved):
This change is fine and more portable than it was previously, but if I'm reading the Vulkan 1.3 spec right the required minimum value for `maxComputeWorkGroupInvocations` is 128.  Would it be most portable to reduce the work-group size to 128, rather than 256?

nikhiljnv (approved):

nikhiljnv:
Merging as discussed on Memory subgroup call of November 14, 2023.

PR #1837, opened 2023-10-26
add MIPS architecture

aharon-abramson:


bashbaug (approved):
LTGM.

All added blocks for `__mips__` are at the end of an `#if` chain, so this shouldn't affect any of the existing builds.

PR #1840, opened 2023-11-04
a few fixes (thread safety & cl_khr_command_buffer UB)

franz:
some fixes we've been carrying in our CTS fork:

* fix UB in `command_buffer_event_sync.cpp`: enqueue of two commands in two separate queues, with both using the same buffer argument, and no synchronization between the commands.
* fix UB in `command_buffer_test_barrier.cpp`: missing synchronization between zeroing command and command-buffer using two separate queues
* make `test_thread_dimensions.cpp` thread-safe to avoid spurious errors.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
@@ -583,7 +583,7 @@ struct CommandBufferEventSync : public BasicCommandBufferTest
 
         // process secondary queue
         error =
-            clEnqueueFillBuffer(queue_sec, in_mem, &pattern_pri, sizeof(cl_int),
+            clEnqueueFillBuffer(queue_sec, in_mem, &pattern_sec, sizeof(cl_int),
                                 0, data_size(), 0, nullptr, nullptr);
         test_error(error, "clEnqueueFillBuffer failed");
 
@@ -593,8 +593,9 @@ struct CommandBufferEventSync : public BasicCommandBufferTest
                    "clEnqueueCommandBufferKHR in secondary queue failed");
 
         // process primary queue
-        error = clEnqueueFillBuffer(queue, in_mem, &pattern_pri, sizeof(cl_int),
-                                    0, data_size(), 0, nullptr, event_ptrs[0]);
+        error =
+            clEnqueueFillBuffer(queue, in_mem, &pattern_pri, sizeof(cl_int), 0,
+                                data_size(), 1, &test_event, event_ptrs[0]);
         test_error(error, "clEnqueueFillBuffer failed");
 
         cl_event wait_list[] = { test_event,
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_barrier.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_barrier.cpp
@@ -93,6 +93,9 @@ struct BarrierWithWaitListKHR : public BasicCommandBufferTest
                                 0, data_size(), 0, nullptr, nullptr);
         test_error(error, "clEnqueueFillBufferKHR failed");
 
+        error = clFinish(queue);
+        test_error(error, "clFinish");
+
         error = clEnqueueCommandBufferKHR(
             0, nullptr, out_of_order_command_buffer, 0, nullptr, &event);
         test_error(error, "clEnqueueCommandBufferKHR failed");
--- a/test_conformance/thread_dimensions/test_thread_dimensions.cpp
+++ b/test_conformance/thread_dimensions/test_thread_dimensions.cpp
@@ -245,10 +245,8 @@ static const char *thread_dimension_kernel_code_not_atomic_not_long =
     "\n"
     "}\n";
 
-char dim_str[128];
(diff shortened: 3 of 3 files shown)


test_conformance/subgroups/subhelpers.h
```diff
@@ -31,7 +31,7 @@
 #define NR_OF_ACTIVE_WORK_ITEMS 4
 
 extern MTdata gMTdata;
-typedef std::bitset<128> bs128;
+typedef std::bitset<1280> bs128;
```
svenvh, line 34: Looking at the discussion in https://github.com/KhronosGroup/OpenCL-CTS/pull/1621#issuecomment-1429974965 the preferred fix seems to be to limit the local work-group sizes of the various non-uniform subgroup tests.

franz: Thanks for pointing out, i will remove that commit.

franz:
Removed the first two commits from the PR; i did not realize there was already an open PR with those.

franz:
rebased on master. should be ready for review.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
```diff
@@ -570,13 +572,13 @@ struct CommandBufferEventSync : public BasicCommandBufferTest
         test_error(error, "clCreateCommandBufferKHR failed");
 
         // record secondary command buffer
-        error = RecordCommandBuffer(command_buffer_sec, kernel);
+        error = RecordCommandBuffer(command_buffer_sec, kernel_sec);
```
bashbaug, line 575: I don't think this is necessarily wrong, but I'm having a tough time figuring out what this test is doing now.

As far as I can tell, for setup:

1. There are two queues (`queue` and `queue_sec`), two command buffers (`command_buffer` and `command_buffer_sec`), two kernels (`kernel` and `kernel_sec`), and two sets of buffers used for kernel inputs and outputs (`in_mem`, `out_mem`, `off_mem`, `in_mem_sec`, `out_mem_sec`, and `off_mem_sec`).
    * Note: `queue_sec` is exclusively in-order.  `queue` may be out-of-order. 
2. `in_mem`, `out_mem`, and `off_mem` are set as kernel arguments for `kernel` and recorded into `command_buffer`.
3. `in_mem_sec`, `out_mem_sec`, and `off_mem_sec` are set as kernel arguments for `kernel_sec` and recorded into `command_buffer_sec`.
4. `command_buffer` records `kernel`, and `command_buffer_sec` records `kernel_sec`.

Then, when the test executes:

5. In `queue_sec`, we initialize `in_mem_sec`, then enqueue `command_buffer_sec`.  We save the event for `command_buffer_sec`.
    * Remember: `command_buffer_sec` recorded `kernel_sec`, which reads `in_mem_sec` and writes `out_mem_sec`.
6. In `queue`, we initialize `in_mem`, then enqueue `command_buffer`, with an event dependency on the initialization and the execution of `command_buffer_sec` in `queue_sec`.  We save the event for `command_buffer`.
    * Remember: `command_buffer` recorded `kernel`, which reads `in_mem` and writes `out_mem`.

To verify the test output:

7. We enqueue a read from `out_mem` with an event dependency on the execution of `command_buffer`.
8. We flush `queue`, then finish `queue_sec`, then finish `queue`.
9. We check the results that we read.

This is all nominally fine, but we never check the results of `out_mem_sec` or anything that executes in `command_buffer_sec`, really.  Is that what is intended?

Another possible solution that wouldn't require a second kernel and a second set of memory objects is to have the initialization of `in_mem` in step (6) depend on the execution of the command buffer in step (5).  This ensures that `in_mem` will not be overwritten while the command buffer is executing.  I'm not sure if that's any better, though...

bashbaug (commented):
Note, this needs another merge/rebase now, to pick up the latest header changes.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_event_sync.cpp
franz: I've looked it up, this test was added in https://github.com/KhronosGroup/OpenCL-CTS/issues/1369, and i think this is meant to be the "Test command-buffer enqueue with an event returned by another queue in the wait-list." -test.

So AFAICT the point of the test is to check the correct order of execution of two events (with command buffers) from different command queues. You have a point and i think your solution makes the test a little more readable. I've updated the commit to reflect this and rebased on main branch.

EwanC: > So AFAICT the point of the test is to check the correct order of execution of two events (with command buffers) from different command queues. 

Yes, I think reflects the intention. Although it looks like from Ben's description of what's happening that there isn't a shared data dependency between the 2 command-buffers. i.e. if each command-bufffer uses it's own kernels and cl_mem buffers then if the ordering was not enforced, would it still lead to incorrect results. If there is a data dependency then we'd only need to validate the results of the final execution (although checking the results of earlier ones wouldn't hurt). Not sure if it's best to file an Issue to refactor this test rather than extending the scope of this PR.

EwanC (approved):

bashbaug:

PR #1850, opened 2023-11-24
Add NULL CL_MEM_HOST_PTR check for all flag combinations

### LINKED ISSUE #1752: Value returned for `CL_MEM_HOST_PTR` not validated when `CL_MEM_USE_HOST_PTR` not present in flags

niranjanjoshi121:
As part of suggestions to #1801, add NULL CL_MEM_HOST_PTR
 check for all flag combinations.

 Fixes #1752


bashbaug (commented):
The change itself looks good to me, though we should close on https://github.com/KhronosGroup/OpenCL-Docs/issues/1008 before merging.

cdai2:
Verified on Intel CPU device.

bashbaug (approved):

bashbaug:
Merging as discussed in the March 12th teleconference.

PR #1852, opened 2023-11-29
Increase the number of work items used by the allocations test

kpet:
The test uses a fixed number of work items to process very large buffers and images. As devices support more and more memory, this leads to an ever-increasing amount of work done in each work item. This results in some implementations and devices hitting timeout issues.

Furthermore, a greater number of work items can provide performance benefits on some devices.

Long term, this test should be redesigned to scale the number of threads dynamically as a function of the max allocation size.

### DIFF
--- a/test_conformance/allocations/allocation_execute.cpp
+++ b/test_conformance/allocations/allocation_execute.cpp
@@ -140,7 +140,7 @@ int check_image(cl_command_queue queue, cl_mem mem) {
 }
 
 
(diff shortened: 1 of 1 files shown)


kpet:
The issue this PR helps with has been seen on the following implementations:
- Arm Mali driver
- clvk on top of NVidia's proprietary Vulkan driver.

Jeremy-Kemp (approved):

kpet:
Merging as agreed in the 2023/12/05 teleconference.

PR #1853, opened 2023-12-01
c11_atomics: Reduce the loopcounter for sequential consistency tests

lakshmih:
Reduce the loop from 1000000 to 500000 since the former value makes the test run too long and cause system issues on certain platforms


bashbaug (approved):
I'm fine with this.

At some point it would be nice to define these limits with a single numeric constant so it would be easier to make the counter smaller (or larger?) on other devices, as needed.

bashbaug:
Merging as discussed in the December 12th teleconference.

It would be good to create an issue to determine what the appropriate value for this counter is.  Could it be made even smaller?

PR #1854, opened 2023-12-02
Updated semaphore tests to use clSemaphoreReImportSyncFdKHR.

joshqti:
Additionally updated common semaphore code to handle spec updates that restrict simultaneous importing/exporting of handles.

See docs issue #939

### DIFF
--- a/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
+++ b/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
@@ -757,7 +757,7 @@ int clExternalSemaphore::wait(cl_command_queue cmd_queue)
     return CL_INVALID_OPERATION;
 }
 
-clExternalSemaphore::~clExternalSemaphore() noexcept(false) = default;
+clExternalSemaphore::~clExternalSemaphore() = default;
 
 clExternalImportableSemaphore::clExternalImportableSemaphore(
     const VulkanSemaphore &semaphore, cl_context context,
@@ -851,7 +851,7 @@ clExternalImportableSemaphore::clExternalImportableSemaphore(
     }
 }
 
-clExternalImportableSemaphore::~clExternalImportableSemaphore() noexcept(false)
+clExternalImportableSemaphore::~clExternalImportableSemaphore()
 {
     cl_int err = clReleaseSemaphoreKHRptr(m_externalSemaphore);
     if (err != CL_SUCCESS)
@@ -927,7 +927,7 @@ clExternalExportableSemaphore::clExternalExportableSemaphore(
     }
 }
 
-clExternalExportableSemaphore::~clExternalExportableSemaphore() noexcept(false)
+clExternalExportableSemaphore::~clExternalExportableSemaphore()
 {
     cl_int err = clReleaseSemaphoreKHRptr(m_externalSemaphore);
     if (err != CL_SUCCESS)
--- a/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
+++ b/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
@@ -36,7 +36,8 @@ pfnclReleaseSemaphoreKHR clReleaseSemaphoreKHRptr;
 pfnclGetSemaphoreHandleForTypeKHR clGetSemaphoreHandleForTypeKHRptr;
 pfnclReImportSemaphoreSyncFdKHR clReImportSemaphoreSyncFdKHRptr;
 
-void init_cl_vk_ext(cl_platform_id opencl_platform)
+void init_cl_vk_ext(cl_platform_id opencl_platform, cl_uint num_devices,
+                    cl_device_id *deviceIds)
 {
     clEnqueueWaitSemaphoresKHRptr =
         (pfnclEnqueueWaitSemaphoresKHR)clGetExtensionFunctionAddressForPlatform(
@@ -81,13 +82,19 @@ void init_cl_vk_ext(cl_platform_id opencl_platform)
                                  "clGetSemaphoreHandleForTypeKHRptr!");
     }
 
+    // Required only if cl_khr_external_semaphore_sync_fd is reported
     clReImportSemaphoreSyncFdKHRptr = (pfnclReImportSemaphoreSyncFdKHR)
         clGetExtensionFunctionAddressForPlatform(
             opencl_platform, "clReImportSemaphoreSyncFdKHR");
-    if (NULL == clReImportSemaphoreSyncFdKHRptr)
+    for (cl_uint i = 0; i < num_devices; i++)
     {
-        throw std::runtime_error("Failed to get the function pointer of "
-                                 "clReImportSemaphoreSyncFdKHRptr!");
+        if (is_extension_available(deviceIds[i],
+                                   "cl_khr_external_semaphore_sync_fd")
+            && (NULL == clReImportSemaphoreSyncFdKHRptr))
+        {
+            throw std::runtime_error("Failed to get the function pointer of "
+                                     "clReImportSemaphoreSyncFdKHR!");
+        }
     }
 }
 
--- a/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
+++ b/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
@@ -757,6 +757,8 @@ int clExternalSemaphore::wait(cl_command_queue cmd_queue)
     return CL_INVALID_OPERATION;
 }
 
+clExternalSemaphore::~clExternalSemaphore() noexcept(false) = default;
+
 clExternalImportableSemaphore::clExternalImportableSemaphore(
     const VulkanSemaphore &semaphore, cl_context context,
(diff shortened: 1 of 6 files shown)


test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -79,6 +80,15 @@ void init_cl_vk_ext(cl_platform_id opencl_platform)
         throw std::runtime_error("Failed to get the function pointer of "
                                  "clGetSemaphoreHandleForTypeKHRptr!");
     }
+
+    clReImportSemaphoreSyncFdKHRptr = (pfnclReImportSemaphoreSyncFdKHR)
```
nikhiljnv, line 84: We should check if cl_khr_external_semaphore_sync_fd extension is supported or not before querying/calling this API.

nikhiljnv (changes requested):

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
joshqti: That sounds correct.  I will make the change.

paulfradgley (dismissed):
These changes look fine to me.

joshqti:
How do we merge with the failing CI on mac?

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -846,7 +867,83 @@ clExternalSemaphore::~clExternalSemaphore() noexcept(false)
     }
 }
 
-int clExternalSemaphore::signal(cl_command_queue cmd_queue)
+int clExternalImportableSemaphore::wait(cl_command_queue cmd_queue)
```
nikhiljnv, line 870: We can now wait on ImportableSemaphore, but need to be able to signal too for other handle types.
Can we extend these to support signal operations?

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -846,7 +867,83 @@ clExternalSemaphore::~clExternalSemaphore() noexcept(false)
...
+{
+    cl_int err = clReleaseSemaphoreKHRptr(m_externalSemaphore);
+    if (err != CL_SUCCESS)
+    {
+        throw std::runtime_error("clReleaseSemaphoreKHR failed!");
+    }
+}
+
+int clExternalExportableSemaphore::signal(cl_command_queue cmd_queue)
 {
```
nikhiljnv, line 947: We can now signal on ExportableSemaphore, but need to be able to wait too for other handle types.
Can we extend these to support wait operations?

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -886,60 +983,7 @@ int clExternalSemaphore::signal(cl_command_queue cmd_queue)
     return err;
 }
 
-int clExternalSemaphore::wait(cl_command_queue cmd_queue)
-{
```
nikhiljnv, line 890: Any reason we need to remove wait and signal operations on base class clExternalSemaphore?

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -742,9 +742,9 @@ int run_test_with_multi_import_same_ctx(
     }
     else
     {
-        clVk2CLExternalSemaphore = new clExternalSemaphore(
+        clVk2CLExternalSemaphore = new clExternalImportableSemaphore(
             vkVk2CLSemaphore, context, vkExternalSemaphoreHandleType, deviceId);
-        clCl2VkExternalSemaphore = new clExternalSemaphore(
+        clCl2VkExternalSemaphore = new clExternalExportableSemaphore(
             vkCl2VkSemaphore, context, vkExternalSemaphoreHandleType, deviceId);
```
nikhiljnv, line 748: This doesn't necessarily be an exportable semaphore for other handle types. Pretty much all semaphores can be created by Vulkan, imported by OpenCL and can be wait on or signaled from either APIs.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -742,9 +742,9 @@ int run_test_with_multi_import_same_ctx(
     }
     else
     {
-        clVk2CLExternalSemaphore = new clExternalSemaphore(
+        clVk2CLExternalSemaphore = new clExternalImportableSemaphore(
             vkVk2CLSemaphore, context, vkExternalSemaphoreHandleType, deviceId);
```
nikhiljnv, line 746: This doesn't necessarily be an importable semaphore. Pretty much all semaphores can be created by OpenCL, imported by Vulkan and can be wait on or signaled from either APIs.

nikhiljnv (changes requested):
Left some comments.

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
joshqti: The choice to not implement both is not related to handle types. The unimplemented methods are never used by any test.

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
joshqti: The choice to not implement both is not related to handle types. The unimplemented methods are never used by any test.

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
joshqti: clExternalSemaphore is an abstract class, wait and signal need to be implemented in clExternalExportableSemaphore or clExternalImportableSemaphore.  The class is called c**lExternal**Semaphore, meaning it is used for communicating with external APIs.  There are only two ways to do that, import or export.  You could implement a semaphore that does neither import or export, but putting that functionality in a class named clExternalSemaphore doesn't make sense.  The spec prohibit simultaneous import/export, which is why clExternalSemaphore got split into two classes.  It improves readability and simplifies the constructors, which would otherwise need to handle both import and export.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
joshqti: It does need to be exportable.  It is passed a Vulkan semaphore in the function signature, and is then signaled on from OpenCL later.  The original code takes a Vulkan semaphore, so it always did an import or export.  The spec now restricts simultaneous importing and export, so all external semaphores must be importable or exportable.  There is no other option.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
joshqti: It does need to be importable.  It is passed a Vulkan semaphore in the function signature, and is then waited on from OpenCL later.

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
nikhiljnv: Unimplemented methods are never used now because clVk2CLExternalSemaphore is made clExternalImportableSemaphore object and clCl2VkExternalSemaphore is made clExternalExportableSemaphore object compared to earlier when both were just plain clExternalSemaphore objects where both wait and signal would have been used. If exporting semaphores is not supported on some platform, there is no way to run the tests in the current form compared to earlier when they could have run with just imported semaphores.

joshqti: Is there a concrete example, or more of a future thing?  I can implement these method, but unless we add a test they will not get used.  Implementing them was not a spec choice, just a choice not to write unused code.  

It is functionally the same as the old system, except it reflects the fact that semaphores cannot be both importable and exportable.  This was necessary, the other option would have been to add a new parameter to clExternalSemaphore() called import_export_mode or something, otherwise the constructor would have no idea what to do with the vulkan semaphore being passed in.

joshqti:
I pushed a change that I believe reflects your comments.  Can we discuss further architectural changes in a follow-up issue?

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -1071,17 +1071,17 @@ int run_test_with_multi_import_diff_ctx(
...
-        clCl2VkExternalSemaphore = new clExternalSemaphore(
+        clCl2VkExternalSemaphore = new clExternalExportableSemaphore(
             vkCl2VkSemaphore, context, vkExternalSemaphoreHandleType, deviceId);
```
nikhiljnv, line 1077: Are we checking somewhere for supported exportable handle types before creating exportable semaphore?
If exportable handle types is empty, may be we should consider falling back to importable semaphore for this semaphore too.

nikhiljnv:
Merging as agreed in memory subgroup call on January 23, 2024.

PR #1855, opened 2023-12-04
geometrics: fix Wsign-compare warnings

svenvh:
Fix all instances of Wsign-compare warnings in the geometrics test, and re-enable the warning so it can catch any new occurrences.


bashbaug (approved):
Thanks, LGTM!

PR #1856, opened 2023-12-04
thread_dimensions: Avoid combinations of very small LWS and very larg…

lakshmih:
…e GWS

Modify the existing condition to include extremely small LWS like 1x1 on large GWS values


bashbaug (approved):
This does reduce test coverage slightly, but the small local work sizes are still tested when the global work size is also reasonably small.

PR #1857, opened 2023-12-08
allocations: Move results array from stack to heap

lakshmih:
Fixes breakage due to #1852

### DIFF
--- a/test_conformance/allocations/allocation_execute.cpp
+++ b/test_conformance/allocations/allocation_execute.cpp
@@ -239,7 +239,9 @@ int execute_kernel(cl_context context, cl_command_queue *queue, cl_device_id dev
     }
 
     // Set the result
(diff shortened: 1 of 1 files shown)


test_conformance/allocations/allocation_execute.cpp
```diff
@@ -167,6 +167,10 @@ int execute_kernel(cl_context context, cl_command_queue *queue, cl_device_id dev
     access_string[0] = '\0';
     kernel_string[0] = '\0';
 
+    cl_uint *returned_results =
+        (cl_uint *)malloc(sizeof(cl_uint) * NUM_OF_WORK_ITEMS);
```
svenvh, line 171: This introduces a memory leak on the various error paths.  Suggesting to use a `std::vector` instead, so that no manual `free` is needed.

lakshmih: Fixed, thanks.

svenvh (approved):

bashbaug (approved):
LGTM.

I seem to remember another large stack array that was causing problems previously but I can't find it after a quick search.  Is there some way to find if we have other large stack arrays that should be converted to heap allocations?

bashbaug:
Merging as discussed in the December 12th teleconference.

PR #1858, opened 2023-12-13
Enable LARGEADDRESSAWARE for 32 bit compilation

### LINKED ISSUE #1842: Random failure for Basic/imagedim_non_pow2 test case on Windows 32bit
OS cannot allocate a contiguous memory space (~256MB) when the test tried to create large 2D images intensively (the free virtual memory is running low, ~300MB available, however, might be fragmented). 

Because the 32 bit test executable is not linked with "LARGEADDRESSAWARE" so only 2GB virtual memory address space is available for the test – which might not be sufficient. If this "LARGEADDRESSAWARE" link option is used, test case passes.

Nuullll:
32-bit executables built with MSVC linker have only 2GB virtual memory address space by default, which might not be sufficient for some tests.

Enable LARGEADDRESSAWARE linker flag for 32-bit targets to allow tests to handle addresses larger than 2 gigabytes.
Fixes #1842

https://learn.microsoft.com/en-us/cpp/build/reference/largeaddressaware-handle-large-addresses?view=msvc-170

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,6 +157,11 @@ if( WIN32 AND "${CMAKE_CXX_COMPILER_ID}" MATCHES "Intel" )
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Qlong-double -Qpc80 /DWIN32 /D_WINDOWS /W3 /GR /EHsc -nologo -Od -D_CRT_NONSTDC_NO_WARNINGS -EHsc -Wall -Qdiag-disable:68,111,177,186,161,869,1028,2259,2553,181,239,265,1188 -fp:strict -fp:source")
 endif()
 
+# To handle addresses larger than 2 gigabytes for 32bit targets
+if( WIN32 AND "${CMAKE_EXE_LINKER_FLAGS}" MATCHES "/machine:X86")
+    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} /LARGEADDRESSAWARE")
+endif()
+
 list(APPEND CLConform_LIBRARIES ${OPENCL_LIBRARIES})
 if(ANDROID)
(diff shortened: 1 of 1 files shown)


cdai2:
@bashbaug  please review.

CMakeLists.txt
```diff
@@ -157,6 +157,11 @@ if( WIN32 AND "${CMAKE_CXX_COMPILER_ID}" MATCHES "Intel" )
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Qlong-double -Qpc80 /DWIN32 /D_WINDOWS /W3 /GR /EHsc -nologo -Od -D_CRT_NONSTDC_NO_WARNINGS -EHsc -Wall -Qdiag-disable:68,111,177,186,161,869,1028,2259,2553,181,239,265,1188 -fp:strict -fp:source")
 endif()
 
+# To handle addresses larger than 2 gigabytes for 32bit targets
+if( WIN32 AND "${CMAKE_EXE_LINKER_FLAGS}" MATCHES "/machine:X86")
```
Nuullll, line 161: Didn't use the condition suggested [here](https://github.com/KhronosGroup/OpenCL-CTS/issues/1842#issuecomment-1804680022)

> if(MSVC AND ${CLConform_TARGET_ARCH} STREQUAL "x86")

1. Users could use a different compiler (e.g. ICX) with MSVC linker
2. Users could cross-compile 32-bit executables on a 64-bit machine (where `${CLConform_TARGET_ARCH} == "x86_64"`)

Nuullll: Thanks @bashbaug for the explanation. It makes sense!

bashbaug (approved):
Thanks, latest version LGTM.

lakshmih (approved):

bashbaug:
Merging as discussed in the January 9th teleconference.

PR #1860, opened 2023-12-15
integer_ops: fix -Wformat warnings

svenvh:
The main sources of warnings were:

 * Printing of a `size_t` which requires the `%zu` specifier.

 * Printing of `cl_long`/`cl_ulong` which is now done using the `PRI*64` macros to ensure portability across 32 and 64-bit builds.

I've split out the clang-formatting of my changes into a separate commit to ease reviewing.

### DIFF
--- a/test_conformance/integer_ops/test_abs.cpp
+++ b/test_conformance/integer_ops/test_abs.cpp
@@ -15,6 +15,8 @@
 //
 #include "harness/compat.h"
 
+#include <cinttypes>
+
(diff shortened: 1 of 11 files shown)


svenvh:
The macos build failure looks unrelated:
```
FAILED: test_conformance/spir/bin/test_spir
...
Undefined symbols for architecture x86_64:
  "reference_add(double, double)", referenced from:
      _functionList in function_list.cpp.o
```

Issue #1639, opened 2023-02-14
Targeted Testing for SPIR-V Versions Greater Than 1.0


bashbaug:
As far as I can tell all of the "spirv_new" tests are currently testing only SPIR-V 1.0 because the assemble_spirv.py script currently passes `--target-env spv1.0` to `spirv-as` (this is probably OK because all of the .spvasm files also say `Version: 1.0`).  I'd like to add targeted testing for newer SPIR-V versions though and I am looking for input how best to do this.

Some options:

1. Parse the .spvasm file to extract the version and pass it to `spirv-as`.
2. Encode the SPIR-V version in the filename somehow, extract the SPIR-V version from the filename and pass it to `spirv-as`.
3. Put all of the .spvasm files for a specific version in a specific directory, pass the version from directory name to `spirv-as`.

Any preference?  Is there another option we should consider?  Thanks!

bashbaug:
I have a POC that does this.  I've gone with (3) for now.

bashbaug:
The overall mechanism for this issue was fixed by #1861, so I am closing it.

I will file separate issues to track testing for specific SPIR-V versions greater than 1.0.

PR #1861, opened 2023-12-16
add SPIR-V version testing

### LINKED ISSUE #1069: testing supported SPIR-V / IL versions
As best I can tell we do not currently test that all reported SPIR-V versions (via `CL_DEVICE_IL_VERSION` or `CL_DEVICE_ILS_WITH_VERSION`) can be used.  At least for the reported SPIR-V versions we should be able to verify this.

Rough test idea:

* Loop over all reported IL versions.
* For non-SPIR-V ILs, do nothing (maybe emit a warning about a skipped IL).
* For each SPIR-V IL:
    * Generate a SPIR-V module with the requested version.
    * Compile the generated SPIR-V module.  Verify that compilation was successful.
    * Execute the generated SPIR-V module.  Verify that execution was successful.  This should require only very light testing - maybe one work-item.

I don't think we'd include targeted testing for each of the "newer" SPIR-V capabilities as part of this test.  The testing for each SPIR-V capability should be done separately and purposefully.  This testing would only be for verifying each SPIR-V version.

bashbaug:
This PR adds basic testing for each of the SPIR-V versions supported by an OpenCL device.  It queries which SPIR-V versions are supported (using `CL_DEVICE_IL_VERSION`), and then compiles a SPIR-V program and executes a simple kernel for each of the reported versions.  Note that none of the new features for each SPIR-V version are tested, but this PR does include infrastructure that could be (and hopefully will be!) used to add targeted testing for version-specific SPIR-V features in the future.

fixes #1069 
see also #1639


haonanya:
@bashbaug, could you please offer run command for added tests? I am not familiar with CTS and having a try to verify it on intel OCL CPURT. Thanks very much.

[haonanya@ spirv_new]$  ./test_spirv_new basic_versions
Reading SPIR-V files from default 'spirv_bin' path.
In case you want to set other directory use '--spirv-binaries-path' argument.
To skip the SPIR-V version check use the '--skip-spirv-version-check' argument.
Compute Device Name = Intel(R) FPGA Emulation Device, Compute Device Vendor = Intel(R) Corporation, Compute Device Version = OpenCL 1.2 , CL C Version = OpenCL C 1.2
sizeof( void*) = 8  (host)
sizeof( void*) = 8  (device)
This device supports SPIR-V offline compilation. SPIR-V version is 1.4
basic_versions...
    testing SPIR-V_1.0...
File spirv_bin//basic.spv64 not found
File /basic not found
ERROR: Unable to build SPIR-V program! (CL_DEVICE_NOT_FOUND from /export/users/haonanya/OpenCLCTS/OpenCL-CTS/test_conformance/spirv_new/test_basic_versions.cpp:92)
basic_versions FAILED
PASSED sub-test.
FAILED test.

bashbaug:
Sure - I always forget how to do this so I have a text file with my notes.  Here's what I do:

Step 1: Build the spirv_new tests:

    make -j20 test_spirv_new

Step 2: Assemble SPIR-V files:
This needs to be done once, or whenever SPIR-V files are added.

    pushd ../test_conformance/spirv_new/ && python3 ./assemble_spirv.py && popd

Step 3: Run the tests:
Be sure to point to the SPIR-V binaries!

    ./test_conformance/spirv_new/test_spirv_new --spirv-binaries-path ../test_conformance/spirv_new/spirv_bin/

bashbaug:
I'd like to add more testing for newer SPIR-V features and I'm going to need some of this infrastructure, so adding "focused review".

haonanya:
> basic_versions

Thank you @bashbaug! Test pass with intel OCL CPURT. Is supported SPIRV-V version is up to 1.4 according to https://github.com/KhronosGroup/SPIRV-LLVM-Translator/blob/main/docs/SPIRVVersionsAndExtensionsHandling.rst?plain=1#L68? Thanks very much.

svenvh (approved):
LGTM

PR #1863, opened 2023-12-18
Fp16 math bruteforce staging

shajder:
Introducing this PR as an intermediate step to merge work on `math_brute_force` to support `cl_khr_fp16` - this is a direct continuation of #1681


bashbaug (approved):
Merging into staging branch, as we discussed.

PR #1864, opened 2023-12-18
Fp16 conversions staging

shajder:
Introducing this PR as an intermediate step to merge work on `conversions` tests to support `cl_khr_fp16` - this is a direct continuation of #1669


bashbaug (approved):
Merging into staging branch, as we discussed.

PR #1865, opened 2023-12-18
fix the license header on the spirv-new tests

### LINKED ISSUE #1471: fix copyright header comments in spirv_new
There are several files in the spirv_new test that appear to be using an older and incorrect copyright header comment:

```
/******************************************************************
Copyright (c) 2016 The Khronos Group Inc. All Rights Reserved.

This code is protected by copyright laws and contains material proprietary to the Khronos Group, Inc.
This is UNPUBLISHED PROPRIETARY SOURCE CODE that may not be disclosed in whole or in part to
third parties, and may not be reproduced, republished, distributed, transmitted, displayed,
broadcast or otherwise exploited in any manner without the express prior written permission
of Khronos Group. The receipt or possession of this code does not convey any rights to reproduce,
disclose, or distribute its contents, or to manufacture, use, or sell anything that it may describe,
in whole or in part other than under the terms of the Khronos Adopters Agreement
or Khronos Conformance Test Source License Agreement as executed between Khronos and the recipient.
******************************************************************/
```

These should be updated to use the proper Apache 2 header comment instead.

bashbaug:
fixes #1471

The source files for the spirv-new tests were using the older Khronos license instead of the proper Apache-2.0 license.  This PR fixes the license header in all source files.

Note, I removed the license header from CMakeLists.txt instead of replacing it because none of the other CMakeLists.txt seemed to have a license header.  If we add a license header to CMakeLists.txt we should add it to all of them.

No functional changes.

### DIFF
--- a/test_common/harness/crc32.h
+++ b/test_common/harness/crc32.h
@@ -1,19 +1,18 @@
-/******************************************************************
-Copyright (c) 2016 The Khronos Group Inc.
-All Rights Reserved.  This code is protected by copyright laws and
-contains material proprietary to the Khronos Group, Inc.
-This is UNPUBLISHED PROPRIETARY SOURCE CODE that may not be disclosed
-in whole or in part to third parties, and may not be reproduced, republished,
-distributed, transmitted, displayed, broadcast or otherwise exploited in any
-manner without the express prior written permission of Khronos Group.
-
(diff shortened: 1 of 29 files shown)


svenvh (approved):

bashbaug:

kpet (approved):

PR #1866, opened 2023-12-18
fix macos builds by avoiding double compilation of function_list.cpp for test_spir

### LINKED ISSUE #1862: macos builds failing after cmake upgrade
The macos presubmit check started failing to build the `test_spir` binary, see e.g. #1860 and #1861 . It turns out there was an update of the runner image (https://github.com/actions/runner-images/pull/9012) which among other things updated cmake from 3.27.8 to 3.28.0.

```
FAILED: test_conformance/spir/bin/test_spir
...
Undefined symbols for architecture x86_64:
  "reference_add(double, double)", referenced from:
      _functionList in function_list.cpp.o
```

This linking error seems to suggest that the `-DFUNCTION_LIST_ULPS_ONLY` preprocessor define wasn't propagated when compiling `function_list.cpp` (which comes from the math_brute_force suite).

bashbaug:
fixes #1862 

This PR modernizes the CMakeLists.txt for test_spir.  It uses target specific commands to add compile options and compile definitions vs. setting COMPILE_FLAGS and adding definitions globally.  This fixes the macos builds, which broke when the runner updated to a new version of CMake.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -38,11 +38,14 @@ jobs:
           sudo apt-get -y install libglu1-mesa-dev freeglut3-dev mesa-common-dev libglew-dev
       - name: Setup MSVC with Ninja
         uses: ilammy/msvc-dev-cmd@v1
+      - name: Get System Info
+        uses: kenchan0130/actions-system-info@master
+        id: system-info
       - name: Setup ccache
         uses: hendrikmuhs/ccache-action@v1.2
         with:
           variant: sccache
-          key: ${{ matrix.os }}-${{ matrix.arch }}
+          key: ${{ matrix.os }}-${{ steps.system-info.outputs.release }}-${{ matrix.arch }}
       - name: Fetch OpenCL Headers
         shell: bash
         run: |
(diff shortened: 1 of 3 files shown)


svenvh (approved):
LGTM, thanks for the fix!

svenvh:
Interestingly, the macos-latest job ran with `cmake version 3.27.7` this time, instead of the problematic `3.28.0`.

bashbaug:
> Interestingly, the macos-latest job ran with `cmake version 3.27.7` this time, instead of the problematic `3.28.0`.

Odd, I'm pretty sure the "older" use of CMake is causing the problem, but I don't have easy access to a Mac and I was counting on CI to tell me the issue was fixed or not.  I wonder if there is a way to trigger a build that will definitely use 3.28.0?  If not, I'll try to find a Mac and do the testing myself.

svenvh:
> I wonder if there is a way to trigger a build that will definitely use 3.28.0? If not, I'll try to find a Mac and do the testing myself.

It seems to be a matter of "luck" whether the runner on which the job lands has the latest image?  I've just rerun the macos-latest job for this PR, and this time it used cmake 3.28.0 and sadly failed like before your change.

bashbaug:
> I've just rerun the macos-latest job for this PR, and this time it used cmake 3.28.0 and sadly failed like before your change.

Darn, back to the drawing board.  I may need to dig up a mac to debug this locally.

bashbaug:
Removing "focused review".  I still think this is a good change, but there's no hurry for it if it is not fixing the CI issues.

svenvh (approved):

bashbaug:
Re-adding "focused review".  We really ought to merge this soon, since we've had failing CI for a while now.

bashbaug:

PR #1868, opened 2023-12-21
Fix -Wformat warnings in multiple tests

svenvh:
All of these warnings stem from printing `size_t` types, which should be done using the `z` length modifier.


bashbaug (approved):

bashbaug:
Merging as discussed in the February 27th teleconference.

PR #1871, opened 2023-12-28
api tests fix: Free MTdata object in test_get_image_info

### LINKED ISSUE #1870: Memory leak in api tests
Valgrind output generated by calling: 

`valgrind  --leak-check=full --show-leak-kinds=all --track-origins=yes --log-file=./api/valgrind-log-file.txt ./api/test_api`

reports the following memory leak:
image

imilenkovic00:
Add call of free_mtdata function in test_get_image_info. This fixes memory leak in api tests.

Fixes #1870.


test_conformance/api/test_mem_object_info.cpp
```diff
@@ -720,6 +720,8 @@ int test_get_image_info( cl_device_id deviceID, cl_context context, cl_mem_objec
         clReleaseMemObject( imageObject );
     }
 
+    free_mtdata(d);
```
svenvh, line 723: An even better solution would be to replace the declaration/initialization of `MTdata d` by `MTdataHolder`, so that no manual `free_mtdata` calls are needed.  That would also fix the leaks on error paths.

imilenkovic00: Thank you for your suggestion. Modified accordingly.

svenvh (approved):

bashbaug:
Merging as discussed in the January 30th teleconference.

PR #1872, opened 2024-01-03
Fix cts build error on mariner OS

jujiang-del:


bashbaug (commented):
This change looks reasonable but I'm having trouble understanding why the previous code was not OK.  Could you please provide more information about the build error?  Thanks!

jujiang-del:
> This change looks reasonable but I'm having trouble understanding why the previous code was not OK. Could you please provide more information about the build error? Thanks!

The error was"/test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp:429:37: error: writing 1 byte into a region of size 0 [-Werror=stringop-overflow=]
  429 |             pattern[pattern_length] = '\0';"

jujiang-del:
!verify

bashbaug:
Thanks, that's helpful.  I'm guessing you're building with some version of gcc?  Seems like this is a known false positive.

@svenvh, you've been most active eliminating warnings and enabling warnings as errors.  How would you like to resolve this?

It's interesting that almost the exact same pattern appears, here, but this doesn't seem to be causing a problem:

https://github.com/KhronosGroup/OpenCL-CTS/blob/27e47e9e878bb58cf1747f71e0813c07eab07f27/test_conformance/extensions/cl_khr_command_buffer/command_buffer_printf.cpp#L338-L342

svenvh:
I've seen various `Wstringop-overflow` false positives over time indeed, affecting only some gcc versions.  If this patch fixes the particular instance I'm okay with merging this.

svenvh:
Closing and reopening to retrigger the buildbots.

svenvh (approved):

PR #1873, opened 2024-01-07
fix return code when readwrite image is not supported

rjodinchr:
This function (do_test) starts by testing write and read individually. Both of them can have errors.

When readwrite image is not supported, the function returns TEST_SKIPPED_ITSELF potentially masking errors leading to the test returning EXIT_SUCCESS even with errors along the way.


bashbaug (approved):
LGTM.

If for some reason no read-only or write-only tests are run then I think this will report "test passed" rather than "test skipped itself", but that'd be a weird configuration and this doesn't seem to be a problem worth worrying about anyhow.  Thanks for the fix!

kpet (approved):

bashbaug:

PR #1874, opened 2024-01-10
bruteforce: Remove unnecessary half to float conversion

lakshmih:


bashbaug (approved):
LGTM.  `func.f_ff` retuns a `double` so it is incorrect to do any sort of "half to float" conversion on it.

I wonder if there's any way to catch errors like these in the future?  Passing a `double` or a `float` to `cl_half_to_float` will almost always be an error.

bashbaug:

PR #1877, opened 2024-01-15
compiler: fix grammar in error message

svenvh:


bashbaug (approved):
The grammar fix is fine.

This seems like a bit of a weird test though.  How could a feature macro be both defined and undefined at the same time?

svenvh:
> This seems like a bit of a weird test though. How could a feature macro be both defined and undefined at the same time?

Agreed. I suppose it is theoretically possible to implement features macros / preprocessor macros wrong, though that would probably lead to plenty of other test failures too.

Issue #1668, opened 2023-03-09
Negative testing for command-buffers


EwanC:
The https://github.com/KhronosGroup/OpenCL-CTS/issues/1369 test plan for [cl_khr_command_buffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_command_buffer) doesn't include any negative testing. However, there are many error codes defined by the extension, it would be nice to verify are reported correctly.

The majority of existing coverage in the CTS does not have negative testing, so this is a request for coverage over and above what we would normally do. However, negative testing does have advantages and we are thinking about adding it in other cases, see https://github.com/KhronosGroup/OpenCL-CTS/issues/516 

Note that if https://github.com/KhronosGroup/OpenCL-Docs/pull/885 gets merged it will mean that we won't need to do Invalid state checks.

kamil-goras-mobica:
Hi, regarding to Command Buffer extension from https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_command_buffer there is 18 functions to cover with tests. Total number of error codes is 108. My plan is to group functions by lets say "family" and possibly error codes. In the end number of tests will be smaller than 108. At this moment it will be about 80. Functions will be grouped like:
**clCreateCommandBufferKHR()** 
_10 error codes_
**clGetCommandBufferInfoKHR()**
_4 error codes_
**clRetainCommandBufferKHR()
clReleaseCommandBufferKHR()**
_3 error codes_
**clFinalizeCommandBufferKHR()**
_4 error codes_
**clEnqueueCommandBufferKHR()**
_13 error codes_
**clCommandNDRangeKernelKHR()**
_9 error codes_
**clCommandBarrierWithWaitListKHR()**
_8 error codes_
**clCommandCopyBufferKHR()
clCommandCopyBufferRectKHR()
clCommandCopyBufferToImageKHR()**
_6 error codes_
**clCommandCopyImageKHR()
clCommandCopyImageToBufferKHR()**
_6 error codes_
**clCommandFillBufferKHR()
clCommandFillImageKHR()**
_6 error codes_
**clCommandSVMMemcpyKHR()
clCommandSVMMemFillKHR()**
_6 error codes_
Thinking about merging **clCommandNDRangeKernelKHR()** and **clCommandBarrierWithWaitListKHR()** because error codes for them looks very similar.
[...]

PR #1879, opened 2024-01-16
Negative test for returning command-buffer mutable handle

EwanC:
See https://github.com/KhronosGroup/OpenCL-Docs/issues/1043 which clarifies that implementations are expected to return `CL_INVALID_VALUE` when an output command-handle is used without a layered extension supporting it.


EwanC:
Closing this PR as superseded by the work from https://github.com/KhronosGroup/OpenCL-CTS/issues/1668

PR #1881, opened 2024-01-22
conversions: Address subnormal test cases for half_float

lakshmih:


lakshmih:
@bashbaug  This fixes conversions for us.

lakshmih:
To pick up where we left off on this, 

the spec does say denorm support is optional (and hence denorms may be flushed to zero) for single precision numbers OpenCL_C section 7.2:

"Support for denormalized numbers with single precision floating-point is optional. Denormalized single precision floating-point numbers passed as input or produced as the output of single precision floating-point operations such as add, sub, mul, divide, and the functions defined in math functions, common functions, and geometric functions may be flushed to zero."

Above statement is not contingent on any compiler option, the compiler option only applies when denorms are supported anyway. 

There is no such statement in the extension spec for half precision but it does say in OpenCL_Ext section 5.2.12:

"The required minimum half precision floating-point capability as implemented by this extension is:
CL_FP_ROUND_TO_ZERO, or CL_FP_ROUND_TO_NEAREST | CL_FP_INF_NAN"

which seems to imply the same optionality. 
Will be nice to make it explicit in the extension spec so the flush to zero behavior is permitted.

test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -326,8 +327,10 @@ int CalcRefValsPat<InType, OutType, InFP, OutFP>::check_result(void *test,
         for (uint32_t i = 0; i < count; i++)
             if (t[i] != c[i] &&
                 // Allow nan's to be binary different
-                !((t[i] & 0x7fff) > 0x7C00 && (c[i] & 0x7fff) > 0x7C00)
-                && !(a[i] != (cl_uchar)0 && t[i] == (c[i] & 0x8000)))
+                !((t[i] & 0x7fff) > 0x7C00 && (c[i] & 0x7fff) > 0x7C00) &&
+                !(a[i] != (cl_uchar)0 && t[i] == (c[i] & 0x8000)) &&
+                // retry per section 7.5.3.2, with ftz on
+                !(IsHalfSubnormal(c[i]) && (HTF(t[i]) == 0.0f)))
```
bashbaug, line 333: The reasoning behind this change seems good but the fix seems slightly wrong.  In short, the `a` array is supposed to track whether the input may be flushed to zero or not, so this case _should_ be handled by the preceding line, which tests if the input may be flushed to zero (`a[i] != (cl_uchar)0`) and then if it actually is flushed to zero, with the proper sign (`t[i] == (c[i] & 0x8000)`).

It looks like there is a problem with the way the `a` array is setup though, for `half` inputs (or outputs).  Specifically, I think we need to define a specific `setAllowZ` for `half`, and call it here:

```c++
        if (gForceFTZ)
        {
            if (inType == kfloat || outType == kfloat)
                setAllowZ((uint8_t *)a, (uint32_t *)s, count);
        }
        if (gForceHalfFTZ)
        {
            if (inType == khalf || outType == khalf)
                setAllowZ((uint8_t *)a, (uint32_t *)s, count);   // <-- needs a half-specific version
        }
```

If this is done then I think this specific change can be reverted.  Can you give this a try and see if it fixes your issue?  Thanks!

bashbaug (changes requested):
I don't think this is quite the right fix - see below.

PR #1882, opened 2024-01-24
Fix testing of half-precision fma.

hvdijk:
Half-precision functions are generally tested against the single-precision reference. This causes double rounding: first to single precision, then from there to half precision. For the most part, it is good enough, but specifically in the case of fma, a correctly rounded result is required and is not obtained, for instance for arguments 0x1.eacp+7, 0x1.3f4p+4, 0x1.c04p+14, which produce an exact result of 0x1.065fffp+15 which should be rounded to half-prefcision 0x1.064p+15, but was previously first rounded to single-precision 0x1.066p+15, and from there to half-precision 0x1.068p+15. Testing against reference_fmal gives us sufficient precision that double rounding does not cause issues.

The f_fma(..., FLUSHED) calls for FTZ testing cannot be updated the same way but do not need to be: these calls all have at least one constant operand of zero. If one operand is zero, double rounding cannot be an issue.

### DIFF
--- a/test_common/harness/errorHelpers.cpp
+++ b/test_common/harness/errorHelpers.cpp
@@ -369,7 +369,7 @@ static float Ulp_Error_Half_Float(float test, double reference)
     return (float)scalbn(testVal - reference, ulp_exp);
 }
 
-float Ulp_Error_Half(cl_half test, float reference)
+float Ulp_Error_Half(cl_half test, double reference)
 {
     return Ulp_Error_Half_Float(cl_half_to_float(test), reference);
 }
--- a/test_common/harness/errorHelpers.h
+++ b/test_common/harness/errorHelpers.h
@@ -185,7 +185,7 @@ static int vlog_win32(const char *format, ...);
 
 extern const char *IGetErrorString(int clErrorCode);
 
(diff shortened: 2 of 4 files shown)


hvdijk:
> The f_fma(..., FLUSHED) calls for FTZ testing cannot be updated the same way but do not need to be: these calls all have at least one constant operand of zero. If one operand is zero, double rounding cannot be an issue.

I do think there is an issue there too, the FTZ handling appears to only check for results that are subnormal in single precision, rather than results that are subnormal in half precision, but that is separate from the issue that I am trying to fix here.

hvdijk:
The "Check code format" is failing because it does a `clang-format` of the diff to `main`, *not* of the diff to the PR's target branch. The errors are issues on the `fp16-staging` branch, not new issues in this PR, and I think I should ignore them?

bashbaug (approved):
This fixed our failure in the `fma` test - thanks!

@lakshmih fyi.

svenvh (approved):

lakshmih (approved):
This doesn't fix our issue with fma but it also doesn't break us so we're good with merging to the staging branch.

bashbaug:
Merging as discussed in the February 6th teleconference.

PR #1884, opened 2024-01-27
fix the reported platform information for math brute force

bashbaug:
When the math brute force test printed the platform version it always printed information for the first platform in the system, which could be different than the platform for the passed-in device.  Fixed by querying the platform from the passed-in device instead.


svenvh (approved):

bashbaug:
Merging as discussed in the January 30th teleconference.

PR #1885, opened 2024-01-31
fix freeglut reinitialization attempt in gl_sharing tests

karolherbst:
Fixes the following error:
`freeglut (./build/test_conformance/gl/test_gl): illegal glutInit() reinitialization attempt`

### DIFF
--- a/test_conformance/gl/main.cpp
+++ b/test_conformance/gl/main.cpp
@@ -250,19 +250,18 @@ int main(int argc, const char *argv[])
         return error;
     }
 
+    // At least one device supports CL-GL interop, so init the test.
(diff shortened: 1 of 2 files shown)


bashbaug (commented):
It looks like the Windows WGLEnvironment has a flag to track whether glut has been initialized or not.  Should we consider porting that solution to Linux instead?

karolherbst:
> It looks like the Windows WGLEnvironment has a flag to track whether glut has been initialized or not. Should we consider porting that solution to Linux instead?

not sure if that actually makes sense as freeglut needs to be initialized at this point anyway. Maybe I read the code wrong, but I think it's expected that some tests will be run at that point.

PR #1886, opened 2024-01-31
Fix cross-context semaphore signals

### LINKED ISSUE #1588: Improvements to the External  Memory and Semaphores conformance tests
1) External Memory Sharing
a)  Write a test that does external memory sharing only and is not dependent on semaphores being supported.
b) Confirm that EnqueueAcquire and EnqueueRelease calls are being made correctly for all tests.

2) Semaphores
 a) Do cross OpenCL context tests for multiple semaphore  handle types such as sync_fd and opaque_fd .
 b) Add a multiple signal - multiple wait test for in-order queues.
  c) Test the following queries - CL_DEVICE_HANDLE_LIST_KHR, CL_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR

joshqti:
Multi-context test was not correctly testing cross-context functionality. Use semaphore import/export mechanism to signal across contexts.  Also add missing acquire/release calls. Fixes outstanding #1588 issues.

### DIFF
--- a/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
+++ b/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
@@ -4,6 +4,7 @@
 #include "opencl_vulkan_wrapper.hpp"
 #include <thread>
 #include <chrono>
+#include <unordered_set>
 
 #define FLUSH_DELAY_S 5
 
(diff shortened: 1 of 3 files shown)


test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
```diff
@@ -174,104 +203,124 @@ int test_external_semaphores_multi_context(cl_device_id deviceID,
...
+    test_error(err, "Could not create command queue");
 
-        // Wait semaphore 2
-        clEventWrapper wait_2_event;
-        err =
-            clEnqueueWaitSemaphoresKHR(queue2, 1, &sema_ext_2.getCLSemaphore(),
-                                       nullptr, 0, nullptr, &wait_2_event);
-        test_error(err, "Could not wait semaphore");
+    if (import_export_handle_types.empty())
+    {
```
nikhiljnv, line 258: This doesn't seem to be a strict requirement. There are different configs in which we could test this scenario.
1. One OpenCL context exporting the semaphore and the other OpenCL context importing the semaphore. --> This would require both exportable and importable semaphores supported.
2. Use Vulkan as intermediate source for creating semaphores, two OpenCL contexts importing two semaphores from Vulkan and communicating across the two contexts using the imported semaphores.

SyncFD would require both exportable and importable semaphore support irrespective of above modes due to additional limitations to signal and wait.

joshqti:
Created new PR for acquire/release https://github.com/KhronosGroup/OpenCL-CTS/pull/1899

nikhiljnv (approved):
LGTM.

PR #1883, opened 2024-01-25
vulkan test coverage for 1d, 3d, 3dlist images

pkaur19:

PR #1887, opened 2024-02-02
Vulkan interop coverage

pkaur19:
Added Code changes for 1D, 1DList, 3D, 3DList and test cases.


nikhiljnv:
@joshqti Please, review this instead of #1883.

pkaur19 (commented):
Code Formatting Completed

test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
```diff
@@ -0,0 +1,206 @@
...
+                                          cl_command_queue _queue,
+                                          int num_elements)
+{
+    cl_int errNum;
+    VulkanDevice vkDevice;
+
+    // Context and command queue creation
+    cl_platform_id platform = NULL;
+    cl_context context = NULL;
+    cl_command_queue cmd_queue = NULL;
```
bashbaug, line 31: Can we use `clContextWrapper` and `clCommandQueueWrapper` instead?  Then we won't have to free these objects "manually".

test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    cl_context_properties contextProperties[] = { CL_CONTEXT_PLATFORM, 0, 0 };
+    errNum = clGetPlatformIDs(1, &platform, NULL);
+    test_error(errNum, "Failed to get platform id");
```
bashbaug, line 35: This makes it impossible to use any platform other than platform zero.  Can we get the platform from the passed-in device ID instead?  Or, even better, do we really need to create a new context and queue here, or can we just use the one that is passed in?

test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
```diff
@@ -0,0 +1,206 @@
...
+    const VulkanMemoryTypeList& memoryTypeList = vkImage1D.getMemoryTypeList();
+    uint64_t totalImageMemSize = vkImage1D.getSize();
+
+    log_info("Memory type index: %u\n", (uint32_t)memoryTypeList[0]);
+    log_info("Memory type property: %d\n",
+             memoryTypeList[0].getMemoryTypeProperty());
+    log_info("Image size : %lu\n", totalImageMemSize);
+
+    VulkanDeviceMemory* vkDeviceMem = new VulkanDeviceMemory(
+        vkDevice, vkImage1D, memoryTypeList[0], vkExternalMemoryHandleType);
```
bashbaug, line 88: I might just be missing it, but where is vkDeviceMem freed?

bashbaug (approved):
I took a look at these tests but I was unable to run them since they seem to be hard-coded to use the first platform on the system.  Do we have an issue already to track moving these tests to use the test harness?

On Linux we'll also need to switch to use the dma_buf extension vs. opaque_fds to test our implementation, but the lack of an ability to pick the platform to run on is the bigger issue for now.

Since this problem is preexisting it's no reason to hold up this PR, but it's really something we ought to fix.

joshqti:
I do not think this change alone is sufficient to fix 1834.  We are missing 2d arrays, and this test only covers creation of images.  We need some testing that uses 1d images and 3d images in a non-trivial way inside OpenCL.

joshqti:
I'm ok with merging this

bashbaug:
Merging as discussed in the May 28th memory TSG.  We will file an issue to fix the remaining PR comments.

PR #1891, opened 2024-02-08
add support for custom devices

aharon-abramson:
enable the CTS to run on custom devices


bashbaug (approved):
I doubt many of the CTS tests would be guaranteed to pass on a custom device, but I can't think of any reason the harness shouldn't allow the CTS tests to run on a custom device, either.

LGTM!

svenvh (approved):

bashbaug:
Merging as discussed in the February 13th teleconference.

Issue #1203, opened 2021-03-26
Subgroups tests - try using bit mask flip while bs128 initialization


gwawiork:
https://github.com/KhronosGroup/OpenCL-CTS/pull/1074#discussion_r554073452

PR #1893, opened 2024-02-12
[NFC] subgroups: clarify bs128 to uint4 conversion

svenvh:
There appears to have been some confusion over `uint_mask` during the initial reviews of the subgroup tests.  Add a comment to clarify the purpose of the mask.

Use a 32-bit all-ones mask (`0xffffffff`) instead of a 64-bit all-ones mask to further clarify the intent, that is, don't delay discarding the upper 32 bits until the assignment but mask them out immediately.

Fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/1203


StuartDBrady (approved):
Looks good to me.

StuartDBrady (approved):
Looks good to me.

PR #1894, opened 2024-02-13
commonfns:initialize the output buffer to zeros

banan328:
Initialize all output buffers to zero to ensure they start with a predictable state. This prevents potential undefined behavior. The reason for this initialization is that in all tests, the size of the buffers exceeds the number of threads in the kernel. Without initialization, the last portion of the output buffer might contain unpredictable data ("junk").


aharon-abramson (approved):

aharon-abramson:
In our testing environment, we check the bit-exactness of output buffers between our compiler and a reference one. Since our framework isn't aware of what part of the output buffer is actually written by a certain kernel, it compares the entire buffer; if that buffer isn't properly initialized and the kernel doesn't write all of it, some of its content is undefined.

banan328:
To continue the suggestion discussed in the meeting about initializing the buffer to zeroes during clCreateBuffer, I would like to express my concern that this approach does not fully address the issue of buffers potentially containing unpredictable data.
Currently, we are using the CL_MEM_COPY_HOST_PTR flag with clCreateBuffer, which copies data from a vector or pointer that might not be initialized.
 This means that even if we initialize the buffer to zeroes initially, the copied data may still be unpredictable.

bashbaug:
Discussed in the June 4th teleconference (I think, may have been the week prior).  Concerns are that this "buffers must be initialized" property is not enforced by the current CTS tests, so even if we merge this PR there is no guarantee that these buffers will continue to be initialized, nor is there a way to have confidence that there are not other tests that would need to be updated to initialize buffers.

Given these concerns, would a better solution be for whatever testing framework that is requiring bit exactness to do its own buffer intialization, vs. doing the buffer intialization explicitly in the tests?

Removing "focused review" while evaluating whether these changes are required.

PR #1895, opened 2024-02-17
Import Vulkan resources via named NT handle

saurabhnv:
Add coverage to import Vulkan resources (memory and semaphore) in Windows via named NT handles.
If no name is given during resource creation, then use NT handles for import.
If a name is given, have an option to either use that name or get the NT handle and use that for import.
Resolves KhronosGroup/OpenCL-Docs#943


test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -550,7 +550,7 @@ clExternalMemory::clExternalMemory(const clExternalMemory &externalMemory)
 clExternalMemory::clExternalMemory(
     const VulkanDeviceMemory *deviceMemory,
     VulkanExternalMemoryHandleType externalMemoryHandleType, uint64_t size,
-    cl_context context, cl_device_id deviceId)
+    cl_context context, cl_device_id deviceId, cl_bool useNameForImport)
```
joshqti, line 553: Why not add new enum?

saurabhnv: Yeah, looks like when I was writing this there was no CL_*_HANDLE_OPAQUE_WIN32_NAME_KHR, but since it is there now, will remove the added param and add them in VulkanExternalMemoryHandleType and VulkanExternalSemaphoreHandleType to use those instead.

joshqti (approved):
lgtm

PR #1898, opened 2024-02-21
Import OpenCL exported semaphore in Vulkan 

saurabhnv:
Except for SYNC_FD, current implementation doesn't import exported OpenCL semaphore in Vulkan and ends up doing signal and wait on essentially two unrelated semaphores (one created in OpenCL and one in Vulkan).
Since OpenCL exports the semaphore, import that in Vulkan to perform signal/wait on the same underlying payload.

### DIFF
--- a/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
+++ b/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
@@ -928,6 +928,67 @@ clExternalExportableSemaphore::clExternalExportableSemaphore(
         throw std::runtime_error(
             "clCreateSemaphoreWithPropertiesKHRptr failed! ");
     }
+
+    if (m_externalHandleType == VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_FD)
+    {
+        err = clGetSemaphoreHandleForTypeKHRptr(
+            m_externalSemaphore, m_device, CL_SEMAPHORE_HANDLE_OPAQUE_FD_KHR,
+            sizeof(int), &fd, nullptr);
+        if (err != CL_SUCCESS)
+        {
+            throw std::runtime_error("Failed to export OpenCL semaphore\n");
+        }
+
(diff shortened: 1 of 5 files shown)


test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -928,6 +936,67 @@ clExternalExportableSemaphore::clExternalExportableSemaphore(
...
+        vkImportSemaphoreFdInfoKHR.flags = 0;
+
+        if (vkImportSemaphoreFdKHR(m_deviceSemaphore.getDevice(),
+                                   &vkImportSemaphoreFdInfoKHR)
+            != VK_SUCCESS)
+        {
+            throw std::runtime_error("Failed to import semaphore in Vulkan\n");
+        }
+    }
+#ifdef _WIN32
```
joshqti, line 967: Use a switch or an else-if branch to explicitly match the windows handle enum type.  If possible eliminate the ifdef, I made an effort to reduce the reliance of ifdefs to select semaphore handle types, and instead query the the platform for supported extensions.  If eliminating the ifdef entirely is not possible, can its scope be reduced?  What is the behavior if not WIN32 but also not opaque fd?

saurabhnv: Sure, actually I wasn't aware of your efforts made in that direction, will try to remove the ifdef.
As for your second question (I guess you are referring to sync_fd?), it still gets re-imported during every signal/wait operation.

saurabhnv: Addressed in recent commit, couldn't eliminate it completely as it uses vkImportSemaphoreWin32HandleKHR, a windows specific def as per vulkan header.

joshqti: Thanks, looks good.  I will test and update if anything fails on my end.

bashbaug (approved):
Merging as discussed in the May 28th memory TSG.

PR #1899, opened 2024-02-24
Add acquire/release

joshqti:
Add acquire/release commands where appropriate.  This is related to 1886.

### DIFF
--- a/test_conformance/vulkan/test_vulkan_interop_buffer.cpp
+++ b/test_conformance/vulkan/test_vulkan_interop_buffer.cpp
@@ -281,6 +281,17 @@ int run_test_with_two_queue(
 
                 cl_event first_launch;
 
+                err = clEnqueueAcquireExternalMemObjectsKHRptr(
+                    cmd_queue1, vkBufferList.size(), buffers, 0, nullptr,
+                    nullptr);
+                test_error_and_cleanup(err, CLEANUP,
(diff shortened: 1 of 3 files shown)


test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -281,6 +281,17 @@ int run_test_with_two_queue(
 
                 cl_event first_launch;
 
+                err = clEnqueueAcquireExternalMemObjectsKHRptr(
+                    cmd_queue1, vkBufferList.size(), buffers, 0, nullptr,
+                    nullptr);
+                test_error_and_cleanup(err, CLEANUP,
+                                       "Failed to acquire buffers");
+                err = clEnqueueAcquireExternalMemObjectsKHRptr(
```
nikhiljnv, line 289: What if these cl_mems are acquired on cmd_queue2 before kernel that uses these on cmd_queue1 which may make access to these on cmd_queue1 invalid. Also, if cmd_queue1 has already acquired these, then acquiring them on cmd_queue2 without releasing from queue2 may be illegal?

joshqti:
Responded to review comments, and moved acquire/release to precede/follow the clEnqueueNDRangeKernel command that uses the relevant mem-objects.

bashbaug:

PR #1902, opened 2024-02-26
remove special-case range clamps for the fp16 divide test

bashbaug:
In the fp32 divide test, the range clamps are needed for the "fast relaxed math" testing, but there is no "fast relaxed math" testing for fp16.  This means that the range clamps are unnecessary and can be removed.


lakshmih:
Need some time to test and review this.

svenvh (approved):
LGTM, good catch!

bashbaug:
All of the CI builds are passing now, except for the known issues with the code formatting in the fp16-staging branch.  Merging as discussed in the April 9th teleconference.

PR #1907, opened 2024-03-07
spirv_new: avoid GLOB for sources in CMakeLists

svenvh:
Using `GLOB` here breaks incremental builds when switching between branches (e.g. when trying out a pull request).  The CMake documentation [discourages use of `GLOB` to collect a list of source files](https://cmake.org/cmake/help/latest/command/file.html#filesystem), so list all source files explicitly.

All other modules already use explicit source file lists; spirv_new was the odd one out.


bashbaug (approved):

bashbaug:
Thanks - I've been annoyed by this in the past!

PR #1913, opened 2024-03-11
fix the type used for bitfield_extract_signed

bashbaug:
We compute the reference values using unsigned types exclusively, even when the input type is signed.  This fixes one place where an signed type was inadvertently used.


StuartDBrady (approved):
Looks good to me.

PR #1914, opened 2024-03-11
conversions: Fix verification of half subnormal cases with FTZ

lakshmih:


test_conformance/conversions/conversions_data_info.h
```diff
@@ -717,6 +726,33 @@ void DataInfoSpec<InType, OutType, InFP, OutFP>::conv_sat(OutType *out,
...
+  // handle the cases that the converted result is subnormal
+  // from double
+  if (std::is_same<OutType, cl_double>::value)
+    *allow |= IsDoubleSubnormal(*out);
+  // from float
+  if (std::is_same<OutType, cl_float>::value)
+    *allow |= IsFloatSubnormal(*out);
+  // from half
+  if (is_out_half())
+    *allow |= IsHalfSubnormal(*out);
```
bashbaug, line 753: Checking: Previously, we were only checking the source data, which is now called `in`.  This section of code is also checking the destination data also, called `out`.  Do we need to check both?

bashbaug (commented):
I don't have an easy way to test this codepath but this looks reasonable to me and I did a quick test and nothing seems to have broken.

I do have one small question though - see below.

test_conformance/conversions/conversions_data_info.h
lakshmih: It's necessary to check the destination data, called out, in class template member function DataInfoSpec<InType, OutType, InFP, OutFP>::set_allow_zero to address the case when the destination data is a denormalized number before rounding, and the execution platform doesn’t support denormalized numbers in destination data type.

lakshmih:
This will get conversions passing for us so will be good to get it in. Thanks.

test_conformance/conversions/conversions_data_info.h
bashbaug: Gotcha, thanks!

bashbaug (approved):

PR #1915, opened 2024-03-12
Added negative tests for function CreateCommandBufferKHR from cl_khr_command_buffer

kamil-goras-mobica:
According to description #1668


test_conformance/extensions/cl_khr_command_buffer/negative_tests/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(MODULE_NAME CL_KHR_COMMAND_BUFFER_NEGATIVE_TESTS)
+
+set(${MODULE_NAME}_SOURCES
+    main.cpp
+    command_buffer_create.cpp
+    ../basic_command_buffer.cpp
+)
+
+set_gnulike_module_compile_flags("-Wno-sign-compare")
```
EwanC, line 9: Since we're adding new tests, it'd be better to keep them warning. Ideally we'd remove this line (which is probably just copied from the other command-buffer test CMAKE) and fix any sign compare warnings.

test_conformance/extensions/cl_khr_command_buffer/negative_tests/command_buffer_create.cpp
```diff
@@ -0,0 +1,318 @@
...
+// limitations under the License.
+//
+#include "../basic_command_buffer.h"
+#include "procs.h"
+
+
+//--------------------------------------------------------------------------
+namespace {
+
+// CL_INVALID_VALUE if num_queues is not one.
```
EwanC, line 23: Can we add a `Skip()` method here that checks `cl_khr_command_buffer_multi_device` is not a supported device extension. See this error language https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#_error_updates_2 of how the extension changes the error "CL_INVALID_VALUE if num_queues is not one." to "CL_INVALID_VALUE if num_queues is zero."

test_conformance/extensions/cl_khr_command_buffer/negative_tests/command_buffer_create.cpp
```diff
@@ -0,0 +1,318 @@
...
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    cl_int Run() override
+    {
+        cl_int error = CL_SUCCESS;
+
+        cl_command_buffer_properties_khr repeated_properties[4] = {
+            CL_COMMAND_BUFFER_FLAGS_KHR, CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR,
+            CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR, 0
```
EwanC, line 100: I think the property *name* here is `CL_COMMAND_BUFFER_FLAGS_KHR`. So this is testing an invalid property name which is good, but think also testing for a repeated but valid property could be done. e.g.
```
cl_command_buffer_properties_khr repeated_properties[5] = {
 CL_COMMAND_BUFFER_FLAGS_KHR, CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR,
 CL_COMMAND_BUFFER_FLAGS_KHR, CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR, 0};
```

bashbaug (commented):
These tests generally look good and they found an issue in the command buffer emulation layer (already fixed), so that's pretty cool.

Did we discuss whether to implement these tests in a separate executable or as a part of the existing command buffer executable?  All else being equal I'd prefer to minimize the number of test executables, and I don't think we've had separate executables for negative tests previously.

EwanC:
> Did we discuss whether to implement these tests in a separate executable or as a part of the existing command buffer executable? All else being equal I'd prefer to minimize the number of test executables, and I don't think we've had separate executables for negative tests previously.

I don't think we have discussed it. I would probably also err on the side of less test executables as I feel like new ones could get missed when people are maintain scripts/csvs for testing, but that's a slight preference rather than a strong one. Prefixing the new tests with `negative_` I think would suitably identify the negative tests in the existing cl_khr_command_buffer executable.

test_conformance/extensions/cl_khr_command_buffer/main.cpp
```diff
@@ -63,7 +63,13 @@ test_definition test_list[] = {
...
     ADD_TEST(event_info_reference_count),
     ADD_TEST(finalize_invalid),
-    ADD_TEST(finalize_empty)
+    ADD_TEST(finalize_empty),
+    ADD_TEST(create_command_buffer_num_queues),
+    ADD_TEST(create_command_buffer_null_queues),
+    ADD_TEST(create_command_buffer_repeated_properties),
+    ADD_TEST(create_command_buffer_not_supported_properties),
+    ADD_TEST(create_command_buffer_queue_without_min_properties),
+    ADD_TEST(create_command_buffer_device_does_not_support_out_of_order_queue),
```
EwanC, line 72: Can we prefix the names of the new tests with `negative_` here and going forward, that way a user can filter to run only negative tests by invoking `$ ./test_conformance/extensions/cl_khr_command_buffer/test_cl_khr_command_buffer negative*`

EwanC (dismissed):

EwanC (approved):

bashbaug:

Issue #1272, opened 2021-06-11
Fix support for CL_UNORM_INT_101010_2 in kernel read write tests


kpet:
See https://github.com/KhronosGroup/OpenCL-CTS/pull/1129#discussion_r647109443

bashbaug:
See also: https://github.com/KhronosGroup/OpenCL-CTS/pull/1917#issuecomment-1995430013

kpet:
I think a good first step would be to grep for areas of the code where `CL_UNORM_INT_101010` is handled but `CL_UNORM_INT_101010_2` is not. In tests where many formats are enumerated individually, especially in multiple places in the code, we should consider whether common abstractions for these list of formats would make sense to introduce.

kpet:
Related to https://github.com/KhronosGroup/OpenCL-CTS/pull/2031

rjodinchr:
Fixed by PR #1917 (commit `0deddf2c`) and PR #2031 (commit `bcfd8f82`), which added complete helper and testing support for `CL_UNORM_INT_101010_2` across `imageHelpers.cpp` and `test_conformance/images/kernel_read_write/`.

PR #1917, opened 2024-03-13
Fix failing image tests for CL_UNORM_INT_101010_2

ahesham-arm:
Add support for `CL_UNORM_INT_101010_2` in the `get_pixel_size` helper function. 

This fixes the following tests:
-  memInfo_image_from_buffer_positive
- image_from_buffer_alignment_negative
- imageInfo_image_from_buffer_positive

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -286,6 +286,7 @@ uint32_t get_pixel_size(const cl_image_format *format)
 
         case CL_FLOAT:
             return get_format_channel_count(format) * sizeof(cl_float);
+        case CL_UNORM_INT_101010_2: return 4;
 
         default: return 0;
     }


svenvh (approved):

test_common/harness/imageHelpers.cpp
```diff
@@ -286,6 +286,7 @@ uint32_t get_pixel_size(const cl_image_format *format)
 
         case CL_FLOAT:
             return get_format_channel_count(format) * sizeof(cl_float);
+        case CL_UNORM_INT_101010_2: return 4;
```
bashbaug, line 289: Nitpick: it'd be more consistent with the rest of the code in this function to move this line above, underneath the case statement for `CL_UNORM_INT_101010`.

bashbaug (commented):
I'm fine merging this as-is and this case is obviously missing here, but if I do a search for `CL_UNORM_INT_101010` it looks like similar changes need to be made in a lot of other places, also.

bashbaug:
> The end goal is to have proper testing for it but I started with a non-controversial change that fixes failures in existing test cases.

I was going to ask to file an issue so we didn't lose track of this, but it looks like we already filed one (#1272).

All good from my side - thanks!

bashbaug (approved):

kpet:
Trivial fix with two approvals, merging.

PR #1920, opened 2024-03-15
[CI] Upgrade to actions/checkout@v4

svenvh:
Node.js 16 actions are [deprecated](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/), so update to v4 which uses Node 20.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -28,7 +28,7 @@ jobs:
             debug: 1
             extra: " debug"
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
LGTM

I've updated a few of my repos to update these actions automatically - see, for example: https://github.com/bashbaug/SimpleOpenCLSamples/pull/108.  I could make a similar change here if we'd like - let me know.

svenvh:
Thanks!  It's probably a good idea to set up dependabot indeed, as we tend to forget about updating dependencies.

PR #1921, opened 2024-03-15
Corrections for mutable arguments tests

shajder:
-added verification of device capabilities against mutable arguments 
-general cleanup

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
@@ -157,9 +157,6 @@ struct MutableDispatchGlobalArguments : public MutableDispatchArgumentsTest
                                           nullptr, nullptr);
         test_error(error, "clEnqueueCommandBufferKHR failed");
 
-        error = clFinish(queue);
-        test_error(error, "clFinish failed");
-
         // check the results of the initial execution
         if (!verify_result(dst_buf_0)) return TEST_FAIL;
 
@@ -194,9 +191,6 @@ struct MutableDispatchGlobalArguments : public MutableDispatchArgumentsTest
         error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
                                           nullptr, nullptr);
         test_error(error, "clEnqueueCommandBufferKHR failed");
-
-        error = clFinish(queue);
-        test_error(error, "clFinish failed");
 
         // Check the results of the modified execution
         if (!verify_result(dst_buf_1)) return TEST_FAIL;
@@ -557,9 +551,6 @@ struct MutableDispatchNullArguments : public MutableDispatchArgumentsTest
                                           nullptr, nullptr);
         test_error(error, "clEnqueueCommandBufferKHR failed");
 
-        error = clFinish(queue);
-        test_error(error, "clFinish failed");
-
         // Check the results of the initial execution
         std::vector<cl_int> dst_data_0(num_elements);
         error = clEnqueueReadBuffer(queue, out_mem, CL_TRUE, 0,
@@ -607,9 +598,6 @@ struct MutableDispatchNullArguments : public MutableDispatchArgumentsTest
         error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
                                           nullptr, nullptr);
         test_error(error, "clEnqueueCommandBufferKHR failed");
-
-        error = clFinish(queue);
-        test_error(error, "clFinish failed");
 
         // Check the results of the modified execution
         std::vector<cl_int> dst_data_1(num_elements);
@@ -810,9 +798,6 @@ struct MutableDispatchSVMArguments : public MutableDispatchArgumentsTest
                                           nullptr, nullptr);
         test_error(error, "clEnqueueCommandBufferKHR failed");
 
-        error = clFinish(queue);
-        test_error(error, "clFinish failed");
-
         // Check the results of the modified execution
         error =
             clEnqueueSVMMap(queue, CL_TRUE, CL_MAP_READ, new_buffer,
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
```diff
@@ -59,48 +76,76 @@ struct MutableDispatchGlobalArguments : public BasicMutableCommandBufferTest
...
 
-        std::vector<cl_int> srcData(num_elements);
+        srcData.resize(num_elements);
         for (size_t i = 0; i < num_elements; i++)
             srcData[i] = (cl_int)genrand_int32(d);
 
-        clMemWrapper srcBuf = clCreateBuffer(context, CL_MEM_COPY_HOST_PTR,
-                                             num_elements * sizeof(cl_int),
-                                             srcData.data(), &error);
+        in_mem = clCreateBuffer(context, CL_MEM_COPY_HOST_PTR,
```
EwanC, line 96: `in_mem` is snake_case while the other variable names here are camelCase, which looks a bit odd. Could we make the variable naming consistent?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
```diff
@@ -59,48 +76,76 @@ struct MutableDispatchGlobalArguments : public BasicMutableCommandBufferTest
...
         test_error(error, "Unable to set initial dst kernel argument");
+        return CL_SUCCESS;
+    }
+
+    // verify the result
+    bool verify_result(const cl_mem &buffer)
+    {
+        cl_int error = CL_SUCCESS;
+        std::vector<cl_int> data(num_elements);
+        error = clEnqueueReadBuffer(queue, buffer, CL_TRUE, 0, data_size(),
```
EwanC, line 124: Prefer to define `error` the first time it's used (and in other places touched by this PR)
```suggestion
        std::vector<cl_int> data(num_elements);
        cl_int error = clEnqueueReadBuffer(queue, buffer, CL_TRUE, 0, data_size(),
```

EwanC (commented):
Since this PR does some cleanup, I'll comment here with a couple ideas of other cleanups I had in mind. But these should be done in a separate PR if we want to do them:

* Most command-buffer tests include that cl_ext.h include both `cl_ext` and `cl.h`, so I think we just need `cl_ext.h` and `cl.h` can be removed - https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl_ext.h#L24
* The command-buffer CMake sets `set_gnulike_module_compile_flags("-Wno-sign-compare")`, it would be nice if we could fix these warning and remove this.

shajder:
> Since this PR does some cleanup, I'll comment here with a couple ideas of other cleanups I had in mind. But these should be done in a separate PR

Do you think I should wait with mentioned cleanup PR until we will have all `cl_khr_command_buffer_mutable_dispatch` work merged to main ?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
shajder: I believe I found all the cases

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
```diff
@@ -572,49 +608,60 @@ struct MutableDispatchNullArguments : public BasicMutableCommandBufferTest
                                           nullptr, nullptr);
         test_error(error, "clEnqueueCommandBufferKHR failed");
 
-        // Check the results of the modified execution
+        error = clFinish(queue);
+        test_error(error, "clFinish failed");
```
EwanC, line 612: To make sure I understand this change correctly, we are inserting a call to `clFinish()` on an in-order queue, between
a call to enqueuing the command-buffer and a blocking `clEnqueueReadBuffer` on the same queue.

Although this isn't an invalid change to the test, I don't think the previous code was incorrect. Due to the in-order queue property the `clEnqueueReadBuffer` should have a dependency on the `clEnqueueCommandBufferKHR` completing.

If this doesn't work with OCK, then I feel it's a bug with OCK that should be reported there rather than weakening the test.

shajder: Indeed, that was the motivation behind that correction so I restored previous approach, BTW shouldn't we then remove then line [744](https://github.com/shajder/OpenCL-CTS/blob/490efa8ddbed9f98c19ef90018f20953c0410829/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp#L744) and [767](https://github.com/shajder/OpenCL-CTS/blob/490efa8ddbed9f98c19ef90018f20953c0410829/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp#L767) as well ? It is followed by blocking `clEnqueueSVMMap`, right ?

EwanC: Yes, I think we could remove those lines you linked as well.

EwanC (approved):

PR #1923, opened 2024-03-18
configure depdendabot github actions version updates

bashbaug:
see #1920 

This sets up dependabot updates to GitHub actions.

I have been using a similar config file in my repos and it has been working well.  Currently, it is setup to check for updates monthly, but we can check for updates more or less frequently as needed.


svenvh (approved):
LGTM, thanks for doing this!

PR #1924, opened 2024-03-19
mem_host_flags: check returned pointer for NULL

svenvh:
The specification states that `clEnqueueMapImage` and `clEnqueueMapBuffer` should return NULL on error.  Ensure this is checked.

This testing gap was caught by a compiler warning about `dataPtr` being written but not read.  With this fix, there are no more Wunused-but-set-variable warnings in this test, so reenable the warning.

### DIFF
--- a/test_conformance/mem_host_flags/CMakeLists.txt
+++ b/test_conformance/mem_host_flags/CMakeLists.txt
@@ -6,6 +6,4 @@ set(${MODULE_NAME}_SOURCES
     mem_host_image.cpp
 )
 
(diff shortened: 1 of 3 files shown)


bashbaug (approved):
LGTM!  Pretty cool how the compiler warning caught this testing gap.

EwanC (approved):

bashbaug:
Discussed in the April 16th teleconference, waiting one more week.

lakshmih (approved):

bashbaug:
There are three approvals on this PR now, so no reason to wait to merge it.

PR #1929, opened 2024-03-26
Added negative tests for clGetCommandBufferInfoKHR

kamil-goras-mobica:
According to description #1668

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_get_info.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_get_info.cpp
@@ -1,0 +1,312 @@
+//
+// Copyright (c) 2024 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_get_info.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+        return TEST_PASS;
+    }
+
+    cl_int RunRefCountInfoTest()
+    {
+        cl_uint ref_count = 0;
+
+        cl_int error = clGetCommandBufferInfoKHR(
+            command_buffer, CL_COMMAND_BUFFER_REFERENCE_COUNT_KHR, 0,
```
EwanC, line 187: Can we use a non-zero `param_value_size` in these tests too? It's possible that an implementation could have a check against zero but not against being less that the return type in general.

For example, here we could use `sizeof(ref_count)-1` as another `param_value_size` input too.

EwanC (approved):
LGTM

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_get_info.cpp
```diff
@@ -0,0 +1,322 @@
...
+    }
+
+    cl_int RunPropArrayInfoTest()
+    {
+        std::vector<cl_command_buffer_properties_khr> properties;
+        properties.resize(number_of_properties);
+
+        cl_int error = clGetCommandBufferInfoKHR(
+            command_buffer, CL_COMMAND_BUFFER_PROPERTIES_ARRAY_KHR,
+            number_of_properties * sizeof(cl_command_buffer_properties_khr) - 1,
```
ahesham-arm, line 226: The result of this calculation is:
`0 * sizeof(cl_command_buffer_properties_khr) - 1` which is `-1`, but this is a `size_t`. This test is incorrect.

PR #1932, opened 2024-03-27
test allocations: restore small number of work items in case of reduction 

gwawiork:


gwawiork:
@kpet This is related to https://github.com/KhronosGroup/OpenCL-CTS/pull/1852
We use reduction argument on simulation then too high number of work items makes test very long and we see timeouts. Would be nice have previous value set. I agree there is a need to implement function which will calculate this number accordingly

kpet:
Merging this PR would result in failures for the implementations listed in #1852. Is the issue you're seeing on a final device or development environment? If the latter, maybe we could introduce command line options to help.

gwawiork:
> Merging this PR would result in failures for the implementations listed in #1852. Is the issue you're seeing on a final device or development environment? If the latter, maybe we could introduce command line options to help.

By default value is set as in #1852  Only when someone use reduction factor (percentage in command line) Do you use it ?
`if (g_reduction_percentage != 100)`

kpet (approved):

kpet:
@gwawiork Apologies I'd missed that when quickly skimming through it. I've now approved the PR. Should be fine.

bashbaug:
Discussed in the April 16th teleconference.  We'll give this one more week for review, but are hoping to merge shortly.

lakshmih (approved):

bashbaug:

PR #1934, opened 2024-03-28
compiler: Fix feature check

lakshmih:
Limit the check for 2.0 specific features to devices that report support for that version

### DIFF
--- a/test_conformance/compiler/test_opencl_c_versions.cpp
+++ b/test_conformance/compiler/test_opencl_c_versions.cpp
@@ -125,7 +125,7 @@ static int test_CL_DEVICE_OPENCL_C_ALL_VERSIONS(cl_device_id device,
     std::vector<cl_name_version> clc_versions(sz / sizeof(cl_name_version));
     error = clGetDeviceInfo(device, CL_DEVICE_OPENCL_C_ALL_VERSIONS, sz,
                             clc_versions.data(), NULL);
-    test_error(error, "Unable to query CL_DEVICE_OPENCL_C_FEATURES");
+    test_error(error, "Unable to query CL_DEVICE_OPENCL_C_ALL_VERSIONS");
 
     for (const auto& clc_version : clc_versions)
     {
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_opencl_c_versions.cpp
```diff
@@ -192,8 +192,27 @@ static int test_CL_DEVICE_OPENCL_C_VERSION_features(cl_device_id device,
...
+                            clc_versions.data(), NULL);
+    test_error(error, "Unable to query CL_DEVICE_OPENCL_C_ALL_VERSIONS");
+    bool supports_clc_20 = false;
+    for (const auto& clc_version : clc_versions)
+    {
+        const unsigned major = CL_VERSION_MAJOR(clc_version.version);
+        const unsigned minor = CL_VERSION_MINOR(clc_version.version);
+        if (major == 2 && minor == 0)
+        {
+            supports_clc_20 = true;
```
bashbaug, line 211: I think this change is OK and debatably more robust to future spec changes, but I don't think it will materially affect testing with the current OpenCL C version requirements.

The current code (before this PR) is checking whether the value returned by [CL_DEVICE_OPENCL_C_VERSION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_VERSION) is OpenCL C 2.0 or greater.  Recall that this is the "highest fully backwards compatible OpenCL C version supported by the compiler for the device".

The updated code (in this PR) is checking whether OpenCL C 2.0 is in [CL_DEVICE_OPENCL_C_ALL_VERSIONS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_OPENCL_C_ALL_VERSIONS).  While this is fine, and the required features should be present when OpenCL C 2.0 is present, because OpenCL C 2.0 is backwards compatible with OpenCL C 1.0, 1.1, and 1.2 I can't think of a scenario where OpenCL C 2.0 would be in CL_DEVICE_OPENCL_C_ALL_VERSIONS but CL_DEVICE_OPENCL_C_VERSION would return OpenCL C 1.2 or below.  Is there a scenario where this would occur in practice?

test_conformance/compiler/test_opencl_c_versions.cpp
```diff
@@ -275,6 +299,13 @@ static int test_CL_DEVICE_OPENCL_C_VERSION_versions(cl_device_id device,
             }
             else
             {
+                if (device_clc_version == Version(3, 0)
+                    && test_clc_version == Version(2, 0))
+                {
+                    // OpenCL C 3.0 is not backwards compatible with OpenCL
+                    // C 2.0, continue the test
+                    continue;
+                }
```
bashbaug, line 308: I don't think this is correct and this change should be reverted.

Since the value returned for CL_DEVICE_OPENCL_C_VERSION describes "the highest fully backwards compatible OpenCL C version supported by the compiler for the device", if a device returns OpenCL C 3.0 for this query then the device must also support OpenCL C 2.0.

In other words, if a device does not support OpenCL C 2.0, then the value returned by CL_DEVICE_OPENCL_C_VERSION should be OpenCL C 1.2 at most, and any higher versions of OpenCL C should be returned by CL_DEVICE_OPENCL_C_ALL_VERSIONS instead.

PR #1935, opened 2024-04-01
Negative test for CL_INVALID_QUEUE_PROPERTIES

EwanC:
Test that verifies `CL_INVALID_QUEUE_PROPERTIES` is returned from `clCreateCommandQueue`, `clCreateCommandQueueWithProperties`, and `clCreateCommandQueueWithPropertiesKHR` to match spec wording

> CL_INVALID_QUEUE_PROPERTIES if values specified in properties are valid but are not supported by the device.

### DIFF
--- a/test_conformance/api/negative_queue.cpp
+++ b/test_conformance/api/negative_queue.cpp
@@ -1,0 +1,174 @@
+//
+// Copyright (c) 2024 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "testBase.h"
+#include "harness/typeWrappers.h"
+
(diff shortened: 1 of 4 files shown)


test_conformance/api/negative_queue.cpp
```diff
@@ -0,0 +1,145 @@
...
+                           "clCreateCommandQueueWithProperties should "
+                           "return CL_INVALID_QUEUE_PROPERTIES if "
+                           "values specified in properties are valid but "
+                           "are not supported by the "
+                           "device.",
+                           TEST_FAIL);
+
+    // Verify the same correct error reporting for
+    // clCreateCommandQueueWithPropertiesKHR
+    if (is_extension_available(deviceID, "cl_khr_create_command_queue"))
```
bashbaug, line 112: This isn't a must-fix, since most implementations have moved to OpenCL 3.0 and hence this won't make much of a difference in practice, but: it might be worth moving this extension part to its own test since it can run with any OpenCL version, whereas the core version requires OpenCL 2.0 or newer.

bashbaug (commented):
I don't have easy access to an implementation that does not support out-of-order queuest to test this, but I took a look through the code and it mostly LGTM.

bashbaug:
Note: there is a merge conflict that will need to be addressed also (should be straightforward).

test_conformance/api/negative_queue.cpp
EwanC: Good shout, I think that makes sense - update to have the KHR in its own test.

bashbaug (approved):

lakshmih (approved):

bashbaug:

PR #1936, opened 2024-04-05
Negative test commandNDRangeKernel

kamil-goras-mobica:

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
@@ -1,0 +1,379 @@
+//
+// Copyright (c) 2024 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "basic_command_buffer.h"
+#include "procs.h"
+
+
+//--------------------------------------------------------------------------
+namespace {
+
+// CL_INVALID_COMMAND_QUEUE if command_queue is not NULL.
+struct CommandNDRangeKernelQueueNotNull : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    cl_int Run() override
+    {
+        cl_int error = clCommandNDRangeKernelKHR(
+            command_buffer, queue, nullptr, kernel, 1, nullptr, &num_elements,
+            nullptr, 0, nullptr, nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_COMMAND_QUEUE,
+                               "clCommandNDRangeKernelKHR should return "
+                               "CL_INVALID_COMMAND_QUEUE",
+                               TEST_FAIL);
+
+        return CL_SUCCESS;
+    }
+};
+
+// CL_INVALID_CONTEXT if the context associated with command_queue,
+// command_buffer, and kernel are not the same.
+struct CommandNDRangeKernelKernelWithDifferentContext
+    : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    cl_int Run() override
+    {
+        cl_int error = CreateKernelWithDifferentContext();
+        test_error(error, "Failed to create kernel");
+
+        error = clCommandNDRangeKernelKHR(
+            command_buffer, nullptr, nullptr, kernel, 1, nullptr, &num_elements,
+            nullptr, 0, nullptr, nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_CONTEXT,
+                               "clCommandNDRangeKernelKHR should return "
+                               "CL_INVALID_CONTEXT",
+                               TEST_FAIL);
+
+        return CL_SUCCESS;
+    }
+
+    cl_int CreateKernelWithDifferentContext()
+    {
+        cl_int error = CL_SUCCESS;
+
+        new_context = clCreateContext(0, 1, &device, nullptr, nullptr, &error);
+        test_error(error, "Failed to create context");
+
+        const char* kernel_str =
+            R"(
+      __kernel void copy(__global int* in, __global int* out, __global int* offset) {
+          size_t id = get_global_id(0);
+          int ind = offset[0] + id;
+          out[ind] = in[ind];
+      })";
+
+        error = create_single_kernel_helper_create_program(
+            new_context, &program, 1, &kernel_str);
+        test_error(error, "Failed to create program with source");
+
+        error = clBuildProgram(program, 1, &device, nullptr, nullptr, nullptr);
+        test_error(error, "Failed to build program");
+
+        kernel = clCreateKernel(program, "copy", &error);
+        test_error(error, "Failed to create copy kernel");
+
+        return CL_SUCCESS;
+    }
+    clContextWrapper new_context;
+};
+
+// CL_INVALID_SYNC_POINT_WAIT_LIST_KHR if sync_point_wait_list is NULL and
+// num_sync_points_in_wait_list is > 0, or sync_point_wait_list is not NULL and
+// num_sync_points_in_wait_list is 0, or if synchronization-point objects in
+// sync_point_wait_list are not valid synchronization-points.
+struct CommandNDRangeKerneSyncPointsNullOrNumZero
+    : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    cl_int Run() override
+    {
+        cl_int error = clCommandNDRangeKernelKHR(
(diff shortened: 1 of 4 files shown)


test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,535 @@
...
+
+        cl_int error = clCommandNDRangeKernelKHR(
+            command_buffer, nullptr, nullptr, kernel, 1, nullptr, &num_elements,
+            nullptr, 0, nullptr, nullptr, &mutable_handle);
+
+        test_failure_error_ret(error, CL_INVALID_VALUE,
+                               "clCommandNDRangeKernelKHR should return "
+                               "CL_INVALID_VALUE",
+                               TEST_FAIL);
+
```
EwanC, line 241: We should skip this test if `cl_khr_command_buffer_mutable_dispatch` is supported, as it relaxes this restriction.

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,535 @@
...
+            clGetDeviceInfo(device, CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR,
+                            sizeof(capabilities), &capabilities, NULL);
+        test_error(error,
+                   "Unable to query CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR");
+
+        bool device_does_not_support_printf =
+            (capabilities & CL_COMMAND_BUFFER_CAPABILITY_KERNEL_PRINTF_KHR)
+            == 0;
+
+        return device_does_not_support_printf;
```
EwanC, line 266: Should this be `!device_does_not_support_printf`, as we want to skip the test is a device **does** support printf, because then we won't get an unsupported error code.

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,535 @@
...
+
+    bool Skip() override
+    {
+        bool has_device_enqueue = false;
+        bool cl_version_2_0_or_higher = false;
+
+        OpenCLCFeatures features;
+        get_device_cl_c_features(device, features);
+
+        const Version clc_version = get_device_cl_c_version(device);
```
EwanC, line 346: Should use `get_device_latest_cl_c_version` here, that will report a higher version if available, and so is more likely to be able to test this feature.

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,525 @@
...
+        test_failure_error_ret(error, CL_INVALID_SYNC_POINT_WAIT_LIST_KHR,
+                               "clCommandNDRangeKernelKHR should return "
+                               "CL_INVALID_SYNC_POINT_WAIT_LIST_KHR",
+                               TEST_FAIL);
+
+        cl_sync_point_khr* invalid_point = nullptr;
+        cl_sync_point_khr* invalid_sync_points[] = { invalid_point };
+        error = clCommandNDRangeKernelKHR(
+            command_buffer, nullptr, nullptr, kernel, 0, nullptr, &num_elements,
+            nullptr, 0, invalid_sync_points[0], nullptr, nullptr);
```
EwanC, line 132: This seems overly complex, `invalid_sync_points[0]` is a `nullptr` - we could just pass a `nullptr` literal here rather than defining `invalid_point` or `invalid_sync_points`.

Doing so would show that passing 0 for `num_sync_points_in_wait_list` and `nullptr` for `sync_point_wait_list` is valid. Since we haven't recorded any commands, at this point there shouldn't be any valid sync-point values and I think you
could just do
```cpp
cl_sync_point_khr invalid_point = 0;  // No sync points defined yet, so any valid should be invalid`
```

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,525 @@
...
+__kernel void enqueue_call_kernel() {
+queue_t def_q = get_default_queue();
+ndrange_t ndrange = ndrange_1D(1);
+enqueue_kernel(def_q, ndrange,
+                   ^{enqueue_call_func();});
+  }
+)";
+        std::string buildOptions = std::string(" ") + cl_std;
+
+        error = create_single_kernel_helper_create_program(
```
EwanC, line 401: Can we use `build_program_create_kernel_helper` here instead which should call `clBuildProgram` and `clCreateKernel` for us and give a build log if the program doesn't build. Same for printf kernel setup.

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,525 @@
...
+};
+
+// CL_INVALID_VALUE if values specified in properties are not valid.
+struct CommandNDRangeKernelInvalidProperties : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    cl_int Run() override
+    {
+        cl_ndrange_kernel_command_properties_khr empty_properties;
```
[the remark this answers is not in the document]
EwanC: This has been initalized to `0`, but I zero is a valid value and represents an empty list (`0` denotes the end of a list). If we want an invalid list we could use `cl_ndrange_kernel_command_properties_khr invalid_property= ~cl_ndrange_kernel_command_properties_khr(0);`

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
EwanC: Apologies Kamil, I realize i've given the wrong advice here. I tested this test out on a device with device enqueue support, and the kernel code is well formed - however we need to use `create_single_kernel_helper` rather than `build_program_create_kernel_helper ` because `build_program_create_kernel_helper` doesn't first create the program.

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+        const char* kernel_str =
+            R"(
+__kernel void enqueue_call_func() {
+  }
+
+__kernel void enqueue_call_kernel() {
+queue_t def_q = get_default_queue();
+ndrange_t ndrange = ndrange_1D(1);
+enqueue_kernel(def_q, ndrange,
```
EwanC, line 395: Think we're missing the `flags` parameter here - https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#built-in-functions-enqueuing-a-kernel

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,520 @@
...
+        error = clCommandNDRangeKernelKHR(
+            command_buffer, nullptr, nullptr, kernel, 0, nullptr, &num_elements,
+            nullptr, 0, sync_points[0], nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_SYNC_POINT_WAIT_LIST_KHR,
+                               "clCommandNDRangeKernelKHR should return "
+                               "CL_INVALID_SYNC_POINT_WAIT_LIST_KHR",
+                               TEST_FAIL);
+
+        cl_sync_point_khr invalid_point = 0;
```
EwanC, line 133: How do we know this is an invalid value and not the same value as `point` returned by `clCommandBarrierWithWaitListKHR`?

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -0,0 +1,520 @@
...
+        cl_sync_point_khr* invalid_sync_points[] = { &invalid_point };
+        error = clCommandNDRangeKernelKHR(
+            command_buffer, nullptr, nullptr, kernel, 0, nullptr, &num_elements,
+            nullptr, 0, invalid_sync_points[0], nullptr, nullptr);
```
EwanC, line 137: Think `num_sync_points_in_wait_list is` should be 1 here otherwise this is testing the same error as on line 126
```suggestion
            nullptr, 1, invalid_sync_points[0], nullptr, nullptr);
```

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
EwanC: I checked a couple of implementations: The emulation layer sets `point` to one, but OCK sets `point` to zero so  `invalid_point = 0` is a valid sync-point and this test fails.

kamil-goras-mobica: @EwanC correction applied

EwanC: thanks

test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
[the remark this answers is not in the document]
kamil-goras-mobica: @EwanC corrected

EwanC (approved):
LGTM

PR #1939, opened 2024-04-07
basic: workaround MSVC compiler bug with post-increment operator

lakshmih:
Prevent the compiler from optimizing away initialization loops

### DIFF
--- a/test_conformance/basic/test_vector_swizzle.cpp
+++ b/test_conformance/basic/test_vector_swizzle.cpp
@@ -516,8 +516,7 @@ static void makeReference(std::vector<T>& ref)
     // single channel lvalue
     for (size_t i = 0; i < N; i++)
     {
-        ref[dstIndex * S + i] = 0;
-        ++dstIndex;
+        ref[dstIndex++ * S + i] = 0;
     }
 
     // normal lvalue


svenvh:
Disabling optimizations is not what `volatile` is intended for.  Are you able to provide more details about what's going wrong?  I've run this test with ubsan and asan just fine, so perhaps it's a miscompilation of your compiler?  If so, it would be better to disable optimizations for a specific compiler (with gcc/clang you could use `#pragma GCC optimize("-O0")` for example, together with `push_options` / `pop_options` to keep it local to a function).

bashbaug:
Discussed in the April 16th teleconference.  We’d prefer not to use “volatile” to work around this issue.  Qualcomm will include more information about the failing case, e.g. which compiler is exhibiting the problem.

lakshmih:
This is with Visual Studio compiler, VS 2022 
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.38.33130\bin\Hostx86\x86\cl.exe
It is probably a compiler bug, not sure. 

We have precedence for using volatile to prevent optimizations (including the commented instance here https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/basic/test_local_kernel_scope.cpp#L114) but if we would prefer to, switching the variable to post-increment seems to help. Would that be acceptable?

--- a/test_conformance/basic/test_vector_swizzle.cpp
+++ b/test_conformance/basic/test_vector_swizzle.cpp
@@ -516,8 +516,7 @@ static void makeReference(std::vector<T>& ref)
     // single channel lvalue
     for (size_t i = 0; i < N; i++)
     {
-        ref[dstIndex * S + i] = 0;
-        ++dstIndex;
+        ref[dstIndex++ * S + i] = 0;
     }

lakshmih:
This is with Visual Studio compiler VS 2022 
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.38.33130\bin\Hostx86\x86\cl.exe

It is probably a compiler bug. It also reproduces pretty consistently in a unit test that just calls that function from main like below:

int main()
{
    std::vector<float> reference(576);
    makeReference<float, 16, 16>(reference);
    for (auto v : reference) {
        printf("%f\n", v);
    }
}


We have precedence for using volatile to prevent optimizations (including the commented instance here https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/basic/test_local_kernel_scope.cpp#L114) but if we would prefer to, switching the variable to post-increment seems to help. Would that be acceptable?

 
--- a/test_conformance/basic/test_vector_swizzle.cpp
+++ b/test_conformance/basic/test_vector_swizzle.cpp
@@ -516,8 +516,7 @@ static void makeReference(std::vector<T>& ref)
     // single channel lvalue
     for (size_t i = 0; i < N; i++)
     {
-        ref[dstIndex * S + i] = 0;
-        ++dstIndex;
+        ref[dstIndex++ * S + i] = 0;
     }

This is breaking conformance for us, so would like to merge some manner of fix for it.

svenvh:
Thanks for the detail.  If it fixes your issue, switching to post-increment would be fine with me.  Otherwise, I suppose you can add something like this to the test's CMakeLists.txt:
```
if (MSVC)
  set_source_files_properties(test_vector_swizzle.cpp PROPERTIES COMPILE_FLAGS "/O0")
endif(MSVC)
```

bashbaug (approved):

svenvh (approved):

bashbaug:

Issue #1058, opened 2020-11-19
printf is very undertested


airlied:
The printf tests don't cover

- multiple arguments
- no arguments
- data before format specifier

franz:
It should also cover more special cases. E.g. both of these two: 

`printf("")`
`printf("%s", "")`

.. should print an empty string. Furthermore, i think the test of printf with NaN is problematic. This test

`{"%f","sqrt(-1.0f)"},`

expects a `-nan` to be printed, however it makes two assumptions, 1) that the sqrt operation returns a negative nan (not positive), and 2) that printf prints the sign of the nan. AFAICT, both are incorrect; the IEEE 754 does not assign any meaning to sign bit of the NaN, and printing out the sign is optional.

pjaaskel:
`printf("%%")` could be interesting to test as well.

pjaaskel:
Can we put this to the "mobica-backlog"? Thanks!

shajder:
Based on previous discussion and spec investigation below is a work plan to extend the current framework of printf tests. I will appreciate any suggestions and corrections:

New set of printf tests with multiple arguments. It will take into account:
- [ ]  various types of arguments test
- [ ]  no arguments test
- [ ]  additional data before/after format specifiers

New tests related to printout formatting:
- [ ] verification of `space` flag prefixing
- [ ] verification of `zero` flag padding

Corrections for previous tests:
- [ ] Complementing vector tests to take into account all vector variants, conversions specifiers and length modifiers
- [ ] Complementing all floating precision printf tests with missing conversions %a, %A, %F
- [ ] Complementing string tests with special cases and precision number
- [ ] Complementing all non-vector conversions with length modifiers tests
- [ ] Correction for nan/inf printf tests

PR #1940, opened 2024-04-08
Corrections for printf test with floating point limits arguments

shajder:
According to work plan from issue #1058

### DIFF
--- a/test_conformance/printf/util_printf.cpp
+++ b/test_conformance/printf/util_printf.cpp
@@ -46,39 +46,41 @@ std::vector<printDataGenParameters> printIntGenParameters = {
 
     //(Minimum)Five-wide,default(right)-justified
 
-    {{"%5d"},"10"},
+    { { "%5d" }, "10" },
 
-        //(Minimum)Five-wide,left-justified
+    //(Minimum)Five-wide,left-justified
 
-    {{"%-5d"},"10"},
+    { { "%-5d" }, "10" },
 
-        //(Minimum)Five-wide,default(right)-justified,zero-filled
+    //(Minimum)Five-wide,default(right)-justified,zero-filled
 
-    {{"%05d"},"10"},
+    { { "%05d" }, "10" },
 
-        //(Minimum)Five-wide,default(right)-justified,with sign
+    //(Minimum)Five-wide,default(right)-justified,with sign
 
-    {{"%+5d"},"10"},
+    { { "%+5d" }, "10" },
 
-         //(Minimum)Five-wide ,left-justified,with sign
+    //(Minimum)Five-wide ,left-justified,with sign
 
-    {{"%-+5d"},"10"},
+    { { "%-+5d" }, "10" },
 
-        //(Minimum)Five-digit(zero-filled in absent digits),default(right)-justified
+    //(Minimum)Five-digit(zero-filled in absent digits),default(right)-justified
 
-    {{"%.5i"},"100"},
+    { { "%.5i" }, "100" },
 
-        //(Minimum)Six-wide,Five-digit(zero-filled in absent digits),default(right)-justified
+    //(Minimum)Six-wide,Five-digit(zero-filled in absent
+    // digits),default(right)-justified
 
-    {{"%6.5i"},"100"},
+    { { "%6.5i" }, "100" },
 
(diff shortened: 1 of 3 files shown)


svenvh:
Thanks!  Would you be able to provide some more details on what corrections were made here?  I see a few changes to the testing scope, but not sure if I've caught all of them.

shajder:
> Thanks! Would you be able to provide some more details on what corrections were made here? I see a few changes to the testing scope, but not sure if I've caught all of them.

Yes, of course:

-applied suggestions from @franz , I've done some brief research and I think Michał has a point at (1) it is not certain that `sqrt(-1.0)` return `-NaN`, thus replaced with `float nan(uint nancode)`. At point (2) Michał has partialy the point because current verification of test result ignores `NaN` sign anyway, thus there is no point to distinguish between positive and negative `NaN`.

-strengthening the test by removing allowance of `Inf`/`1.#IND00`/`NaN` - according to [the spec](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#printf-format-string) only `INF`, `INFINITY`, or `NAN` for uppercase conversion format or `inf`, `infinity`, or `nan` for lowercase conversion format are permissible.

-extending the frameworks functionality to test multiple conversion formats for the same input data with the same expected output, eg. entry:
`{ { "%f", "%e", "%g", "%a" }, "1.0f/0.0f" }`
schedule 4 tests expecting the same result `inf` or `infinity` (the same applies for `single and half float`).

-rest of corrections comes from clang-format sadly, I think there is no way around (except putting clang-format off comments which  I am trying not to abuse)

test_conformance/printf/util_printf.cpp
```diff
@@ -955,48 +1038,58 @@ size_t verifyOutputBuffer(char *analysisBuffer,testCase* pTestCase,size_t testId
...
         return strcmp(eCorrectBuffer,exp);
     }
-    if(!strcmp(pTestCase->_correctBuffer[testId].c_str(),"inf"))
-    return strcmp(analysisBuffer,"inf")&&strcmp(analysisBuffer,"infinity")&&strcmp(analysisBuffer,"1.#INF00")&&strcmp(analysisBuffer,"Inf");
-    if(!strcmp(pTestCase->_correctBuffer[testId].c_str(),"nan") || !strcmp(pTestCase->_correctBuffer[testId].c_str(),"-nan")) {
-       return strcmp(analysisBuffer,"nan")&&strcmp(analysisBuffer,"-nan")&&strcmp(analysisBuffer,"1.#IND00")&&strcmp(analysisBuffer,"-1.#IND00")&&strcmp(analysisBuffer,"NaN")&&strcmp(analysisBuffer,"nan(ind)")&&strcmp(analysisBuffer,"nan(snan)")&&strcmp(analysisBuffer,"-nan(ind)");
-    }
-    return strcmp(analysisBuffer,pTestCase->_correctBuffer[testId].c_str());
+
+    if (!strcmp(pTestCase->_correctBuffer[testId].c_str(), "inf"))
```
svenvh, line 1042: Nice, that's much more readable now!  Optionally, you could also get rid of the `c_str` here and just use `==` string comparisons instead of `!strcmp`.

svenvh (approved):
Thanks for the summary, that's very helpful.  The changes look good to me.

svenvh (dismissed):

test_conformance/printf/util_printf.cpp
shajder: Absolutely, done!

shajder:
@svenvh FYI: the same problem with skipped vector cases applied to this PR as well, I've corrected it but you may want to review again, thanks.

bashbaug (approved):
Thanks, LGTM.

There are a lot of extraneous blank lines in util_printf.cpp.  How would folks feel about removing them?  This might be a good time to do it...

svenvh (approved):

svenvh:
> There are a lot of extraneous blank lines in util_printf.cpp. How would folks feel about removing them? This might be a good time to do it...

+1 for removing them, they're an eyesore.  I would suggest doing that in a separate PR though (maybe after the other printf PR is merged too, to avoid merge conflicts).

PR #1943, opened 2024-04-10
Negative tests for commandCopyImage[ToBuffer]KHR

kamil-goras-mobica:
According to description https://github.com/KhronosGroup/OpenCL-CTS/issues/1668

PR #1944, opened 2024-04-11
Added negative tests for clCommandFill[Buffer, Image]KHR

kamil-goras-mobica:

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
@@ -1,0 +1,457 @@
+//
+// Copyright (c) 2024 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "basic_command_buffer.h"
+#include "procs.h"
+#include <vector>
+
+//--------------------------------------------------------------------------
+namespace {
+
+// CL_INVALID_COMMAND_QUEUE if command_queue is not NULL.
+struct CommandBufferCommandFillQueueNotNull : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    cl_int Run() override
+    {
+        cl_int error = clCommandFillBufferKHR(
+            command_buffer, queue, out_mem, &pattern, sizeof(cl_int), 0,
+            data_size(), 0, nullptr, nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_COMMAND_QUEUE,
+                               "clCommandFillBufferKHR should return "
+                               "CL_INVALID_COMMAND_QUEUE",
+                               TEST_FAIL);
+
+        error = clCommandFillImageKHR(command_buffer, queue, src_image,
+                                      fill_color_1, origin, region, 0, nullptr,
+                                      nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_COMMAND_QUEUE,
+                               "clCommandFillImageKHR should return "
+                               "CL_INVALID_COMMAND_QUEUE",
+                               TEST_FAIL);
+
+        return CL_SUCCESS;
+    }
+
+    cl_int SetUp(int elements) override
+    {
+        cl_int error = BasicCommandBufferTest::SetUp(elements);
+        test_error(error, "BasicCommandBufferTest::SetUp failed");
+
+        src_image = create_image_2d(context, CL_MEM_WRITE_ONLY, &formats, 512,
+                                    512, 0, NULL, &error);
+        test_error(error, "create_image_2d failed");
+
+        return CL_SUCCESS;
+    }
+
+    clMemWrapper src_image;
+    const cl_uint fill_color_1[4] = { 0x05, 0x05, 0x05, 0x05 };
+    const cl_image_format formats = { CL_RGBA, CL_UNSIGNED_INT8 };
+    const size_t origin[3] = { 0, 0, 0 };
+    const size_t region[3] = { 512, 512, 1 };
+    const cl_int pattern = 0xFF;
+};
+
+// CL_INVALID_CONTEXT if the context associated with command_queue,
+// command_buffer, and buffer are not the same.
+struct CommandBufferCommandFillContextNotSame : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    cl_int Run() override
+    {
+        cl_int error = clCommandFillBufferKHR(
+            command_buffer, nullptr, out_mem_ctx, &pattern, sizeof(cl_int), 0,
+            data_size(), 0, nullptr, nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_CONTEXT,
+                               "clCommandFillBufferKHR should return "
+                               "CL_INVALID_CONTEXT",
+                               TEST_FAIL);
+
+        error = clCommandFillImageKHR(command_buffer, nullptr, dst_image_ctx,
+                                      fill_color_1, origin, region, 0, nullptr,
+                                      nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_CONTEXT,
+                               "clCommandFillImageKHR should return "
+                               "CL_INVALID_CONTEXT",
+                               TEST_FAIL);
+
+        command_buffer = clCreateCommandBufferKHR(1, &queue1, 0, &error);
+        test_error(error, "clCreateCommandBufferKHR failed");
+
+        error = clCommandFillBufferKHR(command_buffer, nullptr, out_mem,
+                                       &pattern, sizeof(cl_int), 0, data_size(),
+                                       0, nullptr, nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_CONTEXT,
+                               "clCommandFillBufferKHR should return "
+                               "CL_INVALID_CONTEXT",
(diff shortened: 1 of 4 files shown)


test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    clMemWrapper dst_image;
+    const cl_uint fill_color_1[4] = { 0x05, 0x05, 0x05, 0x05 };
+    const cl_image_format formats = { CL_RGBA, CL_UNSIGNED_INT8 };
+    const size_t origin[3] = { 0, 0, 0 };
+    const size_t region[3] = { 512, 512, 1 };
+    const cl_int pattern = 0xFF;
+};
+};
```
silverclaw, line 409: It seems this is meant to terminate the anonymous namespace.
If I'm right, then there's a trailing semi-colon here.

kamil-goras-mobica: @silverclaw Yes, this is closing anonymous namespace. This is done in the same way as other tests in this directory

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,461 @@
...
+
+        return CL_SUCCESS;
+    }
+
+    cl_int SetUp(int elements) override
+    {
+        cl_int error = BasicCommandBufferTest::SetUp(elements);
+        test_error(error, "BasicCommandBufferTest::SetUp failed");
+
+        src_image = create_image_2d(context, CL_MEM_WRITE_ONLY, &formats, 512,
```
EwanC, line 56: We should check if the device supports with [CL_DEVICE_IMAGE_SUPPORT](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_IMAGE_SUPPORT) before doing this and skip testing `clCommandFillImageKHR` if it's not supported. 

Same comment for the other tests in this file

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,461 @@
...
+                               TEST_FAIL);
+
+        error = clCommandFillImageKHR(command_buffer, nullptr, dst_image,
+                                      fill_color_1, origin, region, 0, nullptr,
+                                      nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_CONTEXT,
+                               "clCommandFillImageKHR should return "
+                               "CL_INVALID_CONTEXT",
+                               TEST_FAIL);
```
EwanC, line 116: It's not clear to me what extra coverage these test give over the what's being tested earlier in the function.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,461 @@
...
+// sync_point_wait_list are not valid synchronization-points.
+struct CommandBufferCommandFillSyncPointsNullOrNumZero
+    : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    cl_int Run() override
+    {
+        cl_sync_point_khr invalid_point = 0;
+        std::vector<cl_sync_point_khr*> invalid_sync_points;
```
EwanC, line 172: Same comment as in another PR about not needing to create a vector here, and just passing `&invalid_point` directly to function.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,461 @@
...
+    clContextWrapper context1;
+    clCommandQueueWrapper queue1;
+    clMemWrapper out_mem_ctx;
+    clMemWrapper dst_image;
+    clMemWrapper dst_image_ctx;
+    const cl_uint fill_color_1[4] = { 0x05, 0x05, 0x05, 0x05 };
+    const cl_image_format formats = { CL_RGBA, CL_UNSIGNED_INT8 };
+    const size_t origin[3] = { 0, 0, 0 };
+    const size_t region[3] = { 512, 512, 1 };
+    const cl_int pattern = 0xFF;
```
EwanC, line 157: If these are identical between all the classes can we define them in a single place, such as a base class that is inherited from.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
EwanC: Since an trailing semi-colon is an error, I don't think we should copy the convention. I'd prefer we didn't have it in new tests and then created a separate PR to remove it from existing tests.

EwanC (changes requested):
A good start but image testing needs skipped when device doesn't support images.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,461 @@
...
+    cl_int Run() override
+    {
+        cl_int error = clCommandFillBufferKHR(
+            command_buffer, queue, out_mem, &pattern, sizeof(cl_int), 0,
+            data_size(), 0, nullptr, nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_COMMAND_QUEUE,
+                               "clCommandFillBufferKHR should return "
+                               "CL_INVALID_COMMAND_QUEUE",
+                               TEST_FAIL);
```
bashbaug, line 37: See comment on https://github.com/KhronosGroup/OpenCL-CTS/pull/1943: this is only an error when `cl_khr_command_buffer_multi_device` is unsupported.

bashbaug (approved):

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
EwanC: Thanks for adding the skip, however because buffers are part of the same test it means that if a device doesn't support images then the buffer tests don't run despite being supported.

I think the file needs refactored in some way so that images & buffers are separate tests, so that if images aren't supported by a device then only the image tests skip.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,574 @@
...
+                        cl_command_queue queue)
+        : BasicCommandBufferTest(device, context, queue)
+    {}
+
+    cl_int SetUp(int elements) override
+    {
+        cl_int error = BasicCommandBufferTest::SetUp(elements);
+        test_error(error, "BasicCommandBufferTest::SetUp failed");
+
+        src_image = create_image_2d(context, CL_MEM_WRITE_ONLY, &formats, 512,
```
EwanC, line 34: We seem to be creating images even when the template `check_image_support` is set to false, we should only do them conditionally when it is set to true.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,574 @@
...
+        dst_image = create_image_2d(context, CL_MEM_READ_ONLY, &formats, 512,
+                                    512, 0, NULL, &error);
+        test_error(error, "create_image_2d failed");
+
+        return CL_SUCCESS;
+    }
+
+    bool Skip() override
+    {
+        if (check_image_support)
```
EwanC, line 47: We aren't calling ` BasicCommandBufferTest::Skip()`  when this is false, which I don't think is correct. As we don't skip without checking the base class conditions

kamil-goras-mobica:
@EwanC Added correction to Skip(). When check_image_support is false then BasicCommandBufferTest::Skip() should be called.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,576 @@
...
+        test_error(error, "Failed to create context");
+
+        out_mem_ctx =
+            clCreateBuffer(context1, CL_MEM_WRITE_ONLY,
+                           sizeof(cl_int) * num_elements, nullptr, &error);
+        test_error(error, "clCreateBuffer failed");
+
+        dst_image_ctx = create_image_2d(context1, CL_MEM_WRITE_ONLY, &formats,
+                                        512, 512, 0, NULL, &error);
+        test_error(error, "create_image_2d failed");
```
EwanC, line 177: We're creating an image here when it's not used in the test.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,576 @@
...
+            data_size(), 0, nullptr, nullptr, nullptr);
+
+        test_failure_error_ret(error, CL_INVALID_COMMAND_QUEUE,
+                               "clCommandFillBufferKHR should return "
+                               "CL_INVALID_COMMAND_QUEUE",
+                               TEST_FAIL);
+
+        error = clCommandFillImageKHR(command_buffer, queue, src_image,
+                                      fill_color_1, origin, region, 0, nullptr,
+                                      nullptr, nullptr);
```
EwanC, line 95: We're using an image API here, but I think this is tested as part of `commandBufferCommandFillImageQueueNotNull`

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -0,0 +1,576 @@
...
+        cl_int error = CommandFillBaseTest::SetUp(elements);
+        test_error(error, "CommandFillBaseTest::SetUp failed");
+
+        context1 = clCreateContext(0, 1, &device, nullptr, nullptr, &error);
+        test_error(error, "Failed to create context");
+
+        out_mem_ctx =
+            clCreateBuffer(context1, CL_MEM_WRITE_ONLY,
+                           sizeof(cl_int) * num_elements, nullptr, &error);
+        test_error(error, "clCreateBuffer failed");
```
EwanC, line 219: we're creating a buffer here that isn't used in the test.

EwanC (approved):
LGTM!

Issue #278, opened 2019-05-22
clcpp cases - current test case registration code have undefined behavior


gwawiork:
It looks that current registration methods for test cases in clcpp/ directory is incorrect. I compiled them using MSVC and was observing that test case names (test_case.exe -list) displayed some random unicode characters instead of test_names. Example:
8.■≈
8.■≈
test_spec_constants_initializing_variables
The problem is in the line:
::autotest::test_suite::global_test_suite().add(test_definition({ptr, name.c_str()}))
name - is the temporary variable which after exit function is not accessible.

kpet:
PR merged, closing.

It would be great if you could use one of the keywords that closes issues automatically for similar cases (see https://help.github.com/en/articles/closing-issues-using-keywords).

PR #1945, opened 2024-04-11
Delete stale files

silverclaw:
These aren't included or built anywhere.


EwanC (approved):
LGTM.

As a bit of due diligence I tried to find references to "autotest" in the repo, and could only find https://github.com/KhronosGroup/OpenCL-CTS/issues/278 which mentions the "clcpp" tests which I think used to live in `test_conformance/clcpp` but no longer exist, so don't see any reason to keep these files around unmaintained.

svenvh (approved):

bashbaug:

PR #1946, opened 2024-04-11
Remove dead code

silverclaw:
This is 7-year old CMake code that never did anything because the function invocation is missing the PROPERTIES keyword. Adding the keyword results in build errors, so just drop the dead code since everything seems to be working without it anyway.


EwanC (approved):
Nice catch

svenvh (approved):

bashbaug:

PR #1947, opened 2024-04-11
test_vulkan: don't throw from destructors

svenvh:
Only report an error (and include the error code), but don't throw an exception as that would call `terminate`.  Failure to release resources is not fatal in other parts of the CTS.

This fixes `-Wterminate` warnings:

    warning: ‘throw’ will always call ‘terminate’ [-Wterminate]
    note: in C++11 destructors default to ‘noexcept’


EwanC (approved):
We don't support these tests, so might want a review from somebody who does, but agree that throwing in destructors is bad practice

bashbaug:
Merging as discussed in the April 16th memory subgroup.

PR #1950, opened 2024-04-18
compiler: fix memory leak

svenvh:
The `test_compiler compiler_defines_for_extensions` test did not free all `malloc`'ed memory.


bashbaug (approved):
Nice catch!

We really ought to rewrite tests like these to use std::strings, but that'd be a much bigger change.

PR #1952, opened 2024-04-19
Fixed incorrect usage of clGetDeviceIDs num_devices

paulfradgley:
clGetDeviceIDs 'num_devices' output parameter is described as:
> num_devices returns the number of OpenCL devices available that match device_type.

but the _test_events out_of_order_event_waitlist_multi_queue_multi_device_ test expects that after calling:
`clGetDeviceIDs(platform, CL_DEVICE_TYPE_ALL, 2, two_device_ids, &number_returned);`

the content of number_returned needs to be 2, but it should be valid to return a larger number.

### DIFF
--- a/test_conformance/events/test_event_dependencies.cpp
+++ b/test_conformance/events/test_event_dependencies.cpp
@@ -97,7 +97,7 @@ int test_event_enqueue_wait_for_events_run_test(
         error = clGetDeviceIDs(platform, CL_DEVICE_TYPE_ALL, 2, two_device_ids,
                                &number_returned);
         test_error(error, "clGetDeviceIDs for CL_DEVICE_TYPE_ALL failed.");
-        if (number_returned != 2)
+        if (number_returned >= 2)
         {
             log_info("Failed to obtain two devices. Test can not run.\n");
             free(two_device_ids);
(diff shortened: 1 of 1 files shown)


test_conformance/events/test_event_dependencies.cpp
```diff
@@ -97,7 +97,7 @@ int test_event_enqueue_wait_for_events_run_test(
         error = clGetDeviceIDs(platform, CL_DEVICE_TYPE_ALL, 2, two_device_ids,
                                &number_returned);
         test_error(error, "clGetDeviceIDs for CL_DEVICE_TYPE_ALL failed.");
-        if (number_returned != 2)
+        if (number_returned >= 2)
```
bashbaug, line 100: Is this check backwards?  I think we'd want to skip this test if we get fewer than two devices, not if we get two or more devices.

```suggestion
        if (number_returned < 2)
```

paulfradgley: Agreed, that is what was intended. Thanks Ben.

bashbaug (approved):

svenvh (approved):

bashbaug:

PR #1953, opened 2024-04-19
Escape subnormal values

shiltian:
Currently we don't escape subnormal values when generating image data. In sampler read tests, we use `!=` to check the two values even when it is floating-point data, which requires the two values are bitwise equal. However, a sampler might flush subnormal values, causing the test case to fail.

In this patch, when generating random image data, we escape subnormal values.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -1125,13 +1125,15 @@ cl_ulong get_image_size_mb(image_descriptor const *imageInfo)
 uint64_t gRoundingStartValue = 0;
 
 
-void escape_inf_nan_values(char *data, size_t allocSize)
+void escape_inf_nan_subnormal_values(char *data, size_t allocSize)
 {
     // filter values with 8 not-quite-highest bits
     unsigned int *intPtr = (unsigned int *)data;
     for (size_t i = 0; i<allocSize>> 2; i++)
     {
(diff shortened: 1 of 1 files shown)


b-sumner:
Section 8.3.3 of the C spec specifically allows sampling to flush denorms, so this change seems to be long overdue.

lakshmih:
Reviewing

bashbaug (approved):
I think I'm OK with this, but noting for completeness: this change removes all fp32 and fp16 denorm values from the input data.  If we want to have _some_ coverage for denorm inputs still, we will want to continue to include denorm inputs, and instead we will want to handle cases where denorms are preserved and when denorms are flushed to zero.

bashbaug:
A bit of spec archaeology, for the Khronos folks:
* A fix similar to this one was proposed in Bugzilla 16324.
* The spec changes were discussed in GitLab issues 30 and 113.

bashbaug (approved):
Re-approving, looks like only whitespace / formatting changed.

shiltian:
I don't have privileges to merge the PR. Can someone do it for me? Thanks!

bashbaug:
It'd be nice to get one more approval before merging - @silverclaw or @lakshmih maybe?

If not, we'll aim to merge in our next teleconference, on Tuesday (June 4th).

silverclaw (approved):
LGTM

lakshmih (approved):

bashbaug:
Thanks!  Merging.

PR #1954, opened 2024-04-19
Update run_conformance.py

y1k2chaitanya:
include retry option for failed tests


y1k2chaitanya (commented):
Hello,
Tried to include retry option for failed subtests

Issue #1703, opened 2023-04-17
cl_khr_command_buffer_multi_device Test Plan


EwanC:
CTS test plan for [cl_khr_command_buffer_multi_device](https://gitlab.khronos.org/opencl/OpenCL-Docs/-/merge_requests/192)

## 1 Platform Info Query

- [ ] Test platform can be queried for `CL_PLATFORM_COMMAND_BUFFER_CAPABILITIES_KHR`

## 2 Device Info Query

- [ ] Test device can be queried for `CL_DEVICE_COMMAND_BUFFER_NUM_SYNC_DEVICES_KHR` &
      `CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR`
	  
## 3 Multiple command-queues, same device

- [ ] Test case for command-buffer recording commands to multiple queues
      targeting the same device:

* Query if a device supports recording a command-buffer to multiple queues with `CL_COMMAND_BUFFER_CAPABILITY_MULTIPLE_QUEUE_KHR`
* If such a device exits, create a command-buffer for that device passing > 1 queues to execute on.
* Record at least one command to each queue using `cl_sync_point_khr` to sync
* Finalize command buffer
* Enqueue
* If simultaneous enqueue is supported by the device, test enqueueing the command-buffer again while the first instance is pending
* Flush and verify results

## 3 Multiple devices with native sync

- [ ] Test case for command-buffer recording commands to command-queues
      attached to different devices that can natively synchronize with `CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR`

* Query to see if device supports recording a command-queue across devices in `CL_DEVICE_COMMAND_BUFFER_SYNC_DEVICES_KHR`
* If so, create a command-buffer passing 1 queue for each device with can be
  recorded against.
* Record at least one command to each queue using `cl_sync_point_khr` to sync
* Finalize command buffer
* Enqueue
* If simultaneous enqueue is supported by all devices, test enqueueing the command-buffer again while the first instance is pending
* Flush and verify results

## 4 Multiple devices with emulated sync
	
- [ ] Test case for command-buffer recording commands to command-queues
      attached to different devices that can synchronize with `CL_COMMAND_BUFFER_PLATFORM_UNIVERSAL_SYNC_KHR`

* Query to see if a platform supports  `CL_COMMAND_BUFFER_PLATFORM_UNIVERSAL_SYNC_KHR`
[...]

PR #1956, opened 2024-04-22
Negative tests clCommandSvmMem[cpy, Fill]KHR

kamil-goras-mobica:


test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -0,0 +1,413 @@
...
+        if (svm_out_mem() == nullptr)
+        {
+            test_error(CL_OUT_OF_RESOURCES, "Unable to allocate SVM memory");
+        }
+
+
+        return CL_SUCCESS;
+    }
+
+    cl_int CreateKernelWithDifferentContext()
```
EwanC, line 123: There's no kernel parameter to these SVM functions, the existence of this error wording is a mistake that should be fixed by https://github.com/KhronosGroup/OpenCL-Docs/pull/1149 So we can remove this kernel creation code.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -0,0 +1,413 @@
...
+            TEST_FAIL);
+
+        return CL_SUCCESS;
+    }
+
+    bool Skip() override
+    {
+        return BasicSVMCommandBufferTest::Skip()
+            || is_extension_available(device,
+                                      "cl_khr_command_buffer_multi_device");
```
EwanC, line 96: Even with the multidevice extension I think this would still be an error, so don't need to skip with the presence of this extension.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+
+        return CL_SUCCESS;
+    }
+
+    bool Skip() override
+    {
+        return BasicSVMCommandBufferTest::Skip()
+            || is_extension_available(device,
+                                      "cl_khr_command_buffer_multi_device");
```
EwanC, line 239: Even with the multidevice extension I think this would still be an error, so don't need to skip with the presence of this extension.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -0,0 +1,413 @@
...
+                               TEST_FAIL);
+
+        return CL_SUCCESS;
+    }
+
+    bool Skip() override
+    {
+        return BasicSVMCommandBufferTest::Skip()
+            || is_extension_available(device,
+                                      "cl_khr_command_buffer_multi_device");
```
EwanC, line 279: Even with the multidevice extension I think this would still be an error, so don't need to skip with the presence of this extension.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -0,0 +1,413 @@
...
+            TEST_FAIL);
+
+        return CL_SUCCESS;
+    }
+
+    bool Skip() override
+    {
+        return BasicSVMCommandBufferTest::Skip()
+            || is_extension_available(device,
+                                      "cl_khr_command_buffer_multi_device");
```
EwanC, line 321: Even with the multidevice extension I think this would still be an error, so don't need to skip with the presence of this extension.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -0,0 +1,413 @@
...
+            "clCommandSVMMemFillKHR should return CL_INVALID_VALUE", TEST_FAIL);
+
+        return CL_SUCCESS;
+    }
+
+    bool Skip() override
+    {
+        return BasicSVMCommandBufferTest::Skip()
+            || is_extension_available(device,
+                                      "cl_khr_command_buffer_multi_device");
```
EwanC, line 360: Even with the multidevice extension I think this would still be an error, so don't need to skip with the presence of this extension.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -0,0 +1,413 @@
...
+            TEST_FAIL);
+
+        return CL_SUCCESS;
+    }
+
+    bool Skip() override
+    {
+        return BasicSVMCommandBufferTest::Skip()
+            || is_extension_available(device,
+                                      "cl_khr_command_buffer_multi_device");
```
EwanC, line 56: Not sure we need this check, we'll have only set a single command-queue when `command_buffer` was created. So I don't think we satisfy the error wording for `CL_INVALID_COMMAND_QUEUE` and would remove this multi-device extension check.


[CL_INVALID_COMMAND_QUEUE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_COMMAND_QUEUE) if the [cl_khr_command_buffer_multi_device](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#cl_khr_command_buffer_multi_device) extension is supported; and either command_queue is NULL and command_buffer was created with more than one queue, or command_queue is not NULL and not a command-queue listed on command_buffer creation.

kamil-goras-mobica: @EwanC so I think this Skip() could be removed, but there could be case it extension is supported, command_queue it not null but command_buffer is created with different command_queue. What do you think about it?

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
kamil-goras-mobica: @EwanC I will remove Skip() for this test.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
kamil-goras-mobica: @EwanC I will remove Skip() for this test.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
kamil-goras-mobica: @EwanC I will remove Skip() for this test.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
kamil-goras-mobica: @EwanC I will remove Skip() for this test.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
kamil-goras-mobica: @EwanC I will remove Skip() for this test.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
kamil-goras-mobica: @EwanC I see that "kernel" word is a mistake as there is no parameter for kernel. Then when INVALID_CONTEXT should be returned? Should I pass not null command_queue created with different context and command_buffer created with different command_queue (with different context) ? What do you think about it?

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -0,0 +1,413 @@
...
+    cl_int Run() override
+    {
+        cl_int error = clCommandSVMMemcpyKHR(
+            command_buffer, nullptr, svm_out_mem(), svm_in_mem(), data_size(),
+            0, nullptr, nullptr, nullptr);
+
+        test_failure_error_ret(
+            error, CL_INVALID_CONTEXT,
+            "clCommandSVMMemcpyKHR should return CL_INVALID_CONTEXT",
+            TEST_FAIL);
```
bashbaug, line 78: FWIW, I think it's ambiguous what should happen in this scenario.  The only error condition described in the spec for [clEnqueueSVMMemcpy](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueSVMMemcpy) for CL_INVALID_CONTEXT is:

> [CL_INVALID_CONTEXT](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_CONTEXT) if the context associated with command_queue and events in event_wait_list are not the same.

I could certainly imagine this would be an error for some implementations (but which error is not clearly specified).  For other implementations though, such as those that support system SVM, it would probably be more work to generate an error in this scenario than it would be for it to "just work".

bashbaug: This test case was removed in the latest commit - works for me.

bashbaug (approved):

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
EwanC: I think it could be a valid test, but I think its scope fits better for when we get round to writing cl_khr_command_buffer_multi_device tests, I've made a note in https://github.com/KhronosGroup/OpenCL-CTS/issues/1703

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
EwanC: > Should I pass not null command_queue created with different context and command_buffer created with different command_queue (with different context) ? What do you think about it?

I would have suggested that, but Ben raised I point I hadn't considered in https://github.com/KhronosGroup/OpenCL-CTS/pull/1956#discussion_r1606114591 so it's fine to leave it out for now.

EwanC (dismissed):

bashbaug:
Discussed in the May 21st teleconference and agreed to merge once the conflicts are resolved.

EwanC (approved):

PR #1957, opened 2024-04-26
Fix build errors related with variable defined array length and gl te…

jujiang-del:
…sts logged error

### DIFF
--- a/test_conformance/api/test_native_kernel.cpp
+++ b/test_conformance/api/test_native_kernel.cpp
@@ -47,7 +47,8 @@ int test_native_kernel(cl_device_id device, cl_context context, cl_command_queue
 
     clMemWrapper streams[ 2 ];
 #if !(defined (_WIN32) && defined (_MSC_VER))
-    cl_int inBuffer[ n_elems ], outBuffer[ n_elems ];
+    cl_int* inBuffer  = new cl_int[n_elems * sizeof(cl_int)];
+    cl_int* outBuffer = new cl_int[n_elems * sizeof(cl_int)];
 #else
     cl_int* inBuffer  = (cl_int *)_malloca( n_elems * sizeof(cl_int) );
     cl_int* outBuffer = (cl_int *)_malloca( n_elems * sizeof(cl_int) );
@@ -109,7 +110,10 @@ int test_native_kernel(cl_device_id device, cl_context context, cl_command_queue
             return 1;
         }
     }
-
+#if !(defined (_WIN32) && defined (_MSC_VER))
+    delete [] inBuffer;
+    delete [] outBuffer;
+#endif
     return 0;
 }
 
--- a/test_conformance/gl/test_buffers.cpp
+++ b/test_conformance/gl/test_buffers.cpp
@@ -128,8 +128,9 @@ int test_buffer_kernel(cl_context context, cl_command_queue queue,
     clMemWrapper streams[3];
     size_t dataSize = numElements * 16 * sizeof(cl_long);
 #if !(defined(_WIN32) && defined(_MSC_VER))
-    cl_long inData[numElements * 16], outDataCL[numElements * 16],
-        outDataGL[numElements * 16];
+    cl_long *inData = (cl_long *)malloc(dataSize);
+    cl_long *outDataCL = (cl_long *)malloc(dataSize);
+    cl_long *outDataGL = (cl_long *)malloc(dataSize);
 #else
     cl_long *inData = (cl_long *)_malloca(dataSize);
     cl_long *outDataCL = (cl_long *)_malloca(dataSize);
--- a/test_conformance/relationals/test_shuffles.cpp
+++ b/test_conformance/relationals/test_shuffles.cpp
@@ -621,9 +621,9 @@ int test_shuffle_dual_kernel(cl_context context, cl_command_queue queue,
     typeSize = get_explicit_type_size( vecType );
 
 #if !(defined(_WIN32) && defined (_MSC_VER))
-    cl_long inData[ inVecSize * numOrders ];
-    cl_long inSecondData[ inVecSize * numOrders ];
-    cl_long outData[ outRealVecSize * numOrders ];
+    cl_long* inData  = new cl_long[inVecSize * numOrders * sizeof(cl_long)];
+    cl_long* inSecondData  = new cl_long[inVecSize * numOrders * sizeof(cl_long)];
+    cl_long* outData = new cl_long[outRealVecSize * numOrders * sizeof(cl_long)];
 #else
     cl_long* inData  = (cl_long*)_malloca(inVecSize * numOrders * sizeof(cl_long));
     cl_long* inSecondData  = (cl_long*)_malloca(inVecSize * numOrders * sizeof(cl_long));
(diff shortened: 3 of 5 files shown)


test_conformance/api/test_native_kernel.cpp
```diff
@@ -46,8 +46,9 @@ int test_native_kernel(cl_device_id device, cl_context context, cl_command_queue
     }
 
     clMemWrapper streams[ 2 ];
-#if !(defined (_WIN32) && defined (_MSC_VER))
-    cl_int inBuffer[ n_elems ], outBuffer[ n_elems ];
+#if !(defined(_WIN32) && defined(_MSC_VER))
```
svenvh, line 49: It's probably better to get rid of the entire `#if`/`#else` block and drop the non-standard `_malloca`, and use `std::vector` instead.  That way we won't need to worry about freeing the memory using `delete[]` either.

jujiang-del: Thanks. Updated.

svenvh: Thanks!

test_conformance/api/test_native_kernel.cpp
```diff
@@ -97,15 +94,19 @@ int test_native_kernel(cl_device_id device, cl_context context, cl_command_queue
...
-    error = clEnqueueReadBuffer( queue, streams[ 1 ], CL_TRUE, 0, n_elems * sizeof(cl_int), outBuffer, 0, NULL, NULL );
+    error = clEnqueueReadBuffer(queue, streams[1], CL_TRUE, 0,
+                                n_elems * sizeof(cl_int), outBuffer.data(), 0,
+                                NULL, NULL);
     test_error( error, "Unable to read results" );
 
     for( int i = 0; i < n_elems; i++ )
     {
-        if( inBuffer[ i ] != outBuffer[ i ] )
+        if (inBuffer[i * sizeof(cl_int)] != outBuffer[i * sizeof(cl_int)])
```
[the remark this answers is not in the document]
jujiang-del: Thanks. Updated.

test_conformance/api/test_native_kernel.cpp
```diff
@@ -46,12 +46,8 @@ int test_native_kernel(cl_device_id device, cl_context context, cl_command_queue
...
 
     clMemWrapper streams[ 2 ];
-#if !(defined (_WIN32) && defined (_MSC_VER))
-    cl_int inBuffer[ n_elems ], outBuffer[ n_elems ];
-#else
-    cl_int* inBuffer  = (cl_int *)_malloca( n_elems * sizeof(cl_int) );
-    cl_int* outBuffer = (cl_int *)_malloca( n_elems * sizeof(cl_int) );
-#endif
+    std::vector<cl_int> inBuffer(n_elems * sizeof(cl_int)),
+        outBuffer(n_elems * sizeof(cl_int));
```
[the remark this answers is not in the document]
jujiang-del: Thanks. Updated.

svenvh (dismissed):

test_conformance/gl/test_buffers.cpp
```diff
@@ -127,14 +127,9 @@ int test_buffer_kernel(cl_context context, cl_command_queue queue,
...
-#if !(defined(_WIN32) && defined(_MSC_VER))
-    cl_long inData[numElements * 16], outDataCL[numElements * 16],
-        outDataGL[numElements * 16];
-#else
-    cl_long *inData = (cl_long *)_malloca(dataSize);
-    cl_long *outDataCL = (cl_long *)_malloca(dataSize);
-    cl_long *outDataGL = (cl_long *)_malloca(dataSize);
-#endif
+    std::vector<cl_long> inData(dataSize), outDataCL(dataSize),
+        outDataGL(dataSize);
```
bashbaug, line 131: We should update the computation of `dataSize` after this change, since we want it to be the data size in number of elements now, not in the number of bytes.  (Basically, remove `* sizeof(cl_long)`.)

test_conformance/relationals/test_shuffles.cpp
```diff
@@ -619,38 +619,33 @@ int test_shuffle_dual_kernel(cl_context context, cl_command_queue queue,
         return error;
 
     typeSize = get_explicit_type_size( vecType );
+    std::vector<cl_long> inData(inVecSize * numOrders * sizeof(cl_long));
+    std::vector<cl_long> inSecondData(inVecSize * numOrders * sizeof(cl_long));
+    std::vector<cl_long> outData(outRealVecSize * numOrders * sizeof(cl_long));
```
bashbaug, line 624: We want to pass the number of elements, not the numbre of bytes.

```suggestion
    std::vector<cl_long> inData(inVecSize * numOrders);
    std::vector<cl_long> inSecondData(inVecSize * numOrders);
    std::vector<cl_long> outData(outRealVecSize * numOrders);
```

bashbaug (commented):
Thanks, this is a nice improvement!  Just a few comments.

test_conformance/gl/test_buffers.cpp
```diff
@@ -168,21 +163,21 @@ int test_buffer_kernel(cl_context context, cl_command_queue queue,
     bufferSize = numElements * vecSize * get_explicit_type_size(vecType);
 
     /* Generate some almost-random input data */
-    gen_input_data(vecType, vecSize * numElements, d, inData);
-    memset(outDataCL, 0, dataSize);
-    memset(outDataGL, 0, dataSize);
+    gen_input_data(vecType, vecSize * numElements, d, inData.data());
+    outDataCL.clear();
+    outDataGL.clear();
```
bashbaug, line 168: One more comment, sorry: `clear` modifies the vector's "size", so I don't think it's what we want to call here.  Note, it doesn't modify the vector's "capacity", and looks like the subsequent part of this test just accesses the vector via `data`, so this "works" but it'd really be better if we didn't modify the vector's size.

https://en.cppreference.com/w/cpp/container/vector/clear

Could we use `assign` or `fill` instead?

https://en.cppreference.com/w/cpp/container/vector/assign
https://en.cppreference.com/w/cpp/algorithm/fill

lakshmih:
Reviewing

bashbaug:
Hi @jujiang-del - did you have a chance to look at the comments above?  We discussed this PR in our teleconference today and ideally they would be addressed before merging - thanks!

jujiang-del:
Thanks for the review. Updated accordingly.

bashbaug (commented):
Mostly LGTM, thanks!

test_conformance/relationals/test_shuffles.cpp
```diff
@@ -618,39 +618,34 @@ int test_shuffle_dual_kernel(cl_context context, cl_command_queue queue,
...
-    cl_long* outData = (cl_long*)_malloca(outRealVecSize * numOrders * sizeof(cl_long));
-#endif
-    memset(outData, 0, outRealVecSize * numOrders * sizeof(cl_long) );
+    typeSize = get_explicit_type_size(vecType);
+    std::vector<cl_long> inData(inVecSize * numOrders);
+    std::vector<cl_long> inSecondData(inVecSize * numOrders);
+    std::vector<cl_long> outData(outRealVecSize * numOrders);
 
-    generate_random_data( vecType, (unsigned int)( numOrders * inVecSize ), d, inData );
+    outData.clear();
```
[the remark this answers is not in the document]
svenvh: Good catch; I'd say it's best to rely on the default initialization and avoid unnecessary `std::fill` calls.

jujiang-del:
Updated. Thanks.

svenvh (approved):

PR #1911, opened 2024-03-09
musl-libc doesn't provide GNU version of strerror_r but uses BSD/XSI one

okias:

PR #1958, opened 2024-04-29
Fix build breakage introduced by #1911

lakshmih:


lakshmih:
1911 is breaking the Android build for us, so would appreciate if this fix or an equivalent one could be reviewed and merged.

svenvh:
@okias since you authored #1911: does this proposed fix still work with your use case?

okias:
@svenvh still working properly for musl-libc, thanks for the ping!

svenvh (approved):

bashbaug:
Merging as discussed in the May 7th teleconference.

Two areas for follow-up:
* Can we do this check in CMake, to choose the right version of `strerror_r`?
* Why wasn't this break found in CI?

PR #1959, opened 2024-05-01
Add special test for generic address space atomics

silverclaw:
The atomic operations are tested with generic pointers but in a way where the compiler can infer the original pointer address space. This commit adds tests specifically for the case where the compiler cannot make inferences.
Test that the correct address is used with atomics when the address is group variant and invariant.

### DIFF
--- a/test_conformance/generic_address_space/atomic_tests.cpp
+++ b/test_conformance/generic_address_space/atomic_tests.cpp
@@ -1,0 +1,224 @@
+//
+// Copyright (c) 2024 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <vector>
+
+#ifdef __APPLE__
+#include <OpenCL/opencl.h>
+#else
+#include <CL/cl.h>
+#endif
+
+#include "errorHelpers.h"
+#include "typeWrappers.h"
+
+namespace {
+// In this source, each workgroup will generate one value.
+// Every other workgroup will use either a global or local
+// pointer on an atomic operation.
+const char* KernelSourceInvariant = R"OpenCLC(
+kernel void testKernel(global atomic_int* globalPtr, local atomic_int* localPtr) {
+    int gid = get_global_id(0);
+    int tid = get_local_id(0);
+    int wgid = get_group_id(0);
+    int wgsize = get_local_size(0);
+
+    if (tid == 0) atomic_store(localPtr, 0);
+
+    barrier(CLK_LOCAL_MEM_FENCE);
+
+    // Initialise the generic pointer to
+    // the global.
+    generic atomic_int* ptr = globalPtr + wgid;
+
+    // In a workgroup-invariant way, select a localPtr instead.
+    if ((wgid % 2) == 0)
+        ptr = localPtr;
+
+    int inc = atomic_fetch_add(ptr, 1);
+
+    // In the cases where the local memory ptr was used,
+    // save off the final value.
+    if ((wgid % 2) == 0 && inc == (wgsize-1))
+        atomic_store(&globalPtr[wgid], inc);
+}
+)OpenCLC";
+
+// In this source, each workgroup will generate two values.
+// Every other work item in the workgroup will select either
+// a local or global memory pointer and perform an atomic
+// operation on that.
+const char* KernelSourceVariant = R"OpenCLC(
+kernel void testKernel(global atomic_int* globalPtr, local atomic_int* localPtr) {
+    int gid = get_global_id(0);
+    int tid = get_local_id(0);
+    int wgid = get_group_id(0);
+    int wgsize = get_local_size(0);
+
+    if (tid == 0) atomic_store(localPtr, 0);
+
+    barrier(CLK_LOCAL_MEM_FENCE);
+
+    // Initialise the generic pointer to
+    // the global.  Two values are written per WG.
+    generic atomic_int* ptr = globalPtr + (wgid * 2);
+
+    // In a workgroup-invariant way, select a localPtr instead.
+    if ((tid % 2) == 0)
+        ptr = localPtr;
+
+    atomic_fetch_add(ptr, 1);
+
+    barrier(CLK_LOCAL_MEM_FENCE);
+
+    // In the cases where the local memory ptr was used,
+    // save off the final value.
+    if (tid == 0)
+        atomic_store(&globalPtr[(wgid * 2) + 1], atomic_load(localPtr));
+}
+)OpenCLC";
+}
+
+int test_generic_atomics_invariant(cl_device_id deviceID, cl_context context,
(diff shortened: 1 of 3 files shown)


test_conformance/generic_address_space/atomic_tests.cpp
```diff
@@ -0,0 +1,231 @@
...
+    clProgramWrapper program = clCreateProgramWithSource(
+        context, 1, &KernelSourceInvariant, nullptr, &err);
+    test_error(err, "clCreateProgramWithSource failed");
+
+    err = clBuildProgram(program, 1, &deviceID, "-cl-std=CL2.0", nullptr,
+                         nullptr);
+    test_error(err, "clBuildProgram failed");
+
+    clKernelWrapper kernel = clCreateKernel(program, "testKernel", &err);
+    test_error(err, "clCreateKernel failed");
```
bashbaug, line 111: It would be best to use `create_single_kernel_helper` here.  Among other things, it will handle passing the right `-cl-std` version, which is important because many OpenCL 3.0 devices do not support OpenCL C 2.0.

test_conformance/generic_address_space/atomic_tests.cpp
```diff
@@ -0,0 +1,231 @@
...
+    clProgramWrapper program = clCreateProgramWithSource(
+        context, 1, &KernelSourceVariant, nullptr, &err);
+    test_error(err, "clCreateProgramWithSource failed");
+
+    err = clBuildProgram(program, 1, &deviceID, "-cl-std=CL2.0", nullptr,
+                         nullptr);
+    test_error(err, "clBuildProgram failed");
+
+    clKernelWrapper kernel = clCreateKernel(program, "testKernel", &err);
+    test_error(err, "clCreateKernel failed");
```
bashbaug, line 177: This should use `create_single_kernel_helper` too.

test_conformance/generic_address_space/atomic_tests.cpp
```diff
@@ -0,0 +1,231 @@
...
+    test_error(err, "clEnqueueNDRangeKernel failed");
+
+    std::vector<cl_int> results(numWGs);
+    err = clEnqueueReadBuffer(queue, buffer, CL_TRUE, 0, bufferSize,
+                              results.data(), 0, nullptr, nullptr);
+    test_error(err, "clEnqueueReadBuffer failed");
+
+    clFinish(queue);
+
+    for (size_t i = 0; i < wgSize; ++i)
```
bashbaug, line 147: I'm not sure if the kernel is incorrect or if this verification code is, but each work-group is writing a value, not each work-item.  One solution might be to verify the number of work-groups, not the work-group size:

```suggestion
    for (size_t i = 0; i < numWGs; ++i)
```

test_conformance/generic_address_space/atomic_tests.cpp
```diff
@@ -0,0 +1,231 @@
...
+                         nullptr);
+    test_error(err, "clBuildProgram failed");
+
+    clKernelWrapper kernel = clCreateKernel(program, "testKernel", &err);
+    test_error(err, "clCreateKernel failed");
+
+    size_t wgSize, retSize;
+    err = clGetKernelWorkGroupInfo(kernel, deviceID,
+                                   CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE,
+                                   sizeof(wgSize), &wgSize, &retSize);
```
bashbaug, line 116: This is probably not going to cause an error or the test to fail, but can you please confirm you are intending to query [CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE) here, and not [CL_KERNEL_WORK_GROUP_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_KERNEL_WORK_GROUP_SIZE)?

test_conformance/generic_address_space/atomic_tests.cpp
```diff
@@ -0,0 +1,231 @@
...
+        ptr = localPtr;
+
+    int inc = atomic_fetch_add(ptr, 1);
+
+    barrier(CLK_LOCAL_MEM_FENCE);
+
+    // In the cases where the local memory ptr was used,
+    // save off the final value.
+    if ((tid % 2) == 0)
+        globalPtr[(wgid * 2) + 1] = inc;
```
bashbaug, line 89: I think there is a race condition here: `(tid % 2) == 0` will be true for many work-items, and they are all going to try to write their own values of `inc` to the same memory location.  Should this be more like the invariant kernel, where it only writes the value to memory when `inc` has a specific value?

lakshmih:
Reviewing and will have some feedback next week.

test_conformance/generic_address_space/atomic_tests.cpp
silverclaw: My objective was to select a workgroup size that the device would execute in a single SIMD unit (commonly referred to as "warp" or "wavefront"). I assume that implementers will return that on the PREFERRED workgroup multiple query.
Alternatively, I could query the subgroup size.

test_conformance/generic_address_space/atomic_tests.cpp
silverclaw: This was a copy-paste error, sorry.

test_conformance/generic_address_space/atomic_tests.cpp
silverclaw: Should be sorted. And yes, it should have been numWGs.

test_conformance/generic_address_space/atomic_tests.cpp
silverclaw: Should be done now

test_conformance/generic_address_space/atomic_tests.cpp
silverclaw: Should be done now

test_conformance/generic_address_space/atomic_tests.cpp
```diff
@@ -0,0 +1,221 @@
...
+    int wgid = get_group_id(0);
+    int wgsize = get_local_size(0);
+
+    if (tid == 0) *localPtr = 0;
+
+    barrier(CLK_LOCAL_MEM_FENCE);
+
+    // Initialise the generic pointer to
+    // the global.  Two values are written per WG.
+    generic int* ptr = globalPtr + (wgid * 2);
```
lakshmih, line 76: Looks like this should be generic atomic_int* since it is used in the atomic_fetch_add function

lakshmih (changes requested):

test_conformance/generic_address_space/atomic_tests.cpp
silverclaw: I patched this up. Had to add some atomic_load and atomic_stores to match.

lakshmih (approved):

test_conformance/generic_address_space/atomic_tests.cpp
bashbaug: I see.  FWIW, I'm aware of some devices where this assumption doesn't hold, but I think the test should be fine, regardless.

I'd suggest adding a comment here just to avoid confusion in the future, but not required, up to you.

bashbaug (dismissed):
Latest version LGTM, thanks!

It would be nice to fix the formatting errors before merging.

silverclaw:
Updated for formatting fixes, and added a requested comment explaining why the PREFERRED workgroup size is being queried.

bashbaug (approved):

bashbaug:

Issue #1691, opened 2023-03-31
A few missing cl_khr_semaphore tests


kpet:
- [ ] `CL_DEVICE_HANDLE_LIST_KHR` as a query using `clGetSemaphoreInfoKHR`
- [ ] `CL_DEVICE_HANDLE_LIST_KHR` as a property passed to `clCreateSemaphoreWithPropertiesKHR`
- [ ] Negative tests for the API error codes of
  - [ ] `clCreateSemaphoreWithPropertiesKHR`
  - [ ] `clEnqueueWaitSemaphoresKHR`
  - [ ] `clEnqueueSignalSemaphoresKHR`
  - [ ] `clGetSemaphoreInfoKHR`
  - [ ] `clReleaseSemaphoreKHR`
  - [ ] `clRetainSemaphoreKHR`

Note that `CL_DEVICE_HANDLE_LIST_KHR` may be renamed depending on what https://github.com/KhronosGroup/OpenCL-Docs/issues/864 concludes.

nikhiljnv:
Have filed https://github.com/KhronosGroup/OpenCL-CTS/issues/1875 for CL_DEVICE_HANDLE_LIST_KHR coverage.

PR #1962, opened 2024-05-07
Added new cl_khr_semaphore tests to verify clCreateSemaphoreWithPropertiesKHR negative results

shajder:
According to work plan from #1691, new clSemaphoreWrapper introduced to avoid duplication of code


test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
```diff
@@ -0,0 +1,672 @@
...
+            (cl_semaphore_properties_khr)CL_SEMAPHORE_HANDLE_OPAQUE_FD_KHR,
+            (cl_semaphore_properties_khr)
+                CL_SEMAPHORE_EXPORT_HANDLE_TYPES_LIST_END_KHR,
+            0
+        };
+
+        semaphore_second =
+            clCreateSemaphoreWithPropertiesKHR(context, sema_2_props, &err);
+        test_failure_error(
+            err, CL_INVALID_DEVICE,
```
saurabhnv, line 593: Shouldn't this be **CL_INVALID_OPERATION**?

shajder: Corrected, thanks!

joshqti (commented):
CL_INVALID_SEMAPHORE_KHR is an error code.  Define an invalid property local to the test, this is not correct usage of the macro

shajder:
> CL_INVALID_SEMAPHORE_KHR is an error code. Define an invalid property local to the test, this is not correct usage of the macro

@joshqti thanks, do you have other values on your mind ? Please keep in mind it supposed to be incorrect usage because this is negative test! `CL_INVALID_SEMAPHORE_KHR` is neither valid name nor property in this case.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
```diff
@@ -0,0 +1,672 @@
...
+            cl_semaphore_properties_khr sema_props[] = {
+                (cl_semaphore_properties_khr)CL_INVALID_SEMAPHORE_KHR,
+                (cl_semaphore_properties_khr)CL_SEMAPHORE_TYPE_BINARY_KHR, 0
+            };
+
+            semaphore =
+                clCreateSemaphoreWithPropertiesKHR(context, sema_props, &err);
+            test_failure_error(
+                err, CL_INVALID_PROPERTY,
+                "Unexpected clCreateSemaphoreWithPropertiesKHR return");
```
bashbaug, line 99: I think there are two possible errors here:

1. There is an invalid property, which would return `CL_INVALID_PROPERTY`.
2. The properties  do not contain the required properties to create a semaphore (specifically, the semaphore type), which would return `CL_INVALID_VALUE`.

We should either allow for both error conditions, or we should disambiguate the errors so only `CL_INVALID_PROPERTY` is returned.

paulfradgley (dismissed):
Looks good to me.

ahesham-arm (approved):

kpet:
2024/07/02 teleconference: @shajder Could you rebase/resolve conflicts please?

nikhiljnv:
Summarizing discussion from memory subgroup call of July 9th 2024.

Per Joshua's comment, using CL_INVALID_SEMAPHORE error code as a property value is confusing.
Ben suggested two options

-     Clarify in a comment that CL_INVALID_SEMAPHORE which is an error code is actually an invalid value that can be passed to test error      condition.
-     Alternatively, pass a different value instead of CL_INVALID_SEMAPHORE to avoid the confusion.

shajder:
> Per Joshua's comment, using CL_INVALID_SEMAPHORE error code as a property value is confusing. Ben suggested two options

corrected according to mentioned suggestion

nikhiljnv (approved):
LGTM.

nikhiljnv:
Merging as discussed in memory subgroup call on August 6th 2024.

PR #1963, opened 2024-05-10
Strengthen address space pointer checks

silverclaw:
Add negative tests for address space casts.

### DIFF
--- a/test_conformance/generic_address_space/basic_tests.cpp
+++ b/test_conformance/generic_address_space/basic_tests.cpp
@@ -281,15 +281,21 @@ int test_casting(cl_device_id deviceID, cl_context context, cl_command_queue que
         NL "    intp = &gint;"
         NL "    failures += !(isFenceValid(get_fence(intp)));"
         NL "    failures += !(to_global(intp));"
+        NL "    failures += (to_local(intp) != NULL);"
+        NL "    failures += (to_private(intp) != NULL);"
         NL "    failures += (*intp != 1);"
         NL
         NL "    intp = &lint;"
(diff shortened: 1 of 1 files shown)


svenvh:
Formatting violations should be ignored here.

lakshmih:
Reviewing

bashbaug (commented):
I'm not convinced these "negative tests" are possible without a spec clarification.

These `to_xxx` functions currently [say](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#address-space-qualifier-functions), for example:

> Returns a pointer that points to a region in the global address space if to_global can cast ptr to the global address space. Otherwise it returns NULL.

... but they don't make any guarantees when the pointer cannot be converted  to the requested address space.  You could imagine some CPU devices, for example, where a generic pointer could be converted to any arbitrary address space and, at least, be safely dereferenced.

FWIW, the same is true for the SPIR-V instruction [OpGenericCastToPtrExplicit](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpGenericCastToPtrExplicit), which says:

> [...] If the cast fails, the instruction result is an OpConstantNull pointer in the Storage Storage Class.

... but never says when the cast may / must / should "fail".  See internal SPIR-V GitLab issue 741.

To be clear, I think it would be great to strengthen the spec so we can add these tests also, but I think we'll need the spec change first.

lakshmih:
Reviewing

bashbaug (changes requested):
Removing "focused review" while we are discussing related spec changes.

PR #1966, opened 2024-05-21
add missing header

rjodinchr:
test_common/harness/integer_ops_test_info.h is using std::vector but is not including the header.

This is breaking on Google internal CI.


rjodinchr:
@bashbaug , Could you have a look please?

bashbaug (approved):
LGTM, merging immediately since this should be non-controversial.

PR #1969, opened 2024-05-23
fix bugs in negative command_buffer tests

franz:
- when calling command buffer APIs, test with `command_queue != NULL` should return `CL_INVALID_VALUE` only if the device doesn't support `cl_khr_command_buffer_multi_device` (added `Skip`)

- some tests enqueued commands with multiple invalid arguments, e.g. `clCommandCopyImageToBufferKHR` with two images and invalid sync points. AFAIK the order of argument checking is not defined, so implementation can return any valid error value for such API calls, but the tests assumed only one particular error would be returned. Fix the API calls to be unambiguous.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
@@ -50,6 +50,12 @@ struct CommandBufferCommandSVMQueueNotNull : public BasicSVMCommandBufferTest
     }
 
     const cl_char pattern_1 = 0x14;
+
+    bool Skip() override
+    {
+        return is_extension_available(device,
+                                      "cl_khr_command_buffer_multi_device");
+    }
 };
 
 // CL_INVALID_SYNC_POINT_WAIT_LIST_KHR if sync_point_wait_list is NULL and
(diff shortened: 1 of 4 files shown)


test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
```diff
@@ -50,6 +50,12 @@ struct CommandBufferCommandSVMQueueNotNull : public BasicSVMCommandBufferTest
     }
 
     const cl_char pattern_1 = 0x14;
+
+    bool Skip() override
+    {
+        return is_extension_available(device,
+                                      "cl_khr_command_buffer_multi_device");
```
EwanC, line 57: Think we should be doing an OR with the base class `Skip()` methods for these new additions, e.g. `BasicSVMCommandBufferTest::Skip()` here. I'm seeing this test not skip anymore on a device without SVM support.

EwanC (commented):
LGTM in general, but one comment about calling `Skip()` in base classes.

As an aside it might be worth checking out a related OpenCL-Docs PR https://github.com/KhronosGroup/OpenCL-Docs/pull/1146 @franz. It's a minor change to make `cl_khr_command_buffer_multi_device` with `clCommandBarrierWithWaitListKHR` error behavior consistent with other entry-points entry-point. Since you implement `cl_khr_command_buffer_multi_device`  it would be good to be aware.

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_svm_mem.cpp
franz: Thanks, should be fixed now

EwanC (approved):
LGTM

PR #1970, opened 2024-05-27
Disable implicit conversion of copysign arguments.

hvdijk:
In binary_float.cpp, copysign is special cased. All the reference functions there take double arguments, except for reference_copysign which takes float. This commit copies that approach to special case copysign in binary_double.cpp as well: all the reference functions there take long double arguments, but this commit changes reference_copysignl to take double. The rationale for this in binary_float applies equally to binary_double: conversions of NAN are not required to preserve its sign bit. On architectures where conversion of NAN resets the sign bit, copysign fp64 would return incorrect reference results.


test_conformance/math_brute_force/binary_double.cpp
```diff
@@ -219,6 +219,7 @@ cl_int Test(cl_uint job_id, cl_uint thread_id, void *data)
     cl_double *r;
     cl_double *s;
     cl_double *s2;
+    cl_int copysign_test = 0;
```
svenvh, line 222: This could be a `bool` of course, but I appreciate that you're following what's in `binary_float.cpp`.  So let's leave it as a `cl_int`, no changes needed here!

(hopefully one day we can deduplicate the binary_*.cpp files, so having fewer differences between them helps)

test_conformance/math_brute_force/reference_math.h
```diff
@@ -205,7 +204,8 @@ long double reference_acoshl(long double x);
 long double reference_asinhl(long double x);
 long double reference_atanhl(long double x);
 long double reference_cbrtl(long double x);
-long double reference_copysignl(long double x, long double y);
+double reference_copysignl(double x, double y);
+long double reference_copysignll(long double x, long double y);
```
svenvh, line 208: This part of the change is counter-intuitive: the reference_math function names follow the suffix convention of the C standard library math functions where a single `l` suffix means `long double`; but your change breaks that for `copysign`.

Instead I would suggest modifying the entry for `copysign` in `functionList`, to explicitly state the reference functions (replacing the `ENTRY` macro for `copysign`).

hvdijk: Here too I was just following what was already done for `copysign`: all the `reference_*` functions for `float` take `double` arguments, except for `copysign`. I agree that it's counter-intuitive though, it shouldn't be too much work to clean that up both for the `float` and for the `double` version, so will make that change.

hvdijk: Done, thanks for the suggestion!

svenvh: Thanks for tidying up the `float` case too!

svenvh (dismissed):

hvdijk:
Rebased to resolve the conflict with the fp16 merge.

svenvh (approved):

bashbaug (approved):
LGTM, matches the logic in binary_float.cpp.

bashbaug:

PR #1971, opened 2024-05-28
Fixes for Image_from_buffer_alignment_negative when alignments are 1

AhmedAmraniAkdi:
The Image_from_buffer_alignment_negative test creates images with incorrect pitches by adding 1 and tests whether the image creation fails.

Devices that return 1 for either of CL_IMAGE_REQUIREMENTS_ROW_PITCH_ALIGNMENT_EXT, CL_IMAGE_REQUIREMENTS_SLICE_PITCH_ALIGNMENT_EXT or CL_IMAGE_REQUIREMENTS_BASE_ADDRESS_ALIGNMENT_EXT will successfully create the image and therefore fail the test.

This change allows to skip the image creation in this case as the error condition (pitch % pitch_alignment != 0) will not be triggered for these devices.

### DIFF
--- a/test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
+++ b/test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
@@ -494,10 +494,11 @@ int image_from_buffer_alignment_negative(cl_device_id device,
                 test_error(err, "Unable to create buffer");
 
                 /* Test Row pitch images */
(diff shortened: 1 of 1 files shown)


kpet (approved):

bashbaug (approved):

lakshmih (approved):

bashbaug:
Merging as discussed in the June 18th teleconference.

PR #1968, opened 2024-05-23
cl_khr_kernel_clock: fix builtin function names

svenvh:
According to the [specification](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#kernel-clock-functions), the work_group and sub_group variants have an `_` in them.

PR #1973, opened 2024-05-28
kernel_clock: Update function names to match spec

lakshmih:
The spec has underscores in work_group and sub_group functions Fix the test to use that.


svenvh:
This looks like a duplicate of #1968

PR #1975, opened 2024-06-03
Add fp16 testing to conversions and bruteforce

svenvh:
Merge the `fp16-staging` branch into `main`, adding fp16 (`half`) testing to the conversions and math bruteforce tests.

### DIFF
--- a/test_common/harness/errorHelpers.cpp
+++ b/test_common/harness/errorHelpers.cpp
@@ -371,7 +371,7 @@ static float Ulp_Error_Half_Float(float test, double reference)
     return (float)scalbn(testVal - reference, ulp_exp);
 }
 
-float Ulp_Error_Half(cl_half test, float reference)
+float Ulp_Error_Half(cl_half test, double reference)
 {
     return Ulp_Error_Half_Float(cl_half_to_float(test), reference);
 }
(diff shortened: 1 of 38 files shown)


svenvh:
As we don't seem to have any remaining issues or concerns with the new fp16 conversions and math bruteforce tests, now would be a good time to merge the staging branch into the `main` branch.  Any further issues with the new tests we can address on `main` as part of normal CTS maintenance.

Most of this work has already been reviewed in other pull requests; I've structured this PR into a merge commit that merges the already-reviewed `fp16-staging` branch, followed by some small "fixup" commits.

bashbaug (approved):
I didn't do a thorough test or review, but I did run the conversions and math tests in "wimpy" mode and it looks like everything is in order.  Given that we've reviewed most of these changes when they went into the fp16-staging branch I'd be OK merging this.  We can always fix any issues we find in the main branch...

svenvh:
> Need to check if any .csv files need to be updated to account for these new tests.

The `opencl_conformance_tests_*.csv` files always list the entire bruteforce and conversions binaries without specifying subtests, so I believe that there's nothing to do for those files.

`test_conformance/spir/khr.csv` does list individual tests, but it would probably need corresponding SPIR files first.  (That's the old SPIR, not SPIR-V.)

Issue #1930, opened 2024-03-26
Strengthen testing of `CL_DEVICE_TYPE` query and use of `CL_DEVICE_TYPE*` enums in device enumeration or context creation


kpet:
- [ ] Add or modify a test that checks that **clGetDeviceInfo** called with `CL_DEVICE_TYPE`  never returns `CL_DEVICE_TYPE_DEFAULT` or `CL_DEVICE_TYPE_ALL`
- [ ] Add a test that (or check that an existing test) uses `CL_DEVICE_TYPE_ALL` to enumerate devices using **clGetDeviceIDs**
- [ ] Add a test that (or check that an existing test) uses `CL_DEVICE_TYPE_ALL` to create a context with **clCreateContextFromType**

- [ ] Add a test that (or check that an existing test) uses `CL_DEVICE_TYPE_DEFAULT` to enumerate devices using **clGetDeviceIDs**
- [ ] Add a test that (or check that an existing test) uses `CL_DEVICE_TYPE_DEFAULT` to create a context with **clCreateContextFromType**

kamil-goras-mobica:
@kpet PR created, please take a look is it what You wanted.

bashbaug:
Closing as discussed in the August 27th teleconference.

PR #1977, opened 2024-06-05
Strengthen testing of CL_DEVICE_TYPE query and use of CL_DEVICE_TYPE*

kamil-goras-mobica:
Issue: #1930


test_conformance/api/test_create_context_from_type.cpp
```diff
@@ -127,4 +127,84 @@ int test_create_context_from_type(cl_device_id deviceID, cl_context context, cl_
...
+    {
+        cl_uint num_devices = 0;
+        error = clGetContextInfo(context_to_test, CL_CONTEXT_NUM_DEVICES, sizeof(cl_uint), &num_devices, nullptr);
+        test_error(error, "clGetContextInfo CL_CONTEXT_NUM_DEVICES failed\n");
+
+        std::vector<cl_device_id> devices(num_devices);
+        error = clGetContextInfo(context_to_test, CL_CONTEXT_DEVICES, num_devices * sizeof(cl_device_id), devices.data(), nullptr);
+        test_error(error, "clGetContextInfo CL_CONTEXT_DEVICES failed\n");
+
+        test_assert_error(devices.size() == 1, "Context must contain exacly one device\n");
```
lakshmih, line 204: nit: exactly (here and some other places)

test_conformance/api/test_platform.cpp
```diff
@@ -230,14 +230,22 @@ int test_get_platform_ids(cl_device_id deviceID, cl_context context, cl_command_
...
     err = clGetDeviceIDs(platforms[p], CL_DEVICE_TYPE_ALL, 0, NULL, &num_devices);
-    test_error(err, "clGetDeviceIDs size failed.\n");
+    test_error(err, "clGetDeviceIDs failed.\n");
+    if(num_devices == 0)
+    {
+        log_error("clGetDeviceIDs must return at least one device\n");
+        total_errors++;
+    }
+    test_assert_error(num_devices > 1,
+                      "clGetDeviceIDs must return at least one device");
```
lakshmih, line 240: Condition should either 'num_devices >= 1' or 'num_devices > 0' and not 'num_devices > 1'.

lakshmih (changes requested):
Need to fix the conditional check in platform.cpp

test_conformance/api/procs.h
```diff
@@ -89,6 +89,8 @@ extern int        test_min_max_language_version(cl_device_id deviceID, cl_contex
 extern int        test_native_kernel(cl_device_id device, cl_context context, cl_command_queue queue, int n_elems );
 
 extern int      test_create_context_from_type(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements);
+extern int      test_create_context_from_type_device_type_all(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements);
+extern int      test_create_context_from_type_device_type_default(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements);
```
lakshmih, line 93: context_from_type goes beyond just creating the context by creating memory objects and enqueueing a kernel etc.
Should these other tests do that as welll?
num_elements seems to be unused in all cases and could be removed.

kamil-goras-mobica: @lakshmih No, I just wanted to create test scenarios with clCreateContextFromType with TYPE_ALL and TYPE_DEFAULT. I did not find any sufficient place so I put it here.
num_elements will be removed.

test_conformance/api/test_create_context_from_type.cpp
kamil-goras-mobica: @lakshmih here according to description:
"Add a test that (or check that an existing test) uses CL_DEVICE_TYPE_DEFAULT to enumerate devices using clGetDeviceIDs
Exactly one device must be returned"
Is here something wrong?

test_conformance/api/test_platform.cpp
kamil-goras-mobica: @lakshmih this will be changed to ">=1"

test_conformance/api/test_create_context_from_type.cpp
```diff
@@ -127,4 +127,84 @@ int test_create_context_from_type(cl_device_id deviceID, cl_context context, cl_
...
+    cl_context_properties properties[3] = {
+      (cl_context_properties)CL_CONTEXT_PLATFORM,
+      (cl_context_properties)platform,
+      0
+    };
+
+    clContextWrapper context_to_test = clCreateContextFromType(properties, CL_DEVICE_TYPE_ALL, notify_callback, NULL, &error);
+    test_error(error, "clCreateContextFromType failed");
+
+    if (context_to_test != nullptr)
```
bashbaug, line 154: I think it should be an error if `context_to_test` is `NULL`.

It'd be safest to check for this explicitly, although if `clCreateContextFromType` does not return an error then the returned context should be non-`NULL`.

test_conformance/api/test_create_context_from_type.cpp
```diff
@@ -127,4 +127,84 @@ int test_create_context_from_type(cl_device_id deviceID, cl_context context, cl_
...
+    {
+        cl_uint num_devices = 0;
+        error = clGetContextInfo(context_to_test, CL_CONTEXT_NUM_DEVICES, sizeof(cl_uint), &num_devices, nullptr);
+        test_error(error, "clGetContextInfo CL_CONTEXT_NUM_DEVICES failed\n");
+
+        std::vector<cl_device_id> devices(num_devices);
+        error = clGetContextInfo(context_to_test, CL_CONTEXT_DEVICES, num_devices * sizeof(cl_device_id), devices.data(), nullptr);
+        test_error(error, "clGetContextInfo CL_CONTEXT_DEVICES failed\n");
+
+        test_assert_error(devices.size() >= 1, "Context must contain at least one device\n");
```
bashbaug, line 164: I'm not quite following what this test is doing.  If all we want to do is to check if the context has more than one device then we only need to query `CL_CONTEXT_NUM_DEVICES`, and do not need to query `CL_CONTEXT_DEVICES`.  Note that we're not doing anything with the contents of the `devices` array here, anyhow.

Since this is creating the context for `CL_DEVICE_TYPE_ALL` should we also be checking that the context has the same number of devices as the number of devices in the platform, or is it OK to check only that the context has one or more devices?  Does it even make sense to have a context for zero devices?

test_conformance/api/test_create_context_from_type.cpp
```diff
@@ -127,4 +127,84 @@ int test_create_context_from_type(cl_device_id deviceID, cl_context context, cl_
...
+    return 0;
+}
+
+int test_create_context_from_type_device_type_default(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
+{
+    cl_device_type type;
+    cl_int error = clGetDeviceInfo(deviceID, CL_DEVICE_TYPE, sizeof(type), &type, NULL);
+    test_error(error, "clGetDeviceInfo for CL_DEVICE_TYPE failed\n");
+
+    if(type == CL_DEVICE_TYPE_ALL || type == CL_DEVICE_TYPE_DEFAULT)
```
bashbaug, line 176: See comment as above about the device type being a bitfield.

test_conformance/api/test_create_context_from_type.cpp
```diff
@@ -127,4 +127,84 @@ int test_create_context_from_type(cl_device_id deviceID, cl_context context, cl_
...
+
+    if (context_to_test != nullptr)
+    {
+        cl_uint num_devices = 0;
+        error = clGetContextInfo(context_to_test, CL_CONTEXT_NUM_DEVICES, sizeof(cl_uint), &num_devices, nullptr);
+        test_error(error, "clGetContextInfo CL_CONTEXT_NUM_DEVICES failed\n");
+
+        std::vector<cl_device_id> devices(num_devices);
+        error = clGetContextInfo(context_to_test, CL_CONTEXT_DEVICES, num_devices * sizeof(cl_device_id), devices.data(), nullptr);
+        test_error(error, "clGetContextInfo CL_CONTEXT_DEVICES failed\n");
```
bashbaug, line 202: Same comment as above - do we need to get the devices here, or do we only care about the number of devices?

If we're getting the devices, should we also check that the device in the context matches the device returned by `clGetDeviceIDs(CL_DEVICE_TYPE_DEFAULT)`?

test_conformance/api/test_create_context_from_type.cpp
bashbaug: I suspect @lakshmih is referring to the spelling mistake:

```suggestion
        test_assert_error(devices.size() == 1, "Context must contain exactly one device\n");
```

test_conformance/api/test_platform.cpp
bashbaug: Do we need both the `log_error` and the `test_assert_error`?  Seems redundant.  Can we remove the `test_assert_error` case?

test_conformance/api/test_platform.cpp
```diff
@@ -281,6 +289,15 @@ int test_get_platform_ids(cl_device_id deviceID, cl_context context, cl_command_
...
     free(devices);
+
+    err = clGetDeviceIDs(platforms[p], CL_DEVICE_TYPE_DEFAULT, 0, NULL, &num_devices);
+    test_error(err, "clGetDeviceIDs failed.\n");
+    if (num_devices != 1)
+    {
+        log_error("clGetDeviceIDs must return exacly one device\n");
+        total_errors++;
+    }
+    test_assert_error(num_devices == 1, "clGetDeviceIDs must return exacly one device");
```
bashbaug, line 300: See above - misspelling + possible redundant conditions.  Can we just do this instead?

```suggestion
    if (num_devices != 1)
    {
        log_error("clGetDeviceIDs must return exactly one device\n");
        total_errors++;
    }
```

test_conformance/device_execution/device_info.cpp
```diff
@@ -99,6 +99,15 @@ int test_device_info(cl_device_id device, cl_context context, cl_command_queue q
         return -1;
     }
 
+    cl_device_type device_type;
+    err_ret = clGetDeviceInfo(device, CL_DEVICE_TYPE, sizeof(cl_device_type), &device_type, &ret_len);
```
bashbaug, line 103: If we're checking for this case here do we also need to check for it in `test_create_context_from_type_device_type_all` and  `test_create_context_from_type_device_type_default`?

If we decide to keep the check here, please see the previous comments about the device type being a bitfield.

bashbaug (changes requested):

test_conformance/api/test_create_context_from_type.cpp
kamil-goras-mobica: @bashbaug ok then I will remove this check.

test_conformance/api/test_create_context_from_type.cpp
```diff
@@ -127,4 +127,84 @@ int test_create_context_from_type(cl_device_id deviceID, cl_context context, cl_
   return 0;
 }
 
+int test_create_context_from_type_device_type_all(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
+{
+    cl_device_type type;
+    cl_int error = clGetDeviceInfo(deviceID, CL_DEVICE_TYPE, sizeof(type), &type, NULL);
+    test_error(error, "clGetDeviceInfo for CL_DEVICE_TYPE failed\n");
+
+    if(type == CL_DEVICE_TYPE_ALL || type == CL_DEVICE_TYPE_DEFAULT)
```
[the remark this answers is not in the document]
kamil-goras-mobica: @bashbaug I will check those bitfields and then create more precise condition

test_conformance/api/test_create_context_from_type.cpp
bashbaug: I'm still not following what this part of the test is doing - can you explain?

* If we only care about the number of devices, we can just use the query for `CL_CONTEXT_NUM_DEVICES`, and do not need to query `CL_CONTEXT_DEVICES`.
* Or, should we compare the set of devices in the context that were queried using `CL_CONTEXT_DEVICES` against something?

Right now we're querying `CL_CONTEXT_DEVICES` but we don't do anything with the result of the query, which seems incorrect.

test_conformance/api/test_create_context_from_type.cpp
bashbaug: We don't look at the contents of the queried `devices` here, either.

I think we should test that the device in the context (there should be only one) is equivalent to the device returned by `clGetDeviceIDs(CL_DEVICE_TYPE_DEFAULT)`.

test_conformance/api/test_platform.cpp
bashbaug: The spelling error still exists in the latest version.

lakshmih (approved):

bashbaug (approved):
Thanks, LGTM!

bashbaug:
Merging as discussed in the August 20th teleconference.

PR #1978, opened 2024-06-05
cl_khr_kernel_clock: use create_single_kernel_helper

svenvh:
`create_single_kernel_helper` invokes `clBuildProgram` and sets `-cl-std` to the latest version supported by the context.

The vast majority of tests are using `create_single_kernel_helper` instead of the `..._create_program` variant, so use the former for the sake of consistency.

### DIFF
--- a/test_conformance/extensions/cl_khr_kernel_clock/kernel_clock.cpp
+++ b/test_conformance/extensions/cl_khr_kernel_clock/kernel_clock.cpp
@@ -117,9 +117,6 @@ public:
                                      sizeof(cl_uint), nullptr, &error);
             test_error(error, "clCreateBuffer failed");
 
-            kernel = clCreateKernel(program, "SampleClock", &error);
-            test_error(error, "Failed to create kernel");
-
             error = clSetKernelArg(kernel, 0, sizeof(out_mem), &out_mem);
             test_error(error, "clSetKernelArg failed");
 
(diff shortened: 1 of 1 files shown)


svenvh:
@AhmedAmraniAkdi please let me know in case your choice for the `..._create_program` variant was deliberate in #1960.

test_conformance/extensions/cl_khr_kernel_clock/kernel_clock.cpp
```diff
@@ -109,14 +109,10 @@ class KernelClockTest {
 
             ptr = kernel_src;
 
-            error = create_single_kernel_helper_create_program(
-                context, &program, 1, &ptr);
+            error = create_single_kernel_helper(context, &program, &kernel, 1,
+                                                &ptr, "SampleClock");
```
bashbaug, line 113: I think we can remove the explicit call to `clCreateKernel` on Line 120 below now, also?

svenvh: Good catch, thanks!  Updated.

bashbaug (approved):

AhmedAmraniAkdi (approved):

bashbaug:
Merging as discussed in the June 18th teleconference.

PR #1980, opened 2024-06-13
spirv_new: fix test_decorate.

hvdijk:
test_decorate was checking for half-precision NAN incorrectly, calling cl_half_from_float where cl_half_to_float was intended, causing a wrong expected result.

test_decorate was also printing the expected and actual results reversed, printing "got -1, want 0", when our implementation already returned 0.


bashbaug (approved):
Fix LGTM.

I was curious how we were passing these tests before this change, but looks like for whatever reason we were already computing `val = 0` when `rhs` was NaN.

svenvh (approved):

bashbaug:

PR #1984, opened 2024-06-19
Array based clUpdateMutableCommandsKHR changes

EwanC:
CTS changes to reflect the spec changes merged in https://github.com/KhronosGroup/OpenCL-Docs/pull/1045 and requires header updates from https://github.com/KhronosGroup/OpenCL-Headers/pull/245

Tested using OCK implementation from https://github.com/codeplaysoftware/oneapi-construction-kit/pull/501

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_arguments.cpp
@@ -164,32 +164,6 @@ struct MutableDispatchGlobalArguments : public MutableDispatchArgumentsTest
 
         cl_mutable_dispatch_arg_khr arg{ 1, sizeof(dst_buf_1), &dst_buf_1 };
 
-#if CL_KHR_COMMAND_BUFFER_MUTABLE_DISPATCH_EXTENSION_VERSION                   \
-    < CL_MAKE_VERSION(0, 9, 2)
-        cl_mutable_dispatch_config_khr dispatch_config{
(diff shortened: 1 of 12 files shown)


bashbaug (commented):
I don't have an easy way to test these changes (unless I pull and build the latest OCK) but they all look reasonable.

I think it's a good proof-point that the versioned extension headers work, but for the CTS tests I think we've generally been OK tracking the latest headers.  Since we don't regularly test (or even build) with older headers I think the older code would eventually break anyhow.

I think it would be a good idea to skip these tests for older versions of the extension though, otherwise they'll likely crash unpredictably if the newer functions run on an older implementation.  Could this be added?

EwanC:
> I don't have an easy way to test these changes (unless I pull and build the latest OCK) but they all look reasonable.
> 
> I think it's a good proof-point that the versioned extension headers work, but for the CTS tests I think we've generally been OK tracking the latest headers. Since we don't regularly test (or even build) with older headers I think the older code would eventually break anyhow.
> 
> I think it would be a good idea to skip these tests for older versions of the extension though, otherwise they'll likely crash unpredictably if the newer functions run on an older implementation. Could this be added?

I've removed the #ifdefs so this PR will require the matching headers to work. Also added in a [CL_DEVICE_EXTENSIONS_WITH_VERSION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_EXTENSIONS_WITH_VERSION) runtime check and skipped testing when a device doesn't support at least 0.9.2 of mutable-dispatch. Although devices which don't support that query will still run into issues

bashbaug (approved):

bashbaug:
Merging as discussed in the September 3rd teleconference + email.

PR #1987, opened 2024-06-20
spirv_new: fix test_decorate to use the device's default rounding

cycheng:
The verification code assumes the hardware uses CL_HALF_RTE, which causes a mismatch computation results when the hardware uses RTZ. Fix to use the hardware's default rounding mode.

### DIFF
--- a/test_conformance/spirv_new/test_decorate.cpp
+++ b/test_conformance/spirv_new/test_decorate.cpp
@@ -324,10 +324,11 @@ int verify_saturated_results(cl_device_id deviceID, cl_context context,
     err = clEnqueueReadBuffer(queue, res, CL_TRUE, 0, out_bytes, &h_res[0], 0, NULL, NULL);
     SPIRV_CHECK_ERROR(err, "Failed to read to output");
 
+    cl_half_rounding_mode half_rounding = get_half_rounding_mode(deviceID);
+
     for (int i = 0; i < num; i++)
     {
-        To val = compute_saturated_output<Ti, Tl, To>(h_lhs[i], h_rhs[i],
-                                                      get_half_rounding_mode(deviceID));
+        To val = compute_saturated_output<Ti, Tl, To>(h_lhs[i], h_rhs[i], half_rounding);
 
         if (val != h_res[i])
         {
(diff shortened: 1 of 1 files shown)


svenvh:
@cycheng there is a related fix #1980 currently under review.  Any chance that solves the issue for your use case too?

cycheng:
@svenvh , I am afraid not because we see the mismatch in
```
f = cl_half_to_float(cl_half_from_float(f, half_rounding));
```
The inputs are: 0xD6F0 (-111.000000) and 0x3B91 (0.945801)
f = -111.0 * 0.945801
RTE hw produces: -105.000000
RTZ hw produces: -104.983887

But I am okay to drop this pull request if #1980 can merge my changes.

svenvh:
> But I am okay to drop this pull request if https://github.com/KhronosGroup/OpenCL-CTS/pull/1980 can merge my changes.

Not sure what @hvdijk thinks about that, but we can also first land #1980 without any changes, and then you can rebase your changes on top.  They seem to be separate issues anyway.

hvdijk:
I agree, they look like separate issues to me too, but I'm happy with whatever you prefer, update my PR to include this, update this PR to include mine, or keep them separate, just let me know if I need to do anything.

cycheng:
@hvdijk , @svenvh, I am happy to close this pull request and let @hvdijk merge this change into #1980. :)

test_conformance/spirv_new/test_decorate.cpp
```diff
@@ -303,9 +324,12 @@ int verify_saturated_results(cl_device_id deviceID, cl_context context,
     err = clEnqueueReadBuffer(queue, res, CL_TRUE, 0, out_bytes, &h_res[0], 0, NULL, NULL);
     SPIRV_CHECK_ERROR(err, "Failed to read to output");
 
+    cl_half_rounding_mode half_rounding = get_half_rounding_mode(deviceID);
```
svenvh, line 327: Ideally we should only call `get_half_rounding_mode` if `std::is_same<Ti, cl_half>::value`.

cycheng: Thank you! Fixed.
I reran the decorate_* tests on our side using RTZ hardware, and all results passed.

svenvh (approved):

Issue #1897, opened 2024-02-21
test plan for CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR


aharon-abramson:
The `CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR` is, as its name suggests, an assertion from the user to the OpenCL runtime. It promises that for a command buffer or some of its ND-range kernel commands, if the user explicitly sets the number of workgroups (by specifying a non-null local work size) to be WG<sub>0</sub> when the ND-range kernel command is recorded, the new number of workgroups - if updated - WG<sub>u</sub>, will not be greater than WG<sub>0</sub>. 

The runtime may use this assertion to choose an implementation optimized for this maximum number of workgroups, or it may choose to ignore it. In any case, this decision should not affect the results of the kernel.

The test items described below shall be run in three configurations:
 - The command buffer is created with `CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR` in its properties.
 - The ND-range command is recorded with `CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR` in its properties.
 - Both the command buffer and ND-range command have `CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR` in their properties.

The test items will follow the same template:
 1. Create a kernel that does some computation and writes the number of its workgroups into an output buffer.
 2. Create the necessary buffers.
 3. Set the kernel's arguments.
 4. Create a command buffer.
 5. Record an ND-range kernel of the kernel above in the command buffer with a non-null local work size so that the resulting number of workgroups will be greater than 1.
 6. Finalize the command buffer.
 7. Enqueue the command buffer.
 8. Wait for the command buffer's completion.
 9. Validate the kernel's results and expected number of workgroups.
 10. Update the command buffer according to the test items in the table below.
 11. Enqueue the command buffer.
 12. Wait for the command buffer's completion.
 13. Validate the kernel's results and the expected number of workgroups.

| Test items | Pass/Fail criteria |
|------------|--------------------|
| WG<sub>u</sub> = WG<sub>0</sub> | Kernel results are valid. Kernel's number of workgroups is WG<sub>u</sub>. |
[...]

PR #1991, opened 2024-06-26
Added new tests for CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR flag with mutable dispatch

banan328:
Added new tests for CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR with mutable dispatch according to this test plan https://github.com/KhronosGroup/OpenCL-CTS/issues/1897

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
@@ -1,0 +1,302 @@
+//
+// Copyright (c) 2024 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include <extensionHelpers.h>
+#include "imageHelpers.h"
+#include "mutable_command_basic.h"
+
+#include <CL/cl.h>
+#include <CL/cl_ext.h>
+
+////////////////////////////////////////////////////////////////////////////////
+// mutable dispatch tests which handle following cases:
+//
+// 1. The command buffer is created with
+// CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR in its properties.
+// 2. The ND-range command is recorded with
+// CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR in its properties.
+// 3. Both the command buffer and ND-range command have
+// CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR in their properties.
+
+struct Configuration
+{
+    const cl_command_buffer_properties_khr *command_buffer_properties;
+    const cl_ndrange_kernel_command_properties_khr *ndrange_properties;
+};
+
+// Define the command buffer properties for each configuration
+const cl_command_buffer_properties_khr command_buffer_properties[] = {
+    CL_COMMAND_BUFFER_MUTABLE_DISPATCH_ASSERTS_KHR,
+    CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR, 0
+};
+
+// Define the ndrange properties
+const cl_ndrange_kernel_command_properties_khr ndrange_properties[] = {
+    CL_MUTABLE_DISPATCH_UPDATABLE_FIELDS_KHR,
+    CL_MUTABLE_DISPATCH_GLOBAL_SIZE_KHR, CL_MUTABLE_DISPATCH_ASSERTS_KHR,
+    CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR, 0
+};
+// Initialize the array of configurations
+Configuration configurations[] = { { command_buffer_properties, nullptr },
+                                   { nullptr, ndrange_properties },
+                                   { command_buffer_properties,
+                                     ndrange_properties } };
+
+template <int test_case>
+struct MutableDispatchWorkGroups : public BasicMutableCommandBufferTest
+{
+
+    MutableDispatchWorkGroups(cl_device_id device, cl_context context,
+                              cl_command_queue queue)
+        : BasicMutableCommandBufferTest(device, context, queue),
+          single_command_buffer(this)
+    {
+        config = configurations[test_case];
+    }
+
+    bool Skip() override
+    {
+        cl_mutable_dispatch_fields_khr mutable_capabilities;
+
+        bool no_additional_wgs_support =
+            !clGetDeviceInfo(
+                device, CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR,
+                sizeof(mutable_capabilities), &mutable_capabilities, nullptr)
+            && mutable_capabilities
+                & CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR;
+
+        return !no_additional_wgs_support
+            || BasicMutableCommandBufferTest::Skip();
+    }
+
+    cl_int SetUp(int elements) override
+    {
+        cl_int error = BasicMutableCommandBufferTest::SetUp(elements);
+        test_error(error, "BasicMutableCommandBufferTest::SetUp failed");
+
+        error = SetUpKernel();
+        test_error(error, "SetUpKernel failed");
+
+        cl_command_buffer_properties_khr properties[5];
+        int index = 0;
+        properties[index++] = CL_COMMAND_BUFFER_FLAGS_KHR;
+        properties[index++] = CL_COMMAND_BUFFER_MUTABLE_KHR;
+        if (config.command_buffer_properties != nullptr)
+        {
+            properties[index++] = *config.command_buffer_properties;
+            properties[index++] = *(config.command_buffer_properties + 1);
+        }
+        properties[index] = 0;
+
+        single_command_buffer =
+            clCreateCommandBufferKHR(1, &queue, properties, &error);
+        test_error(error, "clCreateCommandBufferKHR failed");
+
+        return CL_SUCCESS;
+    }
+
+    cl_int Run() override
+    {
+        const char *num_groups_kernel =
+            R"(
+                __kernel void sample_test(__global int *dst)
+            {
+                size_t tid = get_global_id(0);
+                dst[tid] = get_num_groups(0);
+            })";
+        cl_int error = create_single_kernel_helper(
+            context, &program, &kernel, 1, &num_groups_kernel, "sample_test");
+        test_error(error, "Creating kernel failed");
+
+        clMemWrapper stream = clCreateBuffer(context, CL_MEM_READ_WRITE,
+                                             sizeToAllocate, nullptr, &error);
+
+        error = clSetKernelArg(kernel, 0, sizeof(cl_mem), &stream);
+        test_error(error, "Unable to set indexed kernel arguments");
+
+        // Record an ND-range kernel of the kernel above in the command buffer
+        // with a non-null local work size so that the resulting number of
+        // workgroups will be greater than 1.
+        error = clCommandNDRangeKernelKHR(
+            single_command_buffer, nullptr, config.ndrange_properties, kernel,
+            1, nullptr, &global_work_size, &local_work_size, 0, nullptr,
+            nullptr, &command);
+        test_error(error, "clCommandNDRangeKernelKHR failed");
+
+        error = clFinalizeCommandBufferKHR(single_command_buffer);
+        test_error(error, "clFinalizeCommandBufferKHR failed");
+
+        clEventWrapper events[2];
+        error = clEnqueueCommandBufferKHR(0, nullptr, single_command_buffer, 0,
+                                          nullptr, &events[0]);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        error = clWaitForEvents(1, &events[0]);
+        test_error(error, "clWaitForEvents failed");
+
+        std::vector<cl_int> resultData;
+        resultData.resize(global_work_size);
+        error = clEnqueueReadBuffer(queue, stream, CL_FALSE, 0, sizeToAllocate,
+                                    resultData.data(), 0, nullptr, &events[1]);
+        test_error(error, "clEnqueueReadBuffer failed");
+
+        error = clWaitForEvents(1, &events[1]);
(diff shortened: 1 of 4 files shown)


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,275 @@
...
+// mutable dispatch tests which handle following cases:
+//
+// 1. The command buffer is created with CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR in its properties.
+// 2. The ND-range command is recorded with CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR in its properties.
+// 3. Both the command buffer and ND-range command have CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR in their properties.
+
+typedef struct {
+    const cl_command_buffer_properties_khr *command_buffer_properties;
+    const cl_ndrange_kernel_command_properties_khr *ndrange_properties;
+} Configuration;
```
EwanC, line 34: This is C++ rather than C, so don't need a typedef

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,275 @@
...
+        {command_buffer_properties, ndrange_properties}
+};
+
+template<int test_case>
+struct MutableDispatchWorkGroups : public BasicMutableCommandBufferTest {
+
+    MutableDispatchWorkGroups(cl_device_id device, cl_context context,
+                              cl_command_queue queue)
+            : BasicMutableCommandBufferTest(device, context, queue), out_of_order_queue(nullptr), command_buffer(this) {
+        out_of_order_support = true;
```
EwanC, line 64: Why is this being hardcoded? Will this not beak for devices that don't support `CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR` ?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,275 @@
...
+                        device, CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR,
+                        sizeof(mutable_capabilities), &mutable_capabilities, nullptr)
+                && mutable_capabilities & CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR;
+
+        bool extension_avaliable =
+                is_extension_available(device,
+                                       "cl_khr_command_buffer_mutable_dispatch")
+                == true;
+
+        return !out_of_order_support || !no_additional_wgs_support || !extension_avaliable;
```
EwanC, line 82: I don't think `out_of_order_support` should matter when testing [CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR). Whether you can modify the number of work-groups in a single command is invariant on the in-order/out-of-order properties of the command-buffer.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,275 @@
...
+struct MutableDispatchWorkGroups : public BasicMutableCommandBufferTest {
+
+    MutableDispatchWorkGroups(cl_device_id device, cl_context context,
+                              cl_command_queue queue)
+            : BasicMutableCommandBufferTest(device, context, queue), out_of_order_queue(nullptr), command_buffer(this) {
+        out_of_order_support = true;
+        config = configurations[test_case];
+    }
+
+    bool Skip() override {
```
EwanC, line 68: We should be calling base class Skip in here, `BasicMutableCommandBufferTest::Skip()`

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+        bool extension_avaliable =
+                is_extension_available(device,
+                                       "cl_khr_command_buffer_mutable_dispatch")
+                == true;
+
+        return !out_of_order_support || !no_additional_wgs_support || !extension_avaliable;
+    }
+
+    cl_int SetUp(int elements) override {
```
EwanC, line 85: We should be calling base class Setup in here `BasicMutableCommandBufferTest::SetUp()`

EwanC (changes requested):

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
banan328: I agree I will remove it

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
banan328: you are right, I fixed this

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
banan328: I fixed this, thank you!

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
banan328: done

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
aharon-abramson: The issue here is that our implementation requires the following properties for command buffer queues - `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE | CL_QUEUE_NO_SYNC_OPERATIONS_INTEL`. When we developed this test, we wanted it to be able to run on our runtime; this is why we have this logic in `Skip` and `Setup` (all other command buffer tests are skipped with our implementation).

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+        error = clFinalizeCommandBufferKHR(command_buffer);
+        test_error(error, "clFinalizeCommandBufferKHR failed");
+
+        clEventWrapper events[2];
+        error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
+                                          nullptr, &events[0]);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        clWaitForEvents(1, &events[0]);
```
EwanC, line 144:
```suggestion
        error = clWaitForEvents(1, &events[0]);
```
Same for other `clWaitForEvents` calls

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
EwanC: I understand, but I don't really like forcing the out-of-order requirement on a cross vendor CTS test. I wonder if we could parameterize the test implementation so that there's both an in-order queue test and out-of-order queue test? Since there is only one command the difference should be minimal, and it would mean `CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR` would get test coverage both for yourself and any implementers that don't support out-of-order queues.

As an aside, it would be much better if we could find a way for you to run all the command-buffer tests, but that feels out of scope for this PR.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,300 @@
+//
+// Copyright (c) 2022 The Khronos Group Inc.
```
EwanC, line 2:
```suggestion
// Copyright (c) 2024 The Khronos Group Inc.
```

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
banan328: We just discussed this issue internally and updated the test to match the structure of the other sub-tests

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
banan328: fixed this, thank you

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,300 @@
...
+    cl_int SetUp(int elements) override
+    {
+        cl_int error = BasicMutableCommandBufferTest::SetUp(elements);
+        test_error(error, "BasicMutableCommandBufferTest::SetUp failed");
+
+        error = SetUpKernel();
+        test_error(error, "SetUpKernel failed");
+
+        out_of_order_queue = clCreateCommandQueue(
+            context, device, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &error);
```
EwanC, line 94: I don't think we should mandate out-of-order requirement for this test when the spec doesn't have such a requirement. I wonder if we could parameterize the test implementation so that there's both an in-order queue test and out-of-order queue test? Since there is only one command the difference should be minimal, and it would mean CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR would get test coverage both for yourself and any implementers that don't support out-of-order queues.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
EwanC: My last comment is tangential to the original one about test structure, so i've created a new thread for it https://github.com/KhronosGroup/OpenCL-CTS/pull/1991/files#r1668852431

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
banan328: I think we dont need to override the setUp at all ..., what do you think?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,274 @@
...
+        return CL_SUCCESS;
+    }
+
+    size_t info_global_size = 0;
+    static constexpr size_t test_global_work_size = 64;
+    static constexpr size_t update_global_size = 16;
+    const size_t local_work_size = 8;
+    const size_t sizeToAllocate = 64 * sizeof(cl_int);
+    cl_mutable_command_khr command = nullptr;
+    clCommandBufferWrapper command_buffer;
```
EwanC, line 243: This variable is shadowing the base class `command_buffer` member, we therefore use it in the tests without initializing it

banan328: I fixed this, thank you

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,272 @@
...
+#include "mutable_command_basic.h"
+
+#include <CL/cl.h>
+#include <CL/cl_ext.h>
+
+////////////////////////////////////////////////////////////////////////////////
+// mutable dispatch tests which handle following cases:
+//
+// 1. The command buffer is created with
+// CL_MUTABLE_DISPATCH_ASSERT_NO_ADDITIONAL_WORK_GROUPS_KHR in its properties.
```
EwanC, line 28: I don't see where a command-buffer is being created with this property. For the command itself we use `config.ndrange_properties`, but I don't see any checks for `config.command_buffer_properties`

banan328: You are right, I just removed it by mistake, it was fixed

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -0,0 +1,291 @@
...
+            nullptr // local_work_size
+        };
+
+        cl_mutable_base_config_khr mutable_config{
+            CL_STRUCTURE_TYPE_MUTABLE_BASE_CONFIG_KHR, nullptr, 1,
+            &dispatch_config
+        };
+
+        error =
+            clUpdateMutableCommandsKHR(single_command_buffer, &mutable_config);
```
EwanC, line 212: `single_command_buffer` needs to be created with [CL_COMMAND_BUFFER_MUTABLE_KHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_COMMAND_BUFFER_MUTABLE_KHR) for this to be valid.

EwanC (approved):

PR #1992, opened 2024-06-27
math_brute_force: remove spurious tan skip check

svenvh:
The `skipTestingRelaxed` check suffers the following problems:

 - The use of `skipTestingRelaxed` in the `if` seems reversed: when skipping correctness testing using the `-l` command line option, this variable causes correctness testing to be run for relaxed-mode `tan` regardless.

 - Accuracy testing should only be skipped for derived `tan` implementations.  Non-derived `tan` implementations must still be tested for accuracy, so the condition for setting the `skipTestingRelaxed` variable is incomplete.

 - It is unclear why only `tan` is conditionalized here.  There are other functions such as `tanpi` for which one would expect identical behaviour.

The actual skipping of accuracy checks for derived implementations happens in `Test()`, so just remove `skipTestingRelaxed` as it does not seem to add any value.

### DIFF
--- a/test_conformance/math_brute_force/unary_float.cpp
+++ b/test_conformance/math_brute_force/unary_float.cpp
@@ -489,7 +489,6 @@ int TestFunc_Float_Float(const Func *f, MTdata d, bool relaxedMode)
     cl_int error;
     float maxError = 0.0f;
     double maxErrorVal = 0.0;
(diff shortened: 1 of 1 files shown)


kpet (approved):

bashbaug:

PR #1993, opened 2024-06-28
math_brute_force: skip all enqueues in -l mode

svenvh:
The math_brute_force test has a link-check-only mode to quickly test if all math functions are present.

In link-check-only mode, most tests return immediately after kernel compilation, but some tests also run the kernel and read back the result data before returning.  Running the kernels takes a lot more time, defeating the purpose of the `-l` mode.

Break out at the start of the main `for` loops in link-check-only mode instead of returning directly, so that each test can still log its trailing message (in most cases just a `\n`).

### DIFF
--- a/test_conformance/math_brute_force/unary_two_results_float.cpp
+++ b/test_conformance/math_brute_force/unary_two_results_float.cpp
@@ -68,6 +68,8 @@ int TestFunc_Float2_Float(const Func *f, MTdata d, bool relaxedMode)
 
     for (uint64_t i = 0; i < (1ULL << 32); i += step)
     {
+        if (gSkipCorrectnessTesting) break;
+
         // Init input array
         uint32_t *p = (uint32_t *)gIn;
         if (gWimpyMode)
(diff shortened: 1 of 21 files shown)


test_conformance/math_brute_force/unary_two_results_float.cpp
```diff
@@ -258,12 +260,6 @@ int TestFunc_Float2_Float(const Func *f, MTdata d, bool relaxedMode)
             }
         }
 
-        if (gSkipCorrectnessTesting)
-        {
-            if (isFract && gIsInRTZMode) (void)set_round(oldRoundMode, kfloat);
```
EwanC, line 263: There's an extra `if` statement getting removed here and in `unary_two_results_half.cpp` compared to the other tests, is that accounted for by this change?

EwanC: Makes sense, thanks :)

EwanC (approved):

PR #2000, opened 2024-07-04
Drop linking Vulkan directly

### LINKED ISSUE #1999: Vulkan should be linked using the vulkan-wrapper and not directly
The purpose of `vulkan_wrapper` is to load the Vulkan library (`vulkan-1.dll`, `libvulkan.so`, etc.) dynamically, at runtime, using `LoadLibrary` or `dlopen`.

Currently, targets that link to `vulkan_wrapper` also link to `Vulkan` and even `dl`. This should not be needed, and adds additional, unnecessary link requirements when building.

ahesham-arm:
The `vulkan-wrapper` helper library's purpose is to eliminate the need to link targets directly to Vulkan, and load it dynamically instead at runtime, using `dlopen` and `LoadLibrary`.

Remove the direct linking of the Vulkan library from targets that link to `vulkan-wrapper`.

Fixes #1999

### DIFF
--- a/test_conformance/extensions/cl_khr_external_semaphore/CMakeLists.txt
+++ b/test_conformance/extensions/cl_khr_external_semaphore/CMakeLists.txt
@@ -15,12 +15,7 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 include_directories (${CLConform_INCLUDE_DIR})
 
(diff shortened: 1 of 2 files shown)


kpet (approved):

bashbaug:
Merging as discussed in the July 9th teleconference.

PR #2001, opened 2024-07-05
computeinfo: add CL_DEVICE_HALF_FP_CONFIG

svenvh:
The computeinfo test already checks that the cl_khr_fp16 extension is available before querying `CL_DEVICE_HALF_FP_CONFIG`, but never actually made the `CL_DEVICE_HALF_FP_CONFIG` query.

### DIFF
--- a/test_conformance/computeinfo/main.cpp
+++ b/test_conformance/computeinfo/main.cpp
@@ -188,6 +188,7 @@ config_info config_infos[] = {
     CONFIG_INFO(2, 0, CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT, cl_uint),
 
     CONFIG_INFO(1, 1, CL_DEVICE_MEM_BASE_ADDR_ALIGN, cl_uint),
+    CONFIG_INFO(1, 1, CL_DEVICE_HALF_FP_CONFIG, cl_device_fp_config),
     CONFIG_INFO(1, 1, CL_DEVICE_SINGLE_FP_CONFIG, cl_device_fp_config),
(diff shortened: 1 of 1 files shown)


test_conformance/computeinfo/main.cpp
```diff
@@ -190,6 +190,7 @@ config_info config_infos[] = {
     CONFIG_INFO(1, 1, CL_DEVICE_MEM_BASE_ADDR_ALIGN, cl_uint),
     CONFIG_INFO(1, 1, CL_DEVICE_SINGLE_FP_CONFIG, cl_device_fp_config),
     CONFIG_INFO(1, 1, CL_DEVICE_DOUBLE_FP_CONFIG, cl_device_fp_config),
+    CONFIG_INFO(1, 1, CL_DEVICE_HALF_FP_CONFIG, cl_device_fp_config),
```
kpet, line 193: If this affects the display order, a minor improvement might be to move this line to before the one with `CL_DEVICE_SINGLE_FP_CONFIG` so capabilities are reported in increasing float bit width order.

kpet (dismissed):
LGTM. Left a potential suggestion for improvement.

test_conformance/computeinfo/main.cpp
svenvh: > If this affects the display order

It does.

> so capabilities are reported in increasing float bit width order.

I was following the enum value order, but `CL_DEVICE_DOUBLE_FP_CONFIG` is already breaking that order anyway (and I can see why, as it's nicer to have all *_FP_CONFIG values reported together).  Reporting in increasing float bit width order seems nicer indeed, I'll make the change.

kpet (approved):

bashbaug:

PR #2004, opened 2024-07-07
Use the `--parallel` build option in CI workflows

### LINKED ISSUE #2003: CI should not limit the number of jobs that CMake uses for building
In `presubmit.sh` the actual command that invokes the build is `cmake --build . -j2` or `cmake --build . -j3`. Instead of limiting the number of jobs to an arbitrary number, it should be `cmake --build . --parallel` instead.

ahesham-arm:
Instead of selecting an arbitrary number for the number of parallel build jobs, use CMake's `--parallel` option instead.

Option was introduced in version 3.12, GitHub runners currently use CMake version 3.29.6 so this is supported.

Fixes #2003


bashbaug:
Merging as discussed in the July 9th teleconference.

PR #2008, opened 2024-07-10
[NFC] math_brute_force: reformat function list manually

svenvh:
clang-format has already been disabled here.  Make the function list a bit easier to read.  Make the currently commented out functions stand out better on a line of their own.


EwanC (approved):

svenvh:
Merging immediately as this is non-controversial and brings no functional changes.

PR #2010, opened 2024-07-11
Implement Negative Command Queue Tests

ahesham-arm:
Implement negative tests for the following functions:

- clCreateCommandQueueWithProperties
- clCreateCommandQueue
- clSetDefaultDeviceCommandQueue
- clRetainCommandQueue
- clReleaseCommandQueue
- clGetCommandQueueInfo
- clSetCommandQueueProperties

### DIFF
--- a/test_conformance/api/negative_queue.cpp
+++ b/test_conformance/api/negative_queue.cpp
@@ -16,39 +16,133 @@
 #include "testBase.h"
 #include "harness/typeWrappers.h"
 
+#include <array>
+#include <vector>
+#include <initializer_list>
+
+static bool different_device_available(cl_device_id deviceID,
+                                       cl_device_id* different_device)
+{
+    cl_uint num_platforms = -1;
+    cl_int err = clGetPlatformIDs(0, nullptr, &num_platforms);
+    test_error(err, "clGetPlatformIDs");
+    std::vector<cl_platform_id> platforms(num_platforms);
+    err = clGetPlatformIDs(num_platforms, platforms.data(), nullptr);
+    test_error(err, "clGetPlatformIDs");
+    cl_platform_id device_platform = getPlatformFromDevice(deviceID);
+    for (auto platform : platforms)
+    {
+        if (platform != device_platform)
+        {
+            err = clGetDeviceIDs(platform, CL_DEVICE_TYPE_DEFAULT, 1,
+                                 different_device, nullptr);
+            test_error(err, "clGetDeviceID");
+            return true;
+        }
+        // Check to see if the current device platform supports other devices
+        else
+        {
+            cl_uint num_devices = -1;
+            err = clGetDeviceIDs(platform, CL_DEVICE_TYPE_ALL, 0, nullptr,
+                                 &num_devices);
+            test_error(err, "clGetDeviceIDs");
+            // If the platform only has 1 device which is our device then ignore
+            // it
+            if (num_devices <= 1)
+            {
+                continue;
+            }
+            else
+            {
+                std::vector<cl_device_id> alternate_devices(num_devices);
+                err = clGetDeviceIDs(platform, CL_DEVICE_TYPE_ALL, num_devices,
+                                     alternate_devices.data(), nullptr);
+                test_error(err, "clGetDeviceIDs");
+                for (auto device : alternate_devices)
+                {
+                    if (device != deviceID)
+                    {
+                        *different_device = device;
+                        return true;
+                    }
+                }
+            }
+        }
+    }
+    return false;
+}
+
 int test_negative_create_command_queue(cl_device_id deviceID,
                                        cl_context context,
                                        cl_command_queue queue, int num_elements)
 {
-    cl_command_queue_properties device_props = 0;
-    cl_int error = clGetDeviceInfo(deviceID, CL_DEVICE_QUEUE_PROPERTIES,
-                                   sizeof(device_props), &device_props, NULL);
-    test_error(error, "clGetDeviceInfo for CL_DEVICE_QUEUE_PROPERTIES failed");
+    cl_int err = 0;
+    clCreateCommandQueue(nullptr, deviceID, 0, &err);
+    test_failure_error_ret(
+        err, CL_INVALID_CONTEXT,
+        "clCreateCommandQueue should return CL_INVALID_CONTEXT when: \"context "
+        "is not a valid context\" using a nullptr",
+        TEST_FAIL);
 
-    // CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE is the only optional property to
-    // clCreateCommandQueue, CL_QUEUE_PROFILING_ENABLE is mandatory.
-    const bool out_of_order_device_support =
-        device_props & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE;
-    if (out_of_order_device_support)
+    clCreateCommandQueue(context, nullptr, 0, &err);
+    test_failure_error_ret(
+        err, CL_INVALID_DEVICE,
+        "clCreateCommandQueue should return CL_INVALID_DEVICE when: \"device "
+        "is not a valid device\" using a nullptr",
+        TEST_FAIL);
+
+    cl_device_id different_device = nullptr;
+    if (different_device_available(deviceID, &different_device))
     {
-        // Early return as we can't check correct error is returned for
-        // unsupported property.
-        return TEST_PASS;
+        clCreateCommandQueue(context, different_device, 0, &err);
+        test_failure_error_ret(
+            err, CL_INVALID_DEVICE,
+            "clCreateCommandQueue should return CL_INVALID_DEVICE when: "
+            "\"device is not associated with context\"",
+            TEST_FAIL);
     }
 
-    // Try create a command queue with out-of-order property and check return
-    // code
-    cl_int test_error = CL_SUCCESS;
-    clCommandQueueWrapper test_queue = clCreateCommandQueue(
-        context, deviceID, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &test_error);
-
+    cl_queue_properties invalid_property(-1);
+    clCreateCommandQueue(context, deviceID, invalid_property, &err);
     test_failure_error_ret(
-        test_error, CL_INVALID_QUEUE_PROPERTIES,
-        "clCreateCommandQueue should return CL_INVALID_QUEUE_PROPERTIES if "
-        "values specified in properties are valid but are not supported by "
-        "the "
-        "device.",
+        err, CL_INVALID_VALUE,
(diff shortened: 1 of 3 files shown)


test_conformance/api/negative_queue.cpp
```diff
@@ -172,3 +350,259 @@ int test_negative_create_command_queue_with_properties_khr(
...
+    cl_queue_properties property(CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE);
+    cl_int err = clSetCommandQueueProperty(nullptr, property, CL_TRUE, nullptr);
+    test_failure_error_ret(
+        err, CL_INVALID_COMMAND_QUEUE,
+        "clSetCommandQueueProperty should return CL_INVALID_COMMAND_QUEUE "
+        "when: \"command_queue is not a valid command-queue\" using a nullptr",
+        TEST_FAIL);
+
+
+    err = clSetCommandQueueProperty(reinterpret_cast<cl_command_queue>(context),
```
EwanC, line 590: Think these tests could be over ambitious with the reinterpret cast being an invalid object error on all implementations. There was some discussion [here](https://github.com/KhronosGroup/OpenCL-Docs/issues/482#issuecomment-795685410) about what a valid openCL object is, resulting in merged PR https://github.com/KhronosGroup/OpenCL-Docs/pull/842  that specifies 

> An OpenCL implementation must check for a `NULL` object to determine if an
object is valid.  The behavior for all other invalid objects is implementation-defined.

test_conformance/api/negative_queue.cpp
```diff
@@ -172,3 +350,259 @@ int test_negative_create_command_queue_with_properties_khr(
...
+                TEST_FAIL);
+        }
+    }
+
+    err = clSetDefaultDeviceCommandQueue(nullptr, deviceID, queue);
+    test_failure_error_ret(
+        err, CL_INVALID_CONTEXT,
+        "clSetDefaultDeviceCommandQueue should return CL_INVALID_CONTEXT when "
+        "\"context is not a valid context\" using a nullptr",
+        TEST_FAIL);
```
bashbaug, line 395: I think a lot of these cases need to be updated to allow for `CL_INVALID_OPERATION` as well, when the passed-in device ID does not support a replaceable default on-device queue.  This can happen if the device is a pre-OpenCL 2.0 device, or if the device is an OpenCL 3.0 device that does not support a replaceable default.

test_conformance/api/negative_queue.cpp
```diff
@@ -172,3 +350,259 @@ int test_negative_create_command_queue_with_properties_khr(
...
+        err, CL_INVALID_VALUE,
+        "clGetCommandQueueInfo should return CL_INVALID_VALUE when: \"size in "
+        "bytes specified by param_value_size is < size of return type and "
+        "param_value is not a NULL value\"",
+        TEST_FAIL);
+
+    return TEST_PASS;
+}
+
+int test_negative_set_command_queue_property(cl_device_id deviceID,
```
bashbaug, line 576: I think clSetCommandQueueProperty needs to handle the cases when the device is newer than OpenCL 1.0 and hence can return CL_INVALID_OPERATION.

See discussion and the latest changes in the now-merged https://github.com/KhronosGroup/OpenCL-CTS/pull/1182.

bashbaug (changes requested):

test_conformance/api/negative_queue.cpp
lakshmih: Agreed. Anything other than NULL is implementation defined so we should ideally skip these tests.

test_conformance/api/negative_queue.cpp
```diff
@@ -57,55 +168,122 @@ int test_negative_create_command_queue_with_properties(cl_device_id deviceID,
...
+            properties[3] = max_size + 1;
+            clCreateCommandQueueWithProperties(context, deviceID,
+                                               properties.data(), &err);
+            test_failure_error_ret(
+                err, CL_INVALID_VALUE,
+                "clCreateCommandQueueWithProperties should return "
+                "CL_INVALID_VALUE when: \"values specified in "
+                "properties are not valid\" using a queue size "
+                "greater than CL_DEVICE_QUEUE_ON_DEVICE_MAX_SIZE",
+                TEST_FAIL);
```
lakshmih, line 284: Spec:
[CL_INVALID_VALUE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_VALUE) if values specified in properties are not valid.

[CL_INVALID_QUEUE_PROPERTIES](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_QUEUE_PROPERTIES) if values specified in properties are valid but are not supported by the device.

Looks like this could equally be interpreted as a case for "CL_INVALID_QUEUE_PROPERTIES" since the size itself is "not supported by the device" having exceeded CL_DEVICE_QUEUE_ON_DEVICE_MAX_SIZE. 

Would like to discuss this in the next call.

lakshmih: Agreed on 8/27 to accept either error codes (CL_INVALID_VALUE or CL_INVALID_QUEUE_PROPERTIES) for this case. Arm will make this fix in a subsequent patch.

bashbaug:
Removing the "focused review" label until the comments above have been addressed.

test_conformance/api/negative_queue.cpp
kpet: @ahesham-arm Looks like the PR will need updating to address the request above.

test_conformance/api/negative_queue.cpp
ahesham-arm: Done.

test_conformance/api/negative_queue.cpp
ahesham-arm: Done.

test_conformance/api/negative_queue.cpp
ahesham-arm: Done.

test_conformance/api/negative_queue.cpp
ahesham-arm: Done.

test_conformance/api/negative_queue.cpp
```diff
@@ -57,55 +151,107 @@ int test_negative_create_command_queue_with_properties(cl_device_id deviceID,
...
+        err = clGetDeviceInfo(deviceID, CL_DEVICE_QUEUE_ON_DEVICE_MAX_SIZE,
+                              sizeof(max_size), &max_size, nullptr);
+        test_error(err, "clGetDeviceInfo");
+        if (max_size > 0)
+        {
+            properties[0] = CL_QUEUE_PROPERTIES;
+            properties[1] =
+                CL_QUEUE_ON_DEVICE | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE;
+            properties[2] = CL_QUEUE_SIZE;
+            properties[3] = max_size + 1;
```
haonanya1, line 240: max_size is unsigned int, if max_size obtained from query is UINT_MAX, max_size +1 is 0 and then queue properties is valid.

test_conformance/api/negative_queue.cpp
```diff
@@ -57,55 +151,107 @@ int test_negative_create_command_queue_with_properties(cl_device_id deviceID,
...
+            clCreateCommandQueueWithProperties(context, deviceID,
+                                               properties.data(), &err);
+            if (err != CL_INVALID_VALUE && err != CL_INVALID_QUEUE_PROPERTIES)
+            {
+                log_error(
+                    "clCreateCommandQueueWithProperties should return "
+                    "CL_INVALID_VALUE or CL_INVALID_QUEUE_PROPERTIES when: "
+                    "\"values specified in properties are not valid\" using a "
+                    "queue size greather than "
+                    "CL_DEVICE_QUEUE_ON_DEVICE_MAX_SIZE");
```
haonanya1, line 250: If error occurs, it prints 
negative_create_command_queue_with_properties...
clCreateCommandQueueWithProperties should return CL_INVALID_VALUE or CL_INVALID_QUEUE_PROPERTIES when: "values specified in properties are not valid" using a queue size greather than CL_DEVICE_QUEUE_ON_DEVICE_MAX_SIZEnegative_create_command_queue_with_properties FAILED
                                                                    
can we add "\n" or "." to split the 2 sentences? Same comments for other log_error call.

test_conformance/api/negative_queue.cpp
haonanya1, line 240: if max_size is uint max and then properties[3] will be 0. This is not expected since we want a value greater than max queue size.

test_conformance/api/negative_queue.cpp
```diff
@@ -172,3 +318,223 @@ int test_negative_create_command_queue_with_properties_khr(
...
+        err = clSetDefaultDeviceCommandQueue(context, different_device, queue);
+        if (err != CL_INVALID_OPERATION && err != CL_INVALID_DEVICE)
+        {
+            log_error("clSetDefaultDeviceCommandQueue should return "
+                      "CL_INVALID_OPERATION or CL_INVALID_DEVICE "
+                      "when: \"device is not associated with context\"");
+            return TEST_FAIL;
+        }
+        err = clSetDefaultDeviceCommandQueue(context, deviceID, nullptr);
+        if (err != CL_INVALID_OPERATION && err != CL_INVALID_DEVICE)
```
haonanya1, line 388: typo for CL_INVALID_DEVICE, should be CL_INVALID_COMMAND_QUEUE.

test_conformance/api/negative_queue.cpp
```diff
@@ -172,3 +318,223 @@ int test_negative_create_command_queue_with_properties_khr(
...
+                                             cl_command_queue queue,
+                                             int num_elements)
+{
+    cl_queue_properties property(CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE);
+    cl_int err = clSetCommandQueueProperty(nullptr, property, CL_TRUE, nullptr);
+    if (err == CL_INVALID_OPERATION)
+    {
+        // Implementations are allowed to return an error for
+        // non-OpenCL 1.0 devices. In which case, skip the test.
+        return TEST_SKIPPED_ITSELF;
```
haonanya1, line 519: OpenCL 1.0 device can also return CL_INVALID_OPERATION, how can we know if it is non-OpenCL 1.0 devices.
The case return CL_INVALID_COMMAND_QUEUE since command_queue is nullptr according to https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/main/loader/icd_dispatch_generated.c#L2521.

bashbaug:
Removing "focused review" until the review comments above are addressed, and the merge conflict is resolved.

ahesham-arm:
Closing as it being superseded by #2505

PR #2011, opened 2024-07-11
printf: add missing cstdint include

svenvh:
Since GCC 13 some headers are no longer included transitively through C++ Standard Library headers.


kpet (approved):

svenvh:
Merging immediately as this is non-controversial.

PR #2014, opened 2024-07-16
Use sccache on all CI platforms

ahesham-arm:
MSVC on Windows was previously omitted from the list of CI platforms that use sccache. Unify the CMAKE_CACHE_OPTIONS used for all CI platforms.

In the case of a 100% cache hit, this cuts down the average checks time from 8 minutes to 4 minutes, example:
https://github.com/ahesham-arm/OpenCL-CTS/actions/runs/9896449792


kpet (approved):
Thanks, it seems to be passing the CI so I assume that whatever issue we might have faced in the past is now fixed. We can always revert the change if it's not the case.

bashbaug:

Issue #2009, opened 2024-07-11
test_printf producing false expected results


coldav:
in the oneAPI Construction Kit, our overnight testing on 10th shows repeated fails on the test_printf test:

15)testing printf("%+#21.15E",789456123.0)
verifyOutputBuffer failed with kernel:
__kernel void test3(void)
{
   printf("%+#21.15E\n",789456123.0);
}

 expected: +7.894561280000000E+08
 got:      +7.894561230000000E+08

This was run on x86 in our "host" target.
I believe the 123 part is actually correct, so the test seems wrong.

https://github.com/KhronosGroup/OpenCL-CTS/pull/1988 maybe related as it was committed around the same time.

To show with the oneAPI Construction Kit.
```
git clone git@github.com:codeplaysoftware/oneapi-construction-kit.git
cd oneapi-construction-kit
cmake -GNinja -DCA_MUX_TARGETS_TO_ENABLE="host" -DCA_ENABLE_API=cl       -DCA_LLVM_INSTALL_DIR=<path_to_17_or_18_llvm_install>t -Bbuild
ninja -c build libOpenCL.so

```

coldav:
> @coldav could you tell which subtest this is? This should be printed before the line starting with `0)`.

float...

coldav:
With a little bit more digging if we disable double support we do get the 128 answer. We think it's acceptable to do the conversion to double though so both the 123 and 128 are correct.
As an aside:
Some platforms show
`
/home/harald/OpenCL-CTS/test_conformance/printf/test_printf.cpp: In function ‘_cl_program* {anonymous}::makeMixedFormatPrintfProgram(_cl_kernel**, cl_context, cl_device_id, unsigned int, unsigned int, const std::string&)’:
/home/harald/OpenCL-CTS/test_conformance/printf/test_printf.cpp:300:14: error: ‘uint8_t’ is not a member of ‘std’; did you mean ‘wint_t’?
  300 |         std::uint8_t is_int = genrand_int32(gMTdata) % 2;
      |              ^~~~~~~
      |              wint_t
/home/harald/OpenCL-CTS/test_conformance/printf/test_printf.cpp:303:32: error: ‘is_int’ was not declared in this scope; did you mean ‘u_int’?
  303 |         set_round(deviceRound, is_int != 0 ? kint : kfloat);
      |                                ^~~~~~
      |                                u_int
ninja: build stopped: subcommand failed.
`
so there is likely a missing include.

[...]

PR #2015, opened 2024-07-18
[printf]: input value should be exactly representable as 32-bit float

dneto0:
Fixed: #2009


bashbaug (approved):

Kerilk (approved):
Checked with this Ruby program:
```Ruby
irb(main):001:0> require 'float-formats'
irb(main):002:0> puts Flt::IEEE_binary32(789456128.0).to(Float)
789456128.0
irb(main):003:0> puts Flt::IEEE_binary32(789456123.0).to(Float)
789456128.0
```

PR #2024, opened 2024-07-24
Fixes for basic explicit_s2v and commonfns degrees for cl_half

AhmedAmraniAkdi:
Basic explicit_s2v:

The verification step was always using round to even when converting a float to half even for round to zero cores.

Commonfns degrees:

The verification step was only taking into account infinities and not values that over/underflow. This resulted in an incorrect error calculation. E.g:

double cpu_result = 175668.85998711039;
cl_half gpu_result = 31743; // this is 65504 when converting to float, we overflowed.
float error = (cpu_result - gpu_result) * some_factor;

The fix adds the check if( (cl_half) reference == test ) before calculating the error.

### DIFF
--- a/test_common/harness/conversions.cpp
+++ b/test_common/harness/conversions.cpp
@@ -263,10 +263,11 @@ static Long sLowerLimits[kNumExplicitTypes] = {
         }                                                                      \
         break;
 
-#define TO_HALF_CASE(inType)                                                   \
+#define TO_HALF_CASE(inType, halfRoundingMode)                                 \
     case kHalf:                                                                \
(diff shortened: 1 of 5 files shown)


lakshmih (approved):
Tangentially related, the comments for Ulp_Error_Half_Float() and Ulp_Error_Double() are incorrect. They should mention respectively
 
// if( (half) reference == test )// return 0.0f;
and
// if( (double) reference == test )// return 0.0f;
instead of
// if( (float) reference == test )// return 0.0f;

svenvh (approved):

bashbaug:
Merging as discussed in the August 13th teleconference.

PR #2025, opened 2024-07-24
api: fix -Wformat warnings

svenvh:
The main sources of warnings were:

 * Printing of a `size_t` which requires the `%zu` specifier.

 * Printing of `cl_long`/`cl_ulong` which is now done using the `PRI*64` macros to ensure portability across 32 and 64-bit builds.

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -18,6 +18,7 @@
 #include "harness/testHarness.h"
 #include <ctype.h>
 #include <string.h>
+#include <cinttypes>
 
 const char *sample_single_param_kernel[] = {
     "__kernel void sample_test(__global int *src)\n"
@@ -227,13 +228,13 @@ int test_min_max_work_items_sizes(cl_device_id deviceID, cl_context context,
     {
         if (deviceMaxWorkItemSize[i] < 1)
         {
-            log_error("MAX_WORK_ITEM_SIZE in dimension %d is invalid: %lu\n", i,
+            log_error("MAX_WORK_ITEM_SIZE in dimension %d is invalid: %zu\n", i,
                       deviceMaxWorkItemSize[i]);
             errors++;
         }
         else
         {
-            log_info("Dimension %d has max work item size %lu\n", i,
+            log_info("Dimension %d has max work item size %zu\n", i,
                      deviceMaxWorkItemSize[i]);
         }
     }
@@ -257,7 +258,7 @@ int test_min_max_work_group_size(cl_device_id deviceID, cl_context context,
                             NULL);
     test_error(error, "Unable to get max work group size from device");
 
-    log_info("Reported %ld max device work group size.\n", deviceMaxThreadSize);
+    log_info("Reported %zu max device work group size.\n", deviceMaxThreadSize);
 
     if (deviceMaxThreadSize == 0)
     {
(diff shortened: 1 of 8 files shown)


test_conformance/api/test_mem_object_info.cpp
```diff
@@ -18,21 +18,24 @@
...
-}
+#define TEST_MEM_OBJECT_PARAM(mem, paramName, val, expected, name, type, cast) \
+    error = clGetMemObjectInfo(mem, paramName, sizeof(val), &val, &size);      \
+    test_error(error, "Unable to get mem object " name);                       \
+    if (val != expected)                                                       \
+    {                                                                          \
+        log_error("ERROR: Mem object " name                                    \
+                  " did not validate! (expected " type ", got " type           \
+                  " from %s:%d)\n",                                            \
+                  (cast)expected, (cast)val, __FILE__, __LINE__);              \
```
svenvh, line 29: The only change here besides the formatting is the `(cast)` in front of `expected`.

test_conformance/api/test_queries.cpp
```diff
@@ -380,7 +381,8 @@ int check_get_command_queue_info_params(cl_device_id deviceID,
         clGetDeviceInfo(deviceID, host_queue_query, sizeof(host_queue_props),
                         &host_queue_props, NULL);
     test_error(error, "clGetDeviceInfo failed");
-    log_info("CL_DEVICE_QUEUE_ON_HOST_PROPERTIES is %d\n", host_queue_props);
+    log_info("CL_DEVICE_QUEUE_ON_HOST_PROPERTIES is %" PRId64 "\n",
```
EwanC, line 384: ` cl_queue_properties` is a typedef of `cl_properties`, which is a typdef of `cl_ulong`, so `PRIu64` might be better.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -614,14 +616,16 @@ int test_min_max_mem_alloc_size(cl_device_id deviceID, cl_context context,
...
+            "total memory! (%" PRIu64 " or %" PRIu64
+            "MB, from a total mem size of %" PRIu64 "MB)\n",
             maxAllocSize, (maxAllocSize / 1024) / 1024,
             (requiredAllocSize / 1024) / 1024);
         return -1;
     }
 
-    log_info("Reported max allocation size of %lld bytes (%gMB) and global mem "
-             "size of %lld bytes (%gMB).\n",
+    log_info("Reported max allocation size of %" PRId64
```
EwanC, line 626: We're using a `PRId64` to print `cl_ulong` values, i'd probably prefer `PRIu64` to match the sign.

EwanC (approved):
Nitpicking whether we should use `PRIu64` rather than `PRId64` for some cases, but LGTM in general.

test_conformance/api/test_queries.cpp
svenvh: Thanks for spotting that, fixed.

test_conformance/api/test_api_min_max.cpp
svenvh: Fully agree, I think I've fixed all occurrences with the latest push.

EwanC (dismissed):

svenvh:
Rebased and fixed conflicts.

EwanC (approved):

Issue #1900, opened 2024-02-25
Test for callbacks called upon errors


sanguinariojoe:
Already commented here, but I suppose this deserves an entry here.

Two platforms (rusticl and pocl) had the same issue, both not calling callbacks upon error status. I really think a conformance test shall be added to check it.

bashbaug:
The test was added in #2028, so this issue can be closed (as discussed in the August 27th teleconference).

PR #2028, opened 2024-07-31
Added two tests not calling event calback upon error status

kamil-goras-mobica:
#1900 
Tested on POCL, callback is not called.


test_conformance/events/main.cpp
```diff
@@ -24,39 +24,41 @@
...
-    ADD_TEST(event_flush),
-    ADD_TEST(event_finish_execute),
-    ADD_TEST(event_finish_array),
-    ADD_TEST(event_release_before_done),
-    ADD_TEST(event_enqueue_marker),
-#ifdef CL_VERSION_1_2
-    ADD_TEST(event_enqueue_marker_with_event_list),
-    ADD_TEST(event_enqueue_barrier_with_event_list),
-#endif
+    //    ADD_TEST(event_get_execute_status),
```
bashbaug, line 27: I think these tests were only commented out for testing, and should be re-enabled.

test_conformance/events/test_callbacks.cpp
```diff
@@ -372,3 +383,75 @@ int test_callbacks_simultaneous(cl_device_id deviceID, cl_context context,
...
+    const cl_int pattern_pri = 0xA;
+    clMemWrapper in_mem =
+        clCreateBuffer(context, CL_MEM_READ_ONLY, sizeof(cl_int) * num_elements,
+                       nullptr, &error);
+    test_error(error, "clCreateBuffer failed");
+
+    bool confirmation = false;
+    error = clSetEventCallback(user_event, CL_COMPLETE,
+                               combuf_event_callback_function, &confirmation);
+    test_error(error, "clSetEventCallback failed");
```
bashbaug, line 434: Since this test is putting the callback on the user event it really isn't doing much different than the "simple" test, above.

I would suggest:

1. Including an event on the fill buffer command, below.
    * Note, the fill buffer command should still depend on the user event, also.
3. Set the callback on the fill buffer event, not on the user event.
4. After setting the user event to an error status the error status should propagate to the fill buffer command (because it is dependent on the user event), and the callback should therefore be called.

kamil-goras-mobica: @bashbaug done

test_conformance/events/main.cpp
kamil-goras-mobica: @bashbaug yes it was only for internal testing. Re-enabled again

lakshmih (dismissed):

bashbaug (dismissed):
Thanks, latest version LGTM!

Bikeshedding: The names of the new tests are currently "callback_not_called_simple" and "callback_not_called_enqueue_command", though this is a bit confusing because the tests are verifying that the callbacks are called, not that they are NOT called.  Would it be better to call them something like "callback_on_error_simple" and "callback_on_error_enqueue_command"?

To be clear, I'm fine merging this as-is, and the test behavior looks good.  Thanks!

bashbaug (approved):

lakshmih (approved):

bashbaug:
Merging as discussed in the August 20th teleconference.

PR #2031, opened 2024-08-01
Extend testing of CL_UNORM_INT_101010_2

gorazd-sumkovski-arm:
All existing tests in `test_image_streams`, that are capable of testing image formats using the `CL_UNORM_INT_101010_2` data type, now do so.


bashbaug (commented):
I did a search for `CL_UNORM_INT_101010`, thinking that these are places that likely need to be updated for `CL_UNORM_INT_101010_2`.  Can you check that these don't need updates?

1. `get_32_bit_image_format()`
2. `compare_scanlines()` - this one is probably OK.
3. `get_format_max_int()`
4. `get_format_min_int()`
5. `read_image_pixel()`
6. `get_pixel_bytes()`

There are a few others in other test suites also, but maybe those are intended for another PR.

gorazd-sumkovski-arm:
1. `get_32_bit_image_format()` - Added (although this function isn't used anywhere).
2. `compare_scanlines()` - As you suggested, no need for a special case because all 32 bits in `CL_UNORM_INT_101010_2` are used.
3. `get_format_max_int()` - Added.
4. `get_format_min_int()` - Added.
5. `read_image_pixel<T>()` - Already present in `read_image_pixel_float()`. Seems like `read_image_pixel<T>()` is used for `int` and `uint` while `float` has a separate function. As a side note, it looks like the cases for 101010, 555 and 565 in `read_image_pixel<T>()` are wrong, they are missing the normalisation divisions.
6. `get_pixel_bytes()` - Added.

bashbaug (approved):
I'd be surprised if there weren't a few more places that need updates for 101010_2, but I didn't find anymore cases that are obviously missing.  Thanks!

kpet:
Any objections to merging this PR ahead of the next teleconference? It is blocking the addition of testing for the 2_101010 format added by https://github.com/KhronosGroup/OpenCL-Docs/pull/1223.

bashbaug:
I'm going to go ahead and merge this.  Since these are purely additions I think the risk of problems for implementations that do not support the 101010_2 format is low.  If we find problems we can always revert it.

@lakshmih FYI.

kpet:
Thanks!

PR #2032, opened 2024-08-01
Fix bug in copy image mipmap comparisons

joshqti:
The destination image for mipmap copies is incorrect, use the width of the mip-level instead of mip-level 0. The check if (where < dstImageInfo->width) is not correct, width is the size of level 0, but it should be the width of the destination mip level.

Deletes redundant scanline comparison, which is confusing and uneccesary. compare_scanline and memcmp do the same thing, except compare_scanline correctly take image format into account, whereas memcmp finds all differences.


bashbaug (commented):
This seems reasonable to me, but these tests are hard to follow.

Is there any test execution time difference after this change?  I'd want to check that calling compare_scanlines unconditionally vs. only after calling memcmp doesn't slow down testing, since memcmp will be sufficient for most image formats, and it'll be a lot faster to compare the entire scanline in one call using memcpy vs. pixel-by-pixel as part of compare_scanlines.

bashbaug:
This PR appears to be stuck - it's had "focused review" since August.  What do we need to do to "un-stick" it?

It might help to understand a test command line that failed before this change and that passes after this change, and more detail why it was failing previously.

PR #2033, opened 2024-08-05
Fix snorm

joshqti:
When comparing scanlines for SNORM images, take into account that -1.0 can be exactly represented in two different ways.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -488,7 +488,9 @@ size_t compare_scanlines(const image_descriptor *imageInfo, const char *aPtr,
                 cl_uchar aPixel = *(cl_uchar *)aPtr;
                 cl_uchar bPixel = *(cl_uchar *)bPtr;
                 // -1.0 is defined as 0x80 and 0x81
-                if ((aPixel != bPixel) && ((aPixel | bPixel) != 0x81))
+                aPixel = (aPixel == 0x80) ? 0x81 : aPixel;
+                bPixel = (bPixel == 0x80) ? 0x81 : bPixel;
+                if (aPixel != bPixel)
                 {
                     return column;
                 }
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -501,9 +501,9 @@ size_t compare_scanlines(const image_descriptor *imageInfo, const char *aPtr,
                 cl_ushort aPixel = *(cl_ushort *)aPtr;
                 cl_ushort bPixel = *(cl_ushort *)bPtr;
                 // -1.0 is defined as 0x8000 and 0x8001
-                aPixel = aPixel == 0x8000 ? 0x8001 : aPixel;
-                bPixel = bPixel == 0x8000 ? 0x8001 : bPixel;
-                if ((aPixel != bPixel))
+                aPixel = (aPixel == 0x8000) ? 0x8001 : aPixel;
+                bPixel = (bPixel == 0x8000) ? 0x8001 : bPixel;
+                if (aPixel != bPixel)
                 {
                     return column;
                 }
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -496,8 +496,8 @@ size_t compare_scanlines(const image_descriptor *imageInfo, const char *aPtr,
             break;
 
             case CL_SNORM_INT16: {
-                cl_short aPixel = *(cl_short *)aPtr;
-                cl_short bPixel = *(cl_short *)bPtr;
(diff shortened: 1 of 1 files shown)


test_common/harness/imageHelpers.cpp
```diff
@@ -484,6 +484,28 @@ size_t compare_scanlines(const image_descriptor *imageInfo, const char *aPtr,
...
+                // -1.0 is defined as 0x80 and 0x81
+                if ((aPixel != bPixel) && ((aPixel | bPixel) != 0x81))
+                {
+                    return column;
+                }
+            }
+            break;
+
+            case CL_SNORM_INT16: {
+                cl_short aPixel = *(cl_short *)aPtr;
```
joshqti, line 499: This needs to be ushort

svenvh:
> take into account that -1.0 can be exactly represented in two different ways.

Does this only apply to the Embedded Profile or also to the Full Profile?

bashbaug:
Can you please explain a bit more exactly what scenario this change is trying to fix?  Does the problem happen when copying SNORM data from one image to the other?  If so, does the multiple representations for -1.0 cause problems for the source data, or the destination data, or both?

The reason why I ask is because the spec is pretty clear how the conversions between SNORM <-> float should occur:

[SNORM to float](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#converting-normalized-integer-channel-data-types-to-floating-point-values)
[float to SNORM](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#converting-floating-point-values-to-normalized-integer-channel-data-types)

I _think_ I have an idea how this still could cause a problem (specifically with the two source representations for -1?) but I'd like to be sure I understand it - thanks!

joshqti:
Correct, there are two valid representations on -1.0, it is possible a valid image copy to cause a false failure in the test as-is.

joshqti:
Snorm has two representations of -1.0.  Lets look at CL_R, SNORM_INT8 for example.

Lets sat a source image has the pixel 0x80, this is one of two valid -1.0 representation.  One way to copy this pixel is to first read, say using "x = read_imagef(source image)".  Variable x will be -1.0.  Lets say the device then writes the pixel to a destination image using the forumal from the spec "convert_char_sat_rte(f * 127.0f)".  It will convert -127.0f to -127, which is 0x81.  

The CTS test will notice 0x80 != 0x81.  This change makes CTS aware of snorm's multiple representation of -1.0.

test_common/harness/imageHelpers.cpp
```diff
@@ -484,6 +484,28 @@ size_t compare_scanlines(const image_descriptor *imageInfo, const char *aPtr,
             }
             break;
 
+            case CL_SNORM_INT8: {
+                cl_uchar aPixel = *(cl_uchar *)aPtr;
+                cl_uchar bPixel = *(cl_uchar *)bPtr;
+                // -1.0 is defined as 0x80 and 0x81
+                if ((aPixel != bPixel) && ((aPixel | bPixel) != 0x81))
```
bashbaug, line 491: This might be a bit too clever, since as-written it will pass the test if `aPixel` is `0x81` and `bPixel` is `0`, for example.

Maybe it'd be better to just pick one of the `-1.0` representations instead, something like:

```suggestion
                // -1.0 is defined as 0x80 and 0x81
                aPixel = (aPixel == 0x80) ? 0x81 : aPixel;
                bPixel = (bPixel == 0x80) ? 0x81 : bPixel;
                if (aPixel != bPixel)
```

test_common/harness/imageHelpers.cpp
```diff
@@ -484,6 +484,28 @@ size_t compare_scanlines(const image_descriptor *imageInfo, const char *aPtr,
...
+                    return column;
+                }
+            }
+            break;
+
+            case CL_SNORM_INT16: {
+                cl_ushort aPixel = *(cl_ushort *)aPtr;
+                cl_ushort bPixel = *(cl_ushort *)bPtr;
+                // -1.0 is defined as 0x8000 and 0x8001
+                if ((aPixel != bPixel) && ((aPixel | bPixel) != 0x8001))
```
bashbaug, line 502: See above, and apply whatever fix we make for `CL_SNORM_INT8` to `CL_SNORM_INT16` here.

bashbaug (commented):
Sorry for the delay!  The explanation of the problem makes sense, but I have a small suggestion for the fix.

test_common/harness/imageHelpers.cpp
joshqti: Nice catch, thanks. I like your suggestion.

bashbaug (approved):

bashbaug:

PR #2036, opened 2024-08-06
Fix issues in `test_external_semaphores_cross_queues_io2`

gorazd-sumkovski-arm:
1. Fix memory leak by making `context2` use the context wrapper.
2. Create two separate events to associate with the two calls to `clEnqueueSignalSemaphoresKHR()`, as events cannot be reused.

Also delete the unused helper function
`semaphore_external_cross_queue_helper()` (which was a duplicate of `external_semaphore_cross_queue_helper()` anyway).


kpet (approved):

bashbaug:
Merging as discussed in the August 6th teleconference.

PR #2041, opened 2024-08-07
add SPIR-V 1.4 testing for no integer wrap decorations

bashbaug:
This PR adds targeted testing for SPIR-V 1.4 features. Specifically, this PR adds testing for the NoSignedWrap and NoUnsignedWrap decorations.

We can reuse large parts of the testing for the `cl_khr_spirv_no_integer_wrap_decoration` extension to test the SPIR-V 1.4 feature.

### DIFF
--- a/test_conformance/spirv_new/CMakeLists.txt
+++ b/test_conformance/spirv_new/CMakeLists.txt
@@ -4,7 +4,6 @@ set(${MODULE_NAME}_SOURCES
   main.cpp
   test_basic_versions.cpp
   test_cl_khr_expect_assume.cpp
-  test_cl_khr_spirv_no_integer_wrap_decoration.cpp
   test_decorate.cpp
   test_get_program_il.cpp
   test_linkage.cpp
@@ -27,6 +26,7 @@ set(${MODULE_NAME}_SOURCES
   test_op_vector_extract.cpp
   test_op_vector_insert.cpp
   test_op_vector_times_scalar.cpp
+  test_no_integer_wrap_decoration.cpp
 )
 
 set(TEST_HARNESS_SOURCES
--- a/test_conformance/spirv_new/CMakeLists.txt
+++ b/test_conformance/spirv_new/CMakeLists.txt
@@ -7,6 +7,7 @@ set(${MODULE_NAME}_SOURCES
   test_decorate.cpp
   test_get_program_il.cpp
   test_linkage.cpp
+  test_no_integer_wrap_decoration.cpp
   test_op_atomic.cpp
   test_op_branch_conditional.cpp
   test_op_branch.cpp
@@ -26,7 +27,6 @@ set(${MODULE_NAME}_SOURCES
   test_op_vector_extract.cpp
   test_op_vector_insert.cpp
   test_op_vector_times_scalar.cpp
-  test_no_integer_wrap_decoration.cpp
 )
 
 set(TEST_HARNESS_SOURCES
--- a/test_conformance/spirv_new/spirv_asm/spv1.4/no_integer_wrap_decoration_fadd_int.spvasm32
+++ b/test_conformance/spirv_new/spirv_asm/spv1.4/no_integer_wrap_decoration_fadd_int.spvasm32
@@ -1,0 +1,46 @@
+; SPIR-V
+; Version: 1.4
+; Generator: Khronos SPIR-V Tools Assembler; 0
+; Bound: 23
+; Schema: 0
+               OpCapability Addresses
+               OpCapability Linkage
+               OpCapability Kernel
+          %1 = OpExtInstImport "OpenCL.std"
+               OpMemoryModel Physical32 OpenCL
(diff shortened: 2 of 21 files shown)


test_conformance/spirv_new/test_no_integer_wrap_decoration.cpp
```diff
@@ -183,39 +202,76 @@ int test_ext_cl_khr_spirv_no_integer_wrap_decoration(cl_device_id deviceID,
...
+                     "supported; skipping tests.\n");                          \
+            return TEST_SKIPPED_ITSELF;                                        \
+        }                                                                      \
+        return test_no_integer_wrap_decoration<cl_##TYPE>(                     \
+            deviceID, context, queue,                                          \
+            "ext_cl_khr_spirv_no_integer_wrap_decoration_" #FUNC "_" #TYPE,    \
+            #FUNC, #TYPE);                                                     \
+    }
+
+TEST_FMATH_FUNC_KHR(int, fadd)
```
svenvh, line 245: These names are confusing (fadd suggests floating point addition, but this extension only applies to integer types); iadd would make more sense.  However I am aware you're following existing code, so I'm fine if you want to leave the names as they are.

test_conformance/spirv_new/spirv_asm/spv1.4/no_integer_wrap_decoration_fadd_int.spvasm32
```diff
@@ -0,0 +1,46 @@
+; SPIR-V
+; Version: 1.4
+; Generator: Khronos SPIR-V Tools Assembler; 0
+; Bound: 23
+; Schema: 0
+               OpCapability Addresses
+               OpCapability Linkage
+               OpCapability Kernel
+          %1 = OpExtInstImport "OpenCL.std"
```
svenvh, line 9: Seems unused.

bashbaug: Yeah, it is.  I fixed these in the SPIR-V files I added for this test, but I didn't change any of the existing SPIR-V files, including those used to test the "no integer wrap decoration" via the extension.

I think this is pretty common because the SPIR-V LLVM Translator appears to be adding the OpExtInstImport pretty much unconditionally, so if you start with a file generated by the SPIR-V LLVM Translator you'd need to delete it manually, and I'm not in the habit of doing that (yet).

test_conformance/spirv_new/test_no_integer_wrap_decoration.cpp
bashbaug: Let me know what you'd like to do here.  To fix this we'd need to fix the test names in the .cpp file, and rename all of the .spvasm files.  This isn't too bad and I don't mind making the change, but it could also be somewhat disruptive.

Note, I could just "fix" the SPIR-V 1.4 tests since they're new, but that seems even more confusing, so I'd either like to change the SPIR-V 1.4 tests and the extension tests together, or leave them both as-is.

test_conformance/spirv_new/CMakeLists.txt
```diff
@@ -27,6 +26,7 @@ set(${MODULE_NAME}_SOURCES
   test_op_vector_extract.cpp
   test_op_vector_insert.cpp
   test_op_vector_times_scalar.cpp
+  test_no_integer_wrap_decoration.cpp
```
svenvh, line 29: Sorry one more nit: it would be nice to keep this list in alphabetical order.

test_conformance/spirv_new/spirv_asm/spv1.4/no_integer_wrap_decoration_fadd_int.spvasm32
svenvh: Ah, I wasn't even aware the SPIR-V LLVM Translator adds the OpExtInstImport unconditionally!

test_conformance/spirv_new/test_no_integer_wrap_decoration.cpp
svenvh: Only renaming the SPIR-V 1.4 tests seems like the worst solution indeed.  I'm inclined to leave the renaming of both as a future PR.

test_conformance/spirv_new/CMakeLists.txt
bashbaug: Ah, I didn't realize they were in alphabetical order previously.  Fixed!

test_conformance/spirv_new/test_no_integer_wrap_decoration.cpp
bashbaug: Sounds good.  I've filed https://github.com/KhronosGroup/OpenCL-CTS/issues/2057 to ensure we don't forget about this.

svenvh (approved):

bashbaug:

PR #2043, opened 2024-08-09
cl_fill_image: Use CL_DEVICE_VERSION instead of CL_DEVICE_NUMERIC_VER…

lakshmih:
…SION

CL_DEVICE_NUMERIC_VERSION is only available from OpenCL 3.0 Use CL_DEVICE_VERSION instead.


kpet (changes requested):
I suggest you use `get_device_cl_version` from the harness instead.

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -117,7 +108,8 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
                 cl_mem_flags buffer_flags = CL_MEM_READ_WRITE;
                 if (gEnablePitch)
                 {
-                    if (CL_VERSION_MAJOR(version) == 1)
+                    int major = (version.to_int()) / 10;
```
kpet, line 111: We should probably add `Version::major()` and `Version::minor()` getters to the harness code so users don't have to do math.

bashbaug (dismissed):

kpet (dismissed):
The change looks good but it seems you need to rebase or merge `main`.

bashbaug (approved):
Merging as discussed in the December 3rd teleconference.

PR #2045, opened 2024-08-13
Fix external semaphore test when exportability is not supported

saurabhnv:
An implementation may not support exportable semaphore, subtests available in cl_khr_external_semaphore assumes support for exportable semaphore, resulting in failure on such implementation. Allow implementations to use importable semaphore in such cases.

### DIFF
--- a/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
+++ b/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
@@ -4,7 +4,7 @@
 #include "opencl_vulkan_wrapper.hpp"
 #include <thread>
 #include <chrono>
-#include <unordered_set>
+#include <algorithm>
 
 #define FLUSH_DELAY_S 5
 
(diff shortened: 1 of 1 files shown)


lakshmih (approved):

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
```diff
@@ -96,18 +96,26 @@ static cl_int get_device_semaphore_handle_types(
...
+    if (num_handle_types > 0)
     {
-        handle_types.push_back(handle_type);
+        std::vector<cl_external_semaphore_handle_type_khr>
+            handle_types_query_result(num_handle_types);
+        err =
+            clGetDeviceInfo(deviceID, param,
+                            handle_types_query_result.size()
+                                * sizeof(cl_external_semaphore_handle_type_khr),
+                            handle_types_query_result.data(), nullptr);
```
bashbaug, line 110: Minor: could we read directly into the `handle_types` vector here, rather than reading into a temporary vector and then copying?

bashbaug (approved):
Note, I didn't try running these changes.

bashbaug:
Merging as discussed in the September 3rd memory subgroup.

PR #2046, opened 2024-08-13
Fix vulkan subtest for diffCtx and platform/device info

saurabhnv:
Fixes:
1. Multi import diff ctx subtest which acquires/releases external memory via queue not associated with the context in which memory was imported.
2. Platform/Device info subtests to handle different platforms and availability of the query.

### DIFF
--- a/test_conformance/vulkan/test_vulkan_platform_device_info.cpp
+++ b/test_conformance/vulkan/test_vulkan_platform_device_info.cpp
@@ -16,9 +16,11 @@
 
 #include <CL/cl.h>
 #include <CL/cl_ext.h>
+#include "harness/deviceInfo.h"
 #include "harness/testHarness.h"
 #include <iostream>
 #include <string>
+#include <vector>
 
 typedef struct
 {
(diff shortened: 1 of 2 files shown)


test_conformance/vulkan/test_vulkan_platform_device_info.cpp
```diff
@@ -48,6 +50,41 @@ _info device_info_table[] = {
...
+        != CL_SUCCESS)
+    {
+        throw std::runtime_error("clGetPlatformInfo failed\n");
+    }
+
+    /* The returned string does not include the null terminator. */
+    return std::string(info.data(), size - 1);
+}
+
+bool is_platform_extension_available(cl_platform_id platform, const char* extensionName)
```
bashbaug, line 82: These new functions really ought to go into the harness so they can be reused by other tests.  I'm a little surprised we don't have them already...

test_conformance/vulkan/test_vulkan_platform_device_info.cpp
```diff
@@ -76,7 +114,20 @@ int test_platform_info(cl_device_id deviceID, cl_context _context,
 
     for (i = 0; i < num_platforms; i++)
```
bashbaug, line 115: I think it would be more appropriate here to get the platform to test from the passed-in device handle rather than to loop over all of the platforms in the system.  Rationale: the CTS tests are testing a single device (and hence a single platform), chosen by the harndess.  The test results should be reported for that single device, and should not be affected by any other (possibly buggy or incomplete) platforms that are installed in the system.

test_conformance/vulkan/test_vulkan_platform_device_info.cpp
saurabhnv: Sure.

test_conformance/vulkan/test_vulkan_platform_device_info.cpp
saurabhnv: Yeah, makes sense, will update.

bashbaug (approved):
Thanks, latest version LGTM.

I'm still unable to choose the platform and device to run on for these tests using the usual mechanisms (e.g. `CL_PLATFORM_INDEX`), so some additional work is still needed, but this is a definite step in the right direction.

bashbaug:
Discussed in the September 3rd memory subgroup.  Waiting a bit longer for a review from Qualcomm.

lakshmih (approved):

bashbaug:
Merging as discussed in the September 24th teleconference.

PR #2049, opened 2024-08-15
Update test_event_dependencies.cpp

aharon-abramson:
In test_event_enqueue_wait_for_events_run_test:
* Decrease TEST_COUNT to 10
* Add clFinish to queues

### DIFF
--- a/test_conformance/events/test_event_dependencies.cpp
+++ b/test_conformance/events/test_event_dependencies.cpp
@@ -32,7 +32,7 @@ const char *write_kernels[] = {
 };
 
 #define TEST_SIZE 10000
-#define TEST_COUNT 100
+#define TEST_COUNT 10
 #define RANDOMIZE 1
 #define DEBUG_OUT 0
 
@@ -446,6 +446,14 @@ int test_event_enqueue_wait_for_events_run_test(
                                 &last_loop_event, NULL);
 
     test_error(error, "clEnqueueReadBuffer failed");
+
+    error = clFinish(queues[0]);
+    test_error(error, "clFinish(queues[0]) failed");
+    if (two_queues)
+    {
+        error = clFinish(queues[1]);
+        test_error(error, "clFinish(queues[1]) failed");
+    }
 
     failed = 0;
     for (i = 0; i < (int)TEST_SIZE; i++)


aharon-abramson:
Our implementation holds an internal reference count until a command (or one of its direct or indirect successors) is waited for and released (if the user holds a `cl_event` for it). When `TEST_COUNT` is 100, it means that our implementation must have at least 500 commands alive until a synchronization point for all of them is reached (which is currently also missing - therefore, I'm adding the calls for `clFinish`). I believe decreasing `TEST_COUNT` to 10 is enough for the test's purpose.

test_conformance/events/test_event_dependencies.cpp
```diff
@@ -447,6 +447,14 @@ int test_event_enqueue_wait_for_events_run_test(
 
     test_error(error, "clEnqueueReadBuffer failed");
 
+    error = clFinish(queues[0]);
```
EwanC, line 450: The `clEnqueueReadBuffer` call prior to this to `queues[0]` is blocking, I'm not sure if this `clFinish` is required for correct behaviour of this user code according to the OpenCL spec

aharon-abramson: Yes, but the read buffer command is only dependent on `last_loop_event`, while the other ND-range kernel and the two write buffer commands have no successors.

EwanC: ah makes sense, thanks

EwanC (approved):

test_conformance/events/test_event_dependencies.cpp
bashbaug: Sorry, I'm still not following why the `clFinish` calls are required.

Is this needed to ensure the correct data is read by `clEnqueueReadBuffer`?  (I don't think so, because the calls to `clFinish` occur after the call to `clEnqueueReadBuffer`.)

I suppose some commands may not have executed before the command-queue (or queues) is released, but that should be fine, too, and I don't think there is any requirement that a command-queue is empty before it is released.

Can you please say a little more what the calls to `clFinish` are doing?

aharon-abramson: You are right that there is no such requirement, but if there is no sync point for all commands, they might not complete until function `main` returns, and deleting them when the program exits might cause undefined finalization scenarios.

aharon-abramson: I've opened https://github.com/KhronosGroup/OpenCL-Docs/issues/1238 to discuss the spec issue.

bashbaug: Thanks!

I'm fine merging these changes as-is while the spec issue is under discussion.  Adding in the calls to `clFinish` definitely isn't wrong, though (depending how the spec issue gets resolved) they may be unnecessary.

lakshmih: Same, fine with these merging as-is while spec issue is under discussion.

lakshmih (approved):

bashbaug (approved):

aharon-abramson:
@bashbaug, can you please merge it?

bashbaug:
Merging - has three approved reviews.

PR #2055, opened 2024-08-19
compile Vulkan SPIR-V shaders in CMAKE_CURRENT_BINARY_DIR

### LINKED ISSUE #2040: stop generating Vulkan SPV in source directories
With #1878 we now generate the .spv files for the Vulkan-related CTS tests vs. using pre-compiled binaries, but the generation step puts the generated files in the source directory.  This generally violates CMake best practices, which is to put generated files in build-specific directories ("out-of-source build trees"), and clutters git status since the generated files are un-tracked.

We should generate the files in a build directory instead, likely based off of `CMAKE_CURRENT_BINARY_DIR`, rather than `CMAKE_CURRENT_SOURCE_DIR`.

If we can't do this for whatever reason, we should add the generated files to .gitignore, which will at least avoid the git status clutter.

bashbaug:
fixes #2040 

* Compiles the Vulkan SPIR-V files in a directory based off of CMAKE_CURRENT_BINARY_DIR.
* Changes the search path for the Vulkan SPIR-V files to this directory rather than a semi-arbitrary set of directories.

### DIFF
--- a/test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
+++ b/test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
@@ -50,7 +50,6 @@ const VulkanPhysicalDevice &getVulkanPhysicalDevice()
     cl_uint num_devices = 0;
     cl_uint device_no = 0;
     const size_t bufsize = BUFFERSIZE;
-    char buf[BUFFERSIZE];
     const VulkanInstance &instance = getVulkanInstance();
     const VulkanPhysicalDeviceList &physicalDeviceList =
         instance.getPhysicalDeviceList();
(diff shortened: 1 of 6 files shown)


bashbaug:
Hmm, this is going to need a bit more work on Windows and other multi-config generators, since these executables get built into a directory based on the current configuration.  Converting to a "draft" while I work this out.

bashbaug:
OK, I think I have a solution that works for the multi-config generators now.

kpet (approved):
A step in the right direction :).

lakshmih (approved):

nikhiljnv (approved):
LGTM.

Issue #1578, opened 2022-11-23
api consistency requirements for fp64 and fp16


bashbaug:
We don't currently have any API consistency tests for fp64 (and fp16) but there are enough moving parts here so we probably should... assuming it's clear what the API consistency requirements are (more on this in a bit).

Specifically, it seems that the following queries should be consistent:

If fp64 is supported...
* `cl_khr_fp64` must be in the extension string.
* `CL_DEVICE_DOUBLE_FP_CONFIG` must return a nonzero value.
* `CL_DEVICE_PREFERRED_VECTOR_WIDTH_DOUBLE` must return a nonzero value.
* (not 100% sure about this one) `CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE` must return a nonzero value? What about "soft float" implementations? Can they return zero because they have no "native" double support?

Otherwise, when fp64 is NOT supported...
* `cl_khr_fp64` must not be in the extension string.
* `CL_DEVICE_DOUBLE_FP_CONFIG` must return zero.
* `CL_DEVICE_PREFERRED_VECTOR_WIDTH_DOUBLE` must return zero.
* `CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE` must return zero.

I think the same logic should also hold for fp16, the `cl_khr_fp16` extension, and the analogous `HALF` queries.

This seemed fairly straightforward to me, though I did find this comment in one of the CTS tests:

https://github.com/KhronosGroup/OpenCL-CTS/blob/9137cb66ed01f23d1eec84846635ff61491361cd/test_conformance/api/test_queries.cpp#L575

```c++
    // Note that even if cl_khr_fp64, the preferred width for double can be non-zero.  For example, vendors
    // extensions can support double but may not support cl_khr_fp64, which implies math library support.
```

Does this make sense, or is this a stale comment that can be removed?

kamil-goras-mobica:
@bashbaug added PR: https://github.com/KhronosGroup/OpenCL-CTS/pull/2058

PR #2058, opened 2024-08-23
Added fp16 and fp64 api consistency tests

kamil-goras-mobica:
#1578


test_conformance/api/test_api_consistency.cpp
```diff
@@ -1148,3 +1148,121 @@ int test_consistency_3d_image_writes(cl_device_id deviceID, cl_context context,
...
+                          "CL_DEVICE_DOUBLE_FP_CONFIG must return 0");
+
+        error =
+            clGetDeviceInfo(deviceID, CL_DEVICE_PREFERRED_VECTOR_WIDTH_DOUBLE,
+                            sizeof(value), &value, nullptr);
+        test_error(
+            error,
+            "Unable to get device CL_DEVICE_PREFERRED_VECTOR_WIDTH_DOUBLE");
+        test_assert_error(
+            value > 0, "CL_DEVICE_PREFERRED_VECTOR_WIDTH_DOUBLE must return 0");
```
svenvh, line 1199: Is this wrong?
```suggestion
            value == 0, "CL_DEVICE_PREFERRED_VECTOR_WIDTH_DOUBLE must return 0");
```
(and same for `CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE`)

test_conformance/api/test_api_consistency.cpp
```diff
@@ -1148,3 +1148,124 @@ int test_consistency_3d_image_writes(cl_device_id deviceID, cl_context context,
...
+    cl_int error = CL_SUCCESS;
+    cl_device_fp_config value = 0;
+
+    if (is_extension_available(deviceID, "cl_khr_fp64"))
+    {
+        error = clGetDeviceInfo(deviceID, CL_DEVICE_DOUBLE_FP_CONFIG,
+                                sizeof(value), &value, nullptr);
+        test_error(error, "Unable to get device CL_DEVICE_DOUBLE_FP_CONFIG");
+        test_assert_error(
+            value > 0, "CL_DEVICE_DOUBLE_FP_CONFIG must return nonzero value");
```
svenvh, line 1165: Maybe a question for other reviewers or @bashbaug: the specification states some minimum double precision floating-point capabilities that we could check here instead of only checking that `value > 0`.  Would that be worth adding?
(and likewise for half precision)

svenvh (commented):
The latest fixes LGTM; left a question for other reviewers.

test_conformance/api/test_api_consistency.cpp
bashbaug: I agree - it would be preferable to check for the minimum set of required values.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -1223,11 +1243,27 @@ int test_consistency_requirements_fp16(cl_device_id deviceID,
...
+
         error = clGetDeviceInfo(deviceID, CL_DEVICE_HALF_FP_CONFIG,
                                 sizeof(value), &value, nullptr);
         test_error(error, "Unable to get device CL_DEVICE_HALF_FP_CONFIG");
         test_assert_error(value > 0,
                           "CL_DEVICE_HALF_FP_CONFIG must return nonzero value");
+        if (version < Version(2, 0))
+        {
+            test_assert_error(
+                value & (CL_FP_ROUND_TO_NEAREST | CL_FP_INF_NAN),
```
svenvh, line 1256: The [specification](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_HALF_FP_CONFIG) for `CL_DEVICE_HALF_FP_CONFIG` doesn't explicitly mention that this config is tied to versions < 2.0:

> the minimum half-precision floating-point capability for OpenCL 2.0 or newer devices is:
> 
> [CL_FP_ROUND_TO_ZERO](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_FP_ROUND_TO_ZERO)
> 
> or
> 
> [CL_FP_ROUND_TO_NEAREST](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_FP_ROUND_TO_NEAREST) |
> [CL_FP_INF_NAN](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_FP_INF_NAN).
> 

  The specification for `..._DOUBLE_FP_CONFIG` is unambiguous, but for `..._HALF_FP_CONFIG` the wording could also be interpreted as "either of these configs can be considered a minimum config for OpenCL 2.0 or newer".

I'm not quite sure which interpretation was intended, maybe @bashbaug can advise?

bashbaug: Yeah, I think there's a minor spec bug here, and I just filed https://github.com/KhronosGroup/OpenCL-Docs/issues/1267 to fix it.

My understanding is:

* The OpenCL version supported by the device is irrelevant.
* Devices supporting half-precision (via `cl_khr_fp16`) must either support `CL_FP_ROUND_TO_ZERO` or the combination of `CL_FP_ROUND_TO_NEAREST | CL_FP_INF_NAN`.

This is consistent with the `cl_khr_fp16` description in the last standalone extension doc v3.0.15.

svenvh (approved):

test_conformance/api/test_api_consistency.cpp
```diff
@@ -1151,3 +1151,148 @@ int test_consistency_3d_image_writes(cl_device_id deviceID, cl_context context,
...
+                          "CL_DEVICE_PREFERRED_VECTOR_WIDTH_DOUBLE must return "
+                          "nonzero value");
+
+        error = clGetDeviceInfo(deviceID, CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE,
+                                sizeof(value), &value, nullptr);
+        test_error(error,
+                   "Unable to get device CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE");
+        test_assert_error(
+            value > 0,
+            "CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE must return nonzero value");
```
bashbaug, line 1206: So we go in with eyes open: this change requires `CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE` to be non-zero when fp64 is supported.  This is true even if there is no "native" double precision support, say when `CL_DEVICE_DOUBLE_FP_CONFIG` includes `CL_FP_SOFT_FLOAT`.

No action required, but if we did want to "fix" this, some options might be:

1. Remove the check entirely - the spec doesn't have any requirements for a value for `CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE` when fp64 is supported.  Note, there is a requirement with fp64 is NOT supported, so this check should not change.
2. Have the check only when `CL_DEVICE_DOUBLE_FP_CONFIG` does not include `CL_FP_SOFT_FLOAT`, since there should be some "native" support in this case.

Same comments also apply to fp16.

bashbaug (approved):

test_conformance/api/test_api_consistency.cpp
bashbaug: Discussed in the October 1st teleconference.  We are going to leave this as-is for now.  We will revisit in the future, if needed, if a vendor wants to return zero for `CL_DEVICE_NATIVE_VECTOR_WIDTH_DOUBLE` and still support fp64 for some reason.

bashbaug:

test_conformance/api/test_api_consistency.cpp
```diff
@@ -1151,3 +1151,148 @@ int test_consistency_3d_image_writes(cl_device_id deviceID, cl_context context,
...
+                   "Unable to get device CL_DEVICE_NATIVE_VECTOR_WIDTH_HALF");
+        test_assert_error(
+            value > 0,
+            "CL_DEVICE_NATIVE_VECTOR_WIDTH_HALF must return nonzero value");
+    }
+    else
+    {
+        error = clGetDeviceInfo(deviceID, CL_DEVICE_HALF_FP_CONFIG,
+                                sizeof(value), &value, nullptr);
+        test_error(error, "Unable to get device CL_DEVICE_HALF_FP_CONFIG");
```
hvdijk, line 1277: This requires implementations without the `cl_khr_fp16` extension to recognise `CL_DEVICE_HALF_FP_CONFIG`, but support for `CL_DEVICE_HALF_FP_CONFIG` is only required if that extension is available, is it not? https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_HALF_FP_CONFIG says "provided by the cl_khr_fp16 extension". Our implementation, when FP16 support is disabled, returns `CL_INVALID_VALUE` here rather than setting `value` to `0`, and I think that is correct, since the spec also says it returns:

> [CL_INVALID_VALUE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_VALUE) if param_name is not one of the supported values or if size in bytes specified by param_value_size is < size of return type as specified in the [Device Queries](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#device-queries-table) table and param_value is not a NULL value or if param_name is a value that is available as an extension and the corresponding extension is not supported by the device.

svenvh: @hvdijk I agree with your analysis.  `CL_DEVICE_HALF_FP_CONFIG` is provided by the `cl_khr_fp16` extension, so if that extension is not supported the query should return `CL_INVALID_VALUE` indeed.

(the half query seems different from `CL_DEVICE_DOUBLE_FP_CONFIG`, which does not have a "provided by" clause)

hvdijk: Thanks, created https://github.com/KhronosGroup/OpenCL-CTS/pull/2108 to change that.

PR #2060, opened 2024-08-28
[NFC] Use std::vector to store large size array.

haonanya:
windows default stack size is 1M, BUFFER_SIZE is 2 * 1024 * 1024, use array with BUFFER_SIZE elemets on stack can exceed available stack size limits.


haonanya:
@svenvh, can you please take a look? This align with https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/math_brute_force/unary_two_results_half.cpp#L58.

svenvh (approved):

haonanya:
@svenvh , are we ready to land? I don't have merge permissions, thanks.

svenvh:
Merging as this should be uncontroversial.

PR #2061, opened 2024-08-28
[NFC] math_brute_force: remove warning options

svenvh:
All of these are already set in the top-level `CMakeLists.txt` nowadays, so no need to repeat them in the test's own `CMakeLists.txt`.

### DIFF
--- a/test_conformance/math_brute_force/CMakeLists.txt
+++ b/test_conformance/math_brute_force/CMakeLists.txt
@@ -54,10 +54,6 @@ set(${MODULE_NAME}_SOURCES
     utility.h
 )
 
-# math_brute_force compiles cleanly with -Wall (except for a few remaining
-# warnings), but other tests not (yet); so enable -Wall locally.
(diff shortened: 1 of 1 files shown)


svenvh:
Suspecting there was an update to the Windows runners that broke our Windows builds:
```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.41.34120\include\cmath(831): error C2062: type 'signed char' unexpected
```

Previous runs with `MSVC/14.40.33807` were passing.

bashbaug (approved):
LGTM.

I don't see anywhere we are setting `-Wno-strict-aliasing` anymore, but maybe that's no longer a problem?

svenvh:
> I don't see anywhere we are setting `-Wno-strict-aliasing` anymore, but maybe that's no longer a problem?

Good catch, I should have said "... or no longer needed", as we don't need `-Wno-strict-aliasing` anymore.

svenvh:
Close/reopen to retrigger CI.

EwanC (approved):

PR #2062, opened 2024-08-29
fix generic_address_space, command_buffer_event_sync, test_compiler and images/test_1D_buffer

franz:
fixes several issues:

* `generic_address_space` test: add check for program scope variables. The test uses both generic AS and program-scope variables, however it only checked the generic-AS presence in `clGetDeviceInfo`.
* `compiler/test_compiler_defines_for_extensions.cpp`: add `cl_khr_command_buffer_multi_device` to the list of recognized extensions
* `command_buffer_event_sync.cpp`: add delays for testing `clSetEventCallback` - according to specification, these can be executed asynchronously (in a separate thread) by the OpenCL implementation, hence the event callback is not quaranteed to be called before `clFinish()` returns. Existing test `events/test_callbacks.cpp` also waits for callback with loops of usleep.
* `images/kernel_image_methods/test_1D_buffer.cpp`: fix allocation size being too small for the 1D buffer backing the image

### DIFF
--- a/test_conformance/generic_address_space/main.cpp
+++ b/test_conformance/generic_address_space/main.cpp
@@ -92,7 +92,8 @@ test_status InitCL(cl_device_id device) {
     if (version >= Version(3, 0))
     {
         cl_int error;
-        cl_bool support_generic;
+        cl_bool support_generic = CL_FALSE;
+        size_t max_gvar_size = 0;
 
         error = clGetDeviceInfo(device, CL_DEVICE_GENERIC_ADDRESS_SPACE_SUPPORT,
                                 sizeof(support_generic), &support_generic, NULL);
@@ -103,6 +104,21 @@ test_status InitCL(cl_device_id device) {
         }
 
         if (!support_generic)
+        {
+            return TEST_SKIP;
+        }
+
+        error = clGetDeviceInfo(device, CL_DEVICE_MAX_GLOBAL_VARIABLE_SIZE,
+                                sizeof(max_gvar_size), &max_gvar_size, NULL);
+        if (error != CL_SUCCESS)
+        {
+            print_error(error,
+                        "Unable to MAX_GLOBAL_VARIABLE_SIZE value. "
+                        "Program scope variable support is required");
+            return TEST_FAIL;
+        }
+
+        if (!max_gvar_size)
         {
             return TEST_SKIP;
         }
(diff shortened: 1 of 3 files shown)


franz:
The code check is complaining about the extension list, however i think the existing formatting is more readable.

test_conformance/generic_address_space/main.cpp
```diff
@@ -106,6 +107,21 @@ test_status InitCL(cl_device_id device) {
...
             return TEST_SKIP;
         }
+
+        error = clGetDeviceInfo(device, CL_DEVICE_MAX_GLOBAL_VARIABLE_SIZE,
+                                sizeof(max_gvar_size), &max_gvar_size, NULL);
+        if (error != CL_SUCCESS)
+        {
+            print_error(error,
+                        "Unable to MAX_GLOBAL_VARIABLE_SIZE value. "
+                        "Program scope variable support is required");
```
bashbaug, line 117: Nitpick: the current phrasing for this error message is confusing.  How about:

```suggestion
            print_error(error,
                        "Unable to query CL_DEVICE_MAX_GLOBAL_VARIABLE_SIZE.");
```

test_conformance/images/kernel_image_methods/test_1D_buffer.cpp
```diff
@@ -67,14 +67,13 @@ static int test_get_1Dimage_buffer_info_single(cl_context context,
     if (gDebugTrace)
         log_info(" - Creating 1D image %d ...\n", (int)imageInfo->width);
 
-    buffer = clCreateBuffer(
-        context, flags, imageInfo->width * get_pixel_size(imageInfo->format),
-        NULL, &error);
+    buffer =
+        clCreateBuffer(context, flags, imageValues.getSize(), NULL, &error);
```
bashbaug, line 71: Note, there is a merge conflict here now, caused by the recently merged https://github.com/KhronosGroup/OpenCL-CTS/pull/2073.  It's a little different than the change in this PR, though.  Is there a reason to prefer one fix vs. the other?

franz: No, AFAICT they are both the exact same number (row pitch). I'll drop the patch

franz:
Dropped the conflicting commit & rebased onto main.

bashbaug (approved):

bashbaug:

PR #2064, opened 2024-08-30
Specify precision in the format flag for float4 hexadecimal output

delecui:
This will ensure consistent output on any conforming implementation.
Note that the specification does not define the output style when
precision is missing.


test_conformance/printf/util_printf.cpp
```diff
@@ -1231,7 +1231,12 @@ std::vector<std::string> correctBufferVector = {
 
     "1.23e+03,9.88e+05,5.00e-04",
 
+#ifdef _WIN32
```
delecui, line 1234: See https://godbolt.org/z/YzPsj1W4f and https://godbolt.org/z/hhPMd6zoo
Both outputs conform to OpenCL 3.0 and C99 standards regarding `%a` specifier where precision is missing.

OpenCL spec demands `if the precision is missing, then the precision is sufficient for an exact representation of the value`
See [https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#printf-format-string](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#printf-format-string:~:text=if%20the%20precision%20is%20missing%2C%20then%20the%20precision%20is%20sufficient%20for%20an%20exact%20representation%20of%20the%20value)

delecui:
Hi @svenvh and @bashbaug, could you help review the changes? Thanks.

test_conformance/printf/util_printf.cpp
svenvh: As both results are correct, I think we should not guard this on `_WIN32`.  Instead I would suggest specifying a precision in the entry in `printVectorGenParameters`, so that the output will be the same on any implementation.

delecui: Agreed. Thanks for your suggestion.
`_WIN32` conditional compilation has been removed.
precision `.1` is added to generate format `%.1v4hla` and the reference is updated accordingly.

svenvh: LGTM, thanks for the update!

svenvh (approved):

bashbaug (approved):

delecui:
Hi @svenvh, Could you help merge this PR? Thanks. I don't have permission to merge.

svenvh:
> Hi @svenvh, Could you help merge this PR? Thanks. I don't have permission to merge.

We may want to leave this open for a few more days, so that other vendors have a chance to look at it.

delecui:
> > Hi @svenvh, Could you help merge this PR? Thanks. I don't have permission to merge.
> 
> We may want to leave this open for a few more days, so that other vendors have a chance to look at it.

Understood. Thanks for the update. Will wait for further reviews.

svenvh:
Close/reopen to retrigger CI.

delecui:
Hi @svenvh, Can we land this PR now ? Thanks.

svenvh:
This has been open for almost 3 weeks now, and has 2 approvals, so merging.

Issue #1833, opened 2023-10-21
isnan, isfinite, and isinf are implemented as macros for MSVC


bashbaug:
The CTS tests currently define `isnan`, `isfinite`, and `isinf` macros for MSVC:
https://github.com/KhronosGroup/OpenCL-CTS/blob/40aaa2bcf2460a083b89427cffaacdbd5f707ffc/test_common/harness/compat.h#L141-L151

Unfortunately, this means we cannot use the functions `std::isnan`, `std::isfinite`, and `std::isinf` functions from C++11 without generating build errors on Windows only.

We should switch to the C++ versions exclusively.  Failing that, we should find another way to define these aliases that doesn't preclude using the C++ versions.  Maybe:

```c++
using isnan = std::isnan;
using isfinite = std::isfinite;
using isinf = std::isinf;
```

kamil-goras-mobica:
@bashbaug Could you take a look at this? Is it solving problem? https://github.com/KhronosGroup/OpenCL-CTS/pull/2121

PR #2065, opened 2024-09-02
Fix Windows build with MSVC 19.41

svenvh:
Include `cmath` instead of `math.h` in C++ mode under MSVC, to avoid
build errors inside the header.  Ideally we would not condition this
on `_MSC_VER`, but issue 1833 currently prevents doing so.

### DIFF
--- a/test_common/harness/compat.h
+++ b/test_common/harness/compat.h
@@ -112,6 +112,8 @@ int feclearexcept(int excepts);
 
 #if defined(__INTEL_COMPILER)
 #include <mathimf.h>
+#elif __cplusplus && defined(_MSC_VER)
+#include <cmath>
 #else
 #include <math.h>
 #endif


delecui (approved):
The fix works fine for me. 
I was evaluating removing the macros like `isinf`, `isnan`, and other ones from `compat.h` that conflict with `cmath` in MSVC 19.41.
However, this fix is more concise and safer.

svenvh:
> I was evaluating removing the macros like `isinf`, `isnan`, and other ones from `compat.h` that conflict with `cmath` in MSVC 19.41.

As I don't have access to an MSVC environment, I wasn't quite sure what was causing the trouble (so my fix was a shot in the dark essentially).  Good to know that those macros are the culprit, we should prioritize #1833 then.

bashbaug (approved):
Thanks, LGTM!

#1833 is already on Mobica's backlog so hopefully this is a short-term fix and we can include cmath unconditionally soon.

bashbaug:
I think we should merge this ASAP to unblock CI builds.  @lakshmih and others, any concerns?

svenvh:
Merging to get CI back to green.  If any concerns/issues are reported post-merge, we can consider reverting and investigate an alternative solution.

PR #2068, opened 2024-09-03
Added semaphore test to cover out-of-order operations synchronized with barrier

### LINKED ISSUE #1979: Semaphore plus Barrier plus Out-of-Order Queue Test
See related OpenCL spec issue discussion: https://github.com/KhronosGroup/OpenCL-Docs/issues/1178#issuecomment-2148210553

We should have a targeted semaphore test that mixes semaphores, barriers, and out-of-order queues, since this is likely to be a common pattern for applications using semaphores in an out-of-order queue.

The rough flow will look something like:

```c++
// Producers:
// A producer operation can can be any enqueue operation, such as a kernel or a memory fill.
clEnqueueXXX(producer_queue, ...);
// There can be more than one independent producer operations.
clEnqueueXXX(producer_queue, ...);

// This barrier requires all producer operations to be complete:
clEnqueueBarrierWithWaitList(producer_queue); 

// The semaphore cannot be signaled until the barrier is complete:
clEnqueueSignalSemaphore(producer_queue, semaphore);

// Note: the consumer_queue can be a separate queue or the same as the producer_queue!
// If the consumer_queue and the producer_queue are the same then there needs to be
// an explicit dependency between the semaphore signal and the semaphore wait, which
// could be an event dependency or another barrier.

// Consumers:
// Wait for the producers to signal completion via the semaphore:
clEnqueueWaitSemaphore(consumer_queue, semaphore);

// This barrier ensures all consumers cannot start until the semaphore wait is complete:
clEnqueueBarrierWithWaitList(consumer_queue);

// Now consumers can execute, and should see the producer results.
clEnqueueXXX(consumer_queue, ...);
clEnqueueXXX(consumer_queue, ...);
```

Additional notes:

- Suggest testing multiple independent producer and consumer operations, two each should be fine.
- Suggest testing the same out-of-order queue for both producers and consumers, and separate producer and consumer out-of-order queues.
- For the same out-of-order queue test, it's probably fine to test either a barrier or an explicit event dependency between the semaphore signal and wait.

shajder:
Fixes #1979 according to task description.


shajder:
close/open to retrigger CI

shajder:
> I don't think this test is doing the right thing. Please double-check the flow described in #1979...

Right, I think I see the mistake, sorry for confusion. After correction producer operations do not write destination buffer (only source buffer twice) and consumer operations do not write source buffer (write destination, then copy source to destination). Please let me know if I am missing something else. Thanks

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
```diff
@@ -0,0 +1,321 @@
...
+// limitations under the License.
+//
+
+#include <vector>
+
+#include "semaphore_base.h"
+
+namespace {
+
+int num_elems = 0;
```
bashbaug, line 23: Can we get rid of this non-const global variable?  We should pass the number of events to the test class constructor and store it per-class, instead.

bashbaug (changes requested):
Getting closer, but there are still some problems that need to be fixed.

Note, the barrier test is currently using a single producer and a single consumer.  Ideally we would have two producers and two consumers instead, but if we want to do this in a follow-on PR (or not at all) I would be fine with a single producer and a single consumer, for now.

shajder:
> Getting closer, but there are still some problems that need to be fixed.
> 
> Note, the barrier test is currently using a single producer and a single consumer. Ideally we would have two producers and two consumers instead, but if we want to do this in a follow-on PR (or not at all) I would be fine with a single producer and a single consumer, for now.

@bashbaug I will refactor the test to use two producers/consumers, that will make much more sense.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+        err = clEnqueueWriteBuffer(
+            producer_queue, in_mem, CL_FALSE, 0, sizeof(cl_int) * num_elems,
+            host_buffer.data(), 0, nullptr, &wait_events[0]);
+        test_error(err, "clEnqueueReadBuffer failed");
+
+        err = clEnqueueFillBuffer(producer_queue, in_mem, &pattern_sec,
+                                  sizeof(cl_int), 0, sizeof(cl_int) * num_elems,
+                                  1, &wait_events[0], nullptr);
+        test_error(err, "clEnqueueFillBuffer failed");
```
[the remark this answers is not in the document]
shajder: Replaced with two producer + two consumer operations:

1) each producer operation writing half of input buffer simultaneously
2) each consumer operation copies half input to output buffer simultaneously (with modified kernel with offset argument)

side note: we could replace `clEnqueueBarrierWithWaitList` calls with wait-events - effect the same but slightly cleaner code.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
```diff
@@ -0,0 +1,321 @@
...
+            test_error(err, "Could not wait semaphore");
+        }
+
+        err = clEnqueueBarrierWithWaitList(consumer_queue, 0, nullptr, nullptr);
+        test_error(err, " clEnqueueBarrierWithWaitList ");
+
+        err = clEnqueueFillBuffer(consumer_queue, out_mem, &pattern_pri,
+                                  sizeof(cl_int), 0, sizeof(cl_int) * num_elems,
+                                  0, nullptr, &wait_events[0]);
+        test_error(err, "clEnqueueFillBuffer failed");
```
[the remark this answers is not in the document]
shajder: Refactored with two consumer operations, each copying half of input to output buffer.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+        size_t threads = (size_t)num_elems;
+        err = clEnqueueNDRangeKernel(consumer_queue, kernel, 1, nullptr,
+                                     &threads, nullptr, 1, &wait_events[0],
+                                     &wait_events[1]);
```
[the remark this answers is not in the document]
shajder: Removed reusing of wait events.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
shajder: Done! Side note: as consequence I had to correct other tests due to the fact all uses the same interface.

bashbaug: OK, I see.  As an alternative, if you aren't comfortable changing so many existing tests, consider adding overloads for `MakeAndRunTest`, so there is one version with an `nelems` parameter and one version without.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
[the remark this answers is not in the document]
bashbaug: I'm sorry, I don't think I'm explaining this very well.  Catch me on Slack if you'd like and we can have a more interactive conversation.

By "multiple producers" I'm thinking of two commands in the same queue, not two separate producer queues.  I'm also thinking that each command should produce into a unique memory object, so not writing into the same memory memory object, which is probably a data race.

Here's the original snip from #1979 with a bit more detail what I am looking for.  I am going to use "fill" as my producer and "copy" as my consumer, but other options are also viable:

```c++
// Producers:
// Note, these commands should have no event dependencies and produce no events!
clEnqueueFillBuffer(producer_queue, temp_data_a, pattern_a);
clEnqueueFillBuffer(producer_queue, temp_data_b, pattern_b);

// This barrier requires all producer operations to be complete:
clEnqueueBarrierWithWaitList(producer_queue); 

// The semaphore cannot be signaled until the barrier is complete:
clEnqueueSignalSemaphore(producer_queue, semaphore);

// Note: the consumer_queue can be a separate queue or the same as the producer_queue!
// If the consumer_queue and the producer_queue are the same then there needs to be
// an explicit dependency between the semaphore signal and the semaphore wait, which
// could be an event dependency or another barrier.

// Consumers:
// Wait for the producers to signal completion via the semaphore:
clEnqueueWaitSemaphore(consumer_queue, semaphore);

// This barrier ensures all consumers cannot start until the semaphore wait is complete:
clEnqueueBarrierWithWaitList(consumer_queue);

// Note: these commands should also have no event dependencies and produce no events!
clEnqueueCopyBuffer(consumer_queue, temp_data_a, result_data_a);
clEnqueueCopyBuffer(consumer_queue, temp_data_b, result_data_b);

clEnqueueReadBuffer(consumer_queue, result_data_a, ...); // should read pattern_a
clEnqueueReadBuffer(consumer_queue, result_data_b, ...); // should read pattern_b
```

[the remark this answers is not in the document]
shajder: Understood... I hope :). I reverted single producer/consumer queues and doubled input/output buffers. In order to copy buffers instead of `clEnqueueCopyBuffer` I am using previous kernels, please let me know if that works for you.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
shajder: Indeed, it crossed my mind. After all, the work has already been done, I think it is a bit more consistent with other `MakeAndRun` procedures.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
```diff
@@ -0,0 +1,368 @@
...
+            test_error(err, "clEnqueueReadBuffer failed");
+
+            err = clEnqueueFillBuffer(
+                producer_queue, in_mem_B, &pattern_B, sizeof(cl_int), 0,
+                sizeof(cl_int) * num_elems, 1, &user_event, nullptr);
+            test_error(err, "clEnqueueReadBuffer failed");
+
+            // launch producer operations simultaneously
+            err = clSetUserEventStatus(user_event, CL_COMPLETE);
+            test_error(err, "clSetUserEventStatus failed");
```
bashbaug, line 222: This isn't a must-fix, but I don't think the user event is required here, since I don't think "real-world" code will hold back the producers like this.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
```diff
@@ -0,0 +1,368 @@
...
+            test_error(err, "clSetKernelArg failed");
+
+            err = clEnqueueNDRangeKernel(consumer_queue, kernel, 1, nullptr,
+                                         &threads, nullptr, 1, &user_event,
+                                         &wait_events[1]);
+            test_error(err, "clEnqueueNDRangeKernel failed");
+
+            // launch consumer operations simultaneously
+            err = clSetUserEventStatus(user_event, CL_COMPLETE);
+            test_error(err, "clSetUserEventStatus failed");
```
bashbaug, line 286: Likewise, I don't think a user event is required here, either.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
```diff
@@ -0,0 +1,368 @@
...
+
+            err = clEnqueueSignalSemaphoresKHR(consumer_queue, 1, semaphore,
+                                               nullptr, 2, &wait_events[0],
+                                               nullptr);
+            test_error(err, "Could not signal semaphore");
```
bashbaug, line 291: I'm not following why this semaphore signal is required.  What is it for?

If it can be removed (and I think it can), then we can remove all use of `wait_events` also, since the `clEnqueueReadBuffer` calls to verify results are blocking.

bashbaug (approved):
I'm OK with this as-is and I think the important issues have been addressed, but here are a few nice-to-haves.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
shajder: Corrected

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
shajder: Corrected

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
shajder: Ops, I assumed reading buffer must be synchronized with kernels due to out-of-order type queue. Since that is not the case I have removed event dependency, thanks.

joshqti:
This change looks fine to me

test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
bashbaug: > I assumed reading buffer must be synchronized with kernels due to out-of-order type queue.

Ah, I'm sorry, I led you astray here.  Yes, there does need to be some sort of dependency between the kernels and the reads, otherwise the reads can execute before the kernels complete.  I don't think we should do this with an event dependency on a call to `clEnqueueSignalSemaporeKHR` like we had previously, but we will need either an explicit event dependency or another call to `clEnqueueBarrierWithWaitList` between the kernels and the calls to `clEnqueueReadBuffer`.

bashbaug (approved):
Thanks, LGTM!

bashbaug:
Merging as discussed in the November 19th teleconference.

PR #2050, opened 2024-08-15
Add properties parameter to all command-buffer commands

EwanC:
Make changes to reflect OpenCL-Docs change https://github.com/KhronosGroup/OpenCL-Docs/pull/1215 and requires header change https://github.com/KhronosGroup/OpenCL-Headers/pull/260 to use preprocessor path with extra parameters

Used [CL_DEVICE_EXTENSIONS_WITH_VERSION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_EXTENSIONS_WITH_VERSION) runtime check and skipped testing when a device doesn't support required versions.

Tested using OCK updated with this API change https://github.com/codeplaysoftware/oneapi-construction-kit/pull/519

PR #2072, opened 2024-09-06
Add properties parameter to all command-buffer commands (update)

bashbaug:
This is an update to #2050, with updates from the main branch.


bashbaug:
Merging as discussed in the September 3rd teleconference + email.

PR #2074, opened 2024-09-06
fix a few array-based mutable command buffer issues

bashbaug:
fixes a few minor issues introduced by #1984.

edit: Tested with the updated command buffer emulation layer on a device that supports out-of-order queues and SVM and all of the `test_cl_khr_mutable_dispatch` tests pass.


EwanC (approved):
Thanks for catching these

PR #2075, opened 2024-09-09
vulkan: Propagate fix to multiImport diffCtx

joshqti:
Vulkan descriptors are being programmed incorrectly, use fix from 1676 in other subtest


bashbaug:
Discussed in the October 15th teleconference.  Applies the same fix from https://github.com/KhronosGroup/OpenCL-CTS/pull/1676 to another case.  OK to merge.

PR #2076, opened 2024-09-09
vulkan: Use Vulkan 1.1

joshqti:
Increase Vulkan version to 1.1, which is required to support vkGetBufferMemoryRequirements2.


nikhiljnv (approved):

nikhiljnv:
Merging this as discussed in Memory subgroup call on October 1st, 2024.

PR #2077, opened 2024-09-10
vulkan: Use image row pitch

joshqti:
When importing a Vulkan external image, query and
pass OpenCL a row pitch if OpenCL is assuming linear for the imported external handle type.  Additionally fix a bug where OpenCL is being told to create mipmapped images at all times.

### DIFF
--- a/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
+++ b/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
@@ -482,7 +482,7 @@ getCLImageInfoFromVkImageInfo(const VkImageCreateInfo *VulkanImageCreateInfo,
     img_desc->image_row_pitch = 0; // Row pitch set to zero as host_ptr is NULL
     img_desc->image_slice_pitch =
         img_desc->image_row_pitch * img_desc->image_height;
-    img_desc->num_mip_levels = 1;
+    img_desc->num_mip_levels = 0;
     img_desc->num_samples = 0;
     img_desc->buffer = NULL;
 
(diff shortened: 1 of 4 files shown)


test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -676,6 +677,14 @@ clExternalMemoryImage::clExternalMemoryImage(
     std::vector<cl_mem_properties> extMemProperties1;
     cl_device_id devList[] = { deviceId, NULL };
 
+    VulkanImageTiling vulkanImageTiling =
+        vkClExternalMemoryHandleTilingAssumption(
```
saurabhnv, line 681: Is this query required? VkImageCreateInfo is available, why not use its tiling field which gets populated during Vulkan image creation?

Also, since this change is for linear images, may I suggest a follow up change (or if possible, include in this?) in vulkan/\*consistency_for_{1,3}d*.cpp?
Those are querying opencl linear tiling assumption, which is incorrect because vkCreateImage has a restriction where image type other than 2D is invalid for linear tiled image.

joshqti: We are querying OpenCL's tiling assumption because if OpenCL assumes linear, it implies that it cannot infer the underlaying memory layout from the external memory handle type, and a row pitch should be provided to guarantee OpenCL imports the image correctly.  

If the OpenCL tiling assumption is not linear, a row pitch is not required, even for linear Vulkan images, because the OpenCL implementation can correctly infer the memory layout from the external memory handle.

kpet (approved):

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
nikhiljnv: The comment above doesn't necessarily impact the change here, but just want to make sure that we are not using assume_linear query to imply inference which is supposed to be a different capability in itself.

nikhiljnv:
The change itself looks fine to me, but will double check with @saurabhnv and unblock here.

bashbaug:

PR #2078, opened 2024-09-10
vulkan: Remove redundant negative testing

joshqti:
- negative testing for semaphore functions is accomplished in semaphore tests, as well as create image in the api test

### DIFF
--- a/test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
+++ b/test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
@@ -158,16 +158,6 @@ int test_consistency_external_for_1dimage(cl_device_id deviceID,
     test_error(errNum, "Unable to create Image with Properties");
     image.reset();
 
-    // Passing properties, image_desc and image_format all as NULL
-    image = clCreateImageWithProperties(context, NULL, CL_MEM_READ_WRITE, NULL,
-                                        NULL, NULL, &errNum);
-    test_failure_error(
-        errNum, CL_INVALID_IMAGE_DESCRIPTOR,
-        "Image creation must fail with CL_INVALID_IMAGE_DESCRIPTOR "
-        "when all are passed as NULL");
-
-    image.reset();
-
     // Passing NULL properties and a valid image_format and image_desc
     image =
         clCreateImageWithProperties(context, NULL, CL_MEM_READ_WRITE,
--- a/test_conformance/vulkan/test_vulkan_api_consistency_for_3dimages.cpp
+++ b/test_conformance/vulkan/test_vulkan_api_consistency_for_3dimages.cpp
@@ -162,16 +162,6 @@ int test_consistency_external_for_3dimage(cl_device_id deviceID,
     test_error(errNum, "Unable to create Image with Properties");
     image.reset();
 
-    // Passing properties, image_desc and image_format all as NULL
-    image = clCreateImageWithProperties(context, NULL, CL_MEM_READ_WRITE, NULL,
-                                        NULL, NULL, &errNum);
-    test_failure_error(
-        errNum, CL_INVALID_IMAGE_DESCRIPTOR,
-        "Image creation must fail with CL_INVALID_IMAGE_DESCRIPTOR "
-        "when all are passed as NULL");
-
-    image.reset();
-
     // Passing NULL properties and a valid image_format and image_desc
     image =
         clCreateImageWithProperties(context, NULL, CL_MEM_READ_WRITE,


test_conformance/vulkan/test_vulkan_api_consistency.cpp
```diff
@@ -323,27 +323,6 @@ int test_consistency_external_image(cl_device_id deviceID, cl_context _context,
     test_error(errNum, "Unable to create Image with Properties");
     image.reset();
 
-    // Passing image_format as NULL
-    image = clCreateImageWithProperties(context, extMemProperties.data(),
```
nikhiljnv, line 327: This is not related to semaphore negative tests, but for image import negative testing and should be retained, right?

test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
```diff
@@ -158,47 +158,6 @@ int test_consistency_external_for_1dimage(cl_device_id deviceID,
     test_error(errNum, "Unable to create Image with Properties");
     image.reset();
 
-    // Passing properties, image_desc and image_format all as NULL
-    image = clCreateImageWithProperties(context, NULL, CL_MEM_READ_WRITE, NULL,
```
nikhiljnv, line 162: I think this should be retained for the same reason I mentioned above.

test_conformance/vulkan/test_vulkan_api_consistency_for_3dimages.cpp
```diff
@@ -162,47 +162,6 @@ int test_consistency_external_for_3dimage(cl_device_id deviceID,
     test_error(errNum, "Unable to create Image with Properties");
     image.reset();
 
-    // Passing properties, image_desc and image_format all as NULL
-    image = clCreateImageWithProperties(context, NULL, CL_MEM_READ_WRITE, NULL,
```
nikhiljnv, line 166: Same comment as above.

nikhiljnv (changes requested):

test_conformance/vulkan/test_vulkan_api_consistency_for_3dimages.cpp
joshqti: This test case is ambiguous.  both format and image desc are null, either case is an error, and the OpenCL spec does specify if CL_INVALID_IMAGE_FORMAT_DESCRIPTOR  or CL_INVALID_IMAGE_DESCRIPTOR is to be returned if both are  NULL.

This test case does not make sense for this extension, there is no information being passed that assosciates clCreateImageWithProperties and cl_khr_external_memory.  

If we believe this is important coverage, it would be a better fit in api or basic.

test_conformance/vulkan/test_vulkan_api_consistency.cpp
joshqti: These failure conditions are already described in the OpenCL base specification.  If coverage is missing, they should be added to one of the core tests, such as API or basic.  The exception is if cl_khr_external_memory has a special provision for image_format or image_desc that I am forgetting.

test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
joshqti: This test case does not make sense for this extension, there is no information being passed that assosciates clCreateImageWithProperties and cl_khr_external_memory. 

If we believe this is important coverage, it would be a better fit in api or basic.

bashbaug:
Discussed in the October 15th memory subgroup.  Waiting for @nikhiljnv to respond.

test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
nikhiljnv: I prefer we retain the coverage here until we add it to more appropriate place rather than removing it right away.
We can either add the same to more appropriate tests or move these tests to more appropriate place in a subsequent PR.

nikhiljnv:
@joshqti 
Can we resolve the conflicts on TOT?

test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
joshqti: Sure, we can leave in the two valid negative tests.  I made an issue to cover these https://github.com/KhronosGroup/OpenCL-CTS/issues/2139

nikhiljnv:
Merging as discussed in memory subgroup call on November 5th 2024.

Issue #1982, opened 2024-06-18
Eliminate usage of half type in the CTS test


lakshmih:
Remove usage of half types in favor of float types and write_imagef in the kernels. Refer discussion on  CTS PR #1974

kamil-goras-mobica:
@bashbaug PR created: https://github.com/KhronosGroup/OpenCL-CTS/pull/2079

bashbaug:
PR #2147 fixed some of these tests, but there are still a few others that appear to have the same pattern, see https://github.com/KhronosGroup/OpenCL-CTS/pull/2147#pullrequestreview-2434825373.

Keeping this issue open until we fix the other tests, or until we confirm that they do not have a similar problem.

PR #2079, opened 2024-09-11
Removed usage of half types in CTS test

kamil-goras-mobica:
#1982

### DIFF
--- a/test_conformance/gl/common.h
+++ b/test_conformance/gl/common.h
@@ -51,7 +51,6 @@ static const format common_formats[] = {
   { GL_RGBA16UI_EXT, GL_RGBA_INTEGER_EXT, GL_UNSIGNED_SHORT,               kUShort },
   { GL_RGBA32UI_EXT, GL_RGBA_INTEGER_EXT, GL_UNSIGNED_INT,                 kUInt },
   { GL_RGBA32F_ARB,  GL_RGBA,             GL_FLOAT,                        kFloat },
-  { GL_RGBA16F_ARB,  GL_RGBA,             GL_HALF_FLOAT,                   kHalf }
 };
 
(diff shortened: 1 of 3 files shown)


bashbaug (changes requested):
This isn't the correct fix - sorry for poorly explaining what needs to be done!

We want to keep the testing for `CL_HALF_FLOAT`.  In other words, we shouldn't lose any test covereage after making these changes.

We want to remove the dependency on `cl_khr_fp16` for these tests though, so the tests that currently require `cl_khr_fp16` (which are used to test `CL_HALF_FLOAT`) need to be modified.

See the discussion here: https://github.com/KhronosGroup/OpenCL-CTS/pull/1974

One possible solution would be to eliminate all usage pointers to the `half4` type by switching to pointers to the `half` type instead.  This should allow removal of `cl_khr_fp16`, because pointers to the `half` type used by the `vload_half` and `vstore_half` functions is in core OpenCL.

A better (but more involved) solution is to eliminate all usage of `half4`, `vload_half`, and `vstore_half`, period, and switch to float reference data instead.

Let's chat more tomorrow if you have questions - thanks!

bashbaug:
I'm going to remove the "focused review" label until updates have been made.

kamil-goras-mobica:
@bashbaug I created corrections in separate PR: https://github.com/KhronosGroup/OpenCL-CTS/pull/2147

bashbaug:
I think this can be closed now, especially since #2147 is merged?

kamil-goras-mobica:
@bashbaug Yes, this can be closed. Should I do it?

bashbaug:
I think I confused myself on Tuesday's teleconference, sorry about that!

This PR can (and should) be closed.  I'll take care of that.  Thanks!

PR #2082, opened 2024-09-12
Fix bug in `mutable_command_full_dispatch`

gorazd-sumkovski-arm:
This test did not pass the `-cl-std=` flag when building the program. As a result, for an OpenCL 3.0 device, the program will be "compiled using the highest OpenCL C 1.x language version supported" by the device. However this will force uniform work-group sizes which leads to a `CL_INVALID_WORK_GROUP_SIZE` error.

To fix this, use the `create_single_kernel_helper()` helper function which will automatically get the device version and pass that to `-cl-std=` when building the program.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_full_dispatch.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_full_dispatch.cpp
@@ -296,6 +296,11 @@ struct MutableCommandFullDispatch : InfoMutableCommandBufferTest
         };
 
         size_t work_offset = 0;
+        /* Round the global work size up to nearest multiple of the local work
+         * size to ensure work group uniformity. */
+        num_elements =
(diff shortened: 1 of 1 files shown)


bashbaug:
I'm OK with this change and using the helper functions is good in general, but should we also update this test so it does not require non-uniform work-groups?  I don't see any place where we check for non-uniform work-group support, so even when we compile for OpenCL C 3.0 it's possible it won't work if it runs on a device that does not support non-uniform work-groups.

FWIW also, when I run this test on our devices using the command buffer emulation layer I don't see use of non-uniform work-groups.

EwanC (dismissed):
Change itself LGTM.

Agree with Ben's point that the test is assuming uniform workgroups, and we should either update the kernel to remove that assumption or check the device. Happy for this improvement to be filed as an issue if not wanting to be done in this PR.

gorazd-sumkovski-arm:
I've made it so uniform work groups are always used. If there is a compelling reason to test with non-uniform work groups on devices that support it, I will adjust it again.

EwanC (approved):

bashbaug:

Note, the CLA has already been signed as the submitter has committed to this repo previously.

PR #2083, opened 2024-09-13
Extended printf test to verify length sub-specifiers

shajder:
According to work plan for issue #1058


test_conformance/printf/util_printf.cpp
```diff
@@ -1397,6 +1397,104 @@ testCase testCaseMixedFormat = { TYPE_MIXED_FORMAT_RANDOM,
...
+
+//=============================================================
+
+// length sub-specifier format
+
+//=============================================================
+
+std::vector<printDataGenParameters> printLenSpecGenParameters = {
+
+    { { "%hd" }, "32767" },
```
svenvh, line 1409: Do we want to test `%hd`, `%hx` and `%ho` with a negative value too?

shajder: Good point - it makes sense, my only concern is if this should be added here along with sub-specifiers test or in actual `TYPE_INT`/`TYPE_OCTAL`/`TYPE_HEXADEC`. For some reason neither of those tests take into account negative values.

bashbaug: I'd prefer to test positive and negative values here so all of the length modifier testing is in one place.

We probably should test negative numbers for the `int`, `hexadecimal`, and `octal` tests also, but we can do that in a separate PR.

bashbaug:
Note, has a merge conflict that needs to be resolved.

test_conformance/printf/util_printf.cpp
shajder: Done! Soon I will prepare complementation PR for missing negative values in previous test cases

shajder:
> Note, has a merge conflict that needs to be resolved.

Conflicts solved

lakshmih (approved):

svenvh (approved):

bashbaug (approved):

bashbaug:

PR #2084, opened 2024-09-13
[NFC] relationals: fix -Wformat-security warnings

svenvh:
Fix a "format string is not a string literal (potentially insecure)" warning.

There is no security issue here as the format string argument is constructed using a stringstream right before.  Fix this occurrence anyway to allow enabling the warning globally.


kpet (approved):

kpet:
Trivial change, merging.

PR #2088, opened 2024-09-25
Enable building OpenCL-CTS with Intel DPC++/C++ compiler ICX on Windows

qichaogu:
Currently Intel® C++ Compiler Classic (ICC) is supported to build
OpenCL-CTS on Windows. This compiler has been discontinued since the
second half of 2023. Instead, Intel recommends that users transition to
use the LLVM-based Intel® oneAPI DPC++/C++ Compiler (ICX).
This change is to enable users to build OpenCL-CTS with ICX on Windows.


test_conformance/events/test_userevents.cpp
```diff
@@ -29,6 +29,18 @@
 // CL error checking.
 
 #if defined(_MSC_VER)
+#if defined(__INTEL_LLVM_COMPILER)
```
bashbaug, line 32: There is a single `CL_EXIT_ERROR` macro in "test_bufferreadwriterect.cpp" that doesn't require any ifdefs.  Can we use it here instead?  Even better, could we move it to a shared location and use it in both files?

test_common/harness/msvc9.c
```diff
@@ -714,6 +714,7 @@ int32_t float2int(float fx)
     return u.i;
 }
 
+#if !__has_builtin(__builtin_clz)
```
bashbaug, line 717: Two small questions here:

1. Do we need any specific compiler checks for `__has_builtin`?
2. Would this be more readable with the two conditions on the same line:
`#if !defined(_WIN64) && !__has_builtin(__builtin_clz)`

test_conformance/events/test_userevents.cpp
qichaogu: Thanks for pointing this out.
Yes, it's not necessary to define multiple `CL_EXIT_ERROR` and use ifdefs here. I have removed them.
I don't move `CL_EXIT_ERROR` to a shared location. `CL_EXIT_ERROR` in this file returns -1 for a non-CL_SUCCESS code, while `CL_EXIT_ERROR` in "test_bufferreadwriterect.cpp" does not do the same.

test_common/harness/msvc9.c
qichaogu: 1. Yes, we need checks for `__has_builtin`. Updated.
2. I don't combine the 2 conditions together as `#if !defined(_WIN64) && !__has_builtin(__builtin_clz)` because that will require the `#else` (corresponding to `#if !defined(_WIN64)`) also need a condition `#if !__has_builtin(__builtin_clz)`.

bashbaug (commented):
LGTM, ideally this should have a review from another company before merging.

cdai2:
> LGTM, ideally this should have a review from another company before merging.

Who should be invited to review it？

qichaogu:
> > LGTM, ideally this should have a review from another company before merging.
> 
> Who should be invited to review it？

@bashbaug Could you please invite a reviewer from another company?

bashbaug:
@svenvh or @lakshmih would you mind taking a look at this, especially the non-CMake and non-ifdef'd parts?  Thanks!

qichaogu:
Hi @bashbaug Could you please help merge the pull request? Thanks.

CMakeLists.txt
```diff
@@ -153,8 +153,13 @@ if(MSVC)
...
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Qlong-double -Qpc80 /DWIN32 /D_WINDOWS /W3 /GR /EHsc -nologo -Od -D_CRT_NONSTDC_NO_WARNINGS -EHsc -Wall -Qdiag-disable:68,111,177,186,161,869,1028,2259,2553,181,239,265,1188 -fp:strict -fp:source")
+    if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "IntelLLVM")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Qlong-double -Qpc80 /DWIN32 /D_WINDOWS /W3 /GR /EHsc -nologo -Od -D_CRT_NONSTDC_NO_WARNINGS -EHsc -Wall -fp:strict")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Qlong-double -Qpc80 /DWIN32 /D_WINDOWS /W3 /GR /EHsc -nologo -Od -D_CRT_NONSTDC_NO_WARNINGS -EHsc -Wall -fp:strict")
+    else()
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Qlong-double -Qpc80 /DWIN32 /D_WINDOWS /W3 /GR /EHsc -nologo -Od -D_CRT_NONSTDC_NO_WARNINGS -EHsc -Wall -Qdiag-disable:68,111,177,186,161,869,1028,2259,2553,181,239,265,1188 -fp:strict -fp:source")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Qlong-double -Qpc80 /DWIN32 /D_WINDOWS /W3 /GR /EHsc -nologo -Od -D_CRT_NONSTDC_NO_WARNINGS -EHsc -Wall -Qdiag-disable:68,111,177,186,161,869,1028,2259,2553,181,239,265,1188 -fp:strict -fp:source")
+    endif()
```
kpet, line 162: There are many shared flags here. Maybe the common flags could be added unconditionally and the others only when needed?

kpet (approved):
Fine with the change as is but we could deduplicate some of the flags in the CMake files.

CMakeLists.txt
qichaogu: Good point. Updated.

qichaogu:
@kpet @bashbaug Could you please review again and help merge if no objection?

kpet (approved):

bashbaug:

Issue #1235, opened 2021-05-05
Add testing for CL_KERNEL_LOCAL_MEM_SIZE


kpet:
See https://github.com/KhronosGroup/OpenCL-Docs/issues/612

kamil-goras-mobica:
@bashbaug PR: https://github.com/KhronosGroup/OpenCL-CTS/pull/2089

bashbaug:
Fixed by #2089.

PR #2089, opened 2024-09-25
Add testing for CL_KERNEL_LOCAL_MEM_SIZE #1235

kamil-goras-mobica:
#1235


test_conformance/api/test_api_consistency.cpp
```diff
@@ -1151,3 +1179,103 @@ int test_consistency_3d_image_writes(cl_device_id deviceID, cl_context context,
...
+    if (create_single_kernel_helper(context, &program, &kernel, 1,
+                                    local_param_kernel, "local_param_kernel")
+        != 0)
+    {
+        return -1;
+    }
+
+    size_t elements = 100;
+    size_t sizeToAllocate = elements * sizeof(cl_int);
+    int* localData = (cl_int*)malloc(sizeToAllocate);
```
EwanC, line 1236: Can we use a smart pointer here rather than malloc (e.g. `make_unique`) so that if the function returns early before the `free(localData)` we don't leak.

test_conformance/api/main.cpp
```diff
@@ -161,6 +161,7 @@ test_definition test_list[] = {
     ADD_TEST_VERSION(negative_create_command_queue_with_properties,
                      Version(2, 0)),
     ADD_TEST(negative_create_command_queue_with_properties_khr),
+    ADD_TEST(kernel_local_mem_size),
```
bashbaug, line 164: Nitpick: could we call this test "kernel_local_memory_size" instead, to be consistent with the existing "kernel_private_memory_size" test?

bashbaug (commented):
Note: it seems odd to put this test into `test_api_consistency.cpp`, since it has nothing to do with API consistency checks.  I'd recommend putting it into its own new file `test_kernel_local_memory_size.cpp`, similar to the existing `test_kernel_private_memory_size.cpp`.

test_conformance/api/test_api_consistency.cpp
bashbaug: `localData` is unused.  We may as well remove it entirely.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -1151,3 +1179,103 @@ int test_consistency_3d_image_writes(cl_device_id deviceID, cl_context context,
...
+    }
+
+    cl_ulong kernelLocalUsage = 0;
+    error = clGetKernelWorkGroupInfo(kernel, deviceID, CL_KERNEL_LOCAL_MEM_SIZE,
+                                     sizeof(kernelLocalUsage),
+                                     &kernelLocalUsage, NULL);
+    test_error(error,
+               "clGetKernelWorkGroupInfo for CL_KERNEL_LOCAL_MEM_SIZE failed");
+
+    test_assert_error(kernelLocalUsage > 0, "kernel local mem size failed");
```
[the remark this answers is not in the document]
kamil-goras-mobica: @bashbaug On my NVIDIA card kernelLocalUsage is equal to 1. Checked now using POCL and it is 0. So as this is empty kernel with no __local args then it should be 0. If I change this condition to kernelLocalUsage == 0 then test will be failing. What do you prefer: remove this test or move it to separated test which will be failing on my device, on other maybe not.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -25,6 +25,34 @@ __kernel void test(__global int* dst) {
...
+                               "{\n"
+                               "}\n" };
+
+const char* local_memory_kernel[] = {
+    "__kernel void local_memory_kernel(__local int* ptr)\n"
+    "{\n"
+    "__local float array[10000];\n"
+    "for(int i = 0; i<10000; i++)\n"
+    "    array[i]*=2;\n"
+    "}\n"
```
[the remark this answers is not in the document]
kamil-goras-mobica: @bashbaug My NVIDIA driver gave me wrong results than eg. POCL which which added optimizations (?) ant result is 0 memory as you said. I will add correction.

kamil-goras-mobica:
@bashbaug I have added some corrections, tested under Intel, Nvidia, POCL and SimpleOpenCLSamples Layers. There were discrepancies in results, I wanted to check all scenarios that get into my mind. Please review and tell me which does not make sense.

test_conformance/api/test_kernel_local_memory_size.cpp
```diff
@@ -0,0 +1,259 @@
...
+                                     &kernel_local_usage, NULL);
+    test_error(error,
+               "clGetKernelWorkGroupInfo for CL_KERNEL_LOCAL_MEM_SIZE failed");
+
+    const size_t size = 10;
+    const size_t memory = size * sizeof(cl_int);
+
+    size_t global_work_size[] = { size };
+
+    std::unique_ptr<int[]> data(new int[size]);
```
EwanC, line 136: can this be `int data[size]` and make size `constexpr`, I don't think this needs to be allocated on the heap

test_conformance/api/test_kernel_local_memory_size.cpp
```diff
@@ -0,0 +1,259 @@
...
+    error = clSetKernelArg(kernel, 2, sizeof(streams[1]), &streams[1]);
+    test_error(error, "Unable to set indexed kernel arguments");
+
+    error = clEnqueueNDRangeKernel(queue, kernel, 1, nullptr, global_work_size,
+                                   nullptr, 0, NULL, nullptr);
+    test_error(error, "clEnqueueNDRangeKernel failed.");
+
+    error = clEnqueueReadBuffer(queue, streams[1], CL_TRUE, 0, memory,
+                                data.get(), 0, NULL, NULL);
+    test_error(error, "clEnqueueReadBuffer failed");
```
EwanC, line 197: We're not validate the result that gets written back to `data`. I'm not sure we actually need to be executing kernels at all to be verifying the `CL_KERNEL_LOCAL_MEM_SIZE ` query. But if we are doing it for a more thorough test, then we may as well verify that `data` contains the values we expect.

kamil-goras-mobica: @EwanC added verification

test_conformance/api/test_kernel_local_memory_size.cpp
kamil-goras-mobica: @EwanC done

EwanC (commented):
Looking good but noticed that we don't verify the `param_value_size_ret` output parameter to `clGetKernelWorkGroupInfo`, but a conformant implementation should set that to `sizeof(cl_ulong)`. Can we check that too.

test_conformance/api/test_kernel_local_memory_size.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+    kernel_local_usage = 0;
+    param_value_size_ret = 0;
+    error = clGetKernelWorkGroupInfo(
+        kernel, deviceID, CL_KERNEL_LOCAL_MEM_SIZE, sizeof(kernel_local_usage),
+        &kernel_local_usage, &param_value_size_ret);
+    test_error(error,
+               "clGetKernelWorkGroupInfo for CL_KERNEL_LOCAL_MEM_SIZE failed");
+
+    test_assert_error(param_value_size_ret == sizeof(u_long),
```
EwanC, line 136: Prefer `cl_ulong` to `u_long`

kamil-goras-mobica: @EwanC changed

EwanC (dismissed):

bashbaug:
Discussed in the October 15th teleconference.  We will wait one more week for a few more reviews.

test_conformance/api/test_kernel_local_memory_size.cpp
```diff
@@ -0,0 +1,351 @@
...
+    __local int local_data[10];
+    local_ptr[id] = src[id];
+
+    barrier(CLK_LOCAL_MEM_FENCE);
+    if(id / 2 == 0) {
+        for(size_t i = 0; i < 10; i++)
+            local_data[id] += local_ptr[i];
+    }
+    else
+        local_data[id] = local_ptr[id] * 2;
```
bashbaug, line 77: This isn't safe, for two reasons:

1. In the "if" branch, `local_data` may be read when it is uninitialized, due to `+=`.
1. There is no guarantee which of these branches are executed first, and `local_data` will have different results depending which executes first.  Theoretically, even, both branches could execute simultaneously, in which case there is a data race.

Really though, this kernel looks way too complicated for its purpose.  Can we do something simpler?

test_conformance/api/test_kernel_local_memory_size.cpp
```diff
@@ -0,0 +1,351 @@
...
+    error = clGetKernelWorkGroupInfo(
+        kernel, deviceID, CL_KERNEL_LOCAL_MEM_SIZE, sizeof(kernel_local_usage),
+        &kernel_local_usage, &param_value_size_ret);
+    test_error(error,
+               "clGetKernelWorkGroupInfo for CL_KERNEL_LOCAL_MEM_SIZE failed");
+
+    test_assert_error(param_value_size_ret == sizeof(cl_ulong),
+                      "param_value_size_ret failed");
+
+    test_assert_error(kernel_local_usage >= 0, "kernel local mem size failed");
```
bashbaug, line 118: I'm not going to lose sleep over this, but I don't think these "empty kernel" tests are adding anything, and I'd be fine removing them.

* The check for `param_value_size_ret` is covered by all of the other scenarios.
* Because `kernel_local_usage` is a `cl_ulong` the check for `kernel_local_usage >= 0` will always be true.

test_conformance/api/test_kernel_local_memory_size.cpp
```diff
@@ -0,0 +1,351 @@
...
+        {
+            if (data[i] != testData[i]) return false;
+        }
+        return true;
+    };
+    test_assert_error(local_memory_kernel_verify(),
+                      "local_memory_kernel data verificaion failed");
+
+    test_assert_error(kernel_local_usage >= memory,
+                      "kernel local mem size failed");
```
bashbaug, line 184: FWIW, this check is failing for me for POCL because the queried value is coming back as zero.  I'm not sure if this is a POCL issue or if POCL is being clever and optimizing away the local memory array - maybe @pjaaskel can check?

bashbaug (approved):

test_conformance/api/test_kernel_local_memory_size.cpp
```diff
@@ -0,0 +1,351 @@
...
+            testData[0] += testData[i];
+        }
+        for (size_t i = 0; i < size; i++)
+        {
+            if (data[i] != testData[i]) return false;
+        }
+        return true;
+    };
+    test_assert_error(local_memory_kernel_verify(),
+                      "local_memory_kernel data verificaion failed");
```
bashbaug, line 181:
```suggestion
    test_assert_error(local_memory_kernel_verify(),
                      "local_memory_kernel data verification failed");
```

bashbaug (changes requested):
Gah, my previous review should have been "request changes", sorry about that.

Here is one more nitpick - there are several misspellings of "verification".

lakshmih (dismissed):
Looks good in present form, suggested a modification.

test_conformance/api/test_kernel_local_memory_size.cpp
bashbaug: There are still two additional places with the same misspelling.

bashbaug (approved):
LGTM, I'm OK with this going in as-is.

The POCL failure still exists with the latest code, though this may be a legitimate POCL failure.

test_conformance/api/test_kernel_local_memory_size.cpp
pjaaskel: It shouldn't optimize this case away. I opened a [PoCL issue](https://github.com/pocl/pocl/issues/1629). We'll check what's going on.

test_conformance/api/test_kernel_local_memory_size.cpp
```diff
@@ -0,0 +1,319 @@
...
+        testData[9] += 666;
+
+        for (size_t i = 0; i < size; i++)
+        {
+            if (data[i] != testData[i]) return false;
+        }
+
+        return true;
+    };
+    test_assert_error(
```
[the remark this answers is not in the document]
kamil-goras-mobica: @lakshmih corrected

bashbaug (approved):

lakshmih (approved):

bashbaug:

PR #2090, opened 2024-09-25
kernel_image_methods: Set height and depth for 1D buffer to 1

lakshmih:
Setting to zero causes the size to be zero in subsequent calculation

### DIFF
--- a/test_conformance/images/kernel_image_methods/test_1D_buffer.cpp
+++ b/test_conformance/images/kernel_image_methods/test_1D_buffer.cpp
@@ -187,7 +187,8 @@ int test_get_image_info_1D_buffer(cl_device_id device, cl_context context,
 
     imageInfo.type = CL_MEM_OBJECT_IMAGE1D_BUFFER;
     imageInfo.format = format;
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
It's hard to say what the right fix is here.

FWIW, for 1D buffer images and `cl_image_desc`, the OpenCL spec says that the image height and depth are ignored, so it doesn't necessarily seem "wrong" that these values are set to zero.  A `cl_image_desc` is not the same as the `image_descriptor` used in these tests, though.

It looks like the image height is only used as part of the size calculation, which doesn't look at all right.  So, maybe the best thing to do is to fix the size calculation instead, in which case the image height will never be used?

Regardless, it looks like similar code exists in `test_1D.cpp`, so whatever fix we make here we should also make there.

bashbaug (changes requested):
I'd prefer to fix this more like #2106 instead.

bashbaug:
I think this can be closed now that #2106 is merged.

bashbaug:
I'm going to take off "focused review" at the very least, until we determine whether this PR is still required.

PR #2091, opened 2024-09-25
bruteforce: Fix retry logic for cases with subnormals

lakshmih:
Replace the occurrences of 0.0f == test' with 0.0f == HTF(test)'. The types of 0.0f and test are not the same, so the equality comparison will get undesired result when the test represents a -0.0h (i.e., test is 32768=0x8000). In this situation, 0.0f == test will be false, but 0.0f == HTF(test) will be true.

Revise each if-statement to match the OpenCL s7.5.3 Item 4, specifically modify to check that the result is subnormal instead of checking that it is zero. "If the result of 3. is a sub-normal before rounding, the result may be flushed to zero"

### DIFF
--- a/test_conformance/math_brute_force/ternary_half.cpp
+++ b/test_conformance/math_brute_force/ternary_half.cpp
@@ -300,10 +300,10 @@ int TestFunc_Half_Half_Half_Half(const Func *f, MTdata d, bool relaxedMode)
                     if (fail && ftz)
                     {
                         // retry per section 6.5.3.2  with flushing on
-                        if (0.0f == test
-                            && 0.0f
-                                == f->func.f_fma(HTF(hp0[j]), HTF(hp1[j]),
-                                                 HTF(hp2[j]), FLUSHED))
+                        float r = f->func.f_fma(HTF(hp0[j]), HTF(hp1[j]),
+                                                HTF(hp2[j]), FLUSHED);
+                        cl_half c = HFF(r);
+                        if (0.0f == HTF(test) && IsHalfSubnormal(c))
                         {
                             fail = 0;
                             err = 0.0f;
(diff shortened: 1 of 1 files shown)


test_conformance/math_brute_force/ternary_half.cpp
```diff
@@ -424,28 +426,28 @@ int TestFunc_Half_Half_Half_Half(const Func *f, MTdata d, bool relaxedMode)
...
+
+                                cl_half c5 = HFF(r5);
+                                cl_half c6 = HFF(r6);
+                                cl_half c7 = HFF(r7);
+                                cl_half c8 = HFF(r8);
+                                if (0.0f == HTF(test)
+                                    && (IsHalfSubnormal(c5)
+                                        || IsHalfSubnormal(c6)
+                                        || IsHalfSubnormal(c7)
+                                        || IsHalfSubnormal(c8)))
```
lakshmih, line 446: Pattern of fixes throughout are:

1) Replace the occurrences of `0.0f == test' with `0.0f == HTF(test)'. The types of 0.0f and test are not the same, so the equality comparison will get undesired result when the test represents a -0.0h (i.e., test is 32768=0x8000). In this situation, 0.0f == test will be false, but 0.0f == HTF(test) will be true.

2) Revise each if-statement to match the OpenCL s7.5.3 Item 4, specifically modify to check that the result is subnormal instead of checking that it is zero. "If the result of 3. is a sub-normal before rounding, the result may be flushed to zero"

svenvh: Thanks for this clarification; I've added it to the PR description.

bashbaug:
Discussed in the October 1st teleconference.  Will wait one more week for additional reviews.

bashbaug:

PR #2092, opened 2024-09-26
Fix integer overflow in 1D buffer image tests

qichaogu:


test_conformance/images/clCopyImage/test_copy_1D_buffer.cpp
```diff
@@ -182,7 +182,7 @@ int test_copy_image_set_1D_buffer(cl_device_id device, cl_context context,
             do
             {
                 imageInfo.width =
-                    (size_t)random_log_in_range(16, (int)maxWidth / 32, seed);
+                    (size_t)random_log_in_range(16, (int)(maxWidth / 32), seed);
```
qichaogu, line 185: `size_t` `maxWidth` is the value calculated by `clGetDeviceInfo` with `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE`. It's greater than the max integer on some platform. In such a case, `(int)maxWidth` (truncated from size_t to int) may be a negative integer. As a result, the function `random_log_in_range(int, int, MTdata)` will have a second parameter which is negative so it will yield a negative result integer. After converting this negative result integer to `size_t`, `imageInfo.width` will be a very large integer. Finally, the test will fail when trying to allocating memory of a very big size.
Using `(int)(maxWidth / 32)` instead changes the calculation precedence, which can greatly reduce the risk of integer overflow.

qichaogu:
@bashbaug Could you please have a review? Seems I cannot add reviewers.

bashbaug (approved):

qichaogu:
@bashbaug Thanks for your review. Could you please help merge? I don't have permission to merge pull requests.

bashbaug:
Merging as discussed in the October 15th teleconference.

PR #2037, opened 2024-08-06
Added printf test for long type cases

shajder:
according to work plan for issue https://github.com/KhronosGroup/OpenCL-CTS/issues/1058

PR #2094, opened 2024-09-27
Correction for printing long type variables in 32-bit system

shajder:
Due to change request from #2037

### DIFF
--- a/test_conformance/printf/test_printf.h
+++ b/test_conformance/printf/test_printf.h
@@ -81,7 +81,6 @@ struct printDataGenParameters
 
 // Reference results - filled out at run-time
 static std::vector<std::string> correctBufferInt;
-static std::vector<std::string> correctBufferLong;
 static std::vector<std::string> correctBufferHalf;
 static std::vector<std::string> correctBufferFloat;
 static std::vector<std::string> correctBufferDouble;
--- a/test_conformance/printf/util_printf.cpp
+++ b/test_conformance/printf/util_printf.cpp
@@ -128,33 +128,33 @@ std::vector<printDataGenParameters> printLongGenParameters = {
 
     //(Minimum) fifteen-wide,default(right)-justified
 
-    { { "%5ld" }, "10000000000L" },
+    { { "%5lld" }, "10000000000L" },
 
     //(Minimum) fifteen-wide,left-justified
 
-    { { "%-15ld" }, "-10000000000L" },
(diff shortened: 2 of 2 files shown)


shajder:
hello @hvdijk , could you confirm PR fixes the problem you reported (I don't have 32-bit OS, just went through godbolt but still appreciate your confirmation, thanks)

hvdijk:
Hi, thanks for the quick action! This fixes the host side of things, but this also causes `%lld` to be used on the device side, so it does not quite work yet. I see:
```
long...
0)testing printf("%5lld",10000000000L)
warning: <unknown>:0:0: the 'll' length modifier is invalid

note: <unknown>:0:0: in format string "%5lld\0A\00"

No data read from analysis buffer
verifyOutputBuffer failed with kernel:
__kernel void test1(void)
{
   printf("%5lld\n",10000000000L);
}

 expected: 10000000000
 got:
```
I think the format strings for host and device will need to be different.

shajder:
> Hi, thanks for the quick action! This fixes the host side of things, but this also causes `%lld` to be used on the device side, so it does not quite work yet. I see:
> 
> ```
> long...
> 0)testing printf("%5lld",10000000000L)
> warning: <unknown>:0:0: the 'll' length modifier is invalid
> 
> note: <unknown>:0:0: in format string "%5lld\0A\00"
> 
> No data read from analysis buffer
> verifyOutputBuffer failed with kernel:
> __kernel void test1(void)
> {
>    printf("%5lld\n",10000000000L);
> }
> 
>  expected: 10000000000
>  got:
> ```
> 
> I think the format strings for host and device will need to be different.

ouch, it's kind of a mess. I presume the least intrusive way to solve this problem would be to use a table with reference values, please let me know if this solves the problem, thanks.

hvdijk:
That works!
```
long...
0)testing printf("%5ld",10000000000L)
1)testing printf("%-15ld",-10000000000L)
2)testing printf("%015ld",10000000000L)
3)testing printf("%+15ld",-10000000000L)
4)testing printf("%-+15ld",10000000000L)
5)testing printf("%.15li",10000000000L)
6)testing printf("%-+16.15li",-10000000000L)
long passed
```

svenvh (approved):
> ouch, it's kind of a mess. I presume the least intrusive way to solve this problem would be to use a table with reference values

That seems like the more robust solution indeed. LGTM.

bashbaug:

PR #2096, opened 2024-09-30
Add new extensions to conformance csv

EwanC:
The extension tests in `test_conformance/extensions` have not matching entries in the `opencl_conformance_tests_full.csv` file. This means that they are missed from conformance runs by vendors who use this file.

I've commented out `cl_khr_dx9_media_sharing` and `cl_khr_external_semaphore` as creation of the binaries is conditional on the build environment. I could create an issue to track enabling these, e.g. creating a stub binary which passes when the build conditions aren't met.


test_conformance/opencl_conformance_tests_full.csv
svenvh, line 1: This file is still fairly in sync with the other `opencl_conformance_tests_full_*.csv` files.  Should we propagate your additions there too?

And would any of the new tests qualify for being listed in `opencl_conformance_tests_quick.csv` too (which doesn't seem too different from the _full.csv files)?

EwanC: Thanks for the feedback @svenvh, I've added these tests to the other CSV files. These tests are quick to run, they are not compiling and running a large number of kernels across axis of combinations like bruteforce and conversions, so I've included them in the fast running CSV files too.

EwanC:
> No strong preference, but a separate file seems to be in line with what we do for D3D currently.

Thanks for the feedback, I asked on the 01/10/24 OpenCL teleconference and there was no objections to creating a new CSV. So i've updated this PR to add a new `opencl_conformance_tests_vulkan.csv` file for that.

svenvh (approved):

bashbaug:

PR #2099, opened 2024-10-01
Added test for work-item functions with out-of-range arguments

### LINKED ISSUE #2005: Missing tests for work-item functions with out-of-range arguments
The OpenCL C [work-item functions](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#work-item-functions) have defined results for all values of their _dimidx_ argument, even those that are out-of-range.  For example, for **get_global_size**:

> Valid values of dimindx are 0 to get_work_dim() - 1. For other values of dimindx, get_global_size() returns 1.

This behavior is not tested by the CTS, though, and all tested values are in-bounds.

We should test:

* All relevant work-item functions: **get_global_size**, **get_global_id**, **get_local_size**, **get_enqueued_local_size**, **get_local_id**, **get_num_groups**, **get_group_id**, **get_global_offset**.
* Cases when the out-of-range value is known at compile time, and when the out-of-range value is dynamic and is not known to be out-of-range until runtime.
* Cases with 1D, 2D, and 3D ND-ranges, since this will change when a _dimidx_ is out-of-range.

Mesa has some piglit tests we could use for inspiration, see for example:
https://gitlab.freedesktop.org/mesa/piglit/-/blob/main/tests/cl/program/execute/global-offset.cl

See also: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/2638

shajder:
Fixes #2005 according to task description.


test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,552 @@ struct work_item_data
...
+    cl_int Run() override
+    {
+        cl_int error = SetUp(kernel_source);
+        test_error(error, "SetUp failed");
+
+        size_t localThreads[3] = { 0, 0, 0 };
+
+        for (size_t dim = 1; dim <= 3; dim++)
+        {
+            cl_uint local_workgroup_size[3] = { maxWorkItemSizes[0],
```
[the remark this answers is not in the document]
shajder: Corrected

test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,552 @@ struct work_item_data
...
+
+    size_t threads[3] = { 0, 0, 0 };
+
+    TestWorkItemFnsOutOfRange(cl_device_id deviceID, cl_context context,
+                              cl_command_queue queue)
+        : TestWorkItemBase(deviceID, context, queue, 10240,
+                           outOfRangeWorkItemKernelCode)
+    {}
+
+    virtual cl_int SetUp(const char *kernel_source) override
```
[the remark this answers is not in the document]
shajder: Done

svenvh (approved):

test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,552 @@ struct work_item_data
...
+        ind += (uint)get_global_id(i) * ind_mul;
+        ind_mul *= get_global_size(i);
+    }
+    outData[ind].workDim = (uint)get_work_dim();
+%s
+})";
+
+const char *outOfRangeWorkItemKernelCodeExt =
+    R"(
+    uint dimindx=get_work_dim()+1;
```
bashbaug, line 102: A smart compiler would still be able to prove at compile time that this `dimidx` is out-of-range.

If you want to be sure the compiler cannot prove that the `dimidx` is out-of-range, it'd be better to pass it as a kernel argument.

test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,552 @@ struct work_item_data
...
+    outData[ind].globalSize[0] = (uint)get_global_size(dimindx);
+    outData[ind].globalID[0] = (uint)get_global_id(dimindx);
+    outData[ind].localSize[0] = (uint)get_local_size(dimindx);
+    outData[ind].localID[0] = (uint)get_local_id(dimindx);
+    outData[ind].numGroups[0] = (uint)get_num_groups(dimindx);
+    outData[ind].groupID[0] = (uint)get_group_id(dimindx);
+)";
 
-int test_work_item_functions(cl_device_id deviceID, cl_context context, cl_command_queue queue, int num_elements)
+const char *outOfRangeWorkItemKernelCodeExt11 =
```
bashbaug, line 111: Could we use a single kernel for this stuff, and guard the parts that are specific to newer OpenCL C versions using `__OPENCL_VERSION__` or some other preprocessor mechanism?  I think that would be easier to understand and maintain vs. stitching together these sample kernels with `sprintf`.

test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,552 @@ struct work_item_data
...
 {
-    int error;
+    TestWorkItemBase(cl_device_id did, cl_context c, cl_command_queue q,
+                     cl_uint data_size, const char *src)
+        : device(did), context(c), queue(q), program(nullptr), kernel(nullptr),
+          outData(nullptr), d_holder(gRandomSeed), testData(data_size),
+          kernel_source(src), max_workgroup_size(0)
+    {}
+
+    virtual cl_int SetUp(const char *kernel_source)
```
bashbaug, line 180: Something doesn't seem right here - we have a `kernel_source` member variables and a `kernel_source` function argument.  Do we need both?

test_conformance/basic/test_work_item_functions.cpp
shajder: Argument passed

test_conformance/basic/test_work_item_functions.cpp
shajder: Done, param renamed

bashbaug: These tests are very difficult to understand and maintain currently, so I think we need to rethink how these tests are structured.  The parameter name was just a symptom.  Here are a few concrete suggestions:

1. I strongly suggest either leaving the existing `work_item_functions` test alone, or biting the bullet and refactoring it completely.  Currently, it appears to be split between keeping the same high-level test logic and moving to a test "fixture", which is then preventing the new tests from being implemented cleanly.
2. Having `hardcoded` as a template parameter is confusing.  It is used in exactly one place, to setup the right kernel string for the test.  Could we just pass in the right kernel string instead, from the start?  If not, could we pass it to the constructor to setup the kernel string properly and then be done with it?
3. Having a `kernel_source` member variable and a `src` argument passed to the `Setup` function is confusing, and I still don't know why we have both.  To make matters worse, in some cases we just pass `kernel_source` to `Setup`, and in some cases we don't.
4. Why is there both a constructor and a `Setup` function?  Could we move some of the functionality out of `Setup` into the constructor?  If not, why not?
5. The `work_item_functions` test does not have a `Validate` function, but the `work_item_functions_out_of_range` tests do.  Why is this?

test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,500 @@ struct work_item_data
...
+    }
+
+    cl_int Run() override
+    {
+        cl_int error = SetUp(kernel_source);
+        test_error(error, "SetUp failed");
+
+        cl_int dim_param = 4;
+        error = clSetKernelArg(kernel, 1, sizeof(cl_int), &dim_param);
+        test_error(error, "Unable to set kernel arg");
```
bashbaug, line 452: I think this means that these tests are testing for `dimidx` equal to four exclusively, is that correct?

If so, this is better than nothing, but it means we aren't really checking for proper out-of-range behavior at `dimidx` equal to two for 1D or 2D ranges, or for `dimidx` equal to one for 1D ranges.

bashbaug (commented):
I find these test quite difficult to follow still and I'd like to see some improvements.

I can't help but think we're making this all a little too complicated, too.  What if we did something like this, instead:

1. Always submit a relatively small ND-range, maybe four work-groups consisting of a four of work-items, enough to make one of the indices greater than one to differentiate between default values of zero and one.
    * For example, submit a global work size of 1 x 16 and a local work size of 1 x 4 to test a 2D range.
2. For each work-item function, write out four values.
    * In the "hardcoded" case, write out the result of the function with `dimidx` ranging from `0` to `3`, so at least one value will always be out-of-range.
    * In the non-hardcoded case, write out the result of the function with `dimidx` ranging from `N + 0` to `N + 3`, where `N` is a kernel argument set to zero.
3. Do we even need to worry about the in-range arguments, or can we only test the out-of-range arguments?  The in-range arguments should already be tested by the `work_item_functions` test, correct?

test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,500 @@ struct work_item_data
...
+                    log_error("ERROR: get_global_offset() did not return "
+                              "proper value "
+                              "for %d dimensions (expected %d, got %d)\n",
+                              (int)dim, (int)dim, (int)testData[q].workDim);
+                    return -1;
+                }
+            }
+        }
+
+        return CL_SUCCESS;
```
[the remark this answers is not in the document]
shajder: Corrected as you requested. Previously, for consistency it followed the same schema from existing test.

test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,500 @@ struct work_item_data
...
+                          "for %d dimensions (expected %d, got %d)\n",
+                          (int)dim, (int)dim, (int)testData[q].workDim);
+                return -1;
+            }
+            if (testData[q].globalSize[0] != 1)
+            {
+                log_error("ERROR: get_global_size(%d) did not return "
+                          "proper value for the argument out of range "
+                          "(expected 1, got %d)\n",
+                          (int)dim + 1, (int)testData[q].globalSize[0]);
```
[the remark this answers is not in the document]
shajder: Corrected, indeed that printout could be confusing.

test_conformance/basic/test_work_item_functions.cpp
```diff
@@ -35,146 +39,500 @@ struct work_item_data
...
+{
+    int ind_mul=1;
+    int ind=0;
+    for( uint i = 0; i < get_work_dim(); i++ )
+    {
+        ind += (uint)get_global_id(i) * ind_mul;
+        ind_mul *= get_global_size(i);
+    }
+    outData[ind].workDim = (uint)get_work_dim();
+%s
```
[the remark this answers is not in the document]
shajder: Corrected. As mentioned previously, the idea was to reuse existing structure but now it was replaced with more suitable.

test_conformance/basic/test_work_item_functions.cpp
shajder: Moved to the loop and set to number dimensions + 1. If you think this is still not acceptable I can follow your idea with new approach for the test - number of change requests is already large so if you don't mind I will hold at this point waiting for more feedback.

PR #2102, opened 2024-10-03
Fix printing in test_api_min_max

jwarchul:
before: Filling constant buffer with 553805440 cl_ints (-2079745536 bytes).
after: Filling constant buffer with 553805440 cl_ints (2215221760 bytes).


svenvh (approved):

svenvh:
Non-controversial fix, so merging immediately.

Issue #1010, opened 2020-10-12
test_printf intermittently fails on Windows10


nikhiljnv:
test_printf intermittently fails on Windows10.
We have verified that device side printf is working correctly and driver is handling the same as expected.
The problem is narrowed down to the way test handles redirection of printf on host. The other printf tests which directly print on command-line or to a file are working fine.
This looks like a test issue where we need to fix the re-direction of printf.

PR #2103, opened 2024-10-04
printf: add fflush(stdout) at the end

lakshmih:
Some platforms (Windows in our case) seem to need this.

### DIFF
--- a/test_common/harness/testHarness.cpp
+++ b/test_common/harness/testHarness.cpp
@@ -689,6 +689,7 @@ static void print_results(int failed, int count, const char *name)
             log_error("FAILED %s.\n", name);
         }
     }
+    fflush(stdout);
 }
 
 int parseAndCallCommandLineTests(int argc, const char *argv[],


bashbaug:
Can you say a bit more why this is required?  What are the symptoms before this change, and how does the change fix them?

I know the printf tests do some odd things with stdout, but the tests themselves (specifically, `doTest`) also have calls to `fflush(stdout)`.  Are we missing a call somewhere in `doTest`?  Or, is there some reason why we still need this additional call in the test `main` function for some reason?

nikhiljnv:
@lakshmih 
Are you seeing behavior similar to https://github.com/KhronosGroup/OpenCL-CTS/issues/1010 that you are trying to fix with this change?

lakshmih:
@nikhiljnv  Not sure it is the same issue but essentially we don't get output from the print_results function in testHarness.cpp that prints the cumulative pass/fail result without the fflush. 
@bashbaug The individual doTest messages do get flushed but it is the final pass / fail from the harness (see above) that is missing.

svenvh:
> we don't get output from the print_results function in testHarness.cpp that prints the cumulative pass/fail result without the fflush.

Perhaps it makes more sense to move the `fflush` into the harness's `print_results` function then?

svenvh (approved):

bashbaug (approved):

bashbaug:

PR #2104, opened 2024-10-08
Some events are not released in test_mem_host_flags

aharon-abramson:
In test_mem_host_flags, some events are not released. In fact, they are not used, so they can be removed altogether.

### DIFF
--- a/test_conformance/mem_host_flags/checker_mem_host_no_access.hpp
+++ b/test_conformance/mem_host_flags/checker_mem_host_no_access.hpp
@@ -81,10 +81,9 @@ cl_int cBuffer_check_mem_host_no_access<T>::Setup_Test_Environment()
 template <class T>
 cl_int cBuffer_check_mem_host_no_access<T>::verify_RW_Buffer()
 {
-    cl_event event;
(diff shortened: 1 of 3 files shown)


bashbaug (approved):
This is an improvement, and I'm fine checking this in as-is, but it looks like there are still some unreleased events.  Using the OpenCL Intercept Layer and [LeakChecking](https://github.com/intel/opencl-intercept-layer/blob/main/docs/controls.md#leakchecking-bool), here is the report before this change:

```
No cl_context leaks detected.
No cl_command_queue leaks detected.
No cl_mem leaks detected.
Possible leak of type cl_event!
    Number of Allocations: 672
    Number of Retains:     0
    Number of Releases:    336
```

And after:

```
No cl_context leaks detected.
No cl_command_queue leaks detected.
No cl_mem leaks detected.
Possible leak of type cl_event!
    Number of Allocations: 432
    Number of Retains:     0
    Number of Releases:    336
```

Let me know if you 'd like help tracking down some of the remaining leaks.

bashbaug:
Merging as discussed in the October 22nd teleconference.  We will address the remaining leaks in a subsequent PR.

PR #2105, opened 2024-10-08
Fix for out of bounds access

tomasz-platek:
Rounding up the clusters_results size to avoid out of bounds access when subgroup_size is not a multiple of cluster_size.


svenvh (approved):

bashbaug:

PR #2106, opened 2024-10-08
kernel_image_methods: Fix size checks in 1Dbuffer case

dneto0:
Image height is initialized to 0.  But the size calculation should assume a height of 1.

### DIFF
--- a/test_conformance/images/kernel_image_methods/test_1D_buffer.cpp
+++ b/test_conformance/images/kernel_image_methods/test_1D_buffer.cpp
@@ -262,8 +262,7 @@ int test_get_image_info_1D_buffer(cl_device_id device, cl_context context,
                     imageInfo.rowPitch += extraWidth;
                 } while ((imageInfo.rowPitch % pixelSize) != 0);
 
-                size = (cl_ulong)imageInfo.rowPitch * (cl_ulong)imageInfo.height
-                    * 4;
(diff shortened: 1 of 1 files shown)


test_conformance/images/kernel_image_methods/test_1D_buffer.cpp
```diff
@@ -262,8 +262,7 @@ int test_get_image_info_1D_buffer(cl_device_id device, cl_context context,
                     imageInfo.rowPitch += extraWidth;
                 } while ((imageInfo.rowPitch % pixelSize) != 0);
 
-                size = (cl_ulong)imageInfo.rowPitch * (cl_ulong)imageInfo.height
```
dneto0, line 265: imageInfo.height is initialized to 0 earlier, so the loop exits right away.  The loop has not been doing anything useful.  Change the calculation to use an assumed height of 1.

bashbaug:
I think https://github.com/KhronosGroup/OpenCL-CTS/pull/2090 tries to fix this too, but this looks like a better fix IMHO.

@lakshmih does this fix work for you?

@dneto0 could you please fix the size calculation in test_1D.cpp also?  I think it has the same problem.
https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/images/kernel_image_methods/test_1D.cpp#L232

lakshmih (approved):

bashbaug (approved):

PR #2107, opened 2024-10-08
Fix warning now treated as error

ahesham-arm:
Following the re-enablement of narrowing warnings, this fixes a compilation error when running the `ubuntu 22.04 arm` GitHub action.

### DIFF
--- a/test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_wait_signal.cpp
+++ b/test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_wait_signal.cpp
@@ -88,7 +88,8 @@ template <RunMode mode> struct InvalidCommandQueue : public SemaphoreTestBase
         }
 
         cl_device_partition_property partitionProp[] = {
-            CL_DEVICE_PARTITION_EQUALLY, maxComputeUnits / 2, 0
+            CL_DEVICE_PARTITION_EQUALLY,
+            static_cast<cl_device_partition_property>(maxComputeUnits / 2), 0
         };
 
         cl_uint deviceCount = 0;


bashbaug (commented):
Non-controversial, merging immediately.

(Is this something we could/should have caught in CI?)

bashbaug (approved):

ahesham-arm:
> Non-controversial, merging immediately.
> 
> (Is this something we could/should have caught in CI?)

@bashbaug Thank you for the quick review and merge. I think the problem was that the pull request had passed all the CI checks when it was merged, the compilation error was a from a file that was added after that point. I don't think it would have been caught by the CI except if we had asked to rerun the checks before merging (not unreasonable but will not likely scale well with the volume of pull requests we have open).

ahesham-arm:
> This change breaks 32 bit builds (built on x86_64 linux with -m32):
> 
> `/extensions/cl_khr_semaphore/test_semaphores_negative_wait_signal.cpp:91:58: error: narrowing conversion of '(maxComputeUnits / 2)' from 'cl_uint' {aka 'unsigned int'} to 'cl_device_partition_property' {aka 'int'} [-Werror=narrowing] 91 | CL_DEVICE_PARTITION_EQUALLY, maxComputeUnits / 2, 0`
> 
> Do you need an issue created or do you want to just fix the new error?
> 
> As a side-note we have had other 32 bit fails for similar reasons - should we add 32 bit x86_64 build? (+test?)

Are you sure this change breaks your build? The error you pasted seems to reference the original code before this fix is applied.

coldav:
Ah yes, I think you are right, we saw a fail last night and i checked the latest commits. My apologies.

PR #1428, opened 2022-05-24
Added initial set of tests for the cl_khr_semaphore extension

marcat03:

PR #2110, opened 2024-10-11
Moved sub-test external_semaphores_import_export_fd to test_external_semaphore executable

### LINKED ISSUE #1876: Move test_semaphores_import_export_fd from cl_khr_semaphore to cl_khr_external_sync_fd related tests
Some cl_khr_external_sync_fd related tests are currently present in cl_khr_semaphore.
https://github.com/KhronosGroup/OpenCL-CTS/blob/f924273cdd607db39e8ff5e1490f5cef87242b7e/test_conformance/extensions/cl_khr_semaphore/test_semaphores.cpp#L651C80-L651C80

May be move these to a separate directory specific to cl_khr_external_sync_fd or cl_khr_external_semaphore at least.

### LINKED ISSUE #1989: Semaphore tests should not require clGetSemaphoreHandleForTypeKHR
The semaphore tests (for the `cl_khr_semaphore` extension) currently require an implementation of the function `clGetSemaphoreHandleForTypeKHR`, but this function is part of `cl_khr_external_semaphore` and is not part of `cl_khr_semaphore`.
Therefore, an implementation of this function should not be required for the `cl_khr_semaphore` tests.

shajder:
Fixes #1989 according to task description. Moreover, fixes #1876 as well.


test_conformance/extensions/cl_khr_external_semaphore/main.cpp
```diff
@@ -14,6 +14,7 @@
 //
 #include "procs.h"
 #include "harness/testHarness.h"
+#include "harness/testHarness.h"
```
bashbaug, line 17: We don't need to include the same file twice.

```suggestion
#include "harness/testHarness.h"
```

test_conformance/extensions/cl_khr_external_semaphore/main.cpp
```diff
@@ -26,6 +27,7 @@ test_definition test_list[] = {
     ADD_TEST(external_semaphores_cross_queues_io2),
     ADD_TEST(external_semaphores_multi_signal),
     ADD_TEST(external_semaphores_multi_wait),
+    ADD_TEST_VERSION(external_semaphores_import_export_fd, Version(1, 2)),
```
bashbaug, line 30: I know this line just moved over from the previous test, but it's inconsistent with the other external semaphore tests.

Options:
1. Add the OpenCL 1.2 check to the other external semaphore tests.  This is probably the right thing to do because the spec says that semaphores require OpenCL 1.2 (though I forget why).
2. Remove the OpenCL 1.2 version check from this test.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
```diff
@@ -51,9 +52,11 @@
         }                                                                      \
     } while (false)
 
-static const char* source = "__kernel void empty() {}";
+namespace {
```
bashbaug, line 55: This isn't a hill I'm going to die on, but I have a slight preference for "static" vs. the anonymous namespace, if for no other reason than to reduce diffs.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
```diff
@@ -0,0 +1,122 @@
...
+    GET_PFN(deviceID, clCreateSemaphoreWithPropertiesKHR);
+    GET_PFN(deviceID, clEnqueueSignalSemaphoresKHR);
+    GET_PFN(deviceID, clEnqueueWaitSemaphoresKHR);
+    GET_PFN(deviceID, clGetSemaphoreHandleForTypeKHR);
+    GET_PFN(deviceID, clReleaseSemaphoreKHR);
+
+    // Create ooo queue
+    clCommandQueueWrapper queue = clCreateCommandQueue(
+        context, deviceID, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &err);
+    test_error(err, "Could not create command queue");
```
bashbaug, line 55: I realize this came over from the previous test also, but do we ever check that out-of-order queues are supported?

Does this need to be an out-of-order queue, or would an in-order queue suffice, such as the one provided by the harness?

shajder: Please correct me if I am wrong but I can't spot any contraindications for `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE` to be used along with external semaphore. Spec only mentions "valid command-queue". On the other hand there are in the test source code two events signal/wait which indicates synchronization verification.

I wasn't able to find anything more from original PR #1428. I am leaving as is for now.

test_conformance/extensions/cl_khr_external_semaphore/main.cpp
shajder: Done

test_conformance/extensions/cl_khr_external_semaphore/main.cpp
shajder: Corrected

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
shajder: Corrected

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
bashbaug: It's not so much that out-of-order queues cannot work with external semaphores, it's more that this is an optional feature that may not be supported by all devices supporting external semaphores.  So, if we want to use out-of-order queues, at the very least we'll need to check for out-of-order queue support, and skip the test if out-of-order queues aren't supported.

Note, because the queue is out-of-order, and there are no event dependencies between the signal and the wait in the out-of-order queue, I think (in theory, at least) the wait could execute before the signal.  I haven't fully convinced myself, but it seems like this could be a problem.  If the queue were an in-order queue, we would instead guarantee that the signal occurs before the wait, which I believe is the intent.

shajder: Yep, I think I see your point. `wait_event` is not really waiting for anything so in theory it could execute before the signal. I just pushed verification for `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE` property support, let me know if I should replace out-of-order queue with regular one.

joshqti: I don't think out of order queues pose an issue, the spec says clEnqueueSignalSemaphore will not return until the semaphore is safe to wait.  

I think in-order queues would be better, despite not being a problem.  The test is testing for importing/exporting the semaphore, OOO queues are not required to test this, no reason to not run this test if an implementation only supports in order. 

The value of external semaphores is not queue synchronization, it is sync with other APIs, or perhaps other OpenCL contexts.

bashbaug: I still think it would be better to use an in-order queue instead, so the test can run on more devices, but all of the test logic is correct now so I'm fine with this going in as-is.

bashbaug (approved):

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
joshqti: > I still think it would be better to use an in-order queue instead, so the test can run on more devices, but all of the test logic is correct now so I'm fine with this going in as-is.

I agree with you.  Should we just make the change now, so no one has to do it later?

nikhiljnv:
As discussed in Memory subgroup call on December 17, 2024, requesting Qualcomm, ARM and/or Imagination to review and provide feedback.

nikhiljnv:
@joshqti Are you waiting for any of your comments to be addressed? If all good, can you approve?

nikhiljnv:
Merging as discussed in memory subgroup call on January 7th, 2025.

PR #2114, opened 2024-10-14
vulkan: Choose where Shaders are generated

joshqti:
- generate shader files in VULKAN_TEST_RESOURCES directory
- allows for outside CMakeLists to specify arbitrary destination


bashbaug (approved):
Merging as discussed in the October 15th memory subgroup.

Some additional changes may be needed to find the shader binaries in a different directory, but we will cross that bridge if we come to it.

PR #2063, opened 2024-08-29
Added missing test cases for CL_SEMAPHORE_DEVICE_HANDLE_LIST_KHR queries

shajder:
According to comments from issue #1875

PR #2117, opened 2024-10-15
fixes a narrowing warning (treated as an error) affecting some platforms

bashbaug:
Looks like https://github.com/KhronosGroup/OpenCL-CTS/pull/2063 has a "narrowing" warning that is now treated as an error and is hence causing CI builds to fail.

Applies the same fix as https://github.com/KhronosGroup/OpenCL-CTS/pull/2107 to fix this warning.


bashbaug:
CC @shajder @ahesham-arm @coldav - I can't add you as reviewers, but you'll probably want this fix.

svenvh (approved):

svenvh:
Non-controversial, merging immediately.

PR #2118, opened 2024-10-16
Fix bug in `mutable_command_full_dispatch`

xinjin01:
The test is missing clFinish after clEnqueueSVMUnmap. Add clFinish after clEnqueueSVMUnmap.


bashbaug (approved):
LGTM.  Without this PR there is no call to clFinish between the call to clEnqueueSVMUnmap and clSVMFree, and the description of clSVMFree [says](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clSVMFree.html):

> Note that clSVMFree does not wait for previously enqueued commands that may be using svm_pointer to finish before freeing svm_pointer. It is the responsibility of the application to make sure that enqueued commands that use svm_pointer have finished before freeing svm_pointer. This can be done by enqueuing a blocking operation such as clFinish, clWaitForEvents, clEnqueueReadBuffer or by registering a callback with the events associated with enqueued commands and when the last enqueued command has finished freeing svm_pointer.

For the Khronos folks, note that this is exactly the type of error that a "blocking free" aims to solve.

EwanC (approved):

PR #2121, opened 2024-10-18
Corrected build under msvc compiler #1833

### LINKED ISSUE #1833: isnan, isfinite, and isinf are implemented as macros for MSVC

kamil-goras-mobica:
Fixes #1833 according to task description.

### DIFF
--- a/test_common/harness/compat.h
+++ b/test_common/harness/compat.h
@@ -116,6 +116,15 @@ int feclearexcept(int excepts);
 #include <math.h>
 #endif
 
+#ifdef __cplusplus
+#include <cmath>
+extern "C" {
+using std::isfinite;
+using std::isinf;
+using std::isnan;
+}
+#endif
+
 #ifndef M_PI
 #define M_PI 3.14159265358979323846264338327950288
 #endif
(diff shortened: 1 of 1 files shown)


kamil-goras-mobica:
#1833

kamil-goras-mobica:
@bashbaug I've added correction now with using directive all calls isnan should be using std::isnan, also <cmath> is needed.

bashbaug (approved):
This works for me.  It's definitely a step in the right direction.

Ultimately I'd like to eliminate these "using" lines and just call the std:: versions directly, but we can do that in a follow-on PR.

bashbaug:
Merging as discussed in the November 12th teleconference.

PR #2123, opened 2024-10-21
Re-enable simple_semaphore_2

joshqti:
Re-enable a disabled external semaphore test.
Delete obsolete subtests that remain in comments.


ahesham-arm:
LGTM, cc: @kpet

PR #2125, opened 2024-10-24
remove using namespace std and use std namespace explicitly

bashbaug:
Removes `using namespace std` and adds `std::` explicitly instead, which is usually on calls to `min`.

This is generally best practice, and it also might be helpful when there are the same function names in the std namespace and in the global namespace (e.g. #1833).

### DIFF
--- a/test_common/harness/parseParameters.cpp
+++ b/test_common/harness/parseParameters.cpp
@@ -25,8 +25,6 @@
 #include <sys/stat.h>
 #include <string.h>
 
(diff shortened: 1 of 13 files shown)


svenvh (approved):

EwanC (approved):

bashbaug:
@lakshmih any concerns?  Maybe we can merge this offline?

bashbaug:

PR #2127, opened 2024-10-24
Fix payload value query for external semaphores

ahesham-arm:
When creating a CL semaphore object from a Vulkan semaphore one, we explicitly pass `-1` as the file descriptor value in the case of `VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SYNC_FD`. According to the CL specification:

>     The special value -1 for fd is treated like a valid sync file
>     descriptor referring to an object that has already signaled. The
>     import operation will succeed and the semaphore will have a
>     temporarily imported payload as if a valid file descriptor had
>     been provided.

The test currently checks that the semaphore payload is unsignalled, unconditionally, which is incorrect.

Changed the test to check for the correct expected payload value.

### DIFF
--- a/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
+++ b/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
@@ -197,10 +197,15 @@ int test_external_semaphores_queries(cl_device_id deviceID, cl_context context,
         test_error(err, "Could not release semaphore");
         SEMAPHORE_PARAM_TEST(CL_SEMAPHORE_REFERENCE_COUNT_KHR, cl_uint, 1);
 
-        // Confirm that querying CL_SEMAPHORE_PAYLOAD_KHR returns the unsignaled
+        // Confirm that querying CL_SEMAPHORE_PAYLOAD_KHR returns the correct
         // state
+        cl_semaphore_payload_khr expected_payload_value =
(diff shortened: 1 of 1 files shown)


bashbaug:
Would it be better to get a valid SYNC_FD file descriptor for this test case instead of using the special `-1` file descriptor value?

All of the other handle types seem to be getting a valid handle:

`handle = semaphore.getHandle(externalSemaphoreHandleType);`

ahesham-arm:
> Would it be better to get a valid SYNC_FD file descriptor for this test case instead of using the special `-1` file descriptor value?
> 
> All of the other handle types seem to be getting a valid handle:
> 
> `handle = semaphore.getHandle(externalSemaphoreHandleType);`

I agree that we need an additional test here, because we should be covering two cases:
* Creating a semaphore with a sync fd set to -1, and checking that the payload is signalled (which is what this PR fixes)
* Creating a semaphore with a valid sync fd value other than -1, and checking that the payload is unsignalled (which I can address) in a follow up PR.

joshqti:
This change looks good to me.  I agree with adding another test as a follow-up.

bashbaug (approved):
Works for me.  I'm fine adding the additional test in a follow-on PR.

joshqti (approved):
Approved as discussed

nikhiljnv:
Merging as discussed in memory subgroup call on November 5th 2024.

PR #2129, opened 2024-10-25
Fix CTS mutable_dispatch memory leak

xinjin01:
Current class wrapper of the CTS test framework allows getting the pointer of the private member object. This puts the object at risk of losing its original value if the pointer gets reassigned, causing a memory leak and potentially other problems.

This happens to the "clMemWrapper kernel" used by
mutable_command_work_groups tests, where the "clMemWrapper kernel" gets initialised by the default basic setup function and then it gets reassigned by the build_program_create_kernel_helper() helper  function through pointer.

This patch fixes this issue by updating mutable_command_work_groups tests: instead of calling basic setup function and then initialise the "clMemWrapper kernel" object again in the helper function, it now overrides the basic setup function to make sure the "clMemWrapper kernel" will be assigned only once.


EwanC (approved):
LGTM

bashbaug:

PR #2134, opened 2024-10-30
Added test to verify flush on clReleaseCommandQueue with multiple queues

### LINKED ISSUE #2087: Add a test to verify that `clReleaseCommandQueue` does trigger a flush in multi-queue scenario
See https://github.com/KhronosGroup/OpenCL-Docs/issues/1248 for background.

We should add a test that:

- Creates two command queues
- Enqueues work in command queue A
- Enqueues work in command queue B that depends on work in command queue A
- Calls `clReleaseCommandQueue` on command queue A
- Calls `clFinish` on command queue B
- Checks that all work enqueued to command queue A and command queue B has run correctly

shajder:
Fixes #2087 according to task description.


test_conformance/api/test_queue.cpp
```diff
@@ -59,3 +59,63 @@ int test_queue_flush_on_release(cl_device_id deviceID, cl_context context,
...
+    test_error(err, "clReleaseCommandQueue failed");
+
+    err = clFinish(queue_B);
+    test_error(err, "clFinish failed");
+
+    // Wait for kernel to execute since the queue must flush on release
+    bool success = poll_until(2000, 50, [&event_B]() {
+        cl_int status;
+        cl_int err = clGetEventInfo(event_B, CL_EVENT_COMMAND_EXECUTION_STATUS,
+                                    sizeof(cl_int), &status, nullptr);
```
bashbaug, line 112: Since we've called `clFinish` on `queue_B` here, the `event_B` must already be `CL_COMPLETE`.

bashbaug (commented):
@kpet can you check that the test description in #2087 is correct?

This test follows the test description exactly (thanks!), but I would like to ensure it is properly testing that the flush happsn on `clReleaseCommandQueue` and not somewhere else (e.g. the `clFinish`).

bashbaug:
OK, to be clear, I don't think this test is wrong with either the explicit call to `clFinish()` (or an explicit call to `clFlush()` instead), but I also think it could very likely be testing behavior other than the implicit flush on `clReleaseCommandQueue`.

I think that the proposed single-queue test would be more useful:

> * Enqueue work in a queue and get an event
> * Call `clReleaseCommandQueue` on that queue
> * Poll the execution status of the event, it should transition to `CL_COMPLETE` "soon".

If we really wanted to test the letter of the spec, we should also call `clRetainCommandQueue` on the queue first, so the call to `clReleaseCommandQueue` decrements the reference count but doesn't cause the reference count to reach zero.  This is the behavior we are considering changing - linking in https://github.com/KhronosGroup/OpenCL-Docs/issues/1248#issuecomment-2371816941.

bashbaug (approved):

bashbaug:
Merging as discussed in the January 28th teleconference.

PR #2135, opened 2024-10-30
Fix various Wformat warnings

svenvh:
test_copy_1D.cpp: num_mip_levels is a cl_uint, so don't cast and just print using `%u`.

test_pipe_info.cpp: arg_type_qualifier is a 64-bit wide bitfield, so print in hexadecimal format using the correct length modifier.

test_device_partition.cpp: it is not clear what the bit width of `cl_device_partition_property` should be, so cast the operands to align with the format specifiers.


bashbaug (approved):
Thank you for fixing these!

bashbaug:

PR #2138, opened 2024-10-31
Refactor test copy from and to image to buffer

MichaelRizkalla-arm:
This change refactors the following files:
1. test_imagearraycopy.cpp
2. test_arrayimagecopy.cpp

The purpose of refactoring is to allow passing different buffer and image flags to the test function as well as passing customisable test functions.


joshqti:
The code looks fine, and passes testing

bashbaug (approved):
I'm curious to see how the `test_function_t` function pointers will be used.

PR #2143, opened 2024-11-06
mutable dispatch related fixes

franz:


test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
```diff
@@ -416,6 +416,13 @@ struct CommandBufferCommandFillBufferMutableHandleNotNull
 {
     using CommandFillBaseTest::CommandFillBaseTest;
 
+    bool Skip() override
+    {
+        if (CommandFillBaseTest::Skip()) return true;
+        return is_extension_available(device,
+                                      "cl_khr_command_buffer_mutable_dispatch");
```
EwanC, line 423: The mutable-dispatch extension affects kernel commands, not other types of commands. The spec wording for [clCommandFillBufferKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCommandFillBufferKHR) and other non-kernel command appending entry-points is:

> * mutable_handle returns a handle to the command. This parameter is unused, and must be NULL."

The "must be NULL" part isn't conditional on the support for `cl_khr_command_buffer_mutable_dispatch`, so I don't think skipping here is correct. However, a future [cl_khr_command_buffer_mutable_memory_commands](https://github.com/KhronosGroup/OpenCL-Docs/pull/1065) extension could relax this for other types of commands.

EwanC (changes requested):

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_fill.cpp
franz: You're right, my mistake. I've dropped the commit.

EwanC (commented):
Sorry it's taken me a while to re-review this.

My main thought from looking at this again is whether we need to be using as many as 256K and 16K work items for the test, which is quite a big increase on what we had before. I'm not sure if on embedded devices or simulators that would slow the test time down and not really give us that much extra testing coverage of the extension. Could we use work item sizes that are an order of magnitude less but still resolve the issues you've seen?

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
```diff
@@ -153,9 +153,9 @@ struct MutableDispatchGlobalSize : public InfoMutableCommandBufferTest
     }
 
     size_t info_global_size = 0;
-    const size_t update_global_size = 3;
-    const size_t sizeToAllocate = global_work_size;
-    const size_t num_elements = sizeToAllocate / sizeof(cl_int);
+    const size_t update_global_size = 16 * 1024;
+    const size_t sizeToAllocate = global_work_size * sizeof(cl_int);
+    const size_t num_elements = global_work_size;
```
bashbaug, line 158: The `sizeToAllocate` issue was already fixed by https://github.com/KhronosGroup/OpenCL-CTS/pull/2437

Unless `update_global_size` needs to be modified, I believe these changes are no longer required.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_info.cpp
```diff
@@ -132,7 +132,17 @@ struct PropertiesArray : public InfoMutableCommandBufferTest
...
-        return InfoMutableCommandBufferTest::Skip();
+
+        cl_mutable_dispatch_fields_khr mutable_capabilities;
+        cl_int error = clGetDeviceInfo(
+            device, CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR,
+            sizeof(mutable_capabilities), &mutable_capabilities, nullptr);
+        test_error(error, "clGetDeviceInfo failed");
+
+        if ((mutable_capabilities & CL_MUTABLE_DISPATCH_ARGUMENTS_KHR) == 0)
+            return true;
```
bashbaug, line 143: I believe these changes are all covered elsewhere now.

We removed this `Skip` function in https://github.com/KhronosGroup/OpenCL-CTS/pull/2498, since it was already covered in a base class.

This check for `CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES` has already existed in the `Skip` check in the base class, see:

https://github.com/KhronosGroup/OpenCL-CTS/blob/349973878bd92be01eefc489206d73bdbe5e7aed/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_basic.h#L104-L108

bashbaug (commented):
@franz, are these changes still required?  Other than the global work size modifications, I think all of the other fixes have already been made in other PRs (not ideal, obviously, but it is what it is).

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp
franz: IIRC the `update_global_size` change was the main motivation of the PR. In the original code, the updated global size (3) is not a divisor of the initial global size (4*16). Is this expected to work even when the implementation doesn't support non-uniform workgroups ?

bashbaug: Note that this test is changing the global work size, not the local work size, and that the local work size recorded into the command buffer is initially `nullptr`:

https://github.com/KhronosGroup/OpenCL-CTS/blob/349973878bd92be01eefc489206d73bdbe5e7aed/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp#L74-L76

It is not modified by the mutable command update, so it remains `nullptr`:

https://github.com/KhronosGroup/OpenCL-CTS/blob/349973878bd92be01eefc489206d73bdbe5e7aed/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_global_size.cpp#L89-L100

Therefore, I don't think there is any requirement to support non-uniform work-groups to run this test.  Do you agree?

franz: OK, that part of mutable dispatch was unclear to me. But i agree with your explanation.

Therefore there is no need to keep the `update_global_size` related changes of this PR, the issue is in our implementation.

If all the other changes are already fixed, we can close this PR.

bashbaug: Cool, if you do find that any of these fixes are still needed, feel free to re-open this PR (or open another new PR if it'd be easier).  Thanks!

PR #2147, opened 2024-11-13
Removed usage of half types in CTS gl tests

### LINKED ISSUE #1982: Eliminate usage of half type in the CTS test

kamil-goras-mobica:
see #1982

### DIFF
--- a/test_conformance/gl/test_images_write_common.cpp
+++ b/test_conformance/gl/test_images_write_common.cpp
@@ -37,8 +37,7 @@ static const char *kernelpattern_image_write_1D =
 "}\n";
 
 static const char *kernelpattern_image_write_1D_half =
-"#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n"
-"__kernel void sample_test( __global half4 *source, write_only image1d_t dest )\n"
+"__kernel void sample_test( __global half *source, write_only image1d_t dest )\n"
 "{\n"
 "    uint index = get_global_id(0);\n"
 "    write_imagef( dest, index, vload_half4(index, (__global half *)source));\n"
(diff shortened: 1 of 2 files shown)


test_conformance/gl/test_images_write_common.cpp
```diff
@@ -37,8 +37,7 @@ static const char *kernelpattern_image_write_1D =
 "}\n";
 
 static const char *kernelpattern_image_write_1D_half =
-"#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n"
-"__kernel void sample_test( __global half4 *source, write_only image1d_t dest )\n"
+"__kernel void sample_test( __global half *source, write_only image1d_t dest )\n"
 "{\n"
 "    uint index = get_global_id(0);\n"
 "    write_imagef( dest, index, vload_half4(index, (__global half *)source));\n"
```
bashbaug, line 43: I think we can get rid of all of these casts to `__global half*` now.  For example:

```suggestion
"    write_imagef( dest, index, vload_half4(index, source));\n"
```

bashbaug (approved):
This is a great first step and we should merge this PR as-is.

I don't think we should close #1982 until we've looked at other uses of `supportsHalf` in these tests, though, since I think they have a similar problem.

kamil-goras-mobica:
> This is a great first step and we should merge this PR as-is.
> 
> I don't think we should close #1982 until we've looked at other uses of `supportsHalf` in these tests, though, since I think they have a similar problem.

I saw in other files in gl directory that function supportsHalf() is used which is checking for cl_khr_fp16. I can take a look at them and create separate PR.

test_conformance/gl/test_images_write_common.cpp
kamil-goras-mobica: > I think we can get rid of all of these casts to `__global half*` now. For example:

@bashbaug Of course I will remove such necessary casting.

bashbaug (approved):

lakshmih (approved):

bashbaug:
I've edited the PR comment and I don't think it will automatically close #1982 anymore.

PR #2151, opened 2024-11-19
[RFC] Move logic for test registration to the test harness

kpet:
And use in device_timer suite.


kpet:
@bashbaug (and others) Here's the change I mentioned in the memory subgroup call. I've rebased it on the latest main. It still seems to work. It may be ready to go or may require additional work. This is work I'd done about 2 years ago so don't remember the exact state of it.

bashbaug (approved):
This is great - thank you for doing this!  Once these pieces are in place we can start to shift existing tests to this new test registration model, and new tests can use it.

One new SPIR-V test was added since this PR was created, so a merge or rebase is needed to apply the same changes to the new test.

Otherwise, LGTM!

bashbaug (approved):

PR #2154, opened 2024-11-20
Corrected test_vulkan to use specific platform/device from harness

### LINKED ISSUE #1926: Fix test_vulkan to allow running it on any platform and VkDevice rather than platform 0, device 0
Fix test_vulkan to allow running it on any platform and VkDevice rather than platform 0, device 0.
Use test_harness for common functionality for choosing platform and device.

shajder:
Fixes #1926 according to task description


joshqti:
This is a nice change, it works for us.  The test harness is a good addition.

test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
```diff
@@ -139,10 +111,48 @@ const VulkanPhysicalDevice &getVulkanPhysicalDevice()
...
+{
+    size_t pdIdx;
+    cl_int errNum = 0;
+    cl_uchar uuid[CL_UUID_SIZE_KHR];
+    const VulkanInstance &instance = getVulkanInstance();
+    const VulkanPhysicalDeviceList &physicalDeviceList =
+        instance.getPhysicalDeviceList();
+
+    errNum = clGetDeviceInfo(deviceId, CL_DEVICE_UUID_KHR, CL_UUID_SIZE_KHR,
+                             uuid, nullptr);
```
bashbaug, line 125: Do we ever check that the `cl_khr_device_uuid` extension is supported for these tests?

If not, is it acceptable behavior to throw `std::runtime_error`?  I think it would be better to skip the tests, if `cl_khr_device_uuid` is a dependency.

nikhiljnv:
As discussed on Memory subgroup call of December 17, 2024, we can add uuid checks subsequently.
Waiting for implementers to test the changes and approve if all good.

nikhiljnv:
Merging as discussed in memory subgroup call on January 7th, 2025.

PR #2158, opened 2024-11-24
Use the system package manager to install cross-compilers

ahesham-arm:
On Ubuntu, use the system's package manager, `apt`, to download and install the cross-compilers for Arm and AArch64. This replaces downloading the compilers as tarballs.

To ensure that the correct version of the compiler is used when calling one without the version suffix, e.g. `aarch64-linux-gnu-gcc`, use `update-alternatives`.

On Linux, the compilers for `x86_64`, `arm`, and `aarch64` will now correctly use the filesystem root as their sysroots, i.e. `/usr/include` will be used when searching for headers, instead of each compiler providing its own.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -31,6 +31,17 @@ jobs:
       - uses: actions/checkout@v4
       - name: Setup Ninja
         uses: seanmiddleditch/gha-setup-ninja@master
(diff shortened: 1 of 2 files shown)


kpet (approved):

svenvh (approved):

bashbaug:

PR #2159, opened 2024-11-25
Use `cl_uint` in some SPIR-V tests to avoid UB

ahesham-arm:
The following tests under `spirv_new`:
* `op_branch_conditional`
* `op_phi`
* `op_selection_merge`

were using randomly generated `cl_int` values, and storing the difference between them as a `cl_int`. If one of the generated values is negative and large enough, while the other is positive and large enough, the difference then becomes a larger value that cannot be represented using `cl_int`.

Switch the tests to use `cl_uint` instead, and update the relevant spvasm{32,64} files to use `OpULessThan` instead of `OpSLessThan`.


bashbaug (approved):
It took me a moment to figure out where 65535 came from in the phi tests, but it made sense once I realized the host code changed too.  The rest of the changes are mechanical.  Thanks for fixing this!

bashbaug:

PR #2161, opened 2024-11-26
Corrected math_brute_force for cl_khr_fp16 tests sqrt and divide to use correctly rounded reference

### LINKED ISSUE #1996: correctly rounded divide test for half is not using a correctly rounded reference
The new correctly rounded divide test for half precision, located in binary_operator_half.cpp is using an fptr for its reference function and computing the reference like this:

        s[j] = HTF(p[j]);
        s2[j] = HTF(p2[j]);
        r[j] = HFF(func.f_ff(s[j], s2[j]));

Here func.f_ff works out to reference_divide().  So r[j] starts with the double precision rounded result of the divide, rounds it to single precision and then rounds that to half.  That's 3 roundings instead of the required single rounding.

Shouldn't this test be disabled until a correct reference is used?

### LINKED ISSUE #2146: test for fp16 sqrt is not using a correctly rounded reference

shajder:
Fixes #1996, fixes #2146 according to issue description

### DIFF
--- a/test_conformance/math_brute_force/binary_operator_half.cpp
+++ b/test_conformance/math_brute_force/binary_operator_half.cpp
@@ -102,6 +102,42 @@ const cl_half specialValuesHalf[] = {
 
 constexpr size_t specialValuesHalfCount = ARRAY_SIZE(specialValuesHalf);
 
+cl_float ComputeRefDouble(cl_float p0, cl_float p1, void *data)
+{
+    TestInfo *job = (TestInfo *)data;
+    fptr func = job->f->func;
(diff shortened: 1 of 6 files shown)


svenvh:
Just repeating what you already commented on https://github.com/KhronosGroup/OpenCL-CTS/issues/1996 and https://github.com/KhronosGroup/OpenCL-CTS/issues/2146: I'd like to understand what values are problematic before we merge this, as your experiments suggest we shouldn't need this.

bashbaug:
Closing as discussed in the February 11th teleconference.

PR #2165, opened 2024-11-27
Add Android to the CI build matrix

ahesham-arm:
Support building for Android on Arm and AArch64 platforms.

Modify the build matrix to add to new variations. Both variants download, extract and setup the Android Native Development Kit (NDK) on a Linux runner. Each variant specifies a `android_arch_abi`, which is passed to CMake during configuration as its `CMAKE_ANDROID_ARCH_ABI` option.

The CMake toolchain file provided by the NDK is used when building for Android. The NDK version used is r27c, which is the latest Long-Term Support (LTS) version. `ANDROID_PLATFORM` is intentionally not set, so the NDK can default to the minimum supported version, which is 21. The compiler (Clang) version used by this NDK is 18.0.3.

The NDK ships with its own sysroot, which has the Linux kernel headers of version 6.8.0, or `LINUX_VERSION_CODE 395264`.

CMake
https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#id23 https://cmake.org/cmake/help/latest/variable/CMAKE_ANDROID_ARCH_ABI.html

NDK
https://developer.android.com/ndk/downloads

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -27,6 +27,14 @@ jobs:
             arch: aarch64
             debug: 1
             extra: " debug"
+          - os: ubuntu-22.04
+            mainmatrix: false
+            arch: android-arm
+            android_arch_abi: armeabi-v7a
+          - os: ubuntu-22.04
+            mainmatrix: false
+            arch: android-aarch64
+            android_arch_abi: arm64-v8a
     steps:
       - uses: actions/checkout@v4
       - name: Setup Ninja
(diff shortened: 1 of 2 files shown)


kpet (approved):
LGTM, thanks! We should start looking into migrating more of what `presubmit.sh` does into the workflow itself but the value we get out of Android builds IMO justifies merging this as is.

bashbaug:
Merging as discussed in the December 3rd teleconference.

PR #2166, opened 2024-11-28
Fix more Wformat warnings related to size_t

svenvh:
Printing of a `size_t` requires the `%zu` specifier.

This fixes occurrences where the previous wrong specifier appears to work in a typical 64-bit build, but causes a Wformat warning in 32-bit builds.

### DIFF
--- a/test_conformance/device_partition/test_device_partition.cpp
+++ b/test_conformance/device_partition/test_device_partition.cpp
@@ -336,8 +336,8 @@ int test_device_set(size_t deviceCount, size_t queueCount, cl_device_id *devices
             if( expectedResults[ i ] != outputData[ i ] )
             {
                 log_error("ERROR: Sample data did not verify for queue %d on "
-                          "device %zu (sample %d, expected %d, got %d)\n",
-                          q, q % deviceCount, (int)i, expectedResults[i],
+                          "device %zu (sample %zu, expected %d, got %d)\n",
+                          q, q % deviceCount, i, expectedResults[i],
                           outputData[i]);
                 for (size_t j=0; j<deviceCount; j++) {
                     if (expectedResultsOneDevice[j][i] == outputData[i])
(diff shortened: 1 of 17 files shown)


EwanC (approved):

test_conformance/device_partition/test_device_partition.cpp
```diff
@@ -332,11 +335,15 @@ int test_device_set(size_t deviceCount, size_t queueCount, cl_device_id *devices
         {
             if( expectedResults[ i ] != outputData[ i ] )
             {
-                log_error( "ERROR: Sample data did not verify for queue %d on device %ld (sample %d, expected %d, got %d)\n",
-                    q, q % deviceCount, (int)i, expectedResults[ i ], outputData[ i ] );
+                log_error("ERROR: Sample data did not verify for queue %d on "
+                          "device %zu (sample %d, expected %d, got %d)\n",
+                          q, q % deviceCount, (int)i, expectedResults[i],
+                          outputData[i]);
```
bashbaug, line 341: Nitpick: I think we could remove the cast on `i` here?

```suggestion
                log_error("ERROR: Sample data did not verify for queue %d on "
                          "device %zu (sample %zu, expected %d, got %d)\n",
                          q, q % deviceCount, i, expectedResults[i],
                          outputData[i]);
```

bashbaug (approved):
Thank you for fixing all of these!

test_conformance/device_partition/test_device_partition.cpp
svenvh: Thanks; the old code isn't using the wrong printf specifier for `i` here, which is why I didn't catch this as part of this effort.  But yes, we can get rid of the cast and use `%zu` too; I've updated the PR.

bashbaug (approved):

lakshmih:
Reviewing.

bashbaug:
Merging as discussed in the December 10th teleconference.

PR #2167, opened 2024-11-29
Fix clCopyImage 1D buffer images tests with use_pitches flag

MichaelRizkalla-arm:
This change fixes the following issues when running `clCopyImage` test with `use_pitches` flag for 1D buffer images.

1. Query device CL version using `get_device_cl_version` rather than using `CL_DEVICE_VERSION`. Previously, using `char` may not be enough to store the return value.
2. Pass `host_ptr` to `clCreateBuffer` when creating the buffer for the 1D image. `host_ptr` may not be `nullptr` if use_pitches flag is used. Also, `buffer_flags` will contain `CL_MEM_USE_HOST_PTR` which requires a host pointer to be passed.
3. Conditionally use the `host_ptr` when calling `clCreateImage`, it will not be used with 1D image buffer. This is made to align with the spec, in which if `mem_flags` has `CL_MEM_USE_HOST_PTR` a host pointer must be present, but if the memory flag is not present, host pointer must be `nullptr`.
4. Use the correct free function `free` and `align_free` based on the allocation function that was used.

### DIFF
--- a/test_conformance/images/clCopyImage/test_copy_generic.cpp
+++ b/test_conformance/images/clCopyImage/test_copy_generic.cpp
@@ -16,9 +16,24 @@
 #include "../testBase.h"
 #include <CL/cl.h>
 
-static void CL_CALLBACK free_pitch_buffer( cl_mem image, void *buf )
+struct pitch_buffer_data
 {
-    free( buf );
+    void *buf;
+    bool is_aligned;
+};
+
+static void CL_CALLBACK free_pitch_buffer(cl_mem image, void *data)
+{
+    pitch_buffer_data *d = static_cast<pitch_buffer_data *>(data);
+    if (d->is_aligned)
+    {
+        align_free(d->buf);
+    }
+    else
+    {
+        free(d->buf);
+    }
+    free(d);
 }
 
 static void CL_CALLBACK release_cl_buffer(cl_mem image, void *buf)
@@ -32,6 +47,7 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
     cl_image_desc imageDesc;
     cl_mem_flags mem_flags = CL_MEM_READ_ONLY;
     void *host_ptr = NULL;
+    bool is_host_ptr_aligned = false;
 
     memset(&imageDesc, 0x0, sizeof(cl_image_desc));
     imageDesc.image_type = imageInfo->type;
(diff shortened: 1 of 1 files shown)


test_conformance/images/clCopyImage/test_copy_generic.cpp
```diff
@@ -43,6 +58,26 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
+    {
+        cl_int err = clGetCommandQueueInfo(queue, CL_QUEUE_DEVICE,
+                                           sizeof(device), &device, nullptr);
+        if (err != CL_SUCCESS)
+        {
+            log_error("Error: Could not get CL_QUEUE_DEVICE from queue");
+            return nullptr;
+        }
+        err = clGetDeviceInfo(device, CL_DEVICE_NUMERIC_VERSION,
+                              sizeof(version), &version, nullptr);
```
bashbaug, line 72: Using `CL_DEVICE_NUMERIC_VERSION` here means that these tests will require OpenCL 3.0.  Could we use the test harness function `get_device_cl_version()` instead, which will work for all OpenCL versions?

If so, since the PR description says "the change queries the version similar to clFillImage test binary," maybe we should update the clFillImage test also?

test_conformance/images/clCopyImage/test_copy_generic.cpp
```diff
@@ -158,23 +173,40 @@ cl_mem create_image( cl_context context, cl_command_queue queue, BufferOwningPtr
...
+                            nullptr, error);
+    }
 
     if (gEnablePitch)
     {
+        pitch_buffer_data *data =
+            static_cast<pitch_buffer_data *>(malloc(sizeof(pitch_buffer_data)));
+        data->buf = host_ptr;
+        data->is_aligned = (CL_VERSION_MAJOR(version) != 1)
+            && (imageInfo->type == CL_MEM_OBJECT_IMAGE1D_BUFFER);
```
bashbaug, line 193: Consider setting up `is_aligned` closer to the call to `align_malloc`.  This might simplify the code and make it easier to handle potential additional calls to `align_malloc` in the future.

MichaelRizkalla-arm: I added `is_host_pointer_aligned` variable to track which allocation function is used.

bashbaug (approved):
Thanks!

bashbaug:

PR #2168, opened 2024-12-03
switch SVM tests to the new test registration framework

bashbaug:
Switches the SVM tests to the new test registration framework.

The first commit is the best to review and contains the actual changes.  The second commit purely has formatting changes.

Note that several of these changes were a bit more than mechanical because many of the SVM tests create a new context vs. using the context provided by the harness and passed to each test function.  The previous code named the context provided by the harness differently, and hence could use the name "context" in each test function, but with the new test registration framework this is no longer possible.  Instead, I am creating the new context using the name "contextWrapper" and then assigning it to the "context" passed to the test function, which seems like the best way to avoid using the wrong context unintentionally.  I am open to suggestions to do this differently.

I have verified that the same calls are made before and after these changes, and specifically that there are no context leaks.


kpet (approved):

bashbaug:
Merging to unblock progress on the unified SVM tests.

Issue #1793, opened 2023-08-01
Memory Leakage in test_conformance/atomics/test_atomics.cpp


incognito1729:
Hi all,

I was going through `test_atomics.cpp`  and found lots of memory leakage due to conditional statements(line no. 26,36,47,58).
So I would like to ask, instead of handling manually free function why are we not using `unique_ptr` here?
I understood some function are using aligned memory allocation/deallocation like  `free_mtdata()`. So here I would like to propose to use wrapper class or custom deleter with smart pointer.

 `char *valids` is used for incrementing at particular index(see line 6,10,28). I would to suggest to use heap allocated memory of type `int8_t` instead of `char` for better coding practice and understanding.

Attaching one function for quick overview.

@bashbaug  @nikhiljnv tagging for quick response.

Disclaimer: I am new to this codebase, Apologies if I misunderstood the use of constructs and why they are implemented in the way they are.

```cpp

1. bool test_atomic_xchg_verify_int(size_t size, cl_int *refValues,
2.                                  cl_int finalValue)
3. {
4.     /* For xchg, each value from 0 to size - 1 should have an entry in the ref
5.      * array, and ONLY one entry */
6.     char *valids;
7.     size_t i;
8.     char originalValidCount = 0;
9. 
10.     valids = (char *)malloc(sizeof(char) * size);
11.     memset(valids, 0, sizeof(char) * size);
12. 
13.     for (i = 0; i < size; i++)
14.     {
15.         if (refValues[i] == INT_TEST_VALUE)
16.         {
17.             // Special initial value
18.             originalValidCount++;
19.             continue;
20.         }
21.         if (refValues[i] < 0 || (size_t)refValues[i] >= size)
22.         {
23.             log_error(
24.                 "ERROR: Reference value %zu outside of valid range! (%d)\n", i,
25.                 refValues[i]);
26.             return false;
27.         }
28.         valids[refValues[i]]++;
29.     }
30. 
31.     /* Note: ONE entry will have zero count. It'll be the last one that
32.      executed, because that value should be the final value outputted */
33.     if (valids[finalValue] > 0)
34.     {
35.         log_error("ERROR: Final value %d was also in ref list!\n", finalValue);
36.         return false;
37.     }
38.     else
39.         valids[finalValue] = 1; // So the following loop will be okay
40. 
41.     /* Now check that every entry has one and only one count */
42.     if (originalValidCount != 1)
43.     {
[...]

PR #2169, opened 2024-12-03
Added unique_ptrs and MTdataHolder to test_atomics.cpp

kamil-goras-mobica:
Fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/1793 according to description

### DIFF
--- a/test_conformance/atomics/test_atomics.cpp
+++ b/test_conformance/atomics/test_atomics.cpp
@@ -137,8 +137,8 @@ int test_atomic_function(cl_device_id deviceID, cl_context context,
     int error;
     size_t threads[1];
     clMemWrapper streams[2];
-    std::vector<cl_int> refValues;
-    std::vector<cl_int> startRefValues;
+    std::vector<char> refValues;
+    std::vector<char> startRefValues;
     size_t threadSize, groupSize;
     const char *programLines[4];
     char pragma[512];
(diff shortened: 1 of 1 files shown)


test_conformance/atomics/test_atomics.cpp
```diff
@@ -247,52 +248,50 @@ int test_atomic_function(cl_device_id deviceID, cl_context context,
     log_info("\t(thread count %d, group size %d)\n", (int)threadSize,
              (int)groupSize);
 
-    refValues = (cl_int *)malloc(typeSize * threadSize);
+    refValues.reset(new cl_int[typeSize * threadSize]);
```
svenvh, line 251: It would have been nice to get rid of the manual allocations too (and use `std::vector` for example), but at least getting rid of the manual `free`s is an improvement.

test_conformance/atomics/test_atomics.cpp
```diff
@@ -345,27 +344,28 @@ int test_atomic_function(cl_device_id deviceID, cl_context context,
             {
                 // Int version
                 intVal = testFns.ExpectedValueIntFn(
-                    threadSize, (cl_int *)startRefValues, i);
+                    threadSize, (cl_int *)startRefValues.get(), i);
```
svenvh, line 347: Are these `(cl_int *)` and `(cl_long *)` casts still needed?

test_conformance/atomics/test_atomics.cpp
```diff
@@ -247,52 +248,47 @@ int test_atomic_function(cl_device_id deviceID, cl_context context,
     log_info("\t(thread count %d, group size %d)\n", (int)threadSize,
              (int)groupSize);
 
-    refValues = (cl_int *)malloc(typeSize * threadSize);
+    refValues.resize(typeSize * threadSize);
```
bashbaug, line 251: This isn't wrong per se, but it is allocating more memory than it was before, because the previous code was allocating `typeSize * threadSize` bytes, and the new code is allocating `typeSize * threadSize` integers.

Maybe instead of `std::vector<cl_int>`for  `refValues` we should instead have `std::vector<char>`, like we do for `destItems`?  Or, even better, maybe `std::vector<uint8_t>` or similar?

kamil-goras-mobica: @bashbaug yes you are right, my fault. I will correct it.

lakshmih:
Reviewing

bashbaug (approved):

svenvh (approved):

bashbaug:

PR #2170, opened 2024-12-05
Add support for allocating DMA buffers

ahesham-arm:
This adds support for allocating DMA buffers on systems that support it, i.e. Linux and Android.

On mainline Linux, starting version 5.6 (equivalent to Android 12), there is a new kernel module framework available called [DMA-BUF Heaps](https://github.com/torvalds/linux/blob/master/drivers/dma-buf/dma-heap.c). The goal of this framework is to provide a standardised way for user applications to allocate and share memory buffers between different devices, subsystems, etc. The main feature of interest is that the framework provides device-agnostic allocation; it abstracts away the underlying hardware, and provides a single IOCTL, DMA_HEAP_IOCTL_ALLOC. Mainline implementation provides two heaps that act as character devices that can allocate DMA buffers; system, which uses the buddy allocator, and cma, which uses the [CMA](https://developer.toradex.com/software/linux-resources/linux-features/contiguous-memory-allocator-cma-linux/) (Contiguous Memory Allocator). Both of these are [kernel configuration options](https://github.com/torvalds/linux/blob/master/drivers/dma-buf/heaps/Kconfig) that need to be enabled when building the Linux kernel. Generally, any kernel module implementing this framework is made available under /dev/dma_heaps/<heap_name>, e.g. /dev/dma_heaps/system.

The implementation currently only supports one type of DMA heaps; `system`, the default device path for which is `/dev/dma_heap/system`. The path can be overridden at runtime using an environment variable, `OCL_CTS_DMA_HEAP_PATH_SYSTEM`, if needed. Extending this in the future should be trivial (subject to platform support), by adding an entry to the enum `dma_buf_heap_type`, and an appropriate default path and overriding environment variable name.

The proposed implementation will conditionally compile if the conditions are met (i.e. building for Linux or Android, using kernel headers >= 5.6.0), and will provide a compile-time warning otherwise, and return `-1` as the DMA handle in runtime if not.

To demonstrate the functionality, a new test is added for the `cl_khr_external_memory_dma_buf` extension. If the extension is supported by the device, a DMA buffer will be allocated and used to create a CL buffer, that is then used by a simple kernel.

This should provide a way forward for adding more tests that depend on DMA buffers.


test_common/harness/alloc.cpp
```diff
@@ -0,0 +1,122 @@
...
+    const auto override_path = getenv(entry.env_var);
+    const auto dma_heap_path =
+        (override_path == nullptr) ? entry.default_path : override_path;
+
+    const int dma_heap_fd = open(dma_heap_path, DMA_HEAP_FLAGS);
+    if (dma_heap_fd == -1)
+    {
+        log_error(
+            "Opening the DMA heap device: %s failed with error: %d (%s)\n",
+            dma_heap_path, errno, strerror(errno));
```
bashbaug, line 82: I only had a little time to poke around with this, but I did notice that I cannot access this device by default:

```
Opening the DMA heap device: /dev/dma_heap/system failed with error: 13 (Permission denied)
```

I can get a little further by running with `sudo`, but that's not generally something I'd like to do when running CTS tests.  Is there a way to use this mechanism without needing root access?

If it's relevant, I'm on a relatively stock Ubuntu 24.10 system:

```
Linux bashbaug-newpc 6.11.0-13-generic #14-Ubuntu SMP PREEMPT_DYNAMIC Sat Nov 30 23:51:51 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

Additionally:

```sh
$ ls -l /dev/dma_heap/system 
crw------- 1 root root 249, 0 Dec 28 15:12 /dev/dma_heap/system
```

ahesham-arm: Correct, on a clean installation, `/dev/dma_heap/system` is limited to RW by root only by default. This can be modified by the system administrator (I assume that is also you, since you can `sudo`) to give permissions to specific users or groups (R, W or both) just like any other Unix file.

kpet:
Discussed in the 2025/01/21 memory TSG: we should update the PR to skip the tests when hitting permission issues instead of failing them.

ahesham-arm:
Updated to address the review comments.

joshqti:
We are ok with this change

test_common/harness/alloc.cpp
bashbaug: In case it helps anyone else, I followed instructions similar to these to give permissions to `/dev/dma_heap/system` to a specific group, and then I was able to run the test:

https://raspberrypi.stackexchange.com/questions/141106/how-to-fix-the-libcamera-error-could-not-open-any-dmaheap-device

test_common/harness/alloc.cpp
```diff
@@ -0,0 +1,123 @@
...
+    if (result == -1)
+    {
+        log_info("Failed to close the DMA heap device: %s\n", dma_heap_path);
+    }
+
+    return dma_heap_data.fd;
+#else
+#warning                                                                       \
+    "Kernel version doesn't support DMA buffer heaps (at least v5.6.0 is required)."
+    return -1;
```
bashbaug, line 111: Should this return `TEST_SKIPPED_ITSELF` vs. failing also?

test_common/harness/alloc.cpp
```diff
@@ -0,0 +1,123 @@
...
+#endif // LINUX_VERSION_CODE >= KERNEL_VERSION(5, 6, 0)
+}
+
+#else
+int allocate_dma_buf(uint64_t size, dma_buf_heap_type heap_type)
+{
+    log_error(
+        "OS doesn't have DMA buffer heaps (only Linux and Android do).\n");
+
+    return -1;
```
bashbaug, line 121: One more - should this return `TEST_SKIPPED_ITSELF` vs. failing also?

test_conformance/extensions/cl_khr_external_memory_dma_buf/test_external_memory_dma_buf.cpp
```diff
@@ -0,0 +1,143 @@
...
+    clMemWrapper src_buffer =
+        clCreateBuffer(context, CL_MEM_READ_ONLY | CL_MEM_COPY_HOST_PTR,
+                       buffer_size_bytes, src_data.data(), &error);
+    test_error(error, "Failed to create the source buffer.");
+
+    /* Imported buffer creation */
+    int dma_buf_fd = allocate_dma_buf(buffer_size_bytes);
+    if (dma_buf_fd < 0)
+    {
+        if (dma_buf_fd == TEST_SKIPPED_ITSELF)
```
bashbaug, line 74: This is fine for now, but it's overloading the `int` return type and making an assumption that `TEST_SKIPPED_ITSELF` is a negative value, and it's possible it won't be true in the future - see https://github.com/KhronosGroup/OpenCL-CTS/issues/2276.

bashbaug (approved):
Mostly LGTM.  I left comments for a few nice-to-have fixes, but nothing is must-fix.  Thanks!

test_conformance/extensions/cl_khr_external_memory_dma_buf/test_external_memory_dma_buf.cpp
ahesham-arm: I agree. I think we should aim for using something like `std::optional` if we are sticking with standard C++ options, or an equivalent of Rust's `std::Result`.

bashbaug (approved):

bashbaug:
Merging as discussed in the February 25th memory subgroup.

hvdijk:
Hi, on systems where `LINUX_VERSION_CODE < KERNEL_VERSION(5, 6, 0)` , this breaks the build:
```
/builds/ComputeAorta/ci-github/build/OpenCL-CTS/test_common/harness/alloc.cpp:109:2: warning: #warning "Kernel version doesn't support DMA buffer heaps (at least v5.6.0 is required)." [-Wcpp]
  109 | #warning                                                                       \
      |  ^~~~~~~
/builds/ComputeAorta/ci-github/build/OpenCL-CTS/test_common/harness/alloc.cpp:53:30: error: 'dma_buf_heap_helper_t lookup_dma_heap(dma_buf_heap_type)' defined but not used [-Werror=unused-function]
   53 | static dma_buf_heap_helper_t lookup_dma_heap(dma_buf_heap_type heap_type)
      |                              ^~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```
The warning seems accurate: `lookup_dma_heap` is unconditionally defined, but only used inside `#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 6, 0)`. That function should probably only be defined when it will be used.

Could you take a look at this? Would you like me to?

ahesham-arm:
> Hi, on systems where `LINUX_VERSION_CODE < KERNEL_VERSION(5, 6, 0)` , this breaks the build due to CMakeLists.txt adding `-Werror` and `-Wunused-function` where supported:
> 
> ```
> /builds/ComputeAorta/ci-github/build/OpenCL-CTS/test_common/harness/alloc.cpp:109:2: warning: #warning "Kernel version doesn't support DMA buffer heaps (at least v5.6.0 is required)." [-Wcpp]
>   109 | #warning                                                                       \
>       |  ^~~~~~~
> /builds/ComputeAorta/ci-github/build/OpenCL-CTS/test_common/harness/alloc.cpp:53:30: error: 'dma_buf_heap_helper_t lookup_dma_heap(dma_buf_heap_type)' defined but not used [-Werror=unused-function]
>    53 | static dma_buf_heap_helper_t lookup_dma_heap(dma_buf_heap_type heap_type)
>       |                              ^~~~~~~~~~~~~~~
> cc1plus: all warnings being treated as errors
> ```
> 
> The warning seems accurate: `lookup_dma_heap` is unconditionally defined, but only used inside `#if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 6, 0)`. That function should probably only be defined when it will be used.
> 
> Could you take a look at this? Would you like me to?

Thank you for reporting this, sorry for the break. I have a fix for this but unfortnatuely I cannot test this locall. Can you please try it and comment on the PR if it works for you? Thanks!

https://github.com/KhronosGroup/OpenCL-CTS/pull/2292

PR #2171, opened 2024-12-06
Added object comparability verification for default device queue

### LINKED ISSUE #1485: add test for object comparability

shajder:
Fixes #1485 according to work plan, point CL_QUEUE_DEVICE_DEFAULT

### DIFF
--- a/test_conformance/api/test_api_consistency.cpp
+++ b/test_conformance/api/test_api_consistency.cpp
@@ -355,6 +355,28 @@ int test_consistency_device_enqueue(cl_device_id deviceID, cl_context context,
                 "clSetDefaultDeviceCommandQueue did not return "
                 "CL_INVALID_OPERATION");
         }
+        else
+        {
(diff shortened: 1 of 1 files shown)


bashbaug (changes requested):
Is this test needed?  It looks like this scenario is already tested in `api/test_device_command_queue.cpp`.  See: https://github.com/KhronosGroup/OpenCL-CTS/pull/1551

shajder:
> Is this test needed? It looks like this scenario is already tested in `api/test_device_command_queue.cpp`. See: #1551

Thanks for noticing that. It looks like a unfortunate coincidence. I posted [my analysis](https://github.com/KhronosGroup/OpenCL-CTS/issues/1485#issuecomment-2435103325) (without answer) in october 2024 before first commit in #1551 and I created my PR before #1551 was merged. Pity.

shajder:
> Should we just close this PR then?

Yes

PR #2174, opened 2024-12-10
add a unified SVM consistency check test

bashbaug:
Adds a unified SVM consistency check test, as per the description in #2150.

Note that the target branch for this PR is the cl_khr_unified_svm branch, not the main branch.

### DIFF
--- a/test_conformance/SVM/test_unified_svm_consistency.cpp
+++ b/test_conformance/SVM/test_unified_svm_consistency.cpp
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2020 The Khronos Group Inc.
+// Copyright (c) 2024 The Khronos Group Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/test_conformance/SVM/test_unified_svm_consistency.cpp
+++ b/test_conformance/SVM/test_unified_svm_consistency.cpp
@@ -1,0 +1,152 @@
+//
+// Copyright (c) 2020 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


bashbaug:
Hmm, the CI build checks are going to fail unless we update the scripts to pull the headers from https://github.com/KhronosGroup/OpenCL-Headers/pull/269.

bashbaug:
I've updated the test as we discussed.

I also pulled in the CI changes so the unified SVM headers are used, and now the automated builds are working again.

test_conformance/SVM/test_unified_svm_consistency.cpp
```diff
@@ -0,0 +1,159 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "common.h"
+#include <cinttypes>
+
+REGISTER_TEST(unified_svm_consistency)
+{
+    if (!is_extension_available(deviceID, "cl_khr_unified_svm"))
```
EwanC, line 22: I've not been following the unified svm test plan closely, but is the idea that the extension tests will live in `test_conformance/SVM` rather than `test_conformance/extensions/cl_khr_unified_svm`?

bashbaug: Good question.  I don't think we've discussed this.  I'll create a new test executable if that's what the group prefers, but I think we've generally tried to minimize test executables, so I'm inclined to put the unified SVM tests here unless we have a good reason to do otherwise.

bashbaug: Hmm, I may have found a few good reasons to create a new test executable:

1. The "SVM" tests run an `InitCL` function to decide whether to run the tests, which currently requires OpenCL 2.0 or newer, and requires that the query for `CL_DEVICE_SVM_CAPABILITIES` returns a non-zero value.  The first part is debatably OK, though the extension is currently written to support OpenCL 1.2 devices also.  The second part is not OK though, because a device can support unified SVM without supporting SVM.

2. The "SVM" tests run with `forceNoContextCreation` set to `true`, which means no context or queue are created.  Not the end of the world, but I think the harness context and queue will be sufficient for most of the unified SVM tests, so it'd be nice to keep things simple and use them.

bashbaug:
Merging (to the cl_khr_unified_svm branch), as discussed in the February 11th memory TSG.

PR #2175, opened 2024-12-10
Removed checking for supporting half

kamil-goras-mobica:
Fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/1982 according to description

### DIFF
--- a/test_conformance/gl/test_images_getinfo_common.cpp
+++ b/test_conformance/gl/test_images_getinfo_common.cpp
@@ -23,7 +23,6 @@
 #include <CL/cl_gl.h>
 #endif
 
-extern int supportsHalf(cl_context context, bool *supports_half);
 
 static int test_image_info(cl_context context, cl_command_queue queue,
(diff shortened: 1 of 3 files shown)


test_conformance/gl/test_images_getinfo_common.cpp
```diff
@@ -97,10 +96,6 @@ static int test_image_format_get_info(cl_context context,
     if (fmt->type == kHalf)
     {
         if (DetectFloatToHalfRoundingMode(queue)) return 0;
```
bashbaug, line 98: This isn't something changed in this PR, but we're also calling `DetectFloatToHalfRoundingMode` again a few lines below.  Here, we return 0 (success) if the detection fails.  Below, we return 1 (failure) if the detection fails.

Do we need both calls?  (I don't think we do.)
Should we return success or failure if the detection fails?  (Probably failure?)

It's fine to fix this in a separate PR.

bashbaug (approved):

bashbaug:

PR #2176, opened 2024-12-10
Added object comparability verification for context devices query

### LINKED ISSUE #1485: add test for object comparability

shajder:
Fixes #1485 according to work plan, point `CL_CONTEXT_DEVICES`

### DIFF
--- a/test_conformance/api/test_queries.cpp
+++ b/test_conformance/api/test_queries.cpp
@@ -485,6 +485,112 @@ int test_get_command_queue_info_compatibility(cl_device_id deviceID,
 int test_get_context_info(cl_device_id deviceID, cl_context context, cl_command_queue ignoreQueue, int num_elements)
 {
     int error;
+
+    // #1 query single device context and perform object comparability test
+    {
+        clContextWrapper context_to_test =
+            clCreateContext(NULL, 1, &deviceID, nullptr, nullptr, &error);
+        test_error_ret(error, "Unable to create testing context", CL_SUCCESS);
+
+        cl_uint num_devices = 0;
+        error = clGetContextInfo(context_to_test, CL_CONTEXT_NUM_DEVICES,
+                                 sizeof(cl_uint), &num_devices, nullptr);
+        test_error(error, "clGetContextInfo CL_CONTEXT_NUM_DEVICES failed\n");
+
+        test_assert_error(num_devices == 1,
+                          "Context must contain exactly one device\n");
+
+        cl_device_id comp_device = nullptr;
+        error = clGetContextInfo(context_to_test, CL_CONTEXT_DEVICES,
+                                 sizeof(cl_device_id), &comp_device, nullptr);
+        test_error(error, "clGetContextInfo CL_CONTEXT_DEVICES failed\n");
+
+        test_assert_error(
+            deviceID == comp_device,
+            "Unexpected result returned by CL_CONTEXT_DEVICES query");
+    }
+
+
+    // #2 query multi-device context and perform objects comparability test
+    auto test_multi_device_context = [&]() {
+        cl_uint maxComputeUnits = 0;
+        cl_int err =
+            clGetDeviceInfo(deviceID, CL_DEVICE_MAX_COMPUTE_UNITS,
+                            sizeof(maxComputeUnits), &maxComputeUnits, NULL);
+        test_error_fail(err, "Unable to get maximal number of compute units");
+
+        cl_device_partition_property partitionProp[] = {
+            CL_DEVICE_PARTITION_EQUALLY,
+            static_cast<cl_device_partition_property>(maxComputeUnits / 2), 0
+        };
+
+        cl_uint deviceCount = 0;
+        // how many sub-devices can we create?
+        err = clCreateSubDevices(deviceID, partitionProp, 0, nullptr,
+                                 &deviceCount);
+        if (err != CL_SUCCESS)
+        {
+            log_info("Can't partition device, test not supported\n");
+            return TEST_SKIPPED_ITSELF;
+        }
+
+        if (deviceCount < 2)
+            test_error_ret(
+                CL_INVALID_VALUE,
+                "Multi context test for CL_INVALID_PROPERTY not supported",
+                TEST_SKIPPED_ITSELF);
+
+        // get the list of subDevices
+        SubDevicesScopeGuarded scope_guard(deviceCount);
+        err = clCreateSubDevices(deviceID, partitionProp, deviceCount,
+                                 scope_guard.sub_devices.data(), &deviceCount);
+        if (err != CL_SUCCESS)
+        {
+            log_info("Can't partition device, test not supported\n");
+            return TEST_SKIPPED_ITSELF;
+        }
+
+        /* Create a multi device context */
+        clContextWrapper multi_device_context = clCreateContext(
+            NULL, (cl_uint)deviceCount, scope_guard.sub_devices.data(), nullptr,
+            nullptr, &err);
+        test_error_ret(err, "Unable to create testing context",
+                       TEST_SKIPPED_ITSELF);
+
+        cl_uint num_devices = 0;
+        error = clGetContextInfo(multi_device_context, CL_CONTEXT_NUM_DEVICES,
+                                 sizeof(cl_uint), &num_devices, nullptr);
+        test_error_ret(error,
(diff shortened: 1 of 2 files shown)


test_conformance/api/test_queries.cpp
```diff
@@ -476,30 +476,124 @@ REGISTER_TEST(get_command_queue_info_compatibility)
 REGISTER_TEST(get_context_info)
 {
     int error;
+
+    // query single device context and perform object comparability test
+    clContextWrapper context_to_test =
+        clCreateContext(NULL, 1, &device, nullptr, nullptr, &error);
+    test_error_ret(error, "Unable to create testing context", CL_SUCCESS);
```
bashbaug, line 483: Creating a context is relatively expensive.  Could we query the device from the context provided by the harness instead of creating a new context?

test_conformance/api/test_queries.cpp
```diff
@@ -476,30 +476,124 @@ REGISTER_TEST(get_command_queue_info_compatibility)
...
+
+    cl_device_partition_property partitionProp[] = {
+        CL_DEVICE_PARTITION_EQUALLY,
+        static_cast<cl_device_partition_property>(maxComputeUnits / 2), 0
+    };
+
+    cl_uint num_devices = 0;
+    // how many sub-devices can we create?
+    err = clCreateSubDevices(device, partitionProp, 0, nullptr, &num_devices);
+    if (err != CL_SUCCESS || num_devices < 2)
```
bashbaug, line 548: Possible areas for improvement here:

1. Could we make the right queries to determine if the device is partitionable vs. trying to partition the device and handling any generated errors?  This isn't a negative test, so it would be nice not to rely on error behavior.
2. Could we handle other partitioning types so the test may run on more devices?

test_conformance/api/test_queries.cpp
```diff
@@ -476,30 +476,124 @@ REGISTER_TEST(get_command_queue_info_compatibility)
...
+                   TEST_FAIL);
+
+    test_assert_error(devices.size() == scope_guard.sub_devices.size(),
+                      "Size of devices arrays must be in sync\n");
+
+    for (cl_uint i = 0; i < devices.size(); i++)
+    {
+        test_error_ret(scope_guard.sub_devices[i] == devices[i],
+                       "Unexpected result returned by CL_CONTEXT_DEVICES query",
+                       TEST_FAIL);
```
bashbaug, line 592: Observation: This requires the devices returned by `CL_CONTEXT_DEVICES` to be returned in the same order as the devices passed to `clCreateContext`.  This is likely the behavior most users will expect, but I don't think we require this in the spec.  Should we clarify `CL_CONTEXT_DEVICES` to document this requirement?

test_conformance/api/test_queries.cpp
shajder: Corrected.

test_conformance/api/test_queries.cpp
shajder: Done as suggested, sadly I don't have partitionable device - I will appreciate a proper test, thanks.

test_conformance/api/test_queries.cpp
```diff
@@ -476,30 +478,180 @@ REGISTER_TEST(get_command_queue_info_compatibility)
...
+        { CL_DEVICE_PARTITION_BY_AFFINITY_DOMAIN,
+          CL_DEVICE_AFFINITY_DOMAIN_L4_CACHE, 0, 0, 0 },
+        { CL_DEVICE_PARTITION_BY_AFFINITY_DOMAIN,
+          CL_DEVICE_AFFINITY_DOMAIN_L3_CACHE, 0, 0, 0 },
+        { CL_DEVICE_PARTITION_BY_AFFINITY_DOMAIN,
+          CL_DEVICE_AFFINITY_DOMAIN_L2_CACHE, 0, 0, 0 },
+        { CL_DEVICE_PARTITION_BY_AFFINITY_DOMAIN,
+          CL_DEVICE_AFFINITY_DOMAIN_L1_CACHE, 0, 0, 0 },
+        { CL_DEVICE_PARTITION_BY_AFFINITY_DOMAIN,
+          CL_DEVICE_AFFINITY_DOMAIN_NEXT_PARTITIONABLE, 0, 0, 0 }
```
bashbaug, line 575: We should be fine with just the last entry here, for `NEXT_PARTITIONABLE`.  This will partition by whatever the "next" affinity domain is, meaning we don't need to check for each of the other affinity domains specifically.

test_conformance/api/test_queries.cpp
```diff
@@ -476,30 +478,180 @@ REGISTER_TEST(get_command_queue_info_compatibility)
...
 
+REGISTER_TEST(get_context_info_mult_devices)
+{
+    cl_int err = CL_SUCCESS;
+    size_t size = 0;
+
+    // query multi-device context and perform objects comparability test
+    err = clGetDeviceInfo(device, CL_DEVICE_PARTITION_PROPERTIES, 0, nullptr,
+                          &size);
+    test_error_ret(err, "clGetDeviceInfo failed", TEST_FAIL);
```
bashbaug, line 537: Minor: this could be shortened slightly by using `test_error_fail` rather than `test_error_ret` and explicitly returning `TEST_FAIL`:

```suggestion
    test_error_fail(err, "clGetDeviceInfo failed");
```

Same pattern is used in a few other places in this test.

test_conformance/api/test_queries.cpp
```diff
@@ -476,30 +478,180 @@ REGISTER_TEST(get_command_queue_info_compatibility)
...
+
+    test_error_ret(size == 0, "Can't partition device, test not supported\n",
+                   TEST_SKIPPED_ITSELF);
```
bashbaug, line 540: This causes a slightly scary "error message" to be printed in the cases when a device does not support sub-devices:

```
ERROR: Can't partition device, test not supported
! ((unknown) from /home/bashbaug/git/OpenCL-CTS/test_conformance/api/test_queries.cpp:550)
get_context_info_mult_devices test not supported
```

It looks like most other tests that skip themselves either call `log_info` before returning `TEST_SKIPPED_ITSELF`, or do not log any message at all.  I think we should do something similar here, rather than calling `test_error_ret`.

bashbaug (commented):
Thanks!  I've verified that the multi-device context works with the Intel CPU implementation, which supports device partitioning.

I have a few follow-on comments but this is getting close.

test_conformance/api/test_queries.cpp
shajder: Done!

test_conformance/api/test_queries.cpp
shajder: Done

test_conformance/api/test_queries.cpp
shajder: Corrected

test_conformance/api/test_queries.cpp
shajder: Done

bashbaug (commented):
Sorry, I found one more issue checking for OpenCL object leaks.

test_conformance/api/test_queries.cpp
```diff
@@ -476,30 +478,173 @@ REGISTER_TEST(get_command_queue_info_compatibility)
...
+    {
+        bool found = false;
+        for (auto it = scope_guard->sub_devices.begin();
+             it != scope_guard->sub_devices.end();)
+        {
+            if (*it == devices[i])
+            {
+                found = true;
+                scope_guard->sub_devices.erase(
+                    it); // in case devices contain duplicates
```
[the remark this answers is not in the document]
shajder: That is correct, fixed. 

> I've verified that the multi-device context works with the Intel CPU implementation...

do you know of an easy way to "emulate" partitioning functionality ? None of my devices have multi-device caps and this is real pain to commit similar changes without proper test. Thanks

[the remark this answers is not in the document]
bashbaug: POCL supports device partitioning for CPUs, although it's failing this test (heads up @pjaaskel).

Otherwise, as long as it's an x86 system, I think you can use the Intel CPU OpenCL implementation, even on a non-Intel CPU.

bashbaug (approved):
Thanks, latest version LGTM.

lakshmih (approved):

bashbaug:

Issue #2173, opened 2024-12-09
Memory sanitizer error in arrayimagecopy and imagearraycopy tests


alan-baker:
After integrating #2138 internally, we started seeing some memory sanitizer issues in the following tests:
* basic/arrayimagecopy
* basic/arrayimagecopy3d
* basic/imagearraycopy
* basic/imagearraycopy3d

The failure mode is a free that is smaller than the allocated size. Here is an example:
```
Initializing random seed to 0.
Requesting Default device based on command line for platform index 0 and device index 0
Compute Device Name = Tesla V100-SXM2-16GB, Compute Device Vendor = NVIDIA Corporation, Compute Device Version = OpenCL 3.0 CLVK on Vulkan v1.3.242 driver 2246066368, CL C Version = OpenCL C 1.2 CLVK on Vulkan v1.3.242 driver 2246066368
Device latest conformance version passed: v2023-12-12-00
Supports single precision denormals: NO
sizeof( void*) = 8  (host)
sizeof( void*) = 4  (device)
imagearraycopy...
9031 third_party/tcmalloc/allocation_sampling.cc:321 *** GWP-ASan has detected a memory error ***
9031 third_party/tcmalloc/allocation_sampling.cc:322 Error originates from memory allocated at:
    @     0x557108c19948  tcmalloc::tcmalloc_internal::SampleifyAllocation()
    @     0x557108c070e3  tcmalloc::tcmalloc_internal::alloc_small_sampled_hooks_or_perthread<>()
    @     0x557108b40d0a  create_random_data()
    @     0x557108ae43f4  test_imagearraycopy_single_format()
    @     0x557108ae4c1d  test_imagearraycommon()
    @     0x557108ae4cf9  test_imagearraycopy()
    @     0x557108b4c9ba  callSingleTestFunction()
    @     0x557108b4c44a  callTestFunctions()
    @     0x557108b4bff7  parseAndCallCommandLineTests()
    @     0x557108b4b92a  runTestHarnessWithCheck()
    @     0x557108a9bdca  main
    @     0x7f7690ac23d4  __libc_start_main
    @     0x557108a9bcaa  _start
9031 third_party/tcmalloc/allocation_sampling.cc:339 Mismatched-size-delete of 1 bytes (expected 262144 bytes) at:
    @     0x557108c1a4fc  tcmalloc::tcmalloc_internal::MaybeUnsampleAllocation()
[...]

PR #2177, opened 2024-12-10
Fix using incorrect free function in arrayimagecopy and imagearraycopy basic tests' unique_ptr

MichaelRizkalla-arm:
This change addresses a free function mis-match issue where `delete` is used to free memory allocated via `malloc` within `create_random_data` and in the test function.

The change involves the following tests:
1. arrayimagecopy
2. arrayimagecopy3d
3. imagearraycopy
4. imagearraycopy3d

This should address #2173


svenvh (approved):

MichaelRizkalla-arm:
I believe we can merge this PR as-is as a short-term fix. I have created issue #2182 to implement a long-term proper fix.

dneto0:
I would like to see this fix land.

svenvh:
Merging as this has two approvals already and should be non-controversial.

PR #2178, opened 2024-12-10
Add utility macro to skip test when an extension is not supported

kpet:
Use in cl_khr_external_semaphore suite as an example.

### DIFF
--- a/test_common/harness/testHarness.h
+++ b/test_common/harness/testHarness.h
@@ -150,6 +150,17 @@ template <typename T> T *register_test(const char *name, Version version)
 
 #define REGISTER_TEST(name) REGISTER_TEST_VERSION(name, Version(1, 2))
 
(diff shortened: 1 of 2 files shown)


kpet:
@bashbaug Here's the idea I mentioned.

bashbaug (approved):
Works for me!  Do you want to merge this as-is, or switch more of the tests to use the utility macro first?

Note, we may eventually want a variant that skips the test if the extension version isn't new enough, also, though we can cross that bridge when we come to it.

kpet:
I'd suggest merging as is to make this available. It'll take a while to convert the whole code base.

> we may eventually want a variant that skips the test if the extension version isn't new enough, also, though we can cross that bridge when we come to it.

Agree x2. Almost mentioned this in the commit message. The case with a version check is a lot less common.

lakshmih (approved):

bashbaug:

PR #2180, opened 2024-12-10
update CI to use the cl_khr_unified_svm headers

bashbaug:
Updates CI to use headers with support for the cl_khr_unified_svm extension in the cl_khr_unified_svm branch.

Note that this PR targets the cl_khr_unified_svm branch, not the main branch.


kpet (approved):
LGTM. We should figure out what needs to change to get rid of the code format check issue. Rebase the cl_khr_unified_svm branch?

kpet (approved):
That works!

PR #2183, opened 2024-12-11
Fix errors in `test_vulkan`

ahesham-arm:
This fixes three problems in `test_vulkan`:

1. One negative test is violating the OpenCL specification. A call to `clEnqueue{Wait,Signal}SemaphoresKHR` with an invalid semaphore should return `CL_INVALID_SEMAPHORE_KHR` and not `CL_INVALID_VALUE`.
> [CL_INVALID_SEMAPHORE_KHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_SEMAPHORE_KHR) if any of the semaphore objects specified by sema_objects is not valid.

2. When populating the list of supported external memory handle types for Vulkan, the types are unconditionally added to the list, without checking if the device supports it or not, this fix is namely for `VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD`.

3. If a device does not support an optional extension (that is required for a test), the test should skip, not throw an exception and fail. A test failure should be reserved for the cases where a device claims support for an extension but then fails to execute the test correctly.

### DIFF
--- a/test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
+++ b/test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
@@ -225,7 +225,8 @@ getDefaultVulkanQueueFamilyToQueueCountMap()
 }
 
 const std::vector<VulkanExternalMemoryHandleType>
-getSupportedVulkanExternalMemoryHandleTypeList()
+getSupportedVulkanExternalMemoryHandleTypeList(
+    const VulkanPhysicalDevice &physical_device)
 {
     std::vector<VulkanExternalMemoryHandleType> externalMemoryHandleTypeList;
 
@@ -238,8 +239,23 @@ getSupportedVulkanExternalMemoryHandleTypeList()
     externalMemoryHandleTypeList.push_back(
         VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_KMT);
 #else
-    externalMemoryHandleTypeList.push_back(
-        VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD);
+    VkPhysicalDeviceExternalBufferInfo buffer_info = {};
+    buffer_info.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTERNAL_BUFFER_INFO;
+    buffer_info.handleType = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT_KHR;
+
+    VkExternalBufferProperties buffer_properties = {};
+    buffer_properties.sType = VK_STRUCTURE_TYPE_EXTERNAL_BUFFER_PROPERTIES;
+
+    vkGetPhysicalDeviceExternalBufferProperties(physical_device, &buffer_info,
+                                                &buffer_properties);
+
+    if (buffer_properties.externalMemoryProperties.externalMemoryFeatures
+        & VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT)
+    {
+
+        externalMemoryHandleTypeList.push_back(
+            VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD);
+    }
 #endif
 
     return externalMemoryHandleTypeList;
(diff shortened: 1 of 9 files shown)


test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
```diff
@@ -228,8 +229,13 @@ getSupportedVulkanExternalMemoryHandleTypeList()
     externalMemoryHandleTypeList.push_back(
         VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_KMT);
 #else
-    externalMemoryHandleTypeList.push_back(
-        VULKAN_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD);
+    if (physical_device.hasExtension(
+            VK_KHR_EXTERNAL_MEMORY_CAPABILITIES_EXTENSION_NAME))
```
nikhiljnv, line 233: From Ben: Is this the right extension name to check for?

ahesham-arm: Yes, source: https://registry.khronos.org/vulkan/specs/latest/man/html/VkExternalMemoryHandleTypeFlagBits.html

joshqti: I would prefer we wait on merging this, it introduces a segfault for us.  We would like to debug

starside: This is not the correct place to check for this extension.  We already request it in vulkan_wrapper.cpp: line 146.  This leads to a crash for us because VK_KHR_EXTERNAL_MEMORY_CAPABILITES_EXTENISION_NAME is not detected, and getSupportedVulkanExternalMemoryHandleTypeList() returns a 0 length list.  When getSupportedVulkanExternalMemoryHandleTypeList()[0] is called, a crash is the result.

fredlee12001:
I believe test_consistency_external_image needs the similar change to skip the test if extension not supported.

ahesham-arm:
> I believe test_consistency_external_image needs the similar change to skip the test if extension not supported.
@fredlee12001 Good catch, thank you! This must have been missed in the recent merge conflict resolution.

Updated both `test_consistency_external_image` and `test_consistency_external_semaphore` to match the behaviour of `test_consistency_external_buffer`, i.e. skip the test if the extension is not supported (don't throw or fail the test).

test_conformance/common/vulkan_wrapper/vulkan_utility.cpp
joshqti: This is not the correct place to check for this extension. We already request it in vulkan_wrapper.cpp: line 146. This leads to a crash for us because VK_KHR_EXTERNAL_MEMORY_CAPABILITES_EXTENISION_NAME is not detected, and getSupportedVulkanExternalMemoryHandleTypeList() returns a 0 length list. When getSupportedVulkanExternalMemoryHandleTypeList()[0] is called, a crash is the result.

ahesham-arm: @joshqti You're right. Additionally, from the `VK_KHR_external_memory_capabilities` section in the Vulkan specification:
> Promotion to Vulkan 1.1
All functionality in this extension is included in core Vulkan 1.1, with the KHR suffix omitted. The
original type, enum, and command names are still available as aliases of the core functionality.
We use Vulkan 1.1, so we should not need to check this extension anyway.

That being said, we should have checks for the usage pattern that we intend with this type of memory, in our case it is export (from Vulkan), i.e. check that `VK_EXTERNAL_MEMORY_FEATURE_EXPORTABLE_BIT` is set for external memory that has the handle type: `VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_FD_BIT_KHR`.

I have update the pull request to reflect this. Please let me know if you disagree and if you are continuing to see segmentation faults with this patch applied.

joshqti:
This change works for us

nikhiljnv (approved):
LGTM.

nikhiljnv:
Merged as discussed in OpenCL memory subgroup call on Jan 28th, 2025.

PR #2184, opened 2024-12-11
Add CMake installation rules

ahesham-arm:
Add installation rules for all the binary targets.

Targets are installed under `<CMAKE_INSTALL_PREFIX>/bin/<CONFIG>` where `<CONFIG>` is `CMAKE_BUILD_TYPE` for single-config generators, e.g. Unix Makefiles and Ninja, or the build configuration for multi-config generators, e.g. Ninja Multi-Config and Visual Studio.

This creates the target `install` on Unix and `INSTALL` on Windows.


bashbaug (commented):
Can you please also check if we need to install the Vulkan SPIR-V files?  See: https://github.com/KhronosGroup/OpenCL-CTS/pull/2055

At some point it would be nice to install the OpenCL SPIR-V files in a sensible place also (and ideally, to auto-build them if any of the source spvasm changes), but this is less critical because the path to the OpenCL SPIR-V files can be set using `--spirv-binaries-path`.

ahesham-arm:
> Can you please also check if we need to install the Vulkan SPIR-V files? See: #2055

You're right, yes they do. I note that the CMakeLists.txt for the Vulkan shaders is wrong in two ways:
1. Project configuration and building are two distinct steps. `execute_process` runs in configuration time and should not be used for generating files. Instead, `add_custom_command` and `add_custom_target` should be used to define _build_ targets that get executed at build times.
2. The source directory should never be modified by the configuration or build processes, thanks for fixing that in #2055.

ahesham-arm:
Fixed the way the Vulkan shaders are generated in #2199 as a requirement for having correct installation rules for them.

ahesham-arm:
#2200 handles the automatic generation of the OpenCL SPIR-V files for `spirv_new`.

bashbaug (dismissed):

bashbaug:
Is there any ordering requirement for this PR and #2199 or #2200?

ahesham-arm:
> Is there any ordering requirement for this PR and https://github.com/KhronosGroup/OpenCL-CTS/pull/2199 or https://github.com/KhronosGroup/OpenCL-CTS/pull/2200?

We should merge #2199 and #2200 first (no specific order required), and then I can edit this PR to additionally install those files as well.

bashbaug:
Hi, both #2199 and #2200 are merged now, so I think this is ready for the final updates.

ahesham-arm:
> Hi, both #2199 and #2200 are merged now, so I think this is ready for the final updates.

Thanks for reminding me! Updated the pull request to include installation rules for both SPIR-V binaries and Vulkan shaders.

bashbaug (approved):
I tested this on Linux and everything seems to have been installed in the right place.

I specifically tested:

* test_spirv_new, to ensure that the SPIR-V binaries are picked up properly
* test_compiler, to ensure that the compiler include directories are picked up properly
* test_vulkan, to ensure that the Vulkan SPIR-V shaders are picked up properly.

It'd be nice for somebody to verify Windows as well, since I'm not sure if I'll get to this before the call tomorrow.

Note that there is no requirement to use the install target, so the risk of merging these changes as-is are fairly low.

bashbaug:

PR #2185, opened 2024-12-11
Fix more 32-bit Wformat warnings

svenvh:
This fixes occurrences where the previous wrong specifier appears to work in a typical 64-bit build, but causes a Wformat warning in 32-bit builds.

### DIFF
--- a/test_conformance/computeinfo/main.cpp
+++ b/test_conformance/computeinfo/main.cpp
@@ -736,7 +736,7 @@ void dumpConfigInfo(config_info* info)
             }
             break;
         case type_cl_device_id:
-            log_info("\t%s == %ld\n", info->opcode_name,
+            log_info("\t%s == %" PRIdPTR "\n", info->opcode_name,
                      (intptr_t)info->config.device_id);
             break;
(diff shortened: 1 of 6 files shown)


test_conformance/computeinfo/main.cpp
```diff
@@ -736,7 +736,7 @@ void dumpConfigInfo(config_info* info)
             }
             break;
         case type_cl_device_id:
-            log_info("\t%s == %ld\n", info->opcode_name,
+            log_info("\t%s == %" PRIdPTR "\n", info->opcode_name,
                      (intptr_t)info->config.device_id);
```
bashbaug, line 740: For my own education, is there any reason we don't use `%p` here?

bashbaug (approved):

test_conformance/computeinfo/main.cpp
svenvh: I was wondering the same, but decided not to change the output format as part of this PR (`%p` will print in hexadecimal, while the test currently outputs decimal).

bashbaug:

Issue #2181, opened 2024-12-10
Migrate all suites to the new test registration framework


kpet:
https://github.com/KhronosGroup/OpenCL-CTS/pull/2151 introduced a new centralised test registration framework. We should migrate all the test suites to this framework to enable further centralised improvements of the test framework.

Here is a list of all the suites with the current status:

- [ ] allocations
- [ ] api
- [ ] atomics
- [ ] basic
- [ ] buffers
- [ ] c11_atomics
- [ ] commonfns
- [ ] compiler
- [ ] computeinfo
- [ ] contractions
- [ ] conversions
- [ ] d3d10
- [ ] d3d11
- [ ] device_execution
- [ ] device_partition
- [x] device_timer
- [ ] events
- [ ] extensions
- [ ] generic_address_space
- [ ] geometrics
- [ ] gl
- [ ] gles
- [ ] half
- [ ] images
- [ ] integer_ops
- [ ] math_brute_force
- [ ] mem_host_flags
- [ ] multiple_device_context
- [ ] non_uniform_work_group
- [ ] pipes
- [ ] printf
- [ ] profiling
- [ ] relationals
- [ ] select
- [ ] spir
- [x] spirv_new
- [ ] subgroups
- [x] SVM
- [ ] thread_dimensions
- [ ] vectors
- [ ] vulkan
- [ ] workgroups

PR #2186, opened 2024-12-12
[NFC] Change the name of the default device provided by the new registration framework

kpet:
The code base uses a mix of 'device' and 'deviceID'. I suggest we standardise on 'device' which is shorter and slightly more prevalent.

Contributes to #2181


bashbaug (approved):

bashbaug:
Straightforward change, merging immediately.

PR #2188, opened 2024-12-13
add option to bruteforce test to force count of worker threads

kcencele:
- to be able to have deterministic results it is useful to have a mechanism
  to force the same count of workers
- this commit doesn't change the default settings but expands
  functionality

Signed-off-by: Katarzyna Cencelewska <katarzyna.cencelewska@intel.com>

### DIFF
--- a/test_conformance/math_brute_force/main.cpp
+++ b/test_conformance/math_brute_force/main.cpp
@@ -379,6 +379,7 @@ static int ParseArgs(int argc, const char **argv)
     gTestNames.push_back("");
 
     int singleThreaded = 0;
+    int forcedWorkerThreads = 0;
 
     { // Extract the app name
         strncpy(appName, argv[0], MAXPATHLEN - 1);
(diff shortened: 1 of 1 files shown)


test_conformance/math_brute_force/main.cpp
```diff
@@ -540,7 +546,17 @@ static int ParseArgs(int argc, const char **argv)
...
 
-    if (singleThreaded) SetThreadCount(1);
+    if (singleThreaded)
+    {
+        vlog("*** WARNING: Reduce worker threads to 1                 ***\n");
+        SetThreadCount(1);
+    }
+    else if (forcedWorkerThreads > 0)
+    {
+        vlog("*** WARNING: Reduce worker threads to %d                ***\n",
```
b-sumner, line 556: Why does the warning say reduce instead of force?

kcencele: true better will be use force, thanks for your comment, changed

svenvh (approved):

bashbaug:

PR #2190, opened 2024-12-13
Migrate vectors suite to the new test registration framework

ahesham-arm:
Contributes to #2181


kpet:
Trivial change, merging straight away.

PR #2191, opened 2024-12-13
Migrate allocations suite to the new test registration framework

ahesham-arm:
Contributes to #2181

### DIFF
--- a/test_conformance/allocations/main.cpp
+++ b/test_conformance/allocations/main.cpp
@@ -272,47 +272,27 @@ int doTest(cl_device_id device, cl_context context, cl_command_queue queue,
     return failure_counts;
 }
 
-int test_buffer(cl_device_id device, cl_context context, cl_command_queue queue,
-                int num_elements)
-{
-    return doTest(device, context, queue, BUFFER);
-}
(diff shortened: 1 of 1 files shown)


test_conformance/allocations/main.cpp
```diff
@@ -416,8 +397,8 @@ void printUsage(const char *execName)
              "of the memory objects.\n");
     log_info("\n");
     log_info("Test names (Allocation Types):\n");
-    for (int i = 0; i < test_num; i++)
+    for (int i = 0; i < test_registry::getInstance().num_tests(); i++)
     {
-        log_info("\t%s\n", test_list[i].name);
+        log_info("\t%s\n", test_registry::getInstance().definitions()[i].name);
     }
```
kpet, line 403: Doesn't this code duplicate the corresponding feature in the harness? Can we get rid of it altogether?

ahesham-arm: It does, but this test suite has a custom arguments parsing logic, and a custom help message, `printUsage`. Removing that still prints the general usage help, but does not list additional options such as `randomize`, `single`, `multiple`, etc.

kpet (approved):

PR #2194, opened 2024-12-17
semaphore: fix Wformat warning

svenvh:
`i` is an `int`, so shouldn't use the `%zu` specifier for printing.


bashbaug (approved):

nikhiljnv (approved):
LGTM.

nikhiljnv:
Merging as discussed in memory subgroup call on December 17, 2024.

PR #2195, opened 2024-12-18
cmake: remove global -Wno-format

svenvh:
Move the global `-Wno-format` compiler option to the individual tests that still trigger Wformat warnings.  The majority of the tests now compile cleanly with `-Wformat` enabled.


bashbaug (approved):
Nice - thank you for fixing all of these!

EwanC (approved):

PR #2197, opened 2024-12-18
Migrate commonfns suite to new test registration framework

kpet:


bashbaug (approved):

PR #2199, opened 2024-12-19
Generate the Vulkan shaders at build time

ahesham-arm:
Use `add_custom_command` and `add_custom_target` instead of `execute_process` so the generation of the Vulkan shader is done at build time and not configuration time.

Use `configure_file` instead of string replacement.


kpet (approved):
Nice!

bashbaug (approved):

bashbaug:

PR #2202, opened 2024-12-25
printf: Fix floating-point rounding consistency for RTZ devices

cycheng:
1. In vector test, prepare RTZ answer for RTZ rounding.

2. In mixed_format_random test, for a given float 'arg', the test previously used 'arg' directly to generate ref_str: 
   ```
     ref_str << str_sprintf(format, arg);
   ```

   This approach incorrectly assumes:
   ```
     (float) arg == to_fp(to_str(arg));
   ```

   However, this assumption fails under RTZ rounding. For example: 
   ```
     arg = 0xC642549C 
     to_str(arg) = -12437.152343f 
     to_fp_rtz(-12437.152343f) = 0xC642549B (-0X1.84A936P+13) 
     to_fp_rte(-12437.152343f) = 0xC642549C (-0X1.84A938P+13)
   ```

   To address this, the reference result is now computed based on the literal float string rather than the original 'arg' value.

### DIFF
--- a/test_conformance/printf/test_printf.cpp
+++ b/test_conformance/printf/test_printf.cpp
@@ -317,8 +317,10 @@ cl_program makeMixedFormatPrintfProgram(cl_kernel* kernel_ptr,
         {
             const float max_range = 100000.f;
             float arg = get_random_float(-max_range, max_range, gMTdata);
-            args_str << str_sprintf("%f", arg) << "f, ";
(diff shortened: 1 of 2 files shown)


lakshmih:
Reviewing

lakshmih (approved):

bashbaug:

PR #2207, opened 2024-12-26
fix negative_create_command_buffer_device_does_not_support_out_of_ord…

### LINKED ISSUE #2204: cl_khr_command_buffer negative_create_command_buffer_device_does_not_support_out_of_order_queue wrong behavior
`cl_khr_command_buffer negative_create_command_buffer_device_does_not_support_out_of_order_queue` [setup function](https://github.com/KhronosGroup/OpenCL-CTS/blob/v2024-12-17-00/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp#L272) tries to create a queue with `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE` and expects it to succeed.

It feels incoherent regarding the name of the test. 

The test is skipped if `clGetDeviceInfo(CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR)` does not report `CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR`.

Maybe the intent of the test was the opposite? A device supporting `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE` but not `CL_COMMAND_BUFFER_CAPABILITY_OUT_OF_ORDER_KHR`, and making sure it fails properly?

rjodinchr:
…er_queue

Test should be skipped if device does not support out of order queue or if device supports out of order command buffer.

Fix #2204

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp
@@ -291,7 +291,7 @@ struct CreateCommandBufferDeviceDoesNotSupportOutOfOderQueue
 
         // If device does not support out of order queue or if device supports
         // out of order command buffer test should be skipped
-        return queue_out_of_order_support && !out_of_order_support;
+        return !queue_out_of_order_support || out_of_order_support;
     }
 
     clCommandQueueWrapper out_of_order_queue;
--- a/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp
@@ -289,8 +289,9 @@ struct CreateCommandBufferDeviceDoesNotSupportOutOfOderQueue
     {
         BasicCommandBufferTest::Skip();
 
-        // If device supports out of order queues test should be skipped
-        return out_of_order_support != 0;
+        // If device does not support out of order queue or if device supports
+        // out of order command buffer test should be skipped
+        return queue_out_of_order_support && !out_of_order_support;
     }
 
     clCommandQueueWrapper out_of_order_queue;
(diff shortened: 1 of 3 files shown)


test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp
```diff
@@ -289,8 +289,9 @@ struct CreateCommandBufferDeviceDoesNotSupportOutOfOderQueue
     {
         BasicCommandBufferTest::Skip();
 
-        // If device supports out of order queues test should be skipped
-        return out_of_order_support != 0;
+        // If device does not support out of order queue or if device supports
+        // out of order command buffer test should be skipped
+        return queue_out_of_order_support && !out_of_order_support;
```
EwanC, line 294: This test is a negative test to verify that the correct error is returned when a device doesn't have `out_of_order_support` command-buffer support, but it is attempted to be create one with an out-of-order queue. So I agree we do need the `queue_out_of_order_support` check you've added and skip if a device can't create an out-or-order queue to use for the test.

However I'm not quite following the current logic why we only skip the test if a device has out-of-order queue support and another condition. I think the check could be written as follows to better reflect the updated code comment.
```cpp
return !queue_out_of_order_support || (out_of_order_support !=0)
```

test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_create.cpp
```diff
@@ -289,8 +289,9 @@ struct CreateCommandBufferDeviceDoesNotSupportOutOfOderQueue
     {
         BasicCommandBufferTest::Skip();
```
EwanC, line 290: I'm not sure if this is related to the issue you are fixing, but we're ignoring the return value from this call, and I'm pretty sure it should beshould be
```cpp
        if (BasicCommandBufferTest::Skip()) return true;
```

This can be a separate GitHub issue/PR to fix if it's unrelated to your issue though (might be other occurrences we need to fix as part of that).

rjodinchr: I agree that should be fixed. But I think it is not related to the issue I'm fixing here.

bashbaug:

PR #2211, opened 2025-01-07
adds SPIR-V tests for scalar printf operands

bashbaug:
Adds targeted SPIR-V tests for printf with scalar operands.  See:

* https://github.com/KhronosGroup/OpenCL-Docs/issues/1211
* https://github.com/KhronosGroup/OpenCL-Docs/pull/1236

The fp32 test is likely to be the most interesting, especially on devices that support fp64, because printf with scalar fp32 operands is not generated by default in this case with Clang and the SPIR-V LLVM Translator.

### DIFF
--- a/test_conformance/spirv_new/test_extinst_printf.cpp
+++ b/test_conformance/spirv_new/test_extinst_printf.cpp
@@ -1,0 +1,230 @@
+//
+// Copyright (c) 2025 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "harness/os_helpers.h"
+#include "testBase.h"
+
+#if defined(_WIN32)
+#include <io.h>
+#define streamDup(fd1) _dup(fd1)
+#define streamDup2(fd1, fd2) _dup2(fd1, fd2)
+#else
+#if defined(__APPLE__)
+#include <sys/sysctl.h>
+#endif
+#include <unistd.h>
+#define streamDup(fd1) dup(fd1)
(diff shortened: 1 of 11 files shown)


test_conformance/spirv_new/test_extinst_printf.cpp
```diff
@@ -0,0 +1,230 @@
...
+    bool acquired = false;
+};
+
+template <typename T>
+static int printf_operands_helper(cl_context context, cl_device_id device,
+                                  cl_command_queue queue,
+                                  const char* spirvFileName,
+                                  const char* kernelName,
+                                  const char* expectedResults, T value)
+{
```
svenvh, line 108: The new tests currently fail on our implementation because the output isn't captured (the OpenCL spec leaves the "output stream" implementation defined).  The printf suite already has some extra handling to set up a call-back if the `cl_arm_printf` extension is supported (see #434); any chance you could add such handling to this test?

test_conformance/spirv_new/test_extinst_printf.cpp
```diff
@@ -0,0 +1,230 @@
...
+#define streamDup2(fd1, fd2) dup2(fd1, fd2)
+#endif
+
+#include <fstream>
+#include <vector>
+
+// TODO: Unify with test_printf.
+struct StreamGrabber
+{
+    StreamGrabber(): tempFileName(get_temp_filename()) {}
```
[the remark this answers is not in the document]
bashbaug: Fixed!

FWIW, I was initially thrown by a Windows warning about returning a pointer to a local variable, but I think this was caused by a different bug, where `strdup` wasn't getting called properly in one Windows-specific case.  I've "fixed" this problem for now by calling `strdup` for it as well.

Ideally we'd return a `std::string` or similar to avoid needing to free the pointer `get_temp_filename` returns, but that's a bigger change than I'd like to make in this PR.

test_conformance/spirv_new/test_extinst_printf.cpp
bashbaug: Sure, done - I think.  Can you have a look at the latest changes?  I don't have an implementation to test the callback specifically, but I do see that a new context and command queue are being created for each of the tests now, so it _should_ be working.

svenvh: Thanks!  Yes, this works for our implementation.

svenvh (approved):

bashbaug:
Would anyone else like to review this before merging?  I'll add "focused review" just in case, but because this is a SPIR-V change maybe the one Arm approval is sufficient.

bashbaug:

PR #2212, opened 2025-01-07
Delete read_array_half and write_array_half tests

kpet:
These tests are broken, have been disabled for the longest time, and don't add significant value.


bashbaug (approved):

Issue #2152, opened 2024-11-19
Risky testing of versions of provisional extensions


hvdijk:
OpenCL CTS includes testing for the provisional extensions [cl_khr_command_buffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#cl_khr_command_buffer) and [cl_khr_command_buffer_mutable_dispatch](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#cl_khr_command_buffer_mutable_dispatch). Both of these contain a warning:

> **This is a *provisional* extension and must be used with caution. See the [description](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#boilerplate-provisional-header) of provisional header files for enablement and stability details.**

This warning is because as long as the extension has provisional status, future versions must be anticipated to potentially contain both ABI-incompatible and API-incompatible changes. Any code that wishes to use this extension must pay attention to the extension's version. This is not hypothetical: ABI-incompatible and API-incompatible changes have been made earlier this year.

For cl_khr_command_buffer, the extension version is checked by <https://github.com/KhronosGroup/OpenCL-CTS/blob/918d561c6c42fb7f3e2ffe1d1574c0c37bb2b15b/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.h#L102-L112>

For cl_khr_command_buffer_mutable_dispatch, the extension version is checked by <https://github.com/KhronosGroup/OpenCL-CTS/blob/918d561c6c42fb7f3e2ffe1d1574c0c37bb2b15b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_basic.h#L90-L95> and <https://github.com/KhronosGroup/OpenCL-CTS/blob/918d561c6c42fb7f3e2ffe1d1574c0c37bb2b15b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_info.cpp#L127-L133>

All three checks have the following in common: if the reported extension version is older than what OpenCL CTS tests, skip the test.

[...]

PR #2215, opened 2025-01-08
Check command-buffer extension version for equality

EwanC:
As described in Issue https://github.com/KhronosGroup/OpenCL-CTS/issues/2152 we currently check the provisional extension version supported by a vendor is the same or older than a test specified version. However, it was discussed in the WG that this should be a check for equality to avoid hitting issues when a implementation tests against an older version of the CTS using a lower extension version, with API breaking changes having occurred since.

The tests for the command-buffer family of extensions are the only provisional KHR tests using this versioning check, so this PR updates all those cases to equality.


bashbaug (approved):
LTGM.

At some point we might want to consider an alternative policy, such as failing the test if the reported version is older than the test version (but continuing to skip if the reported version is newer than the test version), but skipping is the safest thing to do initially.

bashbaug:

PR #2217, opened 2025-01-09
Fix derived command-buffer Skip() checks

EwanC:
It was noticed during another PR review https://github.com/KhronosGroup/OpenCL-CTS/pull/2207/files#r1903921283 that there 
was a case where the return value of a `Skip()` check was ignored, this is fixed in this PR.

I've also tracking down occurrences of derived class overriding the `Skip()` test fixture method, but not calling the parents class `Skip()` check inside of the method. I believe omitting this parent skip check wasn't intentional, it's clearer to explicitly respect the parent classes skip conditions, even if we've got away with not needing too due to the way the derived class skip conditions have been defined.


bashbaug (approved):

bashbaug:

PR #2219, opened 2025-01-09
fix several compile issues with Visual Studio toolchains

bashbaug:
Fixes several compile issues I am seeing for my version of Visual Studio related to an ambiguous call to `fpclassify`, which is called by `isnan` and other similar functions, specifically for the `cl_half` type:

```
19>C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt\corecrt_math.h(401,1): error C2668: 'fpclassify': ambiguous call to overloaded function
19>C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt\corecrt_math.h(298,31): message : could be 'int fpclassify(long double) throw()'
19>C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt\corecrt_math.h(293,31): message : or       'int fpclassify(double) throw()'
19>C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt\corecrt_math.h(288,31): message : or       'int fpclassify(float) throw()'
19>C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt\corecrt_math.h(401,1): message : while trying to match the argument list '(_Ty)'
```

Some of these issues seem like differences in compiler behavior, but at least one appears to have identified a legitimate bug.

Specifically, this change:

* Removes the special-case checks for finite half numbers for commonfns, since this is already handled by `UlpFn`.  (test with: `test_commonfns degrees radians`)
* Assigns to temporary variables to eliminate the ambiguous function call for relationals.  (test with: `test_relationals relational*`)
* Properly converts from half to float when checking for NaNs for select.  This is the one that seems like a legitimate bug.  (test with: `test_select select_half_ushort select_half_short`)
* Uses `std::enable_if` to disambiguate a function call for spirv_new.  (test with: `test_spirv_new decorate_saturated*`)

If it's helpful, my specific Visual Studio version is:

```
Microsoft Visual Studio Professional 2019
Version 16.11.20
VisualStudio.16.Release/16.11.20+32929.386
```

I also have the Windows Software Development Kit 10.0.19041.685 installed.

### DIFF
--- a/test_conformance/commonfns/test_base.h
+++ b/test_conformance/commonfns/test_base.h
@@ -174,27 +174,6 @@ template <typename T> inline half conv_to_half(const T &val)
     return 0;
 }
 
-template <typename T> bool isfinite_fp(const T &v)
-{
-    if (std::is_same<T, half>::value)
-    {
-        // Extract FP16 exponent and mantissa
-        uint16_t h_exp = (((half)v) >> (CL_HALF_MANT_DIG - 1)) & 0x1F;
-        uint16_t h_mant = ((half)v) & 0x3FF;
-
-        // !Inf test
-        return !(h_exp == 0x1F && h_mant == 0);
-    }
-    else
-    {
-#if !defined(_WIN32)
-        return std::isfinite(v);
-#else
-        return isfinite(v);
-#endif
-    }
-}
-
 template <typename T> float UlpFn(const T &val, const double &r)
 {
     if (std::is_same<T, half>::value)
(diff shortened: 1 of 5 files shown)


svenvh (approved):
The fixes look okay to me.

Switching to C++17 and using `constexpr if` might also fix some of the issues, but then I don't know whether C++17 support was already stable enough in VS2019.  Safest to go with your current patch I suppose.

lakshmih (approved):

bashbaug:
Merging as discussed in the February 18th teleconference.

PR #2162, opened 2024-11-26
Removed reciprocal test in math_brute_force

kamil-goras-mobica:
Fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/2145 according to description

PR #2221, opened 2025-01-10
Added corrections to re-enable reciprocal test in math_brute_force suite for relaxed math mode

### LINKED ISSUE #2145: Re-enable reciprocal test in math_brute_force suite
... or delete it if not required. It is currently commented out (https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/math_brute_force/function_list.cpp#L349). Identified by @bashbaug .

shajder:
fixes #2145

As suggested by @svenvh reciprocal has different precision requirements than divide. This PR introduces special path for reciprocal for binar_float_operator to test reciprocal with relaxed math. If this PR will get approvals, invalidate PR #2162


bashbaug:
Possibly dumb question: why are we introducing a special path through binary_operator_float to test reciprocal?  Reciprocal is a unary operation, not a binary operation...

shajder:
> Possibly dumb question: why are we introducing a special path through binary_operator_float to test reciprocal?  Reciprocal is a unary operation, not a binary operation...

Yep, good point. I went that way just because I wasn't able to choose a better spot. There is no unary float operator, right? To create one I would have to duplicate a lot of code. Any other clues?

svenvh:
> > Possibly dumb question: why are we introducing a special path through binary_operator_float to test reciprocal?  Reciprocal is a unary operation, not a binary operation...
> 
> Yep, good point. I went that way just because I wasn't able to choose a better spot. There is no unary float operator, right? To create one I would have to duplicate a lot of code. Any other clues?

Treating reciprocal as a binary operator with its first argument being the constant 1 seems the least worst solution for now.  I'm in the process of reducing code duplication in the math bruteforce suite, so eventually we may be able to move it into its own place.  But for now I'm okay with @shajder's chosen solution.

test_conformance/math_brute_force/binary_operator_float.cpp
```diff
@@ -688,6 +720,9 @@ int TestFunc_Float_Float_Float_Operator(const Func *f, MTdata d,
     double maxErrorVal = 0.0;
     double maxErrorVal2 = 0.0;
 
+    // reciprocal differs from divide only in relaxed mode, skip otherwise
+    if ((strcmp(f->name, "reciprocal") == 0) && !relaxedMode) return CL_SUCCESS;
```
bashbaug, line 724: This is true currently, but it might not be true for long, see: https://github.com/KhronosGroup/OpenCL-Docs/pull/1293

I think it would be better to test "reciprocal" in all cases, even those where the required accuracy matches "divide".

shajder: @bashbaug I can't spot any plans to change spec in reference to fp32 reciprocal precision, am I missing something ?

bashbaug: Discussed in the January 21st teleconference.  We would prefer to test all of the cases, even if there may be some overlap with the divide test:

1. In theory, "divide" and "reciprocal" may have different accuracy requirements, though we are currently trending to keep them in sync.
2. Testing "reciprocal" explicitly will give us more coverage for the "1.0 / x" cases explicitly, which we may not get through "divide".

shajder: Thanks, reciprocal activated for all fp16/fp32/fp64 and relaxed cases

test_conformance/math_brute_force/function_list.cpp
```diff
@@ -418,6 +421,21 @@ const Func functionList[] = {
...
+      { (void*)reference_relaxed_reciprocal },
+      2.5f,
+      0.0f,
+      0.0f,
+      3.0f,
+      2.5f,
+      INFINITY,
+      FTZ_OFF,
+      RELAXED_ON,
+      binaryOperatorF },
```
bashbaug, line 438: To be clear, I'm fine with this as-is, but one side effect of testing reciprocal as part of `binaryOperatorF` is that we seem to be testing more than we need to.  As one extreme example, if I only test the fp16 reciprocal, I'm seeing 128K kernel enqueues of 32K work-items.... even though there are only 64K possible fp16 values to take a reciprocal of.  If I test a true fp16 unary function (e.g. asin) instead, I only see two 32K kernel enqueues, which covers all 64K fp16 values.

For a fp32 reciprocal, we're testing a total of 4B work-items (256K kernel enqueues of 16K work-items), but the data is generated randomly, so we're not guaranteed to have full coverage of all possible fp32 inputs (although we will certainly have fairly good coverage).  I think if we properly considered reciprocal to be a "unary function" instead, we could guarantee full coverage of all fp16 and fp32 values.  We will still need random sampling for fp64 testing, but that's unavoidable.

PS: With two input buffers for the "binary operator", we're also transferring twice as much input data as we need to.  For some of the devices I'm testing it takes more time to transfer the data than it does to execute the reciprocal kernel, so the extra transfers are significant.

bashbaug (approved):

lakshmih (approved):

PR #2225, opened 2025-01-13
Added object comparability verification for CL_DEVICE_PLATFORM query

### LINKED ISSUE #1485: add test for object comparability

shajder:
Fixes #1485 according to work plan, point CL_DEVICE_PLATFORM


test_conformance/api/test_queries.cpp
```diff
@@ -656,6 +656,57 @@ REGISTER_TEST(get_device_info)
...
+    for (int p = 0; p < (int)num_platforms && comp_platform == nullptr; p++)
+    {
+        error = clGetDeviceIDs(platforms[p], CL_DEVICE_TYPE_ALL, 0, nullptr,
+                               &num_devices);
+        test_error(error, "clGetDeviceIDs failed");
+
+        std::vector<cl_device_id> devices(num_devices);
+        error = clGetDeviceIDs(platforms[p], CL_DEVICE_TYPE_ALL, num_devices,
+                               devices.data(), nullptr);
+        test_error(error, "clGetDeviceIDs failed");
```
bashbaug, line 685: I think we should ignore platforms where these queries return an error vs. failing the test.  I know this is atypical and not the pattern used by most tests, but I think it is the right thing to do here for two reasons:

1. If we fail the test when these queries return an error, it means that a buggy platform that just so happens to be installed on the system alongside the platform being tested could cause the test to fail, even if the platform being tested is correct and conformant.
2. `clGetDeviceIDs` is a little weird in that it is specified to return an error when no devices are found rather than returning `num_device` equal to zero.  This means that a conformant platform could return an error, which would also cause the test to fail.  Linking in: https://github.com/KhronosGroup/OpenCL-Docs/issues/1269

Since this test is run through the harness, we could also consider using the `CL_PLATFORM_INDEX` environment variable to choose the platform to test against, though this loop is fine too, and it has the added benefit of testing that `clGetDeviceIDs` returns the same set of device IDs if it is called multiple times.

test_conformance/api/test_queries.cpp
```diff
@@ -656,6 +656,57 @@ REGISTER_TEST(get_device_info)
...
+
+        // try to find invalid device
```
bashbaug, line 687: Nitpick: I don't think this is an "invalid device" that we are looking for.

test_conformance/api/test_queries.cpp
```diff
@@ -656,6 +656,57 @@ REGISTER_TEST(get_device_info)
...
+        for (auto did : devices)
+        {
+            if (did == device)
+            {
+                comp_platform = platforms[p];
+                break;
```
bashbaug, line 693: Should we fail the test if no `comp_platform` is found, basically if we get out of this loop and `comp_platform` is still `nullptr`?

test_conformance/api/test_queries.cpp
shajder: Done, moreover I've added fallback path based on `CL_PLATFORM_INDEX`, thanks for the hint.

test_conformance/api/test_queries.cpp
shajder: Done

test_conformance/api/test_queries.cpp
shajder: Corrected, thanks!

lakshmih (approved):

test_conformance/api/test_queries.cpp
```diff
@@ -796,6 +796,72 @@ REGISTER_TEST(get_device_info)
...
+    if (comp_platform == nullptr)
+    {
+        char *env_mode = getenv("CL_PLATFORM_INDEX");
+        if (env_mode != nullptr)
+        {
+            int choosen_platform_index = atoi(env_mode);
+            if (choosen_platform_index < static_cast<int>(platforms.size()))
+                comp_platform = platforms[choosen_platform_index];
+        }
+    }
```
bashbaug, line 848: I don't think we need this "fallback path".  We've already searched through all of the platforms on the system, so we shouldn't find it if we look again at the platform described by `CL_PLATFORM_INDEX`.  We can just remove this entire code block.

test_conformance/api/test_queries.cpp
```diff
@@ -796,6 +796,72 @@ REGISTER_TEST(get_device_info)
...
+
+    test_error_fail(comp_platform == nullptr,
+                    "Test failed to locate platform for comparison!")
+
+        // 2. compare platforms found with and without using query
+        cl_platform_id plat = nullptr;
```
bashbaug, line 854: Nitpick:
```suggestion
    // 2. compare platforms found with and without using query
    cl_platform_id plat = nullptr;
```

test_conformance/api/test_queries.cpp
shajder: Done.

test_conformance/api/test_queries.cpp
shajder: Corrected.

bashbaug (approved):

lakshmih (approved):

bashbaug:

PR #2226, opened 2025-01-13
vulkan: Choose where Shaders are generated

joshqti:
- generate spv files into VULKAN_TEST_RESOURCES directory
- works with change that generates spv at build time


bashbaug (approved):
Merging as discussed in the January 14th memory subgroup.

PR #2230, opened 2025-01-16
Command-buffer queue compatibility test update

EwanC:
Update cl_khr_command_buffer tests to reflect changes from https://github.com/KhronosGroup/OpenCL-Docs/pull/1292

* Moves negative test for `CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR` from command-buffer creation to enqueue.
* Moves negative test for `CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR` from command-buffer creation to enqueue.
* Introduces a negative test for `CL_INVALID_DEVICE` on command-buffer enqueue for new error condition in spec. Although it requires a context to be contain more than 1 device, which I'm not sure if possible in current test framework.
* Introduces a new test that created a command-buffer using a queue without the profiling property set, then enqueues the command-buffer to a queue with the profiling property set.
* Introduces a new test that creates a command-buffer with an in-order queue, enqueued on an out-of-order queue.
* Introduces a new test that creates a command-buffer with an out-of-order queue, enqueued on an in-order queue.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
@@ -252,6 +252,148 @@ struct SubstituteQueueTest : public BasicCommandBufferTest
     clEventWrapper user_event;
 };
 
+// Command-queue substitution tests which handles below cases:
+// * Template param is true - Create a command-buffer with an in-order queue,
+//   and enqueue command-buffer to an out-of-order queue.
+// * Template param is false - Create a command-buffer with an out-of-order
+//   queue, and enqueue command-buffer to an in-order queue.
+template <bool is_ooo_test>
+struct QueueOrderTest : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    QueueOrderTest(cl_device_id device, cl_context context,
+                   cl_command_queue queue)
+        : BasicCommandBufferTest(device, context, queue), ooo_queue(nullptr),
+          ooo_command_buffer(this)
+    {}
+
+    cl_int RecordOutOfOrderCommandBuffer()
+    {
+        cl_sync_point_khr sync_points[2];
+        const cl_int pattern = pattern_pri;
+        cl_int error =
+            clCommandFillBufferKHR(ooo_command_buffer, nullptr, nullptr, in_mem,
+                                   &pattern, sizeof(cl_int), 0, data_size(), 0,
+                                   nullptr, &sync_points[0], nullptr);
+        test_error(error, "clCommandFillBufferKHR failed");
+
+        error = clCommandFillBufferKHR(ooo_command_buffer, nullptr, nullptr,
+                                       out_mem, &overwritten_pattern,
+                                       sizeof(cl_int), 0, data_size(), 0,
+                                       nullptr, &sync_points[1], nullptr);
+        test_error(error, "clCommandFillBufferKHR failed");
+
+        error = clCommandNDRangeKernelKHR(
+            ooo_command_buffer, nullptr, nullptr, kernel, 1, nullptr,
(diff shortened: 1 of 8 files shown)


test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -252,6 +252,146 @@ struct SubstituteQueueTest : public BasicCommandBufferTest
...
+
+        ooo_queue = clCreateCommandQueue(
+            context, device, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &error);
+        test_error(error,
+                   "clCreateCommandQueue with "
+                   "CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE failed");
+
+        ooo_command_buffer =
+            clCreateCommandBufferKHR(1, &ooo_queue, nullptr, &error);
+        return BasicCommandBufferTest::SetUp(elements);
```
bashbaug, line 373: This is going to call `BasicCommandBufferTest::Setup` twice - once here, and once at the top of the function.  Is this intended?

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
```diff
@@ -252,6 +252,146 @@ struct SubstituteQueueTest : public BasicCommandBufferTest
...
+//   queue, and enqueue command-buffer to an in-order queue.
+template <bool is_ooo_test>
+struct QueueOrderTest : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    QueueOrderTest(cl_device_id device, cl_context context,
+                   cl_command_queue queue)
+        : BasicCommandBufferTest(device, context, queue), ooo_queue(nullptr),
+          ooo_command_buffer(this)
```
bashbaug, line 268: I don't quite understand `ooo_command_buffer(this)`.  Should this initialize to `nullptr` instead?  Even better, can we just rely on the wrapper's default constructor behavior?

```suggestion
          ooo_command_buffer(nullptr)
```

bashbaug (commented):
One additional note - I believe the `negative_enqueue_queue_not_compatible` test is invalid now, and can be removed.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
EwanC: It's a bit convoluted https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/extensions/cl_khr_command_buffer/command_buffer_test_base.h#L87-L91

 The `clCommandBufferWrapper` class can't call the `clRetainCommandBufferKHR` and `clRetainCommandBufferKHR` entrypoints without them being queried from the platform associated with the device, so it takes a pointer to the test instance to get that info from. We do this `command_buffer(this)` initalization too in the `BasicCommandBufferTest` class most tests are derived from, so if we have ideas of changing this it would probably be best tracked in an issue for a standalone PR to address.

bashbaug: Ah, gotcha.  Yeah, if we're going to change this, we should do so in a different PR, agreed.

bashbaug (dismissed):
The new tests LGTM.

I still think the `negative_enqueue_queue_not_compatible` test is invalid now, but would you prefer to remove it in a separate PR?

EwanC:
> The new tests LGTM.
> 
> I still think the `negative_enqueue_queue_not_compatible` test is invalid now, but would you prefer to remove it in a separate PR?

Apologies! I overlooked that bit of feedback. I agree, pushed a new commit to remove that test.

bashbaug (approved):

bashbaug:

PR #2236, opened 2025-01-17
printf: Fix printf 'mixed_format_random'

lakshmih:
%a or %A with printf on MSVC platforms have a default precision of 13, which is in contrast to OpenCL C specification for printf which only allows for exact digits required if precision is not specified.

### DIFF
--- a/test_conformance/printf/test_printf.cpp
+++ b/test_conformance/printf/test_printf.cpp
@@ -247,7 +247,7 @@ cl_program makeMixedFormatPrintfProgram(cl_kernel* kernel_ptr,
     };
 
     std::array<std::vector<std::string>, 2> formats = {
-        { { "%f", "%e", "%g", "%a", "%F", "%E", "%G", "%A" },
+        { { "%f", "%e", "%g", "%13a", "%F", "%E", "%G", "%13A" },
           { "%d", "%i", "%u", "%x", "%o", "%X" } }
     };
     std::vector<char> data_before(2 + genrand_int32(gMTdata) % 8);


lakshmih:
https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#printf – OpenCL C Spec

`a,A A double, halfn, floatn or doublen argument representing a floating-point number is converted in the style [-]0xh.hhhh p±d, where there is one hexadecimal digit (which is nonzero if the argument is a normalized floating-point number and is otherwise unspecified) before the decimal-point character [[77](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#_footnotedef_77)] and the number of hexadecimal digits after it is equal to the precision; if the precision is missing, then the precision is sufficient for an exact representation of the value; if the precision is zero and the # flag is not specified, no decimal point character appears. The letters abcdef are used for a conversion and the letters ABCDEF for A conversion. The A conversion specifier produces a number with X and P instead of x and p. The exponent always contains at least one digit, and only as many more digits as necessary to represent the decimal exponent of 2. If the value is zero, the exponent is zero. A double, halfn, floatn or doublen argument representing an infinity or NaN is converted in the style of an f or F conversion specifier.`

https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions?view=msvc-170#precision – Microsoft


`a, A | The precision specifies the number of digits after the point. | Default precision is 13. If precision is 0, no decimal point is printed unless the # flag is used.
`

https://man.openbsd.org/printf - OpenBSD
`
The argument is printed in style [-]0xh.hhhp[±]d where there is one digit before the hexadecimal point and the number after is equal to the precision specification for the argument. When the precision is missing, enough digits are produced to convey the argument's exact double-precision floating-point representation.`

bashbaug (approved):

bashbaug:

PR #2237, opened 2025-01-20
Added object comparability verification for program devices query

### LINKED ISSUE #1485: add test for object comparability

shajder:
Fixes #1485 according to work plan from issue description.


bashbaug (changes requested):
Can you please port the improvements we made to https://github.com/KhronosGroup/OpenCL-CTS/pull/2176 to this PR also?

shajder:
> Can you please port the improvements we made to #2176 to this PR also?

Done. BTW thanks for the hint regarding the "Intel CPU OpenCL implementation" - I just tested above corrections and it works just fine!

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -422,6 +434,134 @@ int test_get_program_info(cl_device_id deviceID, cl_context context, cl_command_
...
+    }
+
+    /* Create a multi device context */
+    clContextWrapper multi_device_context = clCreateContext(
+        nullptr, (cl_uint)num_devices, scope_guard->sub_devices.data(), nullptr,
+        nullptr, &err);
+    test_error_ret(err, "Unable to create testing context",
+                   TEST_SKIPPED_ITSELF);
+
+    cl_program program = nullptr;
```
bashbaug, line 521:
```suggestion
    clProgramWrapper program = nullptr;
```

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -346,12 +351,19 @@ int test_get_program_info(cl_device_id deviceID, cl_context context, cl_command_
...
+
+    /* Since the device IDs are opaque types we check the CL_DEVICE_VENDOR_ID
+     * which is unique for identical hardware. */
+    cl_uint device1_vid, deviceID_vid;
+    error = clGetDeviceInfo(device1, CL_DEVICE_VENDOR_ID, sizeof(device1_vid),
+                            &device1_vid, NULL);
+    test_error(error, "Unable to get device CL_DEVICE_VENDOR_ID");
+    error = clGetDeviceInfo(deviceID, CL_DEVICE_VENDOR_ID, sizeof(deviceID_vid),
+                            &deviceID_vid, NULL);
+    test_error(error, "Unable to get device CL_DEVICE_VENDOR_ID");
```
bashbaug, line 366: It isn't wrong to continue to check the device vendor IDs here, but I also don't think it's needed anymore, since we're checking that the device IDs themselves are equivalent instead.

shajder: Removed.

test_conformance/compiler/test_build_helpers.cpp

bashbaug (approved):
Thanks, LGTM!

bashbaug:

PR #2239, opened 2025-01-22
fix profiling execute_multipass

### LINKED ISSUE #2238: profiling execute_multipass issues
That test is failing using `clvk` and various Vulkan drivers (`llvmpipe`, `swiftshader`, mesa-based drivers).

I am seeing multiple issues in the test itself:

1.  https://github.com/KhronosGroup/OpenCL-CTS/blob/5b3518096ca7b82854daadb2b2fae704fe2d9cb5/test_conformance/profiling/execute_multipass.cpp#L107
`localThreads` type is `size_t[3]`. Thus the size given is not right, and we proper implementation should end up with undefined values in `localThreads[1]` & `localThreads[2]` at least. Most probably also `localThreads[0]` as `cl_uint` might be smaller than `size_t`...

2. each element of `localThreads` is clamp to its max value, except for `localThreads[2]`. I'm not sure I understand why, and would also clamp it

3. We end up with a workgroup using the maximum work-item sizes per dimensions (`CL_DEVICE_MAX_WORK_ITEM_SIZES`). This can be higher than `CL_DEVICE_MAX_WORK_GROUP_SIZE` on some devices, thus we would also need to make sure not to get higher than it.

4. The input and output buffer are declared as `cl_uchar[w * h * d * nChannels]`. But
4.1. https://github.com/KhronosGroup/OpenCL-CTS/blob/5b3518096ca7b82854daadb2b2fae704fe2d9cb5/test_conformance/profiling/execute_multipass.cpp#L136
The output is created with a bigger size (`*sizeof(cl_float)`)
4.2. https://github.com/KhronosGroup/OpenCL-CTS/blob/5b3518096ca7b82854daadb2b2fae704fe2d9cb5/test_conformance/profiling/execute_multipass.cpp#L240
The output is read with a bigger size (`*4`)

This is trigger segfault because the host buffer where the output should be stored is smaller than the CL buffer.

rjodinchr:
- fix clGetDeviceInfo(CL_DEVICE_MAX_WORK_ITEM_SIZES) by using the proper size

- clamp localThreads[2] as for localThreads[0] and localThreads[2]

- clamp all localThreads elements in regard of CL_MAX_WORK_GROUP_SIZE

- fix the size using to create/read the output buffer

Fix #2238


kpet (approved):

bashbaug (approved):
Works for me.

I was initially confused by the allocation, because it's allocating based on `nChannels` even though the kernel unconditionally writes four channels, but I guess it's OK because because we only ever call this function when `nChannels` is four.

bashbaug:
@rjodinchr can you please fix the merge conflict?

test_conformance/profiling/execute_multipass.cpp
```diff
@@ -131,9 +144,9 @@ static int run_kernel( cl_device_id device, cl_context context, cl_command_queue
     }
```
lakshmih, line 144: Minor, can this be fixed to say 3D.

lakshmih (approved):

bashbaug:

PR #2241, opened 2025-01-22
fix profiling profiling_timebase

### LINKED ISSUE #2240: profiling profiling_timebase using NULL as global_work_size in clEnqueueNDRangeKernel
`profiling profiling_timebase` is using NULL as `global_work_size` in `clEnqueueNDRange Kernel`:
https://github.com/KhronosGroup/OpenCL-CTS/blob/5b3518096ca7b82854daadb2b2fae704fe2d9cb5/test_conformance/profiling/profiling_timebase.cpp#L54

I don't think this is allowed reading the specification.

rjodinchr:
Use explicit work_group_size of 1

Fix #2240

### DIFF
--- a/test_conformance/profiling/profiling_timebase.cpp
+++ b/test_conformance/profiling/profiling_timebase.cpp
@@ -51,8 +51,9 @@ REGISTER_TEST(profiling_timebase)
     err = clGetDeviceAndHostTimer(device, &deviceTimeBeforeQueue, &hostTime);
     test_error(err, "Unable to get starting device time");
 
-    err = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, NULL, NULL, 1, &uEvent,
-                                 &kEvent);
+    size_t global_size = 1;
+    err = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, &global_size, NULL, 1,
+                                 &uEvent, &kEvent);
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
I'm OK with this change since I don't think it's an important part of the test to have a `global_work_size` equal to `NULL`, but I do believe this part of the test is correct, and a `global_work_size` equal to `NULL` is valid - at least for OpenCL 2.1 and newer.  I've filed a few issues to get this clarified, and to ensure that the case where the `global_work_size` is equal to `NULL` is tested:

https://github.com/KhronosGroup/OpenCL-Docs/issues/1302
https://github.com/KhronosGroup/OpenCL-CTS/issues/2244

Note, `clGetDeviceAndHostTimer` also requires OpenCL 2.1, and there's no check for OpenCL 2.1 or newer (that I can see at least), so this really ought to be added.  If you'd prefer to add this check in a separate PR, would you mind filing an issue so it doesn't slip through the cracks?  I suspect it will be less work to just add the check, though 😄

rjodinchr:
Alright, I missed that part of the spec. Honestly having something about `global_work_size` being `NULL` not in the `global_work_size` section but the `work_dim` section is really easy to miss.

In that case, I think we should not land that PR. I'll fix `clvk` according to the spec.

Issue #2234, opened 2025-01-17
Building for AArch64 fails due to "strict-aliasing" ,  "maybe-uninitialized" & "stringop-truncation" errors


Antonios-C:
When building the CTS repo for a AArch64 platform, the build failed due to 3 types of warnings, that turned into errors (-Werror enablement in CMakeLists.txt). 

Example "strict-aliasing":
```
../test_conformance/basic/test_numeric_constants.cpp:529:154: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
  529 | #define TEST_FLOAT_ASSERTION( a, msg, f ) if( !( a ) ) { log_error( "ERROR: Float constant failed requirement: %s (bitwise value is 0x%8.8x)\n", msg, *( (uint32_t *)&f ) ); return -1; }
      |                                                                                                                                                        ~~^~~~~~~~~~~~~~~~
../test_conformance/basic/test_numeric_constants.cpp:574:9: note: in expansion of macro 'TEST_FLOAT_ASSERTION'
  574 |         TEST_FLOAT_ASSERTION( intOut[0] == 0, "isinf( MAXFLOAT ) = false", floatOut[0] )
      |         ^~~~~~~~~~~~~~~~~~~~

```
Example "stringop-truncation":
```
../test_conformance/contractions/contractions.cpp:368:16: error: 'char* strncpy(char*, const char*, size_t)' specified bound 4096 equals destination size [-Werror=stringop-truncation]
  368 |         strncpy( appName, argv[0], MAXPATHLEN );
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../test_conformance/contractions/contractions.cpp:373:16: error: 'char* strncpy(char*, const char*, size_t)' specified bound 4096 equals destination size [-Werror=stringop-truncation]
  373 |         strncpy( baseName, argv[0], MAXPATHLEN );
      |         ~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Examples "maybe-uninitialized": 

- ```../test_common/harness/imageHelpers.cpp:3564:56: error: 'src_lod' may be used uninitialized [-Werror=maybe-uninitialized]```
- ```../test_common/harness/imageHelpers.cpp:3610:56: error: 'dst_lod' may be used uninitialized [-Werror=maybe-uninitialized]```
[...]

PR #2242, opened 2025-01-22
Fix Build Warnings for AArch64

Antonios-C:
This commit links to issue (#2234).

When cross-compiling for AArch64, using gcc 13.3, you encounter three warnings types that turn into errors:

- maybe-uninitialized
- stringop-truncation
- strict-aliasing

This commit fixes all the warnings found, in regards to the first two rules. To resolve the warnigns due to strict-aliasing, I am editing the CMake build system.

Signed-off-by: Antonios Christidis <a-christidis@ti.com>


test_conformance/images/clCopyImage/test_loops.cpp
```diff
@@ -111,6 +111,11 @@ int test_image_type( cl_device_id device, cl_context context, cl_command_queue q
             name = "1D buffer -> 1D";
             imageType = CL_MEM_OBJECT_IMAGE1D_BUFFER;
             break;
+        default:
```
svenvh, line 114: This switch is a bit different from the other switches touched by this PR.  Since `testMethod` is an enum, I'd be wary of adding a `default`: not having a default will allow us to catch missing `case`s at compile time when extending the enum.

svenvh (commented):
Thanks for going through all of these!

test_conformance/images/clCopyImage/test_loops.cpp
Antonios-C: I agree, thank you for the feedback! Just pushed a commit with a possible change. Setting name = nullptr and imageType  = 0 outside the switch-case construct. Let me know what you think.

svenvh (approved):

Antonios-C:
@svenvh What is the process for getting pull requests merged after they have been approved ?

PR #2243, opened 2025-01-23
Update semaphore synchronisation used in Vulkan interop tests

gorazd-sumkovski-arm:


test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -1146,6 +1152,14 @@ int clExternalExportableSemaphore::signal(cl_command_queue cmd_queue)
 
     if (m_externalHandleType == VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SYNC_FD)
     {
+        // Finish the queue to signal the semaphore
+        err = clFinish(cmd_queue);
```
joshqti, line 1156: Can you explain why this is needed?

joshqti: This shouldn't be needed.  The OpenCL spec states that after clEnqueueSignalSemaphoreKHR returns, it is safe to wait.  The implication is that it is also safe to import into vulkan, the finish is not necessary here, and defeats the purpose of semaphores.

test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -328,6 +331,19 @@ int run_test_with_two_queue(
...
                 {
+                    if (VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SYNC_FD
+                        == vkExternalSemaphoreHandleType)
+                    {
+                        // Due to the finish that will happen on cmd_queue2 in
+                        // clCl2VkExternalSemaphore->signal(cmd_queue2),
+                        // cmd_queue1 needs to be finished now because
+                        // cmd_queue2 depends on first_launch which is
+                        // completed in cmd_queue1.
+                        err = clFinish(cmd_queue1);
```
joshqti, line 342: I do not think this finish is required, semaphores can operate asynchronously.  I could be convinced a flush may be needed, but I don't think finish is correct.

This change adds a lot of clFinishes.  I think this makes the test synchronous and incorrect.

kpet:
Discussed in 2025/02/25: Qcom like the refactoring but are questioning the addition of many calls to `clFinish`. Could we split the PR into two (pushing the addition of calls to `clFinish` to the second PR)?

gorazd-sumkovski-arm:
Created #2300 for the refactoring patch. Will rebase this once that's merged.

bashbaug:
Removing "focused review" temporarily.  We will re-add after #2300 is merged.

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
gorazd-sumkovski-arm: Apologies for not being able to respond sooner. There are three different actions here: the enqueuing of a semaphore command, the execution of that semaphore command and the side effects caused by the call to `clGetSemaphoreHandleForTypeKHR()`  and the order in which they occur is paramount. We have a loop over `mtIdx` and an inner loop over `iter`, this is the sequence of events:
`mtIdx = 0`
&nbsp; &nbsp; `iter = 0`
&nbsp; &nbsp; &nbsp; &nbsp; Signal on `clCl2VkExternalSemaphore()` is enqueued.
&nbsp; &nbsp; &nbsp; &nbsp; Call to `clGetSemaphoreHandleForTypeKHR()`  acts a wait.
&nbsp; &nbsp; `iter = 1`
&nbsp; &nbsp; &nbsp; &nbsp; Signal on `clCl2VkExternalSemaphore()` is enqueued.
&nbsp; &nbsp; &nbsp; &nbsp; Call to `clGetSemaphoreHandleForTypeKHR()` acts a wait.
    ...
`mtIdx = 1`
&nbsp; Queue is finished, so signals are executed.
&nbsp; &nbsp; `iter = 0`
&nbsp; &nbsp; &nbsp; &nbsp; Signal on `clCl2VkExternalSemaphore()` is enqueued.
        ...

There are three potential problems with this:
1. In the `iter` loop, the signals are enqueued but not executed, so the calls to `clGetSemaphoreHandleForTypeKHR()` can potentially be seen as consecutive waits.
2. When the queue is finished in the second iteration of the `mtIdx` loop, this executes multiple consecutive signals, which were enqueued in the `iter` loop.
3. After the queue is finished the semaphore is in the signalled state, and then a signal is enqueued on that semaphore, with no apparent wait from anywhere.

The problem is that the enqueue of a signal command places that command in a queue to be executed later, but the call to `clGetSemaphoreHandleForTypeKHR()` does not enqueue "wait side effects" next in the queue, rather this happens in-place, when the function is called. Since there is no way to move the "wait side effects" of `clGetSemaphoreHandleForTypeKHR()` into the queue, I have forced the actions in the correct order with a call to `clFinish()`. Replacing this with `clFlush()` will not guarantee the correct order. I'm not sure if anything better can be done for sync_fd semaphores.


> I'm not sure I follow. It looks like after Vulkan submits, we wait:
>
> [err = clVk2CLExternalSemaphore->wait(cmd_queue1);](https://github.com/KhronosGroup/OpenCL-CTS/blob/fef08680298a1cf9fcbe7a6caf0aeccfaa6b9ff6/test_conformance/vulkan/test_vulkan_interop_buffer.cpp#L266C21-L266C70)
>
>Followed by a signal.
>
>https://github.com/KhronosGroup/OpenCL-CTS/blob/fef08680298a1cf9fcbe7a6caf0aeccfaa6b9ff6/test_conformance/vulkan/test_vulkan_interop_buffer.cpp#L347

clVk2CLExternalSemaphore and clCl2VkExternalSemaphore are two different semaphores which have no relation. An operation on one does not affect the other and neither's payload is (or can be) transferred to the other.

joshqti: > 1. In the `iter` loop, the signals are enqueued but not executed, so the calls to `clGetSemaphoreHandleForTypeKHR()` can potentially be seen as consecutive waits.

This is incorrect.  The spec guarantees that once clEnqueueSignalSemaphoreKHR returns, the semaphore is safe-to-wait.  The spec allows the implementation to block if necessary to achieve this.  Getting the handle is not a wait, but it puts the semaphore in a safe-to-signal state.  The purpose is to allow signal to generate a valid pending signal on the sync_fd, extract the sync fd, and then submit a Vulkan workload that depends on the imported sync_fd, which has a valid pending signal.   When the command goes to CL_COMPLETE is when the signaling actually happens.  

Immediately after extracting the sync_fd, the semaphore is in a safe-tosignal state, so reimport is immediately possible.

PR #2246, opened 2025-01-24
Added support to run the CTS with Vulkan validation layers

### LINKED ISSUE #2116: Flow to run the CTS with Vulkan validation layers
The CTS uses Vulkan for some interop tests. We have seen a few bugs (e.g. part of #1890) that could have been prevented by using the Vulkan validation layers. We should make it easy to run the CTS suites that use Vulkan with the Vulkan validation layers and document the process.

shajder:
Fixes #2116 according to task description.


test_conformance/vulkan/main.cpp
```diff
@@ -97,35 +99,47 @@ bool isDeviceSelection(const char *arg)
...
             {
                 useSingleImageKernel = true;
+                argsRemoveNum = 1;
             }
             if (!strcmp(argv[i], "--useDeviceLocal"))
             {
                 useDeviceLocal = true;
+                argsRemoveNum = 1;
+            }
+            if (!strcmp(argv[i], "--useValidationLayers"))
```
bashbaug, line 128: We should add this new command line argument to `printUsage`.

Actually, it looks like a bunch of command line arguments are missing in `printUsage`, and a few that are described are not actually command line arguments...

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -1191,14 +1191,15 @@ cl_external_memory_handle_type_khr vkToOpenCLExternalMemoryHandleType(
 {
     switch (vkExternalMemoryHandleType)
     {
+        default:
```
bashbaug, line 1194: We may want to do something (like log an error message?) in the default case, since this is almost certainly a bug that should be fixed.

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
```diff
@@ -130,6 +152,25 @@ VulkanInstance::VulkanInstance(): m_vkInstance(VK_NULL_HANDLE)
...
+
+        bool found = false;
+        for (const char *layerName : m_validationLayers)
+            for (const auto &layerProps : layers)
+                if (strcmp(layerName, layerProps.layerName) == 0)
+                {
+                    found = true;
+                    break;
+                }
+        m_useValidationLayers = found;
```
bashbaug, line 171: This will enable the validation layers if _any_ of the requested validation layers are found, even if _some_ of the requested validation layers are not found.  Is this the behavior that we want, or do we only want to enable validation layers if _all_ of the requested validation layers are found?

Actually, since it looks like we use all validation layers when we build our VkInstanceCreateInfo, we should definitely be checking that all validation layers are found, or we need to ensure we only enable the validation layers that were found.

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
```diff
@@ -174,18 +218,57 @@ VulkanInstance::VulkanInstance(): m_vkInstance(VK_NULL_HANDLE)
...
+        // VK_DEBUG_UTILS_MESSAGE_SEVERITY_INFO_BIT_EXT
+        // VK_DEBUG_UTILS_MESSAGE_SEVERITY_WARNING_BIT_EXT
+        // VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT
+        debugCreateInfo.messageSeverity |=
+            VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT;
+        debugCreateInfo.sType =
+            VK_STRUCTURE_TYPE_DEBUG_UTILS_MESSENGER_CREATE_INFO_EXT;
+        debugCreateInfo.messageSeverity =
+            VK_DEBUG_UTILS_MESSAGE_SEVERITY_WARNING_BIT_EXT
+            | VK_DEBUG_UTILS_MESSAGE_SEVERITY_ERROR_BIT_EXT;
```
bashbaug, line 253: We should only set `debugCreateInfo.messageSeverity` once.

bashbaug (changes requested):
Note, there is also a merge conflict that needs to be resolved.

Noting for completeness: There are quite a few validation layer erros when the validation layers are enabled, but we can fix those in a separate PR.

test_conformance/vulkan/main.cpp
shajder: Corrected. @nikhiljnv I've complement missing descriptions but I will appreciate your assistance, it looks like you are the author. Thanks

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
shajder: Added log.

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
shajder: Corrected

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
shajder: Done, thanks!

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
```diff
@@ -130,6 +152,32 @@ VulkanInstance::VulkanInstance(): m_vkInstance(VK_NULL_HANDLE)
...
+        {
+            bool found = false;
+            for (const auto &layerProps : layers)
+                if (strcmp(*it, layerProps.layerName) == 0)
+                {
+                    found = true;
+                    break;
+                }
+            if (!found)
+                it = m_validationLayers.erase(it);
```
bashbaug, line 174: Nice-to-have: We may want to print a message here if a validation layer is not found vs. silently removing it.

bashbaug (dismissed):

test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
shajder: Done

lakshmih (approved):

PR #2248, opened 2025-01-26
fix negative_create_command_buffer_not_supported_properties test

### LINKED ISSUE #2247: command buffer negative_create_command_buffer_not_supported_properties is incorrect
I don't believe that the command buffer `negative_create_command_buffer_not_supported_properties` test is correct.

This test is trying to test this specific error code:

> CL_INVALID_PROPERTY if values specified in properties are valid but are not supported by all the devices associated with command-queues in queues.

It does so by trying to find a "valid but unsupported property", specifically:

```c++
        bool skip = true;
        if (!simultaneous_use_support)
        {
            unsupported_prop = CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR;
            skip = false;
        }
        else if (!device_side_enqueue_support)
        {
            unsupported_prop = CL_COMMAND_BUFFER_DEVICE_SIDE_SYNC_KHR;
            skip = false;
        }
```

Note that `device_side_enqueue_support` was queried using:

```c++
    cl_device_command_buffer_capabilities_khr capabilities;
    error = clGetDeviceInfo(device, CL_DEVICE_COMMAND_BUFFER_CAPABILITIES_KHR,
                            sizeof(capabilities), &capabilities, NULL);
    // snip
    device_side_enqueue_support =
        (capabilities & CL_COMMAND_BUFFER_CAPABILITY_DEVICE_SIDE_ENQUEUE_KHR)
        != 0;
```

There are two problems here:

1. For implementations that do not support `cl_khr_command_buffer_multi_device`, the `CL_COMMAND_BUFFER_DEVICE_SIDE_SYNC_KHR` enum does not exist.  This means that this property is "not valid" rather than "not supported", therefore the error code should be `CL_INVALID_VALUE` rather than `CL_INVALID_PROPERTY`.
2. The `device_side_enqueue_support` capability is unrelated to `CL_COMMAND_BUFFER_DEVICE_SIDE_SYNC_KHR`.

Recommended fixes are:

1. Remove the codepath that tests the "unsupported property" `CL_COMMAND_BUFFER_DEVICE_SIDE_SYNC_KHR`.  This means that the test will skip when simultaneous use is supported, because the only valid property is supported.
2. Revisit all of the cases that check `device_side_enqueue_support` to be sure this was the intended check.

bashbaug:
fixes #2247 

* For the `negative_create_command_buffer_not_supported_properties` test, the only property we can check for is simultaneous use.  All other properties are part of other extensions and hence will generate `CL_INVALID_VALUE`, not `CL_INVALID_PROPERTY`.
* Checks whether the `cl_khr_command_buffer_multi_device` extension is supported when using `CL_COMMAND_BUFFER_DEVICE_SIDE_SYNC_KHR`, instead of `device_side_enqueue_support`.
* If the `cl_khr_command_buffer_multi_device` extension is NOT supported and the `CL_COMMAND_BUFFER_DEVICE_SIDE_SYNC_KHR` command buffer creation flag is used, the expected error code is `CL_INVALID_VALUE`, not `CL_INVALID_PROPERTY`.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
@@ -225,7 +225,7 @@ struct MultiFlagCreationTest : public BasicCommandBufferTest
         }
 
         // If we can't find multiple supported flags, still set a bitfield but
-        // expect CL_INVALID_PROPERTY to be returned on creation.
+        // expect CL_INVALID_VALUE to be returned on creation.
         if (num_flags_set < 2)
         {
             flags = CL_COMMAND_BUFFER_SIMULTANEOUS_USE_KHR
@@ -246,8 +246,8 @@ struct MultiFlagCreationTest : public BasicCommandBufferTest
         else
         {
             test_failure_error_ret(
-                error, CL_INVALID_PROPERTY,
-                "clCreateCommandBufferKHR should return CL_INVALID_PROPERTY",
+                error, CL_INVALID_VALUE,
+                "clCreateCommandBufferKHR should return CL_INVALID_VALUE",
                 TEST_FAIL);
         }
 
--- a/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
@@ -210,7 +210,8 @@ struct MultiFlagCreationTest : public BasicCommandBufferTest
             num_flags_set++;
         }
 
-        if (device_side_enqueue_support)
+        if (is_extension_available(
+                device, CL_KHR_COMMAND_BUFFER_MULTI_DEVICE_EXTENSION_NAME))
         {
             flags |= CL_COMMAND_BUFFER_DEVICE_SIDE_SYNC_KHR;
             num_flags_set++;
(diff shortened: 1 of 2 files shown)


test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -224,7 +225,7 @@ struct MultiFlagCreationTest : public BasicCommandBufferTest
         }
 
         // If we can't find multiple supported flags, still set a bitfield but
-        // expect CL_INVALID_PROPERTY to be returned on creation.
+        // expect CL_INVALID_VALUE to be returned on creation.
```
EwanC, line 228: Looking at this test again, I'm thinking that now we have specific negative tests for command-buffer creation properties it might be better to remove the negative testing component from this test and just try to create a command-buffer with as many flags as are valid. In theory that shouldn't reduce test coverage, but simplify this test.

bashbaug: This makes sense to me, and I don't mind making the change in this PR if we all agree.

To be clear, what this will look like is: instead of having this special code block checking for `num_flags_set < 2`, we will just find as many flags as we can set, and try to create a command buffer with those flags.  This command buffer creation must succeed, so no `CL_INVALID_VALUE` or `CL_INVALID_PROPERTY` check will be needed.

bashbaug: I've made this change and this test is much simpler now - please take a look.  Thanks!

PR #2251, opened 2025-01-28
Enabled tgamma test for bruteforce suite

### LINKED ISSUE #507: bruteforce tests missing functions on master
It seems some built-in functions that were tested with the version on cl12_trunk are no longer tested with the master branch version. Here are a few I could spot:
- erf
- erfc
- reciprocal
- tgamma

We should review the full list against the spec.

shajder:
Fixes #507 according to work plan from issue description.


test_conformance/math_brute_force/reference_math.cpp
```diff
@@ -2111,6 +2111,8 @@ int reference_not(double x)
     return r;
 }
 
+double reference_tgamma(double x) { return tgamma(x); }
```
svenvh, line 2114: I am not sure if we can rely on libc's tgamma for the reference computation?  [Known Maximum Errors in Math Functions](https://www.gnu.org/software/libc/manual/html_node/Errors-in-Math-Functions.html) seems to indicate that most tgamma libc implementations have a rather large ULP value.

shajder: Indeed, thanks! Computing reference values for `tgamma` seems challenging. As far as I understand, each implementation is subject to non-zero error. Please correct me if I am wrong at any point, I see few directions to follow:

- pre-computed high-precision values taken from `MatLab` or similar software. This approach will definitely impact the source code with a separate path for `tgamma` processing.
- limiting tgamma arguments for integral values which will result in zero error outcome. As with the previous proposition, this approach will result as a separate path for `tgamma` processing.
- implementing a procedure to compute `tgamma` with maximum known-error. According to [this paper](https://inria.hal.science/hal-03141101/) OpenLibm contains [tgamma implementation](https://github.com/JuliaMath/openlibm/blob/1eeb1395b4014884608b1f37d9653c1efe4dffce/ld80/e_tgammal.c#L217-L218) with error<1ULP. Moreover, license looks friendly for other open source projects. Then we could take into account the known-error by enlarging the error from [table 65](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#relative-error-as-ulps).

I will appreciate discussion over this subject, thanks!

lakshmih:
Still reviewing/testing this. Thanks.

test_conformance/math_brute_force/reference_math.cpp
bashbaug: @svenvh is this still an issue or is the OpenLibm implementation good enough for a reference?

test_conformance/math_brute_force/unary_double.cpp
```diff
@@ -398,6 +411,14 @@ int TestFunc_Double_Double(const Func *f, MTdata d, bool relaxedMode)
         }
     }
 
+    test_info.isRangeLimited = 0;
+    test_info.tgamma_arg_limit = 0.f;
+    if (0 == strcmp(f->name, "tgamma"))
+    {
+        test_info.isRangeLimited = 1;
```
svenvh: Why the argument range restriction?  There is no such restriction in the OpenCL C Specification as far as I can tell, so I believe the CTS should just test for infinities for too large inputs.

test_conformance/math_brute_force/reference_math.cpp
svenvh: Thanks for the ping and apologies for the late response!

If we are comfortable incorporating the OpenLibm tgamma implementation then that looks like a good solution to me (but I'm not a lawyer, so this deserves some scrutiny).

test_conformance/math_brute_force/unary_double.cpp
shajder: We could either limit argument range or handle special cases while comparing test and reference value. This solution seemed more decent since tgamma works stable is specific ranges but if you think it is better to handle huge values while testing I can apply the correction, thanks.

lakshmih:
Need more time before this gets merged, still evaluating.

paulfradgley:
Our implementation won't pass this test. We'll need some time to put some compiler engineers onto fixing it.

lakshmih:
We’re now able to compute the correct reference values. However, we had a question regarding the required precision for the host reference:

Is 64-bit precision sufficient for this use case, or is 128-bit precision necessary? We’ve noticed that using 128-bit significantly increases execution time, so we’d like to confirm whether that level of precision is truly required.

shajder:
> We’re now able to compute the correct reference values. However, we had a question regarding the required precision for the host reference:
> 
> Is 64-bit precision sufficient for this use case, or is 128-bit precision necessary? We’ve noticed that using 128-bit significantly increases execution time, so we’d like to confirm whether that level of precision is truly required.

My perspective:

1) The 128-bit version of the reference function was significantly easier to port - 230 vs. 878 lines of code.

2) The assumption was that the 128-bit version offers higher precision, which seems desirable.

shajder:
> We’ve noticed that using 128-bit significantly increases execution time, so we’d like to confirm whether that level of precision is truly required.

@lakshmih as discussed in teleconference here is [draft PR](https://github.com/KhronosGroup/OpenCL-CTS/pull/2572) for comparison purpose

lakshmih (approved):

test_conformance/math_brute_force/unary_double.cpp
bashbaug: Sorry for the late comment, but I think I have the same question as Sven:  Is there some reason we are only testing tgamma inputs for this limited range?  Where did the limited range come from?

I agree that values outside of this range will probably result in infinities, but isn't that OK?  Do we need this special case handling?

shajder: @bashbaug For large `tgamma` arguments, both `Intel` and `NVIDIA` `OpenCL` implementations return `NaN`.
According to `C99`, an overflow in `tgamma` should result in `HUGE_VAL`, which maps to +`INFINITY` on `IEEE-754` systems.

I couldn’t find any `OpenCL` specification text that explicitly defines the return value for overflow in `tgamma`. The spec appears not to require vendors to detect overflow before intermediate computations produce `NaN`. The only relevant passage I found is:

> The accuracy (and in some cases the input range(s)) of these functions is implementation-defined.

This seems to leave room for implementations where `NaN` is returned instead of +`INFINITY` for large arguments.

Given this ambiguity, I’ve limited the supported parameter range to avoid these edge cases. Please let me know if you think we should instead accept `NaN` results; however, that would require introducing `tgamma`-specific handling in `unary_float.cpp`.

bashbaug: We should figure out what to do here before merging.

shajder: Okay, I reviewed the spec and the test results once again. Unless I’m missing something obvious, the reference behavior looks correct (infinity):

> ERROR: tgamma: nan ulp error at -0x1.ffp+127 (0xff7f8000): *inf vs. nan

Based on this, I removed the input range limitation. Let’s wait for the feedback.

bashbaug: Hmm, I think something still might be wrong here.

Looking at the output above for the float value with bits 0xff7f8000, this represents the float value -339617750000000000000000000000000000000, so it is a very large negative integer.  I'm not a tgamma expert, but all of the documents I've seen about it say something to the effect of:

> If the argument is a negative integer, NaN is returned and FE_INVALID is raised.

This particular text was taken from: https://en.cppreference.com/w/cpp/numeric/math/tgamma.html

Since the input is a negative integer, shouldn't the reference result be NaN, not INF?

shajder: It is because the new reference function uses a `long double` parameter. In this specific case, the value is no longer treated as a large integer - `long double` has enough precision to represent fractional values at the 10^38 scale. Indeed, the reference function should take this into account. I will work on handling such cases properly.

shajder: @bashbaug I just pushed the correction which handles two special cases not supported by `OpenLibM` implementation. I can still see one problem with `FP64` emulation implementation (`IGC_EnableDPEmulation`, `OverrideDefaultFP64Settings`) but in this case I believe reference function returns a correct value:

`ERROR: tgamma: inf ulp error at -0x1.f800000000001p+51 (0xc32f800000000001): *-0x0.0000000000000p+0 vs. inf`

-0x1.f800000000001p+51 which is -4.4332308831928325e+15, a negative half-integer should make denominator of tgamma equation to overflow to `+INF` causing the result to collapse to `-0`. The device returning `INF` is incorrect.

bashbaug: Interesting, I'm able to reproduce this problem on our GPU devices that support fp64 natively, also.  The issue seems to be based on our tgamma implementation, which is based on the libclc implementation.  Interestingly, the libclc implementation has a special case specifically for when the overflow occurs, see:

https://github.com/llvm/llvm-project/blob/6891a6ef0e2466ffb8a921d7f58934ef464f137a/libclc/clc/lib/generic/math/clc_tgamma.cl#L48

If I disable this special case, I can get a little further, but there are still some failures.

The tests seem to be correct now, and we're passing with our CPU devices that support fp64, so I think we're OK to merge now.  We'll need to fix our GPU fp64 tgamma implementation though, and we ought to fix the libclc implementation while we're at it.

lakshmih (approved):

bashbaug:
Merging as discussed in the March 31st teleconference.

PR #2256, opened 2025-01-31
Fix profiling execute_multipass failure with segmentation fault

jujiang-del:
The buffer size input to the test function clEnqueueReadBuffer was incorrect, which cause segmentation  fault. And it didn't match the size for the host allocation outptr


svenvh (approved):

gwawiork (approved):
LGTM

svenvh:
Merging immediately as this is non-controversial and has 2 approvals.

PR #2258, opened 2025-02-02
add SPIR-V testing for integer dot products

bashbaug:
This PR adds testing for integer dot products via SPIR-V rather than via OpenCL C.  The overall logic for this test largely follows the same testing logic as the `test_integer_ops` test.

The test checks SPIR-V modules using both the SPV_KHR_integer_dot_product extension (works with any SPIR-V version) and SPIR-V 1.6, without the SPV_KHR_integer_dot_product extension.

### DIFF
--- a/test_conformance/spirv_new/CMakeLists.txt
+++ b/test_conformance/spirv_new/CMakeLists.txt
@@ -6,6 +6,7 @@ set(${MODULE_NAME}_SOURCES
   test_cl_khr_expect_assume.cpp
   test_decorate.cpp
   test_get_program_il.cpp
+  test_integer_dot_product.cpp
   test_linkage.cpp
   test_no_integer_wrap_decoration.cpp
   test_op_atomic.cpp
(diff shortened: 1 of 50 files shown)


bashbaug:
Note: The majority of the changes are all of the different permutations of SPIR-V files:

* There are two different combinations of SPIR-V files that are supported: SPV_KHR_integer_dot_product and SPIR-V 1.6.
* There are three different type combinations that are supported: `U`, `S`, and `SU`.
* There are two different opcodes that are tested: OpXDot and OpXDotAccSat.
* There are two different operand types that are supported: 4 x 8-bit vectors and packed 32-bit integers.
* There are two different bit widths that are supported: 64-bit and 32-bit.

Right now, each permutation is in a separate file, so there are a total of 2 x 3 x 2 x 2 x 2 = 48 different SPIR-V files.

I want a separate kernel for each test scenario, but I can try to consolidate similar variants into the same file, if desired.  Is this worthwhile?

PR #2259, opened 2025-02-03
Refactor buffer ReadWrite and Copy tests

MichaelRizkalla-arm:
This change refactors the following tests to use RAII to clean-up allocated resources on exit, and adds additional changes as mentioned below:

- test_arrayreadwrite
    - Allow different `cl_mem_flags` to be passed to the test.
- test_bufferreadwriterect:
    - Allow different `cl_mem_flags` to be passed to the test.
    - Customisable copy, read and write functions.
- test_buffer_copy
    - Fill the destination buffer with `invalid_ptr` instead of `out_ptr` if created with `CL_MEM_(USE/COPY)_HOST_PTR`.
- test_buffer_partial_copy
    - Fill the destination buffer with `invalid_ptr` instead of `out_ptr` if created with `CL_MEM_(USE/COPY)_HOST_PTR`.


test_conformance/buffers/test_buffer_copy.cpp
```diff
@@ -37,210 +38,250 @@ static int verify_copy_buffer(int *inptr, int *outptr, int n)
...
+        {
             log_info("Testing with cl_mem_flags src: %s dst: %s\n", flag_set_names[src_flag_id], flag_set_names[dst_flag_id]);
 
-            for (i=0; i<num_elements; i++){
-                int_input_ptr[i] = (int)genrand_int32( d );
-                int_output_ptr[i] = 0xdeaddead; // seed with incorrect data
+            memset(invalid_ptr.get(), 0xdeaddead,
+                   num_elements); // seed with incorrect data
+            memset(out_ptr.get(), 0xdeadbeef,
+                   num_elements); // seed with incorrect data
```
lakshmih, line 89: This code only sets a quarter of the buffer, because memset only writes 1 byte, so the first quarter of invalid_ptr will be filled with 0xad and the rest will be unknown values.

So this would be better included in the for loop that is next:

for (int i = 0; i < num_elements; i++)
{
    invalid_ptr[i] = (int)0xdeaddead;
    out_ptr[i] = (int)0xdeadbeef;
    reference_ptr[i] = (int)genrand_int32(d);
}

MichaelRizkalla-arm: Thanks for catching it.
I have updated the commit.

test_conformance/basic/test_arrayreadwrite.cpp
```diff
@@ -20,34 +20,44 @@
...
     num_elements = 1024 * 1024 * 4;
-    int                 i, j, err;
-    MTdata              d;
+    MTdataHolder d(gRandomSeed);
 
-    inptr = (cl_uint*)malloc(num_elements*sizeof(cl_uint));
-    outptr = (cl_uint*)malloc(num_elements*sizeof(cl_uint));
+    std::vector<cl_uint> reference_ptr(num_elements);
+    std::vector<cl_uint> inptr(num_elements);
+    std::vector<cl_uint> outptr(num_elements);
```
bashbaug, line 40: Nitpick: none of these are "ptrs".  I understand why `inptr` and `outptr` have the same names that they had previously to minimize diffs, but maybe we can use a better name for the new `reference_ptr`?

test_conformance/basic/test_arrayreadwrite.cpp
```diff
@@ -61,15 +71,20 @@ test_arrayreadwrite(cl_device_id device, cl_context context, cl_command_queue qu
...
+            queue, buffer, CL_TRUE, offset * sizeof(cl_uint),
+            cb * sizeof(cl_uint), &outptr[offset], 0, nullptr, nullptr);
         test_error(err, "clEnqueueReadBuffer failed");
 
-        for (j=offset; j<offset+cb; j++)
+        const cl_uint* expected_buffer_values = reference_ptr.data();
+        for (int j = offset; j < offset + cb; j++)
         {
-            if (inptr[j] != outptr[j])
+            if (expected_buffer_values[j] != outptr[j])
```
bashbaug, line 87: Minor: Can we use `reference_ptr` directly here vs. extracting the raw pointer `expected_buffer_values`?  This will give us bounds checking on `operator[]`.

test_conformance/basic/test_bufferreadwriterect.cpp
```diff
@@ -344,9 +344,22 @@ void CL_CALLBACK mem_obj_destructor_callback( cl_mem, void *data )
...
 }
 
+using test_fn = int (*)(size_t, size_t[3], size_t[3], size_t, size_t[3],
+                        size_t[3]);
+struct TestFunctions
+{
+    test_fn copy;
+    test_fn read;
+    test_fn write;
+};
```
bashbaug, line 354: Could you provide a bit more background what the plan is for these `TestFunctions`?

MichaelRizkalla-arm: I'll be adding another set of copy and write test functions to be used with `CL_MEM_IMMUTABLE_EXT` memory flag.
With this change, we can re-use `test_bufferreadwriterect_impl` and pass to it the new test functions.

lakshmih (dismissed):

test_conformance/basic/test_arrayreadwrite.cpp
MichaelRizkalla-arm: Done

test_conformance/basic/test_arrayreadwrite.cpp
MichaelRizkalla-arm: I renamed this instance to `reference_vals`.

bashbaug (approved):

lakshmih (approved):

bashbaug:
Merging (offline) as discussed in the June 24th teleconference.

PR #2210, opened 2025-01-05
add initial unified SVM capability tests

bashbaug:
These tests are passing on many devices using the unified SVM emulation layer.

Specifically, adds tests for:

* CL_SVM_CAPABILITY_SINGLE_ADDRESS_SPACE_KHR
* CL_SVM_CAPABILITY_DEVICE_UNASSOCIATED_KHR
* CL_SVM_CAPABILITY_HOST_READ_KHR
* CL_SVM_CAPABILITY_HOST_WRITE_KHR
* CL_SVM_CAPABILITY_HOST_MAP_KHR
* CL_SVM_CAPABILITY_DEVICE_READ_KHR
* CL_SVM_CAPABILITY_DEVICE_WRITE_KHR
* CL_SVM_CAPABILITY_DEVICE_ATOMIC_ACCESS_KHR
* CL_SVM_CAPABILITY_INDIRECT_ACCESS_KHR

Still TODO:

* CL_SVM_CAPABILITY_SYSTEM_ALLOCATED_KHR
* CL_SVM_CAPABILITY_DEVICE_OWNED_KHR
* CL_SVM_CAPABILITY_CONTEXT_ACCESS_KHR
* CL_SVM_CAPABILITY_HOST_OWNED_KHR
* CL_SVM_CAPABILITY_CONCURRENT_ACCESS_KHR
* CL_SVM_CAPABILITY_CONCURRENT_ATOMIC_ACCESS_KHR

PR #2261, opened 2025-02-04
add initial unified SVM API tests

bashbaug:
This PR includes tests for `clSVMAllocWithPropertiesKHR` and `clGetSVMPointerInfoKHR` for valid pointers, as per the unified SVM test plan.  These tests are passing on many devices with the unified SVM emulation layer.

Note, these tests use the same test fixture and unified SVM wrapper as the unified SVM capability tests in #2210.


bashbaug:
Merging (to the cl_khr_unified_svm branch), as discussed in the February 11th memory TSG.

PR #2262, opened 2025-02-04
Deduplicate create_image from Copy/Fill image tests

MichaelRizkalla-arm:
1. Remove duplicate `create_image` code that is in both clFillImage and clCopyImage test directories.
2. Unify how pitch buffer's memory is deallocated; The buffer can be allocated with either `malloc` or `align_malloc` and the free function is pre-set in `pitch_buffe_data`'s member variable `free_fn` and used when the buffer is deallocated. With this, the change removes `is_aligned` conditional variable that was used to select the appropriate free function.


lakshmih (approved):

bashbaug:

PR #2268, opened 2025-02-10
bugfix nan test for basic fpmath

zzk0:
This PR fixes the validation logic for cases where the data type is not half. Because the variable nan_test is always false, types like float never trigger a validation failure.

### DIFF
--- a/test_conformance/basic/test_fpmath.cpp
+++ b/test_conformance/basic/test_fpmath.cpp
@@ -110,6 +110,7 @@ int verify_fp(std::vector<T> (&input)[2], std::vector<T> &output,
             both_nan = std::isnan(r) && std::isnan(output[i]);
         }
 
+        // If not both nan, check if the result is the same
         if (!both_nan && (r != output[i]))
         {
             log_error("FP math test for type: %s, vec size: %zu, failed at "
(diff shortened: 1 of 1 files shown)


test_conformance/basic/test_fpmath.cpp
```diff
@@ -98,14 +98,14 @@ int verify_fp(std::vector<T> (&input)[2], std::vector<T> &output,
...
+        bool nan_test = true;
 
         T r = test.ref(inA[i], inB[i]);
 
         if (std::is_same<T, cl_half>::value)
-            nan_test = !(isHalfNan(r) && isHalfNan(output[i]));
+            nan_test = (isHalfNan(r) != isHalfNan(output[i]));
 
-        if (r != output[i] && nan_test)
+        if (r != output[i] && !nan_test)
```
bashbaug, line 108: Hi!  This seems like it has the same problem as before.  Since `nan_test` is true for all types other than `half`, it still skips verification for `float` and `double`, correct?

Instead of trying to fix the NaN checks, can we tear it all out completely?  It's curious to me that we have special handling for `half` NaNs, but we don't have any special handling for `float` or `double` NaNs.

zzk0: > This seems like it has the same problem as before. 

😅... Yes, should check the code again. 

> can we tear it all out completely? 

Use std::isnan for float/double NaN handling.

bashbaug (approved):
Note, at least on my system, `both_nan` is never true.  I think this is because the inputs for these test cases are chosen so NaNs will not occur.  So, I still think we could remove all of this NaN handling if we'd like.... or we could keep it as it is here, just in case it becomes needed in the future.

bashbaug:

Issue #2224, opened 2025-01-10
math_brute_force's fma test seems to strict for RTZ device


cycheng:
Hi,

It seems that the half-precision's FMA test is too strict in verification in the current code (I've tested on the latest commit).
The test case:
```
fma(0x8400, 0x0001, 0x7bff)
= fma_f32(-0.0000610351, 0.0000000596, 65504.0)
= fma_f32(0xb8800000, 0x33800000, 0x477fe000)

RTE HW = 65504.0 = 0x477fe000
RTZ HW = 65503.996093 = 0x477fdfff

65504 = 0x7bff (half) 
65503.996093 = 7bfe (half)
```
This causes 1 ULP error on my device:
```
        Testing half precision? YES
        Is Embedded? YES
        Running in RTZ mode? YES
        Tininess is detected before rounding? NO
        Worker threads: 10
        Testing vector sizes:   2
        Verbose? NO

fma...
  1:             fma fp16
ERROR: fma2: -1.000000 ulp error at {-0x1p-14, 0x1p-24, 0x1.ffcp+15} ({0x8400, 0x0001, 0x7bff}): *0x1.ffcp+15 vs. 0x1.ff8p+15
```

Any comments would be appreciated!
CY

PR #2269, opened 2025-02-10
Correct reference for fma for RTZ devices for half

AhmedAmraniAkdi:
The reference was always being calculated using round to nearest even for fma when testing half precision.

Fixes https://github.com/KhronosGroup/OpenCL-CTS/issues/2224


lakshmih (approved):

bashbaug:

PR #2270, opened 2025-02-11
bruteforce: Check both input values for nan

lakshmih:


test_conformance/math_brute_force/reference_math.cpp
```diff
@@ -1641,6 +1641,7 @@ double reference_expm1(double x)
 double reference_fmax(double x, double y)
 {
     if (isnan(y)) return x;
+    if (isnan(x)) return y;
```
svenvh, line 1644: For consistency, we should probably do the same in:
 - `reference_fmin`
 - `reference_fmaxl`
 - `reference_fminl`

lakshmih (commented):
Addressed

svenvh (approved):

bashbaug (approved):

bashbaug:

PR #2271, opened 2025-02-11
Fix signalling and waiting on semaphore from two queues

MichaelRizkalla-arm:
`semaphores_ooo_ops_cross_queue` uses two OOO command queues to run the test. In one queue, a semaphore is signalled and the semahpore is waited on in the other queue.

The CL specification requires the application to synchronize the queues if objects are shared.

### DIFF
--- a/test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
+++ b/test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
@@ -266,6 +266,15 @@ struct SemaphoreOutOfOrderOps : public SemaphoreTestBase
         err = clEnqueueBarrierWithWaitList(consumer_queue, 0, nullptr, nullptr);
         test_error(err, " clEnqueueBarrierWithWaitList ");
 
+        if (!single_queue)
+        {
+            // Both producer and consumer queues run independently.
+            // A blocking call to clEnqueueReadBuffer in consumer_queue will not
+            // flush the producer queue.
+            err = clFinish(producer_queue);
+            test_error(err, "clFinish failed");
+        }
+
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_semaphore/test_semaphores_cross_queue.cpp
```diff
@@ -266,6 +266,15 @@ struct SemaphoreOutOfOrderOps : public SemaphoreTestBase
         err = clEnqueueBarrierWithWaitList(consumer_queue, 0, nullptr, nullptr);
         test_error(err, " clEnqueueBarrierWithWaitList ");
 
+        if (!single_queue)
+        {
+            // Both producer and consumer queues run independently.
+            // A blocking call to clEnqueueReadBuffer in consumer_queue will not
+            // flush the producer queue.
+            err = clFinish(producer_queue);
+            test_error(err, "clFinish failed");
```
bashbaug, line 275: Would a `clFlush` be sufficient here, or does this need to be a `clFinish`?

I believe both are correct, but the test is a little stronger if it's a `clFlush` and not a `clFinish`.  With a `clFinish`, the semaphore is a NOP for an implementation that maximally batches commands.

joshqti: I agree, flush is a bit more appropriate here.

MichaelRizkalla-arm: I've updated the commit to use `clFlush`

bcalidas:
Looks good for Qualcomm

bashbaug (approved):

bashbaug:

PR #2272, opened 2025-02-11
Added test to verify program queries after recompiling and relinking

### LINKED ISSUE #2163: Missing coverage for `CL_PROGRAM_*` queries whose value can change after a rebuild
The following queries to `clGetProgramInfo` require a built program and their value can change when the program is rebuilt. We do not test these cases adequately. 
- `CL_PROGRAM_NUM_KERNELS`
- `CL_PROGRAM_KERNEL_NAMES`
- `CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT` (Added in 2.2, deprecated in 3.0)
- `CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT` (Added in 2.2, deprecated in 3.0)

We should extend/replace the `test_get_program_info_kernel_names` test to cover the following:

- Create a program from source (the source contains multiple kernels with at least one of them defined under a preprocessor macro)
- Query `CL_PROGRAM_NUM_KERNELS` and check that it fails with `CL_INVALID_PROGRAM_EXECUTABLE`
- Query `CL_PROGRAM_KERNEL_NAMES` and check that it fails with `CL_INVALID_PROGRAM_EXECUTABLE`
- Build the program with the preprocessor macro undefined
- Query `CL_PROGRAM_NUM_KERNELS` and check that the correct number is returned
- Query `CL_PROGRAM_KERNEL_NAMES` and check that the right kernel names are returned
- Build the program again with the preprocessor macro defined
- Query `CL_PROGRAM_NUM_KERNELS` and check that the correct number is returned
- Query `CL_PROGRAM_KERNEL_NAMES` and check that the right kernel names are returned

I suggest we do not cover `CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT` and `CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT` under this issue given OpenCL support for program scope constructors and destructors is only required on 2.2, which nobody really supports.

Also consider adding a variant using clCompileProgram/clLinkProgram.

shajder:
Fixes #2163 according to issue description


test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -422,6 +451,134 @@ int test_get_program_info(cl_device_id deviceID, cl_context context, cl_command_
...
+
+        error = clGetProgramInfo(program, CL_PROGRAM_KERNEL_NAMES, 0, nullptr,
+                                 &kernel_names_len);
+        test_error(error, "clGetProgramInfo failed");
+
+        const size_t len = kernel_names_len + 1;
+        std::vector<char> kernel_names(len, '\0');
+        error =
+            clGetProgramInfo(program, CL_PROGRAM_KERNEL_NAMES, kernel_names_len,
+                             kernel_names.data(), &kernel_names_len);
```
bashbaug, line 506: If you're interested, it's possible to read into a `std::string` directly (for C++11 and newer, which we are), vs. reading into a `std::vector` and then creating a `std::string` from it.

Something like:

```suggestion
        std::string kernel_names(kernel_names_len, '\0');
        error =
            clGetProgramInfo(program, CL_PROGRAM_KERNEL_NAMES, kernel_names_len,
                             &kernel_names[0], nullptr);
        kernel_names.pop_back(); // remove the final '\0'
```

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -422,6 +451,134 @@ int test_get_program_info(cl_device_id deviceID, cl_context context, cl_command_
...
+        test_error(error, "Unable to get kernel names list.");
+        std::string program_names = kernel_names.data();
+
+        std::vector<std::string> expected_names = { "sample_test_A",
+                                                    "sample_test_C" };
+        for (const auto &name : expected_names)
+        {
+            test_assert_error(program_names.find(name.c_str())
```
bashbaug, line 514: Minor: I don't think we need `.c_str()` here, do we?

```suggestion
            test_assert_error(program_names.find(name)
```

Same pattern exists a few other places in this file.

test_conformance/compiler/test_build_helpers.cpp
shajder: Thanks. Obviously I can apply your suggestion but some time ago I was asked by one of reviewers here in CTS project to use `std::vector<char>` in such cases due to the concerns around spec requirement for `std::string` to have internally continuous chunk of memory. As far as I learnt later it was cleared up at some point in C++11 or 14 spec, still it seems like `std::vector<char>` delivers full control over underlying buffer. Just let me know to do the change.

test_conformance/compiler/test_build_helpers.cpp
shajder: Corrected

lakshmih (approved):

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -87,6 +86,35 @@ __kernel void sample_test_C(__global float *src, __global int *dst)
 }
 )";
 
+const char *sample_multi_kernel_code_AB_with_makro = R"(
```
lakshmih, line 89: nit: was this macro?

shajder: Ops, thanks. Corrected

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -550,6 +578,128 @@ REGISTER_TEST(get_program_info_kernel_names)
...
+        std::vector<std::string> expected_names = {
+            "sample_test_A", "sample_test_B", "sample_test_C", "sample_test_D"
+        };
+
+        const size_t len = kernel_names_len + 1;
+        std::vector<char> kernel_names(len, '\0');
+        error =
+            clGetProgramInfo(program, CL_PROGRAM_KERNEL_NAMES, kernel_names_len,
+                             kernel_names.data(), &kernel_names_len);
+        test_error(error, "Unable to get kernel names list.");
```
lakshmih, line 691: Minor: Could you change this to "Could not find expected kernel name" to better reflect the operation"

lakshmih:
Sorry @shajder should have collated my comments but didnt click submit until late!

test_conformance/compiler/test_build_helpers.cpp

lakshmih (approved):

PR #2273, opened 2025-02-11
Vulkan multiimport diffctx fix

dcrawleyqti:
The multi import diff ctx portion of the Vulkan test mistakenly instantiates clExternalImportableSemaphore objects as wrappers for semaphores that will be exported (vkCl2VkSemaphore). This change corrects this and mirrors the correct behavior found in the multi import same ctx test.


kpet:
May be superseded by https://github.com/KhronosGroup/OpenCL-CTS/pull/2243

dcrawleyqti:
superseded by #2246

PR #2274, opened 2025-02-12
spirv_new: print build log in get_program_with_il

svenvh:
Print the build log when building the program in `get_program_with_il` fails, to make it easier to investigate spirv_new test failures.

Factor out a global helper function `OutputBuildLog` for printing the build log for a single device.

### DIFF
--- a/test_common/harness/errorHelpers.cpp
+++ b/test_common/harness/errorHelpers.cpp
@@ -555,6 +555,39 @@ float Ulp_Error_Double(double test, long double reference)
     return result;
 }
 
+cl_int OutputBuildLog(cl_program program, const cl_device_id device)
+{
+    size_t size_ret;
+
+    // Get the build status
+    cl_build_status build_status;
+    cl_int error =
+        clGetProgramBuildInfo(program, device, CL_PROGRAM_BUILD_STATUS,
+                              sizeof(build_status), &build_status, &size_ret);
+    test_error(error, "Unable to query build status");
+
+    // If the build failed then print the status, obtain the build log and
+    // print it.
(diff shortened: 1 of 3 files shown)


test_common/harness/errorHelpers.cpp
```diff
@@ -594,33 +627,7 @@ cl_int OutputBuildLogs(cl_program program, cl_uint num_devices,
...
-                test_error(error, "Unable to query build log size");
-                char *build_log = (char *)malloc(size_ret);
-                error = clGetProgramBuildInfo(program, device_list[i],
-                                              CL_PROGRAM_BUILD_LOG, size_ret,
-                                              build_log, &size_ret);
-                test_error(error, "Unable to query build log");
-                log_error("ERROR: CL_PROGRAM_BUILD_LOG:\n%s\n", build_log);
-                free(build_log);
-            }
+            OutputBuildLog(program, device_list[i]);
```
bashbaug, line 630: If `OutputBuildLog` generates an error do we want to propagate it back to the caller, or are we OK "swallowing" the error and returning `CL_SUCCESS`?

Perhaps this isn't terribly critical because `OutputBuildLog` only generates an error if it is unable to query the build log, and does NOT return an error if the build failed, so generating an error will be a very exceptional condition, but as written this will be a small behavioral difference compared to the previous code.

svenvh: Good point; it seems other parts of the code propagate build log query errors upwards, so I've updated the patch to do so here and in `get_program_with_il`.

bashbaug:
Merging as discussed in the February 18th teleconference.

PR #2280, opened 2025-02-17
add tests for clGetSVMPointerInfoKHR with a non-USVM pointer

bashbaug:
Note, this PR is setup to merge to the `cl_khr_unified_svm` branch.

Querying a non-USVM pointer should return default values and should not return an error.  This is the next item on the `cl_khr_unified_svm` test plan.


bashbaug:
Merging as discussed in the March 18th memory subgroup.

PR #2281, opened 2025-02-18
math_brute_force: treat reciprocal as unary function

svenvh:
Treat reciprocal as a unary function, instead of handling it through the binary function testing mechanism and special-casing it there.

This addresses two shortcomings of the previous implementation:

 - Testing took significantly longer as the entire input domain was tested many times (e.g. fp16 reciprocal has only 2^16 possible input values, but binary function testing iterates over 2^16 * 2^16 input values).

 - The reciprocal test kernel was identical to the divide kernel. Thus the device compiler would see a regular divide operation instead of a reciprocal operation and would be unlikely to emit a specialized reciprocal sequence.

This reverts all of the changes in binary_operator*.cpp made by bcfa1f7c2 ("Added corrections to re-enable reciprocal test in math_brute_force suite for relaxed math mode (#2221)", 2025-02-04).


test_conformance/math_brute_force/function_list.cpp
```diff
@@ -427,9 +427,8 @@ const Func functionList[] = {
     // basic operations
     OPERATOR_ENTRY(add, "+", 0.0f, 0.0f, 0.0f, 0.0f, FTZ_OFF, binaryOperatorF),
     OPERATOR_ENTRY(subtract, "-", 0.0f, 0.0f, 0.0f, 0.0f, FTZ_OFF, binaryOperatorF),
-    //ENTRY(reciprocal, 1.0f, 1.0f, FTZ_OFF, unaryF),
     { "reciprocal",
-      "/",
+      "1.0/",
```
bashbaug, line 431: This doesn't work on several devices that I've tested, unfortunately.

* Devices that do not support fp64 are generating a warning:

```
18:16:23: warning: double precision constant requires cl_khr_fp64 and __opencl_c_fp64, casting to single precision
        RETTYPE res = 1.0/(a);
```

* Some other cases are generating an error:

```
17:12:17: error: scalar operand type has greater rank than the type of the vector element. ('float' and 'half4' (vector of 4 'half' values))
    out[i] = 1.0/(in1[i]);
             ~~~^~~~~~~~~
```

Changing this to `(RETTYPE)(1.0)/` fixed the error, but not the warning.

Changing to `(RETTYPE)(1.0f)/` fixed both, and I think it's still doing what we want it to do, but it leaks details into this function list that feel undesirable.

Do you have any ideas how to fix this cleanly?  Could we maybe implement reciprocal as a clearly identified macro in the test program, then move all of this logic into the macro, for example?

bashbaug (changes requested):
I really like this direction and this is something we should continue pursuing, but unfortunately this isn't working for me in its current form.

test_conformance/math_brute_force/function_list.cpp
svenvh: Thanks for trying this out on different devices!  I'll see if I can make this more robust (and clean).

Removing the focused-review label in the meantime.

svenvh: I've made some changes to the unary_*.cpp files instead.  This allows a type-dependent 1.0 literal suffix for each data type; I wasn't too hopefull about the unsuffixed `1.0` constant in the previous patch anyway.

shajder:
Works fine with my nvidia and intel devices.

bashbaug (approved):
Thanks!  This works for the device I've tested, and in some cases it's a significant performance improvement.

bashbaug:
Discussed in the February 25th teleconference.  Waiting one more week for additional review approvals.

lakshmih (approved):

bashbaug:

PR #2283, opened 2025-02-19
Refactor clCopyImage and clFillImage tests

MichaelRizkalla-arm:
This change mainly extends `clFillImage` and `clCopyImage` test function to include memory flags to be used during creating the image instead of hard-coding these values. The memory flags are also different parameters for source and destination images in `clCopyImage` tests.

### DIFF
--- a/test_conformance/images/clFillImage/test_loops.cpp
+++ b/test_conformance/images/clFillImage/test_loops.cpp
@@ -18,19 +18,38 @@
 
 extern int gTypesToTest;
 
-extern int test_fill_image_set_1D( cl_device_id device, cl_context context, cl_command_queue queue, cl_image_format *format, ExplicitType outputType );
-extern int test_fill_image_set_2D( cl_device_id device, cl_context context, cl_command_queue queue, cl_image_format *format, ExplicitType outputType );
-extern int test_fill_image_set_3D( cl_device_id device, cl_context context, cl_command_queue queue, cl_image_format *format, ExplicitType outputType );
-extern int test_fill_image_set_1D_array( cl_device_id device, cl_context context, cl_command_queue queue, cl_image_format *format, ExplicitType outputType );
-extern int test_fill_image_set_2D_array( cl_device_id device, cl_context context, cl_command_queue queue, cl_image_format *format, ExplicitType outputType );
-extern int test_fill_image_set_1D_buffer(cl_device_id device,
-                                         cl_context context,
-                                         cl_command_queue queue,
-                                         cl_image_format *format,
-                                         ExplicitType outputType);
+extern int test_fill_image_set_1D(cl_device_id device, cl_context context,
+                                  cl_command_queue queue,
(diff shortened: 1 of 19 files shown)


lakshmih (approved):

bashbaug:

test_conformance/images/clFillImage/test_loops.cpp
```diff
@@ -128,8 +147,12 @@ int test_image_type( cl_device_id device, cl_context context, cl_command_queue q
 int test_image_set( cl_device_id device, cl_context context, cl_command_queue queue, MethodsToTest testMethod )
 {
     int ret = 0;
-
-    ret += test_image_type( device, context, queue, testMethod, CL_MEM_READ_ONLY );
+    cl_mem_flags flags = CL_MEM_READ_ONLY;
+    if (gEnablePitch && testMethod != k1DBuffer)
+    {
+        flags |= CL_MEM_USE_HOST_PTR;
```
karolherbst, line 153: So, I'm curious about this change. Specifying this flag when creating an image is valid, however, I'm not sure it's valid when passing it into `clGetSupportedImageFormats`

> flags is a bit-field that is used to specify usage information about the image formats being queried and is described in the [Memory Flags](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#memory-flags-table) table. flags may be [CL_MEM_READ_WRITE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_READ_WRITE) to query image formats that may be read from and written to by different kernel instances when correctly ordered by event dependencies, or [CL_MEM_READ_ONLY](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_READ_ONLY) to query image formats that may be read from by a kernel, or [CL_MEM_WRITE_ONLY](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_WRITE_ONLY) to query image formats that may be written to by a kernel, or [CL_MEM_KERNEL_READ_AND_WRITE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_KERNEL_READ_AND_WRITE) to query image formats that may be both read from and written to by the same kernel instance. flags may also be [CL_MEM_IMMUTABLE_EXT](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MEM_IMMUTABLE_EXT) to query image formats that may be used to create immutable images. Please see [Image Format Mapping](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#image-format-mapping) for clarification.

I think the other question is, is `clGetSupportedImageFormats` allowed to return different formats when `CL_MEM_USE_HOST_PTR` is specified? Do we need to clarify the spec here?

PR #2284, opened 2025-02-19
basic: fix -Wformat-security warnings

svenvh:
Fix warnings such as:

    test_vloadstore.cpp:330:49: error: format string is not a string literal (potentially insecure)

There were no security issues here as the format string arguments do not contain any conversion specifiers and are never written to.  Make that latter fact explicit to avoid the warnings.


bashbaug (approved):
This works for me.

I've also seen a pattern like this used instead:

```
sprintf(..., "%s", store_pattern);
```

lakshmih (approved):

bashbaug:
Merging as discussed in the March 18th teleconference.

PR #2287, opened 2025-02-20
Fix profiling_timebase error, the global work size shouldn't be NULL,…

jujiang-del:
… according to OpenCL specification


bashbaug:
See https://github.com/KhronosGroup/OpenCL-CTS/pull/2241

jujiang-del:
Fixed in OpenCL

PR #2288, opened 2025-02-20
Remove uuid comparison.  This is not in the spec

joshqti:
Remove comparison of Vulkan and OpenCL UUID.  This is not required by the specification for external import of images.  I think we should consider defining this behavior in the spec to ensure portable application code.


joshqti:
Please don't test yet, I am working on refinements.

PR #2289, opened 2025-02-21
Corrected procedure to collect proper size of image from VkImageCreateInfo

### LINKED ISSUE #2155: Wrong cl_image_desc setup collected from VkImageCreateInfo
Procedure `getCLImageInfoFromVkImageInfo` returns image width/height/depth 256x256x10 despite the fact image has 256x16x10 dimensions. Source of the problem is invalid usage of `get2DImageDimensions` which computes [height](https://github.com/KhronosGroup/OpenCL-CTS/blob/918d561c6c42fb7f3e2ffe1d1574c0c37bb2b15b/test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp#L445) without taking into accout depth value:

    height = totalImageSize / row_pitch;

shajder:
Fixes #2155 according to issue description

Additional remark: The image size was previously calculated based on the memory size, which seems unusual. Due to Vulkan's configuration capabilities, the size of memory allocated for a specific texture may differ from what would be expected based on the texture dimensions. Thus, calculating the image dimensions back from the memory size of a Vulkan texture can be challenging.


test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -469,25 +485,58 @@ getCLImageInfoFromVkImageInfo(const VkImageCreateInfo *VulkanImageCreateInfo,
...
+    {
+        error = clGetDeviceInfo(device, CL_DEVICE_IMAGE2D_MAX_WIDTH,
+                                sizeof(width), &width, NULL);
+        test_error(error, "Unable to get CL_DEVICE_IMAGE2D_MAX_WIDTH");
+        error = clGetDeviceInfo(device, CL_DEVICE_IMAGE2D_MAX_HEIGHT,
+                                sizeof(height), &height, NULL);
+        test_error(error, "Unable to get CL_DEVICE_IMAGE2D_MAX_HEIGHT");
+    }
+
+    if (0 == (height | width | depth)) return CL_IMAGE_FORMAT_NOT_SUPPORTED;
```
joshqti, line 513: use height == 0 && width == 0 && depth == 0, it is not obvious to me that this is what is intended.  Also, is this condition possible?  One of the two if branches will be executed, and either the test will fail, or width and height will be non-zero.  This check can probably be deleted.

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -469,25 +485,58 @@ getCLImageInfoFromVkImageInfo(const VkImageCreateInfo *VulkanImageCreateInfo,
...
 
     img_desc->image_depth =
         static_cast<size_t>(VulkanImageCreateInfo->extent.depth);
     img_desc->image_array_size = 0;
-    img_desc->image_row_pitch = 0; // Row pitch set to zero as host_ptr is NULL
-    img_desc->image_slice_pitch =
-        img_desc->image_row_pitch * img_desc->image_height;
+    // If image_row_pitch is zero and the image is created from an external
+    // memory handle, then the image row pitch is implementation-defined
+    img_desc->image_row_pitch = 0;
```
joshqti, line 531: I am ok with this because it is an improvement.   It would be better if row-pitch and slice pitch are specified, if assume linear is set.

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
shajder: Removed. FYI the condition followed harness [helper](https://github.com/KhronosGroup/OpenCL-CTS/blob/5a652786133defe33e5f47510c8b57e62ae580c9/test_common/harness/kernelHelpers.cpp#L1422).

test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
shajder: Correction pushed. As far as I understand we can set row and slice pitches as long as tiling mode is linear. For optimal tiling I left zeros. I have limited ability to test this fix due to the fact my devices do not support image 3D with linear tiling, I will appreciate proper test. Thanks!

joshqti: It is currently crashing for us.  We need a little time to debug

bashbaug:
@shajder is this ready for review?  I see you removed the "focused review" label - thanks!

shajder:
> @shajder is this ready for review?  I see you removed the "focused review" label - thanks!

yes - it is ready for review. I removed the label "focused review" because @joshqti mentioned he needed more time for debugging.

joshqti:
I still see a segfault in vkGetImageSubresourceLayout

joshqti:
`VUID-vkGetImageSubresourceLayout-mipLevel-01716(ERROR / SPEC): msgNum: -1906823971 - Validation Error: [ VUID-vkGetImageSubresourceLayout-mipLevel-01716 ] Object 0: handle = 0x90c9290000000714, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x8e582cdd | vkGetImageSubresourceLayout(): pSubresource.mipLevel (1) must be less than 1. The Vulkan spec states: The mipLevel member of pSubresource must be less than the mipLevels specified in VkImageCreateInfo when image was created (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkGetImageSubresourceLayout-mipLevel-01716)
    Objects: 1
        [0] 0x90c9290000000714, type: 10, name: NULL`

shajder:
> `VUID-vkGetImageSubresourceLayout-mipLevel-01716(ERROR / SPEC): msgNum: -1906823971 - Validation Error: [ VUID-vkGetImageSubresourceLayout-mipLevel-01716 ] Object 0: handle = 0x90c9290000000714, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x8e582cdd | vkGetImageSubresourceLayout(): pSubresource.mipLevel (1) must be less than 1...

This one is fixed, still it didn't cause crash at my machine (nvidia driver version: 550.120)

test_conformance/vulkan/test_vulkan_api_consistency_for_1dimages.cpp
```diff
@@ -156,14 +155,12 @@ struct ConsistencyExternalImage1DTest : public VulkanTestBase
...
-        errNum = getCLImageInfoFromVkImageInfo(&VulkanImageCreateInfo,
-                                               totalImageMemSize, &img_format,
-                                               &image_desc);
-        if (errNum != CL_SUCCESS)
-        {
-            log_error("getCLImageInfoFromVkImageInfo failed!!!");
-            return TEST_FAIL;
-        }
+        auto layout = vkImage1D.getSubresourceLayout();
+        errNum = getCLImageInfoFromVkImageInfo(
```
joshqti, line 159: This is causing a test failure in consistence 1d/3d (line 180 for 3D), where host_ptr is NULL and row_pitch/slice_pitch are non zero.  The OpenCL spec requires row_pitch/slice_pitch to be zero when hostptr is NULL.

shajder: > I am ok with this because it is an improvement. It would be better if row-pitch and slice pitch are specified, if assume linear is set. This should be a follow-up PR if not done in a follow-up patch set.

@joshqti it seems like we're going in circles here. I just removed the computations related to row_pitch/slice_pitch, but please note they were added recently in response to the previous request.

joshqti: The issue is that particular subtest passes a NULL host pointer, which is a violation of the OpenCL spec to pass row-pitch or slice pitch with NULL host pointer, and not using image import.

`       
        // Passing NULL properties and a valid image_format and image_desc
        image = clCreateImageWithProperties(context, NULL, CL_MEM_READ_WRITE,
                                            &img_format, &image_desc, NULL,
                                            &errNum);
        test_error(errNum,
                   "Unable to create image with NULL properties "
                   "with valid image format and image desc");
`

I think the problem isn't your change in this case, I think this test case should be deleted.  The properties are NULL, and therefore this check has nothing to to with OpenCL-Vulkan interop.

shajder: I see. I have restored changes from last commit and removed mentioned subtests. Let me know if that works.

joshqti: Looks good to me, no issues running

PR #2290, opened 2025-02-21
printf: Fix the format specifier for %A to add the decimal point

lakshmih:


lakshmih:
We'd made an error in adding the format specifier, fixed with this change.
Would appreciate a quick review/merge since the current code does not capture the WG's intent properly.

bashbaug (approved):

lakshmih:
Ben, can we merge this? I don't want to merge our own PR. Thanks!

bashbaug:

PR #2291, opened 2025-02-27
Fix cl_khr_semaphore semaphores_ooo_ops_cross_queue

AhmedAmraniAkdi:
The test calls clEnqueueSignalSemaphoresKHR using the producer queue, and clEnqueueWaitSemaphoresKHR using the consumer queue.

It only flushes only the consumer queue (blocking clEnqueueReadBuffer), but never the producer queue which causes the test to never finish as the semaphore is never signalled.


AhmedAmraniAkdi:
I believe the "Presubmit / Build macos-latest" failure is not due to my change.

bashbaug (approved):

joshqti:
This looks good

PR #2292, opened 2025-02-27
Fix compilation failure caused by DMA heaps

ahesham-arm:
Project fails to build on systems with a kernel version older than 5.6.0 because of `-Wunused-function` combined with `-Werror`.

Expand the conditional compilation guard to include the offending code.


hvdijk:
Can confirm this fixes the build for me, thank you!

svenvh (approved):

bashbaug (approved):

bashbaug:
Merging immediately, seems uncontroversial and fixes a build break.

PR #2294, opened 2025-03-03
Use max work item size instead

starlahuang:
The test is using CL_DEVICE_MAX_WORK_GROUP_SIZE as first dimension of local work size. But it can be bigger than the first dimension of CL_DEVICE_MAX_WORK_ITEM_SIZEs which results in failure.

This patch corrects it to query and use the first dimension of CL_DEVICE_MAX_WORK_ITEM_SIZES instead.


bashbaug (approved):
I'd be a little surprised if this were the only place in the CTS tests with this problem.

bashbaug:
Discussed in the March 4th teleconference.  We'll wait one more week for additional reviewers.

PR #2295, opened 2025-03-03
spirv_new: ensure int64 printf test uses 64 bits

svenvh:
The `extinst_printf_operands_scalar_int64` test could fail on 32-bit platforms with `CL_INVALID_ARG_SIZE`, because the helper function was not guaranteed to be instantiated using a 64-bit integer template type.

### DIFF
--- a/test_conformance/spirv_new/test_extinst_printf.cpp
+++ b/test_conformance/spirv_new/test_extinst_printf.cpp
@@ -30,6 +30,8 @@
 #define streamDup2(fd1, fd2) dup2(fd1, fd2)
 #endif
 
+#include <cstdint>
+
 #include <fstream>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


test_conformance/spirv_new/test_extinst_printf.cpp
```diff
@@ -232,8 +234,9 @@ lX = 4
         return TEST_SKIPPED_ITSELF;
     }
 
-    return printf_operands_helper(device, "printf_operands_scalar_int64",
-                                  "printf_operands_scalar_int64", expected, 4L);
+    return printf_operands_helper<int64_t>(
+        device, "printf_operands_scalar_int64", "printf_operands_scalar_int64",
+        expected, 4L);
```
bashbaug, line 239: Ah, good catch.

I guess we could also do something like this?  I don't know if this makes it any more obvious what's supposed to be happening:

```suggestion
    return printf_operands_helper(device, "printf_operands_scalar_int64",
                                  "printf_operands_scalar_int64", expected,
                                  static_cast<cl_long>(4));
```

bashbaug (approved):

test_conformance/spirv_new/test_extinst_printf.cpp
svenvh: That may also work, but to me it feels a little less obvious and more fragile than just making the template type explicit.

svenvh: Thanks, waiting until the call tomorrow is fine!

bashbaug:

PR #2296, opened 2025-03-04
Fix inconsistent variable name in REGISTER_TEST and REQUIRE_EXTENSION

MichaelRizkalla-arm:
Both `REGISTER_TEST` and `REQUIRE_EXTENSION` expect cl_device_id variable but the variable name is inconsistent which makes both macros unusable together.

This change renames `deviceID` in `REQUIRE_EXTENSION` to `device` to be consistent with `REGISTER_TEST`.


bashbaug (approved):

bashbaug:

PR #2298, opened 2025-03-05
Fix test_printf - any nan format is allowed by OpenCL C spec

gwawiork:
OpenCL C says that printing nan is implementation-defined for floating-point numbers - https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#printf-format-string

### DIFF
--- a/test_conformance/printf/util_printf.cpp
+++ b/test_conformance/printf/util_printf.cpp
@@ -19,7 +19,7 @@
 #include "test_printf.h"
 #include <assert.h>
 #include <CL/cl_half.h>
-
+#include <regex>
 
 // Helpers for generating runtime reference results
 static void intRefBuilder(printDataGenParameters&, char*, const size_t);
@@ -1743,10 +1743,14 @@ size_t verifyOutputBuffer(char *analysisBuffer,testCase* pTestCase,size_t testId
     else if (pTestCase->_correctBuffer[testId] == "INF")
         return strcmp(analysisBuffer, "INF")
             && strcmp(analysisBuffer, "INFINITY");
-    else if (pTestCase->_correctBuffer[testId] == "nan")
-        return strcmp(analysisBuffer, "nan") && strcmp(analysisBuffer, "-nan");
-    else if (pTestCase->_correctBuffer[testId] == "NAN")
-        return strcmp(analysisBuffer, "NAN") && strcmp(analysisBuffer, "-NAN");
+    else if (pTestCase->_correctBuffer[testId] == "nan"
+             || pTestCase->_correctBuffer[testId] == "NAN")
+    {
+        std::string pattern =
+            R"(-?)" + pTestCase->_correctBuffer[testId] + R"((\(.*\))?)";
+        std::regex nanRegex(pattern);
+        return !std::regex_match(analysisBuffer, nanRegex);
+    }
 
     return strcmp(analysisBuffer, pTestCase->_correctBuffer[testId].c_str());
 }


test_conformance/printf/util_printf.cpp
```diff
@@ -1744,9 +1744,13 @@ size_t verifyOutputBuffer(char *analysisBuffer,testCase* pTestCase,size_t testId
...
     else if (pTestCase->_correctBuffer[testId] == "nan")
-        return strcmp(analysisBuffer, "nan") && strcmp(analysisBuffer, "-nan");
+        return strcmp(analysisBuffer, "nan") && strcmp(analysisBuffer, "-nan")
+            && strcmp(analysisBuffer, "nan(ind)")
+            && strcmp(analysisBuffer, "-nan(ind)");
     else if (pTestCase->_correctBuffer[testId] == "NAN")
-        return strcmp(analysisBuffer, "NAN") && strcmp(analysisBuffer, "-NAN");
+        return strcmp(analysisBuffer, "NAN") && strcmp(analysisBuffer, "-NAN")
+            && strcmp(analysisBuffer, "NAN(IND)")
+            && strcmp(analysisBuffer, "-NAN(IND)");
```
lakshmih, line 1753: As per OpenCL C spec "NaN is converted in one of the styles [-]nan or [-]nan(n-char-sequence)  — which style, and the meaning of any n-char-sequence, is implementation-defined" 

Since the character sequence has to be implementation-defined, should the comparison use a regex instead of the hard-coded 'IND'?

gwawiork: > As per OpenCL C spec "NaN is converted in one of the styles [-]nan or [-]nan(n-char-sequence)  — which style, and the meaning of any n-char-sequence, is implementation-defined"
> 
> Since the character sequence has to be implementation-defined, should the comparison use a regex instead of the hard-coded 'IND'?

Thanks. Please check now solution with regex

test_conformance/printf/util_printf.cpp
```diff
@@ -1743,10 +1743,12 @@ size_t verifyOutputBuffer(char *analysisBuffer,testCase* pTestCase,size_t testId
...
         return strcmp(analysisBuffer, "INF")
             && strcmp(analysisBuffer, "INFINITY");
-    else if (pTestCase->_correctBuffer[testId] == "nan")
-        return strcmp(analysisBuffer, "nan") && strcmp(analysisBuffer, "-nan");
-    else if (pTestCase->_correctBuffer[testId] == "NAN")
-        return strcmp(analysisBuffer, "NAN") && strcmp(analysisBuffer, "-NAN");
+    else if (pTestCase->_correctBuffer[testId] == "nan"
+             || pTestCase->_correctBuffer[testId] == "NAN")
+    {
+        std::regex nanRegex(R"(-?nan(\(.*\))?)", std::regex_constants::icase);
```
lakshmih, line 1749: Looks like this will accept  Nan, nAn, nAN, ... , however I think the appropriate regex is:

std::regex nanRegex(R"(-?)" + pTestCase->_correctBuffer[testId] + R"((\(.*\))?)"); 

Could you check, please?

gwawiork: I've modified with your regex but added backslashes in last part  R"((\(.*\))?)"); I've verified it works.

lakshmih: Almost there but think std::regex_constants::icase has to be tossed because it makes the regex case insensitive.

gwawiork:
> One more fix needed

Done :)

lakshmih (approved):

bashbaug:

PR #2300, opened 2025-03-06
Refactor external semaphore scaffolding

gorazd-sumkovski-arm:
Remove the `CREATE_OPENCL_SEMAPHORE` macro and use derived class instantiations of the `clExternalSemaphore` class, rather than base pointers to derived class objects.

Remove the default argument for `queryParamName` in `check_external_semaphore_handle_type()`.

Move `check_external_semaphore_handle_type()` checks to constructors of `clExternalImportableSemaphore` and `clExternalExportableSemaphore`, rather than manually making the check before creating an external semaphore.


test_conformance/common/vulkan_wrapper/opencl_vulkan_wrapper.cpp
```diff
@@ -540,40 +540,46 @@ cl_int check_external_semaphore_handle_type(
...
+            break;
         }
     }
-    log_error("cl_khr_external_semaphore extension is missing support for %d\n",
-              requiredHandleType);
 
-    return CL_INVALID_VALUE;
+    if (!found)
+    {
+        log_error("cl_khr_external_semaphore extension is missing support for"
```
joshqti, line 575: Add space after for

test_conformance/vulkan/test_vulkan_api_consistency.cpp
```diff
@@ -376,6 +376,10 @@ struct ConsistencyExternalSemaphoreTest : public VulkanTestBase
         for (VulkanExternalSemaphoreHandleType semaphoreHandleType :
              supportedExternalSemaphores)
         {
+            check_external_semaphore_handle_type(
+                device, semaphoreHandleType,
```
joshqti, line 380: This fails for us.  The problem is semaphoreHandleType is expected to be an opencl type, cl_external_semaphore_handle_type_khr, but VulkanExternalSemaphorteHandleType is passed.

joshqti:
No objections, however we do not support multi-device, so I cannot test

test_conformance/vulkan/test_vulkan_api_consistency.cpp
gorazd-sumkovski-arm: Translated the type with `getCLSemaphoreTypeFromVulkanType()`.

joshqti:
This passes for us, looks good

joshqti (approved):
Approving as discussed April 8

nikhiljnv:
Merging as agreed in memory subgroup call on April 15th, 2025.

PR #2303, opened 2025-03-06
fix error message in work_item_functions_out_of_range

rjodinchr:
The value printed in the error message is not the correct one (used in the comparison statement)


svenvh (approved):

kpet (approved):

kpet:
Obvious fix, merging.

PR #2306, opened 2025-03-07
Modernization of semaphores_negative_create_multi_device_property related to creating sub-devices

shajder:
Due to the improvements developed during work on #2237 with @bashbaug


joshqti:
No objections, however I cannot test because we do not support multi-device

shajder:
@ahesham-arm this is a very similar improvement as one discussed in #2307. I just applied your corrections. Could you be so kind and take a look? Thanks

ahesham-arm:
> @ahesham-arm this is a very similar improvement as one discussed in #2307. I just applied your corrections. Could you be so kind and take a look? Thanks

LGTM, thanks for including the changes here as well.

ahesham-arm (approved):

bashbaug (approved):

PR #2307, opened 2025-03-07
Modernization of semaphores_negative_create_invalid_device related to creating sub-devices

shajder:
Due to the improvements developed during work on #2237 with @bashbaug

### DIFF
--- a/test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
+++ b/test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
@@ -230,7 +230,7 @@ struct CreateInvalidDevice : public SemaphoreTestBase
         err = clGetDeviceInfo(device, CL_DEVICE_PARTITION_PROPERTIES,
                               supported_props.size()
                                   * sizeof(cl_device_partition_property),
-                              supported_props.data(), &size);
+                              supported_props.data(), nullptr);
         test_error_fail(err, "clGetDeviceInfo failed");
 
         if (supported_props.empty() || supported_props.front() == 0)
--- a/test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
+++ b/test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
@@ -219,7 +219,7 @@ struct CreateInvalidDevice : public SemaphoreTestBase
                                      nullptr, &size);
         test_error_fail(err, "clGetDeviceInfo failed");
 
-        if (size == 0)
+        if ((size / sizeof(cl_device_partition_property)) == 0)
         {
             log_info("Can't partition device, test not supported\n");
             return TEST_SKIPPED_ITSELF;
@@ -233,7 +233,7 @@ struct CreateInvalidDevice : public SemaphoreTestBase
                               supported_props.data(), nullptr);
         test_error_fail(err, "clGetDeviceInfo failed");
 
-        if (supported_props.empty() || supported_props.front() == 0)
+        if (supported_props.front() == 0)
         {
             log_info("Can't partition device, test not supported\n");
             return TEST_SKIPPED_ITSELF;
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
```diff
@@ -211,38 +212,77 @@ struct CreateInvalidDevice : public SemaphoreTestBase
...
+        }
+
+        cl_uint maxComputeUnits = 0;
+        err =
+            clGetDeviceInfo(device, CL_DEVICE_MAX_COMPUTE_UNITS,
+                            sizeof(maxComputeUnits), &maxComputeUnits, nullptr);
+        test_error_ret(err, "Unable to get maximal number of compute units",
+                       TEST_FAIL);
+
+        std::vector<std::array<cl_device_partition_property, 5>>
```
ahesham-arm, line 249: what do we gain by using a `std::array` instead of a C-style array here? i.e.
```
std::vector<cl_device_partition_property[5]>
```

test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
```diff
@@ -211,38 +212,77 @@ struct CreateInvalidDevice : public SemaphoreTestBase
...
-            test_error_ret(
-                CL_INVALID_VALUE,
-                "Multi context test for CL_INVALID_PROPERTY not supported",
-                TEST_SKIPPED_ITSELF);
+        std::vector<cl_device_partition_property> supported_props(
+            size / sizeof(cl_device_partition_property), 0);
+        err = clGetDeviceInfo(device, CL_DEVICE_PARTITION_PROPERTIES,
+                              supported_props.size()
+                                  * sizeof(cl_device_partition_property),
+                              supported_props.data(), &size);
```
ahesham-arm, line 233: `&size` is not used after this line, might as well not pass it.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
```diff
@@ -211,38 +212,77 @@ struct CreateInvalidDevice : public SemaphoreTestBase
...
+        test_error_fail(err, "clGetDeviceInfo failed");
 
-        // get the list of subDevices
-        SubDevicesScopeGuarded scope_guard(deviceCount);
-        err = clCreateSubDevices(device, partitionProp, deviceCount,
-                                 scope_guard.sub_devices.data(), &deviceCount);
-        if (err != CL_SUCCESS)
+        if (supported_props.empty() || supported_props.front() == 0)
```
ahesham-arm, line 236: `supported_props.empty()` can never happen,`size` was already checked that if it is `0`

test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
shajder: In C++, C-style arrays cannot be directly stored in STL containers such as std::vector. C-style arrays are neither copyable nor assignable, which is required by std::vector. Does it build for you if you use suggested code ?

test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
shajder: corrected

test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
ahesham-arm: that would result in a vector of size zero, i.e. we pass `0` to `clGetDeviceInfo` (which smaller than `1` or whatever the value we got back was)

test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
ahesham-arm: I see. No, you're right, thanks for clarifying.

test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp
shajder: Okay, this is also scenario for skipping the test instead of failing it. Correction applied.

ahesham-arm (approved):
LGTM, thanks for addressing the review comments.

bashbaug (approved):

PR #2310, opened 2025-03-07
Remove .gitattributes

kpet:
We don't use git lfs. We have no .7z files in the repo.


bashbaug (approved):
No need to wait to merge this one.

PR #2311, opened 2025-03-08
Migrate printf suite to the new test registration framework

ahesham-arm:
Contributes to #2181.


kpet (approved):
All 17 tests are still correctly registered. The macOS bot passed but did not report its status. Merging.

PR #2316, opened 2025-03-09
Migrate basic suite to the new test registration framework

ahesham-arm:
Contributes to #2181.


test_conformance/basic/test_arraycopy.cpp
```diff
@@ -31,14 +30,13 @@ const char *copy_kernel_code =
...
-int
-test_arraycopy(cl_device_id device, cl_context context, cl_command_queue queue, int n_elems)
+REGISTER_TEST(arraycopy)
 {
     cl_uint    *input_ptr, *output_ptr;
     cl_mem                streams[4], results;
     cl_program          program;
     cl_kernel            kernel;
-    unsigned            num_elements = 128 * 1024;
+    unsigned num_elems = 128 * 1024;
```
bashbaug, line 39: If we wanted fewer diffs, we could assign to the `num_elements` function argument, vs. defining a new `num_elems` variable here.

bashbaug (approved):
Noting for completeness: I think the same tests are run after this change, however they are run in a different order.  This shouldn't be a problem, and this isn't something I think we need to fix, but in theory at least it could cause a difference if implementations or scripts are expecting the tests to be run in a particular order.

test_conformance/basic/test_arraycopy.cpp
ahesham-arm: Good point, done!

bashbaug (approved):

PR #2318, opened 2025-03-09
Migrate c11_atomics suite the new test registration framework

ahesham-arm:
Contributes to #2181.


ahesham-arm:
Note that listing the tests before this change by passing `-h` lists them in (almost) an alphaebtical order (it is actually using the order of calls to `ADD_TEST`), but after this change the tests are listed in the order of calls to `REGISTER_TEST`.

This can be trivially changed if we wish to list tests in the future in alphabetical order instead of registration order after all the test registration is centralised.

kpet (approved):
All 34 tests still correctly registered. Merging.

PR #2319, opened 2025-03-10
Migrate compiler suite to the new test registration framework

ahesham-arm:
Contributes to #2181.


bashbaug (approved):
LGTM.  I verified that the same tests enumerate before and after this change, albeit in a different order.

PR #2325, opened 2025-03-10
Migrate multiple_device_context suite to the new test registration framework

ahesham-arm:
Contributes to #2181.


bashbaug (approved):

test_conformance/multiple_device_context/testBase.h
```diff
@@ -16,16 +16,8 @@
...
 
-#include "harness/compat.h"
-
-#include <stdio.h>
-#include <string.h>
-#include <sys/types.h>
-#include <sys/stat.h>
-
-#include "procs.h"
+#include "harness/conversions.h"
```
bashbaug, line 19: Not going to lose sleep over this, but it does seem a little odd to include the conversions header here.

kpet: Harness headers are a mess generally. They will need fixing but this is better tackled with a dedicated PR IMO.

kpet (approved):

PR #2330, opened 2025-03-11
Always enable beta extensions

kpet:
`CL_ENABLE_BETA_EXTENSIONS` was just introduced in https://github.com/KhronosGroup/OpenCL-Headers/pull/276
to guard provisional and other experimental extensions. The CTS currently assumes
all definitions are always present.


kpet:
It seems `CL_EXPERIMENTAL` is completely unused. Do we want to remove it?

bashbaug:
Should the CTS enable beta extensions by default?  I don't think some of the tests for newer extensions are setup to handle the case when the headers aren't present - I'm thinking of things like the new semaphore and command buffer tests.

kpet:
Yes, and we might want to change this. I'll enable it by default in this PR and create an issue for a more comprehensive solution.

bashbaug (approved):

kpet:
See https://github.com/KhronosGroup/OpenCL-CTS/issues/2331

PR #2332, opened 2025-03-11
Migrate integer_ops suite to the new test registration framework

MichaelRizkalla-arm:
Contributes to https://github.com/KhronosGroup/OpenCL-CTS/issues/2181.


test_conformance/integer_ops/test_intmad24.cpp
```diff
@@ -196,13 +193,13 @@ int test_integer_mad24(cl_device_id device, cl_context context, cl_command_queue
     cl_kernel kernel[2*NUM_PROGRAMS];
     size_t threads[1];
 
-    int                num_elements;
+    int n_elems;
```
bashbaug, line 196: Minor: I think we could reduce diffs and eliminate some possible confusion about the difference between `n_elems` and `num_elements` by removing `n_elems` and assigning to the `num_elements` function argument instead.

test_conformance/integer_ops/test_intmul24.cpp
```diff
@@ -182,13 +179,13 @@ int test_integer_mul24(cl_device_id device, cl_context context, cl_command_queue
     cl_kernel kernel[NUM_PROGRAMS*2];
     size_t threads[1];
 
-    int                num_elements;
+    int n_elems;
```
bashbaug, line 182: Same comment about `n_elems` and `num_elements` here.

bashbaug (approved):

kpet (approved):

PR #2334, opened 2025-03-12
Migrate extensions suites to the new test registration framework

ahesham-arm:
This covers:
* cl_ext_cxx_for_opencl
* cl_khr_command_buffer
* cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch
* cl_khr_dx9_media_sharing
* cl_khr_external_semaphore
* cl_khr_kernel_clock
* cl_khr_semaphore

Contributes to #2181.


test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -473,36 +472,31 @@ struct InterleavedEnqueueTest : public BasicCommandBufferTest
 
 } // anonymous namespace
 
-int test_multi_flag_creation(cl_device_id device, cl_context context,
-                             cl_command_queue queue, int num_elements)
+REGISTER_TEST(multi_flag_creation)
```
bashbaug, line 475: Since this file gets compiled and linked into both the `test_cl_khr_comand_buffer` and `test_cl_khr_mutable_dispatch` test executables, these tests are getting enumerated by both executables now, which we probably don't want.  I think we'll need a bit of refactoring so they only enumerate for the `test_cl_khr_command_buffer` executable, like they did previously.  Maybe they could move to a different test, or to a new file that is only built for `test_cl_khr_command_buffer`?

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
```diff
@@ -119,9 +119,7 @@ static cl_int get_device_semaphore_handle_types(
 }
 
 // Confirm the semaphores can be successfully queried
-int test_external_semaphores_queries(cl_device_id device, cl_context context,
-                                     cl_command_queue defaultQueue,
-                                     int num_elements)
+REGISTER_TEST_VERSION(external_semaphores_queries, Version(1, 2))
```
bashbaug, line 122: Minor (here and elsewhere): `Version(1, 2)` is the default, so we could just use `REGISTER_TEST` here.  I know this is porting over the explicit check for `Version(1, 2)` that existed previously, though...

bashbaug (changes requested):
Please check the duplicate test enumerations.  Cleaning up the explicit version checks is a nice-to-have, but not a must-have.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
ahesham-arm: I can do a sweep after all the tests are migrated to find and remove all of these unnecessary versioned test registrations.

bashbaug: Works for me.  Thanks!  I'll resolve this conversation.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
ahesham-arm: Moved to a separate file that is only included in the correct binary. I have made the changes in a separate commit for easier reviewing.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer_tests.cpp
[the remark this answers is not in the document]
ahesham-arm: I assume you meant adding the registration to `basic_command_buffer.cpp`, please check the conversation here: https://github.com/KhronosGroup/OpenCL-CTS/pull/2334#discussion_r1999899247 on why this was needed.

bashbaug (approved):

kpet (approved):

PR #2336, opened 2025-03-13
allocations: Scale number of work-items in relation to input size

lakshmih:
Make the number of work-items proportional to size of the allocation so we launch more work-items and do less work per work-item as buffer sizes go up with device capabilities. 

To test:
test_allocations multiple 5 buffer
test_allocations single 5 buffer

### DIFF
--- a/test_conformance/allocations/main.cpp
+++ b/test_conformance/allocations/main.cpp
@@ -26,6 +26,8 @@ typedef long long unsigned llu;
 
 #define REDUCTION_PERCENTAGE_DEFAULT 50
 
+#define BYTES_PER_WORK_ITEM 2048ULL
+
 int g_repetition_count = 1;
 int g_reduction_percentage = REDUCTION_PERCENTAGE_DEFAULT;
 int g_write_allocations = 1;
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
Works for me.

This does change the way this test works slightly, but I don't think coverage is negatively impacted.

test_conformance/allocations/main.cpp
```diff
@@ -183,12 +185,16 @@ int doTest(cl_device_id device, cl_context context, cl_command_queue queue,
...
-        number_of_work_items = 8192 * 2;
     }
 
     // Round to nearest MB.
     g_max_size &= (size_t)(0xFFFFFFFFFF00000ULL);
 
+    // Scales the number of work-items to keep the amount of bytes processed
+    // per work-item the same.
+    number_of_work_items =
+        std::max(g_max_size / BYTES_PER_WORK_ITEM, 8192ULL * 2ULL);
```
kpet, line 196: You could assert that `g_max_size` is a multiple of `BYTES_PER_WORK_ITEM` maybe to guard against changes to these assumptions in the future.

kpet (approved):

bashbaug:

PR #2341, opened 2025-03-18
Fix broken CI on MacOS

ahesham-arm:
LunarG has changed the SDK file hosted for MacOS v1.3.275.0 from a .zip file to a .dmg file with the old hyperlink transparently redirecting to the new one.

The script expects a .zip archive so it fails when it downloads a .dmg file.

Bump the Vulkan SDK version to 1.4.309.0, which is the latest version and is provided as a .zip archive.


bashbaug (approved):
We should merge this ASAP to unblock CI.

bashbaug:
Merging as discussed in the March 18th teleconference.

PR #2342, opened 2025-03-18
fix test_api min_max_image_buffer_size

### LINKED ISSUE #2245: test_api min_max_image_buffer_size
In that test we are comparing `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE` with `CL_DEVICE_MAX_MEM_ALLOC_SIZE/MAX_DEVICE_MEMORY_SIZE_DIVISOR`.

https://github.com/KhronosGroup/OpenCL-CTS/blob/5b3518096ca7b82854daadb2b2fae704fe2d9cb5/test_common/harness/deviceInfo.h#L53
```
/* We need to use a portion of available alloc size,
 * divide it to leave some memory for the platform. */
```

thus it means that if a device reports `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE = CL_DEVICE_MAX_MEM_ALLOC_SIZE/2 + 1`, the test is failing, while it feels that it should still be passing.

leaving half of the space of the platform is quite a lot, and it feels quite arbitrary.

Am I missing something?

rjodinchr:
This is fixing it for device exposing a big
CL_DEVICE_IMAGE_MAX_BUFFER_SIZE (more than
CL_DEVICE_MAX_MEM_ALLOC_SIZE/2)

Fix #2245

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -1066,8 +1066,7 @@ REGISTER_TEST(min_max_image_buffer_size)
     PASSIVE_REQUIRE_IMAGE_SUPPORT(device);
 
     /* Get the max memory allocation size, divide it */
-    maxAllocSize = get_device_info_max_mem_alloc_size(
-        device, MAX_DEVICE_MEMORY_SIZE_DIVISOR);
+    maxAllocSize = get_device_info_max_mem_alloc_size(device, 1);
 
     /* Get the max image array width */
     error =


lakshmih:
Reviewing this, might need additional changes.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1066,8 +1066,7 @@ REGISTER_TEST(min_max_image_buffer_size)
     PASSIVE_REQUIRE_IMAGE_SUPPORT(device);
 
     /* Get the max memory allocation size, divide it */
-    maxAllocSize = get_device_info_max_mem_alloc_size(
-        device, MAX_DEVICE_MEMORY_SIZE_DIVISOR);
+    maxAllocSize = get_device_info_max_mem_alloc_size(device, 1);
```
bashbaug, line 1069: I'm OK with this as-is, but if we wanted to be slightly more conservative, we could keep this divisor as-is and instead handle the case where the `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE` is greater than the (scaled) `CL_DEVICE_MAX_MEM_ALLOC_SIZE` slightly differently:

```diff
    pixelBytes = maxAllocSize / maxDimensionPixels;
    if (pixelBytes == 0)
    {
-        log_error("Value of CL_DEVICE_IMAGE_MAX_BUFFER_SIZE is greater than "
-                  "CL_MAX_MEM_ALLOC_SIZE so there is no way to allocate image "
-                  "of maximum size!\n");
-        return -1;
+        log_info(
+            "Note, the value of CL_DEVICE_IMAGE_MAX_BUFFER_SIZE is %zu pixels, "
+            "therefore the size of the allocated image may be larger than the  "
+            "scaled CL_DEVICE_MAX_MEM_ALLOC_SIZE of %zu bytes.\n",
+            maxDimensionPixels, maxAllocSize);
+        pixelBytes = 1;
    }
```

bashbaug (approved):

test_conformance/api/test_api_min_max.cpp
rjodinchr: That would fix my issue but I'm not sure I understand the idea behind it.
The current CL is weakening a bit the CTS to make sure we don't fail for a wrong reason. This proposal is weakening the CTS even more as we won't be able to fail at all (for the considered usecases), even for a good reason. Instead we would just get a message that  I find not very clear. Especially the `scaled` bit of it, which while being correct is not precise at all, and makes for a very mysterious message to users of the CTS.

rjodinchr: I think that this test aims at checking if we can create an image from buffer with the maximum size with the biggest number of channels.
I don't see why we should divide by 2 at the beginning. It feels that the specification says it should be possible. Thus we should test it and not put yourself in a smaller configuration to make sure everyone passes that test.

ahesham-arm: @rjodinchr Have you checked https://github.com/KhronosGroup/OpenCL-CTS/pull/1172 for the history behind this change?

rjodinchr: I think I disagree with the explanation in #1172.

Especially for that test (`min_max_image_buffer_size`). The test is really just allocating a buffer and then the image. It uses `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE` to know the size to use. We should not divide the size by 2 just to help an implementation.

To me it feels like some implementation may be exposing `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE` matching an hardware property related to the max buffer size, and assuming there are no limitation for any buffer to be associated to an image buffer. But there is one, `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE` should be equal to the max buffer + the size for the image.
If a driver is not considering the size of an image, it feels like a bug.

rjodinchr: Talking about it with @kpet , it seems that `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE` is the size above which the user should expect not to be able to create the image. But no guarantee that below it the user can create an image.
Then maybe the easiest way to fix the issue is to go forward with @bashbaug 's proposal.

A question remains: If the user cannot expect to create an image1d_buffer with a size below `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE`, why is the CTS testing for `CL_DEVICE_IMAGE_MAX_BUFFER_SIZE/2`. This is arbitrary. It feels like we should either explicit in the spec what is expected or remove that part of the test (testing the maximum).

rjodinchr:
@ahesham-arm @bashbaug

bashbaug (approved):

ahesham-arm (approved):

Issue #2142, opened 2024-11-05
missing testing for cl_ext_float_atomics


bashbaug:
This issue is to track missing tests for the `cl_ext_float_atomics` extension.

There is some testing for these extensions in the Vulkan CTS.

There are also some tests in the Intel Cassian test framework, see: https://github.com/intel/cassian/commit/239de5d8630bee09c0f3e136d4dab6e508bf80e3

A good first step would be to evaluate what is tested by the other frameworks, see what may be useful for the OpenCL CTS, and write up a brief test plan.

shajder:
Below is the proposition for a work plan on `cl_ext_float_atomics` and a few questions, I will appreciate discussions and your answers, thanks!

1. Adaptation of `c11_atomics` test suite in order to add support for `cl_ext_float_atomics` by including new generic cases for data types `fp16`/`fp32`/`fp64` in classes:

- [x] [CBasicTestStore](https://github.com/KhronosGroup/OpenCL-CTS/blob/9f63decb9c85875c3f285005a508037233f530b9/test_conformance/c11_atomics/test_atomics.cpp#L27) with atomic_half
- [x] [CBasicTestLoad](https://github.com/KhronosGroup/OpenCL-CTS/blob/9f63decb9c85875c3f285005a508037233f530b9/test_conformance/c11_atomics/test_atomics.cpp#L295) with atomic_half
- [x] [CBasicTestExchange](https://github.com/KhronosGroup/OpenCL-CTS/blob/9f63decb9c85875c3f285005a508037233f530b9/test_conformance/c11_atomics/test_atomics.cpp#L470) with atomic_half
- [x] [CBasicTestFetchAdd](https://github.com/KhronosGroup/OpenCL-CTS/blob/9f63decb9c85875c3f285005a508037233f530b9/test_conformance/c11_atomics/test_atomics.cpp#L1055) with atomic_float
- [x] [CBasicTestFetchAdd](https://github.com/KhronosGroup/OpenCL-CTS/blob/9f63decb9c85875c3f285005a508037233f530b9/test_conformance/c11_atomics/test_atomics.cpp#L1055) with atomic_double
- [x] [CBasicTestFetchAdd](https://github.com/KhronosGroup/OpenCL-CTS/blob/9f63decb9c85875c3f285005a508037233f530b9/test_conformance/c11_atomics/test_atomics.cpp#L1055) with atomic_half
[...]

PR #2345, opened 2025-03-20
Added support for cl_ext_float_atomics in CBasicTestFetchAdd with atomic_float

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchAdd with support for atomic_float.

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -176,12 +176,12 @@ public:
     {
         return false;
     }
-    virtual bool VerifyExpected(const HostDataType &expected,
-                                const HostAtomicType *const testValue,
-                                cl_uint whichDestValue)
+    virtual bool
+    IsTestNotAsExpected(const HostDataType &expected,
+                        const std::vector<HostAtomicType> &testValues,
+                        cl_uint whichDestValue)
     {
-        if (testValue != nullptr) return expected != testValue[whichDestValue];
-        return true;
+        return expected != testValues[whichDestValue];
     }
     virtual bool GenerateRefs(cl_uint threadCount, HostDataType *startRefValues,
                               MTdata d)
@@ -1451,7 +1451,7 @@ int CBasicTest<HostAtomicType, HostDataType>::ExecuteSingleTest(
                            startRefValues.size() ? &startRefValues[0] : 0, i))
             break; // no expected value function provided
 
-        if (VerifyExpected(expected, destItems.data(), i))
+        if (IsTestNotAsExpected(expected, destItems, i))
         {
             std::stringstream logLine;
             logLine << "ERROR: Result " << i
--- a/test_conformance/c11_atomics/test_atomics.cpp
+++ b/test_conformance/c11_atomics/test_atomics.cpp
@@ -1232,20 +1232,21 @@ public:
 
         return true;
     }
-    bool VerifyExpected(const HostDataType &expected,
-                        const HostAtomicType *const testValue,
-                        cl_uint whichDestValue) override
+    bool IsTestNotAsExpected(const HostDataType &expected,
+                             const std::vector<HostAtomicType> &testValues,
+                             cl_uint whichDestValue) override
     {
         if (std::is_same<HostDataType, HOST_ATOMIC_FLOAT>::value)
         {
-            if (whichDestValue == 0 && testValue != nullptr)
+            if (whichDestValue == 0)
                 return std::abs((HOST_ATOMIC_FLOAT)expected
-                                - testValue[whichDestValue])
+                                - testValues[whichDestValue])
                     > max_error_fp32;
         }
         return CBasicTestMemOrderScope<
-            HostAtomicType, HostDataType>::VerifyExpected(expected, testValue,
-                                                          whichDestValue);
+            HostAtomicType, HostDataType>::IsTestNotAsExpected(expected,
+                                                               testValues,
+                                                               whichDestValue);
     }
     bool VerifyRefs(bool &correct, cl_uint threadCount, HostDataType *refValues,
                     HostAtomicType *finalValues) override


test_conformance/c11_atomics/common.h
```diff
@@ -174,6 +176,13 @@ class CBasicTest : CTest {
     {
         return false;
     }
+    virtual bool VerifyExpected(const HostDataType &expected,
+                                const HostAtomicType *const testValue,
+                                cl_uint whichDestValue)
+    {
+        if (testValue != nullptr) return expected != testValue[whichDestValue];
+        return true;
+    }
```
bashbaug: This function confused me - it returns `true` when the results do NOT verify.  I think we should pick a different name and think carefully when it should return `true` and when it should return `false`.

Additionally:
* Do we ever expect to call this function with `testValue` equal to `nullptr`, or does this mean that something went horribly wrong?
* Can we pass an `std::vector` for `testValue`, specifically a `const std::vector&`, or are there still cases where we need to pass a raw pointer?
* `testValue` is really an array of test values, so it should be plural.
* (Minor) Since the size of our expected value is small, we can probably just pass it by value vs. a const reference.

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1068,61 +1071,209 @@ REGISTER_TEST(svm_atomic_compare_exchange_weak)
 template <typename HostAtomicType, typename HostDataType>
 class CBasicTestFetchAdd
     : public CBasicTestMemOrderScope<HostAtomicType, HostDataType> {
+
+    double min_range;
+    double max_range;
+    double max_error_fp32;
+    std::vector<HostDataType> ref_vals;
```
bashbaug, line 1078: It doesn't look like `ref_vals` is used outside of `GenerateRefs`.  Does it need to be a class member variable?

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1068,61 +1071,209 @@ REGISTER_TEST(svm_atomic_compare_exchange_weak)
...
+            {
+                ref_vals.resize(threadCount);
+
+                for (cl_uint i = 0; i < threadCount; i++)
+                    ref_vals[i] = get_random_float(min_range, max_range, d);
+
+                memcpy(startRefValues, ref_vals.data(),
+                       sizeof(HostDataType) * ref_vals.size());
+
+                // Estimate highest possible summation error for given set.
```
bashbaug, line 1114: I don't understand what this is doing - can you please explain and then add a comment to the code for future maintainers?

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1068,61 +1071,209 @@ REGISTER_TEST(svm_atomic_compare_exchange_weak)
...
-            + ");\n";
+
+        if constexpr (std::is_same_v<HostDataType, HOST_ATOMIC_FLOAT>)
+        {
+            return "  atomic_fetch_add" + postfix + "(&destMemory[0], ("
+                + DataType().AddSubOperandTypeName() + ")oldValues[tid]"
+                + memoryOrderScope + ");\n"
+                + "  oldValues[tid] = atomic_fetch_add" + postfix
+                + "(&destMemory[tid], (" + DataType().AddSubOperandTypeName()
+                + ")0" + memoryOrderScope + ");\n";
```
bashbaug, line 1167: I'm a bit surprised to see this difference.  Is there a reason why the float atomic add kernel is fundamentally different than the non-float types?

Assuming the difference is intended, will we at least handle the HALF and DOUBLE atomics the same way as we handle these FLOAT atomics?

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1068,61 +1071,209 @@ REGISTER_TEST(svm_atomic_compare_exchange_weak)
...
+        }
+        return CBasicTestMemOrderScope<
+            HostAtomicType, HostDataType>::ExecuteSingleTest(deviceID, context,
+                                                             queue);
+    }
+    cl_uint NumResults(cl_uint threadCount, cl_device_id deviceID) override
+    {
+        if constexpr (std::is_same_v<HostDataType, HOST_ATOMIC_FLOAT>)
+        {
+            return threadCount;
```
bashbaug, line 1295: This is another difference I'm surprised to see.  Why is there a different number of results for float atomics vs. non-float atomics?

test_conformance/c11_atomics/test_atomics.cpp
shajder: I've created a fork kernel for a couple of reasons, first of all, please take a look at `atomic_int` version of kernel:

```
__kernel void test_atomic_kernel(uint threadCount, uint numDestItems, __global int *finalDest, __global int *oldValues, volatile __local atomic_int *destMemory)
{
  uint  tid = get_global_id(0);


              // initialize atomics not reachable from host (first thread
              // is doing this, other threads are waiting on barrier)
              if(get_local_id(0) == 0)
                for(uint dstItemIdx = 0; dstItemIdx < numDestItems; dstItemIdx++)
                {
                atomic_store_explicit(destMemory+dstItemIdx,
                                      finalDest[dstItemIdx],
                                      memory_order_relaxed,
                                      memory_scope_work_group);    }
  barrier(CLK_LOCAL_MEM_FENCE);

  oldValues[tid] = atomic_fetch_add_explicit(&destMemory[0], (int)tid + 3, memory_order_relaxed, memory_scope_work_group);
  atomic_fetch_add_explicit(&destMemory[0], (int)tid + 3, memory_order_relaxed, memory_scope_work_group);
  atomic_fetch_add_explicit(&destMemory[0], (int)tid + 3, memory_order_relaxed, memory_scope_work_group);
  atomic_fetch_add_explicit(&destMemory[0], ((int)tid + 3) << (sizeof(int)-1)*8, memory_order_relaxed, memory_scope_work_group);

  // Copy final values to host reachable buffer
  barrier(CLK_LOCAL_MEM_FENCE);
  if(get_local_id(0) == 0) // first thread in workgroup
    for(uint dstItemIdx = 0; dstItemIdx < numDestItems; dstItemIdx++)

                finalDest[dstItemIdx] =
                                         memory_order_relaxed,
                                         memory_scope_work_group);
}

```

Obviously it was implemented for integer atomic values, the original implementation adds a hardcoded value of `tid + 3` to a single atomic variable, `destMemory[0]` (even though `threadCount` kernels are launched, they all read/write this one atomic variable).

I have extended this logic to use an array of atomic floats of size `threadCount`. This is in connection with your question about modifying `NumResults` - we are now verifying the fetch operation across all threads.

In other words, instead of adding the same hardcoded integer `threadCount` times, I assumed that we want to utilize the full range of an array with random values and store these in a single atomic variable. Let me know if this approach fits better to `atomic_float`. I will pause further changes based on your other remarks, as this aspect is crucial for subsequent steps.

bashbaug: OK, I think the equivalent kernel for `atomic_float` looks something like this (run through clang-format to clean up the whitespace):

```c++
__kernel void test_atomic_kernel(uint threadCount, uint numDestItems,
                                 __global float *finalDest,
                                 __global float *oldValues,
                                 volatile __local atomic_float *destMemory) {
  uint tid = get_global_id(0);

  // initialize atomics not reachable from host (first thread
  // is doing this, other threads are waiting on barrier)
  if (get_local_id(0) == 0)
    for (uint dstItemIdx = 0; dstItemIdx < numDestItems; dstItemIdx++) {
      atomic_store_explicit(destMemory + dstItemIdx, finalDest[dstItemIdx],
                            memory_order_relaxed, memory_scope_work_group);
    }
  barrier(CLK_LOCAL_MEM_FENCE);

  atomic_fetch_add_explicit(&destMemory[0], (float)oldValues[tid],
                            memory_order_relaxed, memory_scope_work_group);
  oldValues[tid] = atomic_fetch_add_explicit(&destMemory[tid], (float)0,
                                             memory_order_relaxed,
                                             memory_scope_work_group);

  // Copy final values to host reachable buffer
  barrier(CLK_LOCAL_MEM_FENCE);
  if (get_local_id(0) == 0) // first thread in workgroup
    for (uint dstItemIdx = 0; dstItemIdx < numDestItems; dstItemIdx++)

      finalDest[dstItemIdx] =
          atomic_load_explicit(destMemory + dstItemIdx, memory_order_relaxed,
                               memory_scope_work_group);
}
```

Is this what's expected?  Note:

* The first `atomic_fetch_add_explicit` is still updating the single atomic variable at `destMemory[0]`.
* The second `atomic_fetch_add_explicit` that's setting `oldValues` **is** atomically updating a different atomic variable per work-item (`destMemory[tid]`), **but** because it's atomically adding zero it's a nop except for some very specific floating-point values.

To be clear, this might be just fine, especially if we're reusing the same logic for half-precision and double-precision floats.  The difference surprised me, though, so I figured it's worth confirming that the difference is intentional.

shajder: Okay, I see one problem, let me explain what was the intention of each line: 

- First `atomic_fetch_add_explicit` modifies the same destMemory[0] but uses random values stored in oldValues[tid], that was the intention indeed.
- Second `atomic_fetch_add_explicit` was intended to test **only fetch** operation for all threads. At a closer look I see this operation is not verified on host side which may be confusing. Below correction will fix that:

```
  oldValues[tid] = atomic_fetch_add_explicit(&destMemory[tid], (float)oldValues[tid],
                                             memory_order_relaxed,
                                             memory_scope_work_group);
  if (tid != 0)
      atomic_store_explicit(&destMemory[tid], oldValues[tid],
                        memory_order_relaxed, memory_scope_work_group);

```
explanation: in the first line start value (0) lands in oldValues[tid] and destMemory[tid] is now different from start value, next for non-zero `tid`s we store fetched values which will be tested on the host side. Does it make sense ? If there is no point in testing fetch operation itself we may remove the second part and revert `NumResults` to default which will certainly simplify the code.

test_conformance/c11_atomics/test_atomics.cpp
shajder: With one `EXECUTE_TEST` call, several test cases are launched using different parameters, such as the number of threads. The `ref_vals` container stores previous generations to avoid repeating the same process (as long as requested number of threads is the same as previously).

test_conformance/c11_atomics/test_atomics.cpp
shajder: With one `EXECUTE_TEST` call, several test cases are launched using different parameters, such as the number of threads - to avoid repeating generation of **start reference values** over and over (mostly due to computation of maximum error) `ref_vals` stores result of previous launch and if possible copies those values for next tests.

test_conformance/c11_atomics/test_atomics.cpp
shajder: Just to prevent the point from getting lost, here is a summary of the background investigation:

The existing logic for testing integral atomic types lacks verification of the fetch operation:

`  oldValues[tid] = atomic_fetch_add_explicit(&destMemory[0], (int)tid + 3, memory_order_relaxed, memory_scope_work_group);`

The `oldValues[tid]` is never verified on the host side. Furthermore, with the current structure, it is impossible to verify the reference values, so a correction is necessary. The same issue applies to `CBasicTestFetchSub`, `CBasicTestFetchMin`, and `CBasicTestFetchMax`. This can be fixed using a similar approach as proposed in this PR.

As for this PR, CBasicTestFetchAdd::VerifyRefs is now overridded to verify fetch operation of atomic_fetch_add_explicit calls.

test_conformance/c11_atomics/test_atomics.cpp
shajder: Related to the discussion in the [thread](https://github.com/KhronosGroup/OpenCL-CTS/pull/2345#discussion_r2176496112) above.

test_conformance/c11_atomics/common.h
shajder: Corrected

bashbaug:
Discussed in the August 12th teleconference.  OK to merge once the discussions have been resolved.

PR #2346, opened 2025-03-25
Increase column limit to improve readability given

joshqti:


EwanC:
Leaving the discussion about if 80 or 120 char line length is preferable to the side. If we make this change I think we should consider whether we clang-format the whole repo to the new line line length as part of this PR. If we don't then we'll end up with a mix of line lengths in the codebase as new PRs will be formatted to the longer line length to pass clang-format pre-commit CI, but the rest of the code will still be on the old line length. If we do clang-format the whole repo, then it'll introduce conflicts with existing PRs, and affect git blame.

svenvh:
> Leaving the discussion about if 80 or 120 char line length is preferable to the side. If we make this change I think we should consider whether we clang-format the whole repo to the new line line length as part of this PR. If we don't then we'll end up with a mix of line lengths in the codebase as new PRs will be formatted to the longer line length to pass clang-format pre-commit CI, but the rest of the code will still be on the old line length. If we do clang-format the whole repo, then it'll introduce conflicts with existing PRs, and affect git blame.

Thanks for this comment, I fully agree with your concerns.  Some tests (e.g. bruteforce) fully comply with the current clang-format rules thanks to a previous reformatting commit; changing the line length would break that until another reformatting is committed.

If we decide to go for a big reformat, we might want to revisit the entire .clang-format configuration (e.g. better aligning the various `BraceWrapping`, `AllowShort*` settings).

joshqti:
As discussed on April 8th, conversions is a specific case where 120 characters improves readability significantly. https://github.com/KhronosGroup/OpenCL-CTS/pull/1555/files?diff=split

bashbaug:
The conversions PR referenced above doesn't looks too bad to me, honestly.

If there are places that would benefit from a larger column limit, could we just wrap them in `// clang-format off` and `// clang-format on`?  I think this would be much less disruptive than increasing the column limit globally, especially now that most of the codebase has been reformatted to 80 characters.

bashbaug:
I think we should close this PR, personally, but I'm at least going to take the "focused review" label off for now so it comes off of our lists.

bashbaug:
Discussed in the August 26th teleconference.  We are not planning to pursue this further.

PR #2349, opened 2025-03-25
Stop using optimally-tiled images in external memory tests

kpet:
Support for optimally-tiled images and tiling inference in general is implementation-defined. It should not be relied upon in the CTS.

Also build the code base as C++17 for std::optional. Many Khronos projects are now using C++17 and the transition in the CTS is IMO overdue.


kpet:
As discussed in the 2025/03/25 memory TSG, here's an RFC PR to remove all Vulkan/CL interop testing using optimally-tiled images. This PR seems to pass the few simple tests I've done so far.

joshqti (commented):
Change looks good and passes for us

nikhiljnv (approved):
LGTM.

nikhiljnv:
Merging as discussed in memory subgroup call on April 15th, 2025.

PR #2354, opened 2025-03-31
make fp16 consistency more friendly

### LINKED ISSUE #2339: test_api consistency_requirements_fp16
`consistency_requirements_fp16` & `consistency_requirements_fp64` tests do not behave the same way when the extension is not supported for `CL_DEVICE_HALF_FP_CONFIG` & `CL_DEVICE_DOUBLE_FP_CONFIG`.

`fp16` expects `clGetDeviceInfo` to return an error, while `fp64` expects a return value of `0`.

It's not clear to me in the specification that `fp16` needs to return an error, or that `fp64` should not return an error.

Could we clarify/unify the behaviors?

rjodinchr:
Allow a implementation not exposing `cl_khr_fp16` to support `CL_DEVICE_HALF_FP_CONFIG`. But it that case, it has to return zero, as it is the case for `cl_khr_fp64`.

Fix #2339


kpet:
I'm reaching the same conclusion as you after a quick scan through the specs. There is nothing that requires an error to be returned by `clGetDeviceInfo` for `CL_DEVICE_HALF_FP_CONFIG` when `cl_khr_fp16` is not supported, so no requirement for implementations to behave in a certain way that the CTS should test.

The `CL_DEVICE_HALF_FP_CONFIG` enum is required by the `cl_khr_fp16` extension in the XML but, in theory at least, another extension or core version could require it and we don't generally test that all info enums that are not required by any supported extension are rejected. IMO it makes sense to keep things that way. The purpose of negative tests is to verify that implementations behave correctly in specified error conditions. This is not one of them.

rjodinchr:
So, should we land this PR as it is, or should I modify it to just remove the whole "else" section?

kpet:
I'd suggest removing testing of `CL_DEVICE_HALF_FP_CONFIG` in the "else". Checking the two other queries return 0 when the extension is not supported is required by the spec.

kpet (approved):

bashbaug (approved):

bashbaug:

PR #2356, opened 2025-04-01
Require C++17

kpet:
This enables many useful newer C++ features. C++17 is already required by many components of the Khronos ecosystem.


svenvh (approved):

bashbaug (approved):
Merging as discussed in the April 8th teleconference.

PR #2357, opened 2025-04-02
Added support for cl_ext_float_atomics in CBasicTestFetchMin/Max with atomic_half

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchMin/CBasicTestFetchMax with support for atomic_half.


test_conformance/c11_atomics/common.cpp
```diff
@@ -194,14 +194,28 @@ template<> cl_int AtomicTypeExtendedInfo<cl_int>::MinValue() {return CL_INT_MIN;
 template<> cl_uint AtomicTypeExtendedInfo<cl_uint>::MinValue() {return 0;}
 template<> cl_long AtomicTypeExtendedInfo<cl_long>::MinValue() {return CL_LONG_MIN;}
 template<> cl_ulong AtomicTypeExtendedInfo<cl_ulong>::MinValue() {return 0;}
-template<> cl_float AtomicTypeExtendedInfo<cl_float>::MinValue() {return CL_FLT_MIN;}
+template <> cl_half AtomicTypeExtendedInfo<cl_half>::MinValue()
+{
+    return cl_half_from_float(CL_HALF_MIN, gHalfRoundingMode);
```
bashbaug, line 199: Sorry, missed this previously: should this return `CL_HALF_MIN`, or `-CL_HALF_MAX`?

`CL_HALF_MIN` is the smallest normalized positive half value, so it is a number very close to zero.  `-CL_HALF_MAX` is the half value closest to negative infinity.  I think we want `-CL_HALF_MAX` here, and similar changes for floats and doubles, but I'm not 100% sure.

I'm fine fixing this in a separate PR, since it will affect all three floating-point types.

shajder: @bashbaug I presume the same comment refers to `float` and `double` cases - all fixed [here](https://github.com/KhronosGroup/OpenCL-CTS/pull/2361/commits/8f4feccd6b4f035f3678443ace5dd03a9f7e8d56)

PR #2358, opened 2025-04-02
SPIR-V tests for scalar printf - fix for Windows OS

gwawiork:


test_conformance/spirv_new/test_extinst_printf.cpp
```diff
@@ -94,7 +94,12 @@ struct StreamGrabber
                 results.clear();
                 results.resize(filesize);
                 is.read(&results[0], filesize);
-
+                // Normalize line endings: replace \r\n with \n
```
bashbaug, line 97: Hmm, I think this is problem is happening because we are writing to this file as a text file ("w") and reading from it as a binary file.  Can you try removing `std::ios::binary` above, and see if this eliminates the need for this "fixup" code?

Note that the analogous code in "test_printf.cpp" is opening its file as a text file ("r"), albeit using C APIs rather than C++ APIs.

If we need a more thorough line ending solution - maybe to normalize the reference strings in addition to the result strings? - I think I'd like to see this refactored into a separate function, but let's only cross that bridge if we need to.

gwawiork: Hi. Removing `std::ios::binary` has nothing change. Still errors. I've added separate function to normalize result and reference strings.

bashbaug (approved):

bashbaug:
@svenh would you mind giving this a quick review?  We're going to need this fix for an upcoming CTS submission - thanks!

svenvh (approved):

PR #2359, opened 2025-04-03
Add test_spir installation target

MichaelRizkalla-arm:
`test_spir` and its data are not installed correctly when building the installation target.
This change makes sure both the binary and the test data are installed correctly to `${CMAKE_INSTALL_BINDIR}/$<CONFIG>`.


bashbaug (approved):

PR #2365, opened 2025-04-07
test_vulkan: switch defaults on device local memory

joshqti:
Switch the test_vulkan default to use device local memory


bashbaug:
Discussed in the June 3rd memory subgroup:

* @nikhiljnv can you please check that this is OK on your implementation?
* @joshqti will check to ensure that when device local memory is used the right checks for host visibility are present.

PR #2367, opened 2025-04-10
Added support for cl_ext_float_atomics in CBasicTestFetchSub with atomic_float

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchSub with support for atomic_float.

### DIFF
--- a/test_conformance/c11_atomics/test_atomics.cpp
+++ b/test_conformance/c11_atomics/test_atomics.cpp
@@ -16,10 +16,13 @@
 #include "harness/testHarness.h"
 #include "harness/kernelHelpers.h"
 #include "harness/typeWrappers.h"
+#include "harness/conversions.h"
 
 #include "common.h"
 #include "host_atomics.h"
 
+#include <algorithm>
+#include <numeric>
 #include <sstream>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1811,6 +1892,11 @@ static int test_atomic_fetch_sub_generic(cl_device_id deviceID,
 
     if (gFloatAtomicsSupported)
     {
+        CBasicTestFetchSub<HOST_ATOMIC_FLOAT, HOST_FLOAT> test_float(
+            TYPE_ATOMIC_FLOAT, useSVM);
```
bashbaug, line 1896: This is a fairly minor issue in the grand scheme of things, but a "summary" gets printed for each of the atomic types for each atomic test.  Example:

```
  222 tests executed successfully for atomic_ptrdiff_t
  222 tests executed successfully for atomic_size_t
  222 tests executed successfully for atomic_uintptr_t
  222 tests executed successfully for atomic_intptr_t
  222 tests executed successfully for atomic_ulong
  222 tests executed successfully for atomic_long
  222 tests executed successfully for atomic_uint
  222 tests executed successfully for atomic_int
```

This gets printed by the test object destructor for a specific type.  See:

https://github.com/KhronosGroup/OpenCL-CTS/blob/79d98433cc168a3918b2163ddf7612125211d387/test_conformance/c11_atomics/common.h#L166-L168

Since we're declaring these test objects for float atomics in their own scope, though, the "summary" for them gets printed before a bunch of specific test output for `size_t`, `ptrdiff_t`, etc, and not with all of the other atomic types.

Could we:

* At least declare these test objects in the outer-most scope, even if we keep the `EXECUTE_TEST` part inside of this "if" check?
* Or, since we need to check that the specific atomic operations are supported anyhow, could we eliminate this outermost "if" check entirely?

(I feel like I may have written this already, my apologies if so!)

bashbaug:
Merging as discussed in the October 28th teleconference.

PR #2369, opened 2025-04-11
conversions: use C++17 to avoid Wformat warnings

svenvh:
Using C++17's `if constexpr` avoids Wformat warnings in this template function, as it is now made explicit that the various format strings are only used with appropriate types.


kpet (approved):
A nice first use of C++17 :).

bashbaug (approved):

kpet:
Trivial PR with two approvals. Merging.

PR #2376, opened 2025-04-18
WIP: add a compiler test for extension defines

bashbaug:
Here is a quick test I wrote up to check if OpenCL implementations are setting their defines for extensions properly, see: https://github.com/KhronosGroup/OpenCL-Docs/issues/1328

I'll keep this as a draft while we figure out exactly how these extension defines should behave, and in particular while we figure out if some extensions should be defined only for specific OpenCL C versions.

This may ultimately replace the existing "compiler defines for extensions" test, but I have it as a separate test, for now.

### DIFF
--- a/test_conformance/compiler/CMakeLists.txt
+++ b/test_conformance/compiler/CMakeLists.txt
@@ -10,6 +10,7 @@ set(${MODULE_NAME}_SOURCES
     test_opencl_c_versions.cpp
     test_image_macro.cpp
     test_compiler_defines_for_extensions.cpp
+    test_compiler_defines_for_extensions_new.cpp
     test_pragma_unroll.cpp
     test_unload_platform_compiler.cpp
     test_feature_macro.cpp
--- a/test_conformance/compiler/test_compiler_defines_for_extensions_new.cpp
+++ b/test_conformance/compiler/test_compiler_defines_for_extensions_new.cpp
@@ -1,0 +1,193 @@
+//
+// Copyright (c) 2025 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "testBase.h"
+#include <vector>
+
+// clang-format off
+// These are OpenCL C extensions that should (or should not) have OpenCL C
+// defines, depending whether the extension is supported.
+static const std::vector<std::string> opencl_c_extensions = {
+    "cl_khr_3d_image_writes",
+    "cl_khr_byte_addressable_store",
+    "cl_khr_depth_images",
+    "cl_khr_device_enqueue_local_arg_types",
+    "cl_khr_extended_async_copies",
+    "cl_khr_extended_bit_ops",
+    "cl_khr_fp16",
+    "cl_khr_fp64",
+    "cl_khr_gl_depth_images",
+    "cl_khr_gl_msaa_sharing",
+    "cl_khr_global_int32_base_atomics",
+    "cl_khr_global_int32_extended_atomics",
+    "cl_khr_int64_base_atomics",
+    "cl_khr_int64_extended_atomics",
+    "cl_khr_integer_dot_product",
+    "cl_khr_kernel_clock",
+    "cl_khr_local_int32_base_atomics",
+    "cl_khr_local_int32_extended_atomics",
+    "cl_khr_mipmap_image",
+    "cl_khr_mipmap_image_writes",
+    "cl_khr_select_fprounding_mode",
+    "cl_khr_srgb_image_writes",
+    "cl_khr_subgroup_ballot",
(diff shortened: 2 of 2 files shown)


karolherbst:
Looks like rusticl defines `cles_khr_int64` in a non embedded platform.. oops

The other thing that stands out is `cl_khr_depth_images` for OpenCL C 1.1, but that looks like a clang "bug": https://github.com/llvm/llvm-project/blob/main/clang/include/clang/Basic/OpenCLExtensions.def#L77

So apparently when somebody uses the opencl-extension option, clang checks against the target OpenCL C and doesn't enable an extension. So if OpenCL C 1.1 is set, then clang won't advertise `cl_khr_depth_images` even if `cl_khr_depth_images` was enabled via passed arguments. How should we deal with that?

karolherbst:
> I don't think the issue is specific to `cl_khr_depth_images`, either. Example: https://godbolt.org/z/rjT8Y341K
> 
> Is this whole notion of when an extension is "available" mistaken, or at the very least, overused?
> 
> ```c++
> // Declaration helpers
> #define OPENCL_EXTENSION(ext, pragma, avail) OPENCL_GENERIC_EXTENSION(ext, pragma, avail, 0U, 0U)
> ...
> // OpenCL 1.2.
> OPENCL_EXTENSION(cl_khr_depth_images, true, 120)
> OPENCL_EXTENSION(cl_khr_gl_msaa_sharing,true, 120)
> ```

yeah it's not. I just didn't figure out where the actual code is checking for it. Though I think there might be valid reasons to do it, even if it's just for simplification. E.g. `image1d_t` and other image types only got added with 1.2. So if an extension adds builtins using those types, certain overloads shouldn't be advertised I guess?

So I'd assume this is mostly done to make it simpler, but one could advertise a subset of those extensions, it's probably just more complex to do than a simple version check.

bashbaug:
OK, yeah, I think pretty much all of the extensions that are only "available" in newer OpenCL C versions are suspect.  Here is an updated example: https://godbolt.org/z/edsdjd85W

Some things I think we need to decide are:

1. Do we actually want any of these extensions to be tied to newer OpenCL C versions, or should we consider this a bug and support the extensions (to whatever extent is possible) on all OpenCL C versions?
2. Assuming this is a bug, can we simply fix Clang, or are there backward compatibility concerns?

We may want to move this discussion back to the spec issue for more visibility.

svenvh:
> 1. Do we actually want any of these extensions to be tied to newer OpenCL C versions, or should we consider this a bug and support the extensions (to whatever extent is possible) on all OpenCL C versions?

I suppose we'll have to see per extension how far we can stretch its range of OpenCL C versions; but if an extension doesn't restrict itself to a specific version clang ought to support it for all versions.

> 2. Assuming this is a bug, can we simply fix Clang, or are there backward compatibility concerns?

No objections from my side as I don't expect that a frontend allowing extensions to be used on older OpenCL versions would be problematic (as long as allowing the extension on earlier versions makes sense of course).

karolherbst:
Yeah, I agree that runtimes should be free to advertise as much as possible of such extensions for applications targeting older OpenCL C versions, e.g. ignore `image1d_t` overloads in e.g. `cl_khr_mipmap_image`, but still allow lods for `image2d_t`.

I think it might lead to runtimes having to interpret the spec a bit to see how far they can stretch it, which could lead to some inconsistency. But I think that's still better than not having the feature available. Granted for anything below 1.2 it will practically not matter much, but I think it's important to clarify for future extensions and future OpenCL C versions.

Could either update the spec/extensions to specify what won't be advertised on older versions, or just hope for the best.

wenju-he:
> Here is an updated example: https://godbolt.org/z/edsdjd85W

cl_intel_subgroups* and cl_intel_device_side_avc_motion_estimation extension specs state that OpenCL 1.2 is required.

bashbaug:
> cl_intel_subgroups* and cl_intel_device_side_avc_motion_estimation extension specs state that OpenCL 1.2 is required.

This is a bit nitpicky, but we should probably clarify that the requirement is for OpenCL **C** 1.2 **or newer**.

I'll add checks for these extensions so they are not expected for earlier OpenCL C versions.  I assume that's what we want to do?  I can't think of a reason why these extensions wouldn't work pre-OpenCL C 1.2, but the OpenCL (C) 1.2 requirement has been explicitly documented for many years now, so it's probably safest to just keep it.

bashbaug:
> I'll add checks for these extensions so they are not expected for earlier OpenCL C versions.

Actually, it looks like I don't need to change anything in the current test for these extensions, because the current test is only checking for KHR OpenCL C extensions.  If we ever expanded the tests to cover vendor extension, though, we would need version checks.

Note, I do not see OpenCL (C?) version requirements documented for any of the KHR OpenCL C extensions.

wenju-he:
> Somewhat related, looking at:
> 
> https://github.com/llvm/llvm-project/blob/d4f39b9bd64c17d1b0d8374159879b9c3a625bf2/clang/include/clang/Basic/OpenCLExtensions.def#L80
> 
> There are a bunch of other extensions that are currently tied to OpenCL C 2.0. Why is this? I don't see the OpenCL C 2.0 requirement listed anywhere in the OpenCL spec, and it means that these extensions are not exposed for earlier OpenCL C versions.
> 
> Here is an even more up-to-date example showing this issue: https://godbolt.org/z/nKfGdnhP9

I have created PR to make khr extension available for all OpenCL C versions: https://github.com/llvm/llvm-project/pull/175120

bashbaug:
Most of Intel extensions should be updated on the registry to remove the OpenCL 1.2 dependency now.  There were a few really old ones that still have the text, but I don't think they impact OpenCL C, so I've left them alone.  Please let me know if you find any others that should be updated.  Thanks!

PR #2377, opened 2025-04-18
fixes test_opencl_c_versions logic error

bashbaug:
Found while developing #2376.

Parts of "test_opencl_c_versions" were intending to build programs for different OpenCL C versions, but unfortunately the helper function the test was calling only created the program and did not build it.  Switch to a helper function that builds the program, instead.


kpet (approved):
Good catch!

lakshmih (approved):

bashbaug:
Merging (offline) as discussed in the June 24th teleconference.

PR #2378, opened 2025-04-18
Add missing coverage to api::zero_sized_enqueue

kpet:
- Cases with a NULL global_work_size were not tested
- The specification states that zero-sized enqueues behave similarly to marker commands. Test that events for NDRange kernel commands have the correct CL_COMMAND_NDRANGE_KERNEL command type to guard against implementations naively using a marker command.
- Tidy up the code a bit and add missing error checking.

### DIFF
--- a/test_conformance/api/test_zero_sized_enqueue.cpp
+++ b/test_conformance/api/test_zero_sized_enqueue.cpp
@@ -1,6 +1,6 @@
 //
-// Copyright (c) 2017 The Khronos Group Inc.
-// 
+// Copyright (c) 2017-2025 The Khronos Group Inc.
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


bashbaug (approved):

lakshmih:
Reviewing

hvdijk:
Hi, I came across this as our implementation fails this test, but I believe we are doing the right thing. Although the spec includes "kernels executed over ranges of no work-items" as an example of a command that may be submitted and effectively do nothing, the spec for `clEnqueueNDRangeKernel` still says:

> [clEnqueueNDRangeKernel](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueNDRangeKernel) returns [CL_SUCCESS](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_SUCCESS) if the kernel-instance was successfully queued. Otherwise, it returns one of the following errors:
>
> [...]
>
> [CL_INVALID_GLOBAL_WORK_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_GLOBAL_WORK_SIZE) if global_work_size is NULL or if any of the values specified in global_work_size[0], …​ global_work_size[work_dim - 1] are 0. Returning this error code under these circumstances is [deprecated by](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#unified-spec) version 2.1.

[The meaning of "deprecation"](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#_glossary) is:

> **Deprecation**
>
> Existing features are marked as deprecated if their usage is not recommended as that feature is being de-emphasized, superseded and may be removed from a future version of the specification.

My reading of this is that implementations are required to return an error for this. Applications should not rely on it being an error because a future version of OpenCL may change that, but that change has not yet been done, the 3.0 spec still says it's an error for now.

PR #2381, opened 2025-04-22
Update README after PR 2200

AlexGalazin-IMG:
PR2200 made the presence of spirv-as and spirv-val mandatory.

Reflect this fact in the Readme and update the example build instructions


kpet (dismissed):
LGTM, thanks! That's a nice step forward. We probably should update the CI to be explicit as to where SPIR-V tools are obtained from.

ahesham-arm (approved):

bashbaug (approved):

bashbaug:

PR #2382, opened 2025-04-24
Update local workgroup size to generate desired input

neoming:
The `local_workgroup_size` is changed by `get_max_common_work_group_size`. But the input data still use the original `local_workgroup_size`, which will cause the check logic to failed.

### DIFF
--- a/test_conformance/subgroups/subhelpers.h
+++ b/test_conformance/subgroups/subhelpers.h
@@ -1609,6 +1609,7 @@ template <typename Ty, typename Fns, size_t TSIZE = 0> struct subgroup_test
 
         // Generate the desired input for the kernel
         test_params.subgroup_size = subgroup_size;
+        test_params.local_workgroup_size = local;
         Fns::gen(idata.data(), mapin.data(), sgmap.data(), test_params);
 
         test_status status = TEST_FAIL;


neoming:
Here is the failed case I am debugging:

- The `test_parameter.subgroup_size` is 64. 
- The `test_parameter.global_workgroup_size` is 2000
- The `test_parameter.local_workgroup_size` is 200
- The actual `local` passed to `clEnqueueNDRangeKernel` is 125

The `subgroup 1` in `workgroup 0` is always **failed**, because for `local` is 125, there always be 3 inactive work-items.
failed example

neoming:
Hi @kpet , could you please help to review this pull request? I think this is the true issue for non-uniform subgroup test.

svenvh (approved):

bashbaug (approved):
Thank you for your patience.

LGTM - I'm able to reproduce the problem by forcing `get_max_common_work_group_size` to return a smaller value (125), and this patch fixes the problem.

neoming:
> Thank you for your patience.
> 
> LGTM - I'm able to reproduce the problem by forcing `get_max_common_work_group_size` to return a smaller value (125), and this patch fixes the problem.

Hi @bashbaug 

Thanks for your quickly support! Can this pull request be merged to main repo?

lakshmih:
Reviewing and will comment by next week, thanks

lakshmih (approved):

bashbaug:
Merging offline with multiple approvals in place, as discussed in the July 1st teleconference.

neoming:
Thanks all!

PR #2383, opened 2025-04-24
Sanitize flags before passing them into clGetSupportedImageFormats

karolherbst:
Some tests call get_format_list with e.g. CL_MEM_USE_HOSTPTR, but this flag isn't valid clGetSupportedImageFormats, but it's not clear if implementations are allowed to accept it.

This appears to be a problem since !2283 as now we get CL_MEM_USE_HOSTPTR passed into get_format_list.

### DIFF
--- a/test_conformance/images/common.cpp
+++ b/test_conformance/images/common.cpp
@@ -127,6 +127,9 @@ int get_format_list(cl_context context, cl_mem_object_type imageType,
                     std::vector<cl_image_format> &outFormatList,
                     cl_mem_flags flags)
 {
+    flags &= CL_MEM_READ_WRITE | CL_MEM_READ_ONLY | CL_MEM_WRITE_ONLY
+        | CL_MEM_KERNEL_READ_AND_WRITE | CL_MEM_IMMUTABLE_EXT;
+
     cl_uint formatCount;
     int error = clGetSupportedImageFormats(context, flags, imageType, 0, NULL,
                                            &formatCount);


ahesham-arm:
Going through the spec for `clGetSupportedImageFormats`, my reading is that the flags listed are expected to be supported by the implementation, but additional flags are valid, i.e. implementations that don't make use of e.g. `CL_MEM_USE_HOST_PTR` should ignore it and not return `CL_INVALID_VALUE`.

I'm leaning on my understanding here in that the wording for the `flags` parameter used  the "may" keyword:
```
5. MAY   This word, or the adjective "OPTIONAL", mean that an item is
   truly optional.  One vendor may choose to include the item because a
   particular marketplace requires it or because the vendor feels that
   it enhances the product while another vendor may omit the same item.
   An implementation which does not include a particular option MUST be
   prepared to interoperate with another implementation which does
   include the option, though perhaps with reduced functionality. In the
   same vein an implementation which does include a particular option
   MUST be prepared to interoperate with another implementation which
   does not include the option (except, of course, for the feature the
   option provides.)
```

Source: https://www.ietf.org/rfc/rfc2119.txt

What do you think?

karolherbst:
> Going through the spec for `clGetSupportedImageFormats`, my reading is that the flags listed are expected to be supported by the implementation, but additional flags are valid, i.e. implementations that don't make use of e.g. `CL_MEM_USE_HOST_PTR` should ignore it and not return `CL_INVALID_VALUE`.
> 
> I'm leaning in my understanding here on the wording for the `flags` parameter using the "may" keyword:
> 
> ```
> 5. MAY   This word, or the adjective "OPTIONAL", mean that an item is
>    truly optional.  One vendor may choose to include the item because a
>    particular marketplace requires it or because the vendor feels that
>    it enhances the product while another vendor may omit the same item.
>    An implementation which does not include a particular option MUST be
>    prepared to interoperate with another implementation which does
>    include the option, though perhaps with reduced functionality. In the
>    same vein an implementation which does include a particular option
>    MUST be prepared to interoperate with another implementation which
>    does not include the option (except, of course, for the feature the
>    option provides.)
> ```
> 
> Source: https://www.ietf.org/rfc/rfc2119.txt
> 
> What do you think?

In that case, what would invalid flags be that warrant returning `CL_INVALID_VALUE` as specified in the error code section?

ahesham-arm:
> In that case, what would invalid flags be that warrant returning `CL_INVALID_VALUE` as specified in the error code section?
A flag that is not one of the [supported memory flag values](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#memory-flags-table)

karolherbst:
> > In that case, what would invalid flags be that warrant returning `CL_INVALID_VALUE` as specified in the error code section?
> 
> A flag that is not one of the [supported memory flag values](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#memory-flags-table)

I think maybe it makes sense to change the specification there and simply specify, that the same flags as `clCreateImage`/`clCreateImageWithProperties` are supported and that runtimes can be more fine grained about returned flags.

I don't disagree that e.g. passing in `CL_MEM_USE_HOSTPTR` could be useful, and runtimes could returned a different set of formats, so if we want to go with that, might be better to make it more clear in the spec that this is expected behavior?

PR #2386, opened 2025-05-08
Added support for cl_ext_float_atomics in CBasicTestFetchAddSpecialFloats with atomic_half

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchAddSpecialFloats with support for atomic_half.

### DIFF
--- a/test_conformance/c11_atomics/test_atomics.cpp
+++ b/test_conformance/c11_atomics/test_atomics.cpp
@@ -1439,6 +1439,12 @@ public:
     {
         if constexpr (std::is_same_v<HostDataType, HOST_HALF>)
         {
+            // StartValue is used as an index divisor in the following test
+            // logic. It is set to the number of special values, which allows
+            // threads to be mapped deterministically onto the input data array.
+            // This enables repeated add operations arranged so that every
+            // special value is added to every other one (“all-to-all”).
+
             auto spec_vals = GetSpecialValues();
             StartValue(cl_half_from_float(spec_vals.size(), gHalfRoundingMode));
             CBasicTestMemOrderScope<HostAtomicType,
@@ -1518,6 +1524,14 @@ public:
 
         if constexpr (std::is_same_v<HostDataType, HOST_HALF>)
         {
+            // The start_value variable (set by StartValue) is used
+            // as a divisor of the thread index when selecting the operand for
+            // atomic_fetch_add. This groups threads into blocks corresponding
+            // to the number of special values and implements an “all-to-all”
+            // addition pattern. As a result, each destination element is
+            // updated using different combinations of input values, enabling
+            // consistent comparison between host and device execution.
+
             return std::string(DataType().AddSubOperandTypeName())
                 + " start_value = atomic_load_explicit(destMemory+tid, "
                   "memory_order_relaxed, memory_scope_work_group);\n"


shajder:
> Can this be rebased, thanks.

Done

test_conformance/c11_atomics/main.cpp
```diff
@@ -36,6 +36,7 @@ bool gFloatAtomicsSupported = false;
 cl_device_fp_atomic_capabilities_ext gHalfAtomicCaps = 0;
 cl_device_fp_atomic_capabilities_ext gDoubleAtomicCaps = 0;
 cl_device_fp_atomic_capabilities_ext gFloatAtomicCaps = 0;
+cl_device_fp_config gHalfCaps = 0;
```
bashbaug: Nitpick: These aren't "half caps", rather they're the "half FP config".


```suggestion
cl_device_fp_config gHalfFPConfig = 0;
```

lakshmih (dismissed):

shajder:
> Can you please take a look and see what might be happening?

I can't find any implementation that supports either the `CL_DEVICE_LOCAL_FP_ATOMIC_ADD_EXT` or `CL_DEVICE_GLOBAL_FP_ATOMIC_ADD_EXT` capabilities, so it's difficult to test this feature. I’ll try it on an Arc B570 machine tomorrow, but if you know of any simpler way to access a device that exposes these capabilities, I would appreciate the information. Thanks!

test_conformance/c11_atomics/main.cpp
shajder: Done

test_conformance/c11_atomics/common.h
```diff
@@ -91,6 +96,26 @@ extern cl_int getSupportedMemoryOrdersAndScopes(
     cl_device_id device, std::vector<TExplicitMemoryOrderType> &memoryOrders,
     std::vector<TExplicitMemoryScopeType> &memoryScopes);
 
+inline bool IsHalfNaN(const cl_half v)
```
bashbaug: I don't think this function `IsHalfNaN` is used anymore.

test_conformance/c11_atomics/common.h
```diff
@@ -91,6 +96,26 @@ extern cl_int getSupportedMemoryOrdersAndScopes(
...
+{
+    // Extract FP16 exponent and mantissa
+    uint16_t h_exp = (((cl_half)v) >> (CL_HALF_MANT_DIG - 1)) & 0x1F;
+    uint16_t h_mant = ((cl_half)v) & 0x3FF;
+
+    // NaN test
+    return (h_exp == 0x1F && h_mant != 0);
+}
+
+inline bool IsHalfInfinity(const cl_half v)
```
bashbaug: I don't think this function `IsHalfInfinity` is used anymore, either.

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1572,6 +1753,19 @@ static int test_atomic_fetch_add_generic(cl_device_id deviceID,
...
     {
+        auto spec_vals_halfs =
+            CBasicTestFetchAddSpecialFloats<HOST_ATOMIC_HALF,
+                                            HOST_HALF>::GetSpecialValues();
+        int num_elems = spec_vals_halfs.size();
+
+        for (auto &elem : spec_vals_halfs)
+        {
+            CBasicTestFetchAddSpecialFloats<HOST_ATOMIC_HALF, HOST_HALF>
+                test_half(TYPE_ATOMIC_HALF, elem, useSVM);
```
bashbaug: Is this the best approach?  AFAICT this is essentially running the special values test N times, where N is the number of special values.  This wouldn't be a huge problem, but "running the test" requires things like kernel compilations and buffer creations, so even though each kernel runs quickly, running the tests N times takes considerable time.

Could we move the loop inside of the test instead, to reuse the compiled kernels and created buffers?

Timing results on my laptop (CPU OpenCL):
* Running a single special values iteration: about 6 seconds.
* Running all special value iterations (17 special values): about 100 seconds, so nearly linear scaling.
* Running all atomic_fetch_add tests, across all types: about 220 seconds, so adding tests for special values roughly doubled the test execution time.

shajder: Understood, I’ll think about it. I have one idea, but it’s a bit hackish (it ignores the start value and requires one additional call to atomic_store_explicit).

bashbaug: Thinking out loud, since we're already testing the overall atomicity of these operations via other tests, could this special values test be much simpler?  Rough idea:

1. Create an N x N source buffer, where N is the number of special values to test.  Populate the source buffer with the special values in the order AAAA, BBBB, CCCC, DDDD, ...
2. Create another N x N source buffer, but populate this source buffer with with the special values in the order ABCD, ABCD, ABCD, ABCD, ...
3. Atomically add the two source buffers together, so every special value gets added to every other special value.
4. Check that each atomic add was performed correctly.

shajder: > Thinking out loud, since we're already testing the overall atomicity of these operations via other tests, could this special values test be much simpler?

In general the logic behind `SpecialFloat` tests is rather simple and it is following [suggested schema](https://github.com/intel/cassian/commit/239de5d8630bee09c0f3e136d4dab6e508bf80e3) mentioned in parent [issue](https://github.com/KhronosGroup/OpenCL-CTS/issues/2142).

> * Create another N x N source buffer, but populate this source buffer with the special values in the order ABCD, ABCD, ABCD, ABCD, ...

yep, I think that was the hackish idea I was thinking about - now it is implemented and I believe it is more performant. Implementation is still hot, I will appreciate closer look before I will distribute that schema to other SpecialFloats tests, thanks!

test_conformance/c11_atomics/common.h
shajder: Done

test_conformance/c11_atomics/common.h
shajder: Done

test_conformance/c11_atomics/test_atomics.cpp
bashbaug: Thanks, the new implementation is much faster and I don't think we've lost any test coverage.  Nice work!

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1547,6 +1547,205 @@ class CBasicTestFetchAdd
...
+        {
+            return std::string(DataType().AddSubOperandTypeName())
+                + " start_value = atomic_load_explicit(destMemory+tid, "
+                  "memory_order_relaxed, memory_scope_work_group);\n"
+                  "  atomic_store_explicit(destMemory+tid, oldValues[tid], "
+                  "memory_order_relaxed, memory_scope_work_group);\n"
+                  "  atomic_fetch_add"
+                + postfix + "(&destMemory[tid], ("
+                + DataType().AddSubOperandTypeName()
+                + ")oldValues[tid/(int)start_value]" + memoryOrderScope
```
bashbaug, line 1543: This `tid/(int)start_value` is very tricky: I think it's right, but it's basically putting the divisor as the "start value", encoded as a "half", and then converting to an "int" when performing the division.

I don't think we need to change this necessarily, but I do think it'd be worth adding some comments, either in the test source, or in the kernel source, or perhaps even in both places.

bashbaug (approved):
I still think these tests are still quite hard to follow and perhaps overly clever, but I think they're correct now, and they execute fairly promptly.

test_conformance/c11_atomics/test_atomics.cpp
shajder: Indeed, comments added as requested!

PR #2388, opened 2025-05-12
Added support for cl_ext_float_atomics in CBasicTestFetchAddSpecialFoats with atomic_double

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchAddSpecialFloats with support for atomic_double.

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -22,6 +22,8 @@
 
 #include "host_atomics.h"
 
+#include <iomanip>
+#include <limits>
 #include <vector>
 #include <sstream>
 
@@ -71,6 +73,9 @@ extern int
     gMaxDeviceThreads; // maximum number of threads executed on OCL device
 extern cl_device_atomic_capabilities gAtomicMemCap,
     gAtomicFenceCap; // atomic memory and fence capabilities for this device
+extern bool gFloatAtomicsSupported;
+extern cl_device_fp_atomic_capabilities_ext gDoubleAtomicCaps;
+extern cl_device_fp_config gDoubleCaps;
 
 extern const char *
 get_memory_order_type_name(TExplicitMemoryOrderType orderType);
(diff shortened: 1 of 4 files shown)


test_conformance/c11_atomics/common.h
```diff
@@ -76,9 +76,8 @@ extern int
     gMaxDeviceThreads; // maximum number of threads executed on OCL device
 extern cl_device_atomic_capabilities gAtomicMemCap,
     gAtomicFenceCap; // atomic memory and fence capabilities for this device
-extern cl_half_rounding_mode gHalfRoundingMode;
-extern bool gFloatAtomicsSupported;
-extern cl_device_fp_atomic_capabilities_ext gHalfAtomicCaps;
+
+extern cl_device_fp_config gDoubleCaps;
```
bashbaug: Nitpick:

```suggestion
extern cl_device_fp_config gDoubleFPConfig;
```

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1575,27 +1632,50 @@ class CBasicTestFetchAddSpecialFloats
...
     {
-
         if constexpr (std::is_same_v<HostDataType, HOST_HALF>)
         {
             return static_cast<cl_half>(expected) != testValues[whichDestValue];
         }
+        else
+        {
+            if (std::isnan(testValues[whichDestValue]) && std::isnan(expected))
+                return false;
```
bashbaug, line 1665: Out of curiosity, why do we need this special handling for NaN for fp64, but not for fp16?

bashbaug (approved):

test_conformance/c11_atomics/common.h
shajder: Done

test_conformance/c11_atomics/test_atomics.cpp
shajder: To make long story short, `fp64` is a native floating-point type handled by the FPU, which explicitly defines `NaN != NaN` as `true`. In contrast, `cl_half` is treated as a 16-bit integral alias, the comparison is performed on the underlying bit representation.

PR #2391, opened 2025-05-15
Added support for cl_ext_float_atomics in CBasicTestFetchMinSpecialFloats with atomic_float

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchMinSpecialFloats with support for atomic_float.

### DIFF
--- a/test_conformance/c11_atomics/test_atomics.cpp
+++ b/test_conformance/c11_atomics/test_atomics.cpp
@@ -2146,6 +2146,217 @@ public:
     }
 };
 
+template <typename HostAtomicType, typename HostDataType>
+class CBasicTestFetchMinSpecialFloats
+    : public CBasicTestMemOrderScope<HostAtomicType, HostDataType> {
+
+    std::vector<HostDataType> ref_vals;
+
+public:
+    using CBasicTestMemOrderScope<HostAtomicType, HostDataType>::StartValue;
+    using CBasicTestMemOrderScope<HostAtomicType, HostDataType>::DataType;
+    using CBasicTestMemOrderScope<HostAtomicType, HostDataType>::MemoryOrder;
+    using CBasicTestMemOrderScope<HostAtomicType,
+                                  HostDataType>::MemoryOrderScopeStr;
+    using CBasicTestMemOrderScope<HostAtomicType, HostDataType>::LocalMemory;
+    CBasicTestFetchMinSpecialFloats(TExplicitAtomicType dataType,
+                                    const HostDataType &special, bool useSVM)
+        : CBasicTestMemOrderScope<HostAtomicType, HostDataType>(dataType,
+                                                                useSVM)
+    {
+        if (std::is_same<HostDataType, HOST_ATOMIC_FLOAT>::value)
+        {
+            StartValue(special);
+            CBasicTestMemOrderScope<HostAtomicType,
+                                    HostDataType>::OldValueCheck(false);
+        }
+    }
+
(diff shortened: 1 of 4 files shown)


test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3123,6 +3132,251 @@ class CBasicTestFetchMin
...
+        : CBasicTestMemOrderScope<HostAtomicType, HostDataType>(dataType,
+                                                                useSVM)
+    {
+        // StartValue is used as an index divisor in the following test
+        // logic. It is set to the number of special values, which allows
+        // threads to be mapped deterministically onto the input data array.
+        // This enables repeated add operations arranged so that every
+        // special value is added to every other one (“all-to-all”).
+
+        if (std::is_same_v<HostDataType, HOST_FLOAT>)
```
bashbaug:
```suggestion
        if constexpr (std::is_same_v<HostDataType, HOST_FLOAT>)
```

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3148,6 +3402,16 @@ static int test_atomic_fetch_min_generic(cl_device_id deviceID,
...
     {
+        auto spec_vals =
+            CBasicTestFetchMinSpecialFloats<HOST_ATOMIC_FLOAT,
+                                            HOST_FLOAT>::GetSpecialValues();
+
+        CBasicTestFetchMinSpecialFloats<HOST_ATOMIC_FLOAT, HOST_FLOAT>
+            test_spec_float(TYPE_ATOMIC_FLOAT, useSVM);
+        EXECUTE_TEST(
+            error,
+            test_spec_float.Execute(deviceID, context, queue, num_elements));
```
bashbaug, line 3536:
```suggestion
        EXECUTE_TEST(
            error,
            test_spec_float.Execute(deviceID, context, queue, spec_vals.size() * spec_vals.size()));
```

shajder: indeed, it make sense but plese keep in mind the logic of this test uses whole buffer size ([wrapped](https://github.com/shajder/OpenCL-CTS/blob/017546b546c0a9f6bf80cafeea78fb39b8b8f5dc/test_conformance/c11_atomics/test_atomics.cpp#L3363-L3371)) which is not that big (512/1024). To be consistent with rest of the test I left is as is. My guess is that it has no significant impact on performance, right? If you think it is worth to align that size just let me know and I will adapt remaining tests.

test_conformance/c11_atomics/test_atomics.cpp
shajder: Done

test_conformance/c11_atomics/test_atomics.cpp
bashbaug: Yeah, you're right, it seems like most of the execution time for these tests is spent compiling kernels vs. executing kernels.  Still, it seems odd to have the inconsistency between the `add` tests and these `min` tests.  If we're going to use `num_elements` in this test, should we use it here, also?

https://github.com/KhronosGroup/OpenCL-CTS/blob/c3d9c85743f29e71b0333fd1b88870da0ba49e2f/test_conformance/c11_atomics/test_atomics.cpp#L1794-L1799

shajder: Thanks, corrected. I will review remaining PRs as well!

PR #2393, opened 2025-05-16
Added support for cl_ext_float_atomics in CBasicTestFetchMaxSpecialFloats with atomic_float

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchMaxSpecialFloats with support for atomic_float.


test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -2280,6 +2280,217 @@ class CBasicTestFetchMax
...
+                    // max(x, qNaN) = max(qNaN, x) = x,
+                    // max(qNaN, qNaN) = qNaN,
+                    if (is_qnan(StartValue())
+                        && is_qnan(startRefValues[whichDestValue]))
+                        return !is_qnan(testValue[whichDestValue]);
+                    else if (is_qnan(StartValue()))
+                        return testValue[whichDestValue]
+                            != startRefValues[whichDestValue];
+                    else
+                        return testValue[whichDestValue] != StartValue();
```
shajder, line 2431: @bashbaug the same problem as `atomic_fetch_min` - As per spec:

`max(x, qNaN) = max(qNaN, x) = x`

atomic_fetch_max( qNaN, -0.f) -> NaN

test_conformance/c11_atomics/common.h
```diff
@@ -96,6 +97,37 @@ extern cl_int getSupportedMemoryOrdersAndScopes(
     cl_device_id device, std::vector<TExplicitMemoryOrderType> &memoryOrders,
     std::vector<TExplicitMemoryScopeType> &memoryScopes);
 
+union FloatIntUnion {
```
svenvh, line 103: All of this is already on `main` now I think; perhaps you need to rebase.

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3571,6 +3583,252 @@ class CBasicTestFetchMax
...
+                                  HostDataType>::MemoryOrderScopeStr;
+    using CBasicTestMemOrderScope<HostAtomicType, HostDataType>::LocalMemory;
+    CBasicTestFetchMaxSpecialFloats(TExplicitAtomicType dataType, bool useSVM)
+        : CBasicTestMemOrderScope<HostAtomicType, HostDataType>(dataType,
+                                                                useSVM)
+    {
+        // StartValue is used as an index divisor in the following test
+        // logic. It is set to the number of special values, which allows
+        // threads to be mapped deterministically onto the input data array.
+        // This enables repeated add operations arranged so that every
```
svenvh: s/add/max?

There are a few other stale comments about "add".

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3571,6 +3583,252 @@ class CBasicTestFetchMax
...
+        // special value is added to every other one (“all-to-all”).
```
svenvh: Prefer plain ASCII:
```suggestion
        // special value is added to every other one ("all-to-all").
```

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3571,6 +3583,252 @@ class CBasicTestFetchMax
...
+
+        if constexpr (std::is_same_v<HostDataType, HOST_FLOAT>)
+        {
+            auto spec_vals = GetSpecialValues();
```
svenvh:
```suggestion
            const auto &spec_vals = GetSpecialValues();
```

Same for other `GetSpecialValues` calls.

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3571,6 +3583,252 @@ class CBasicTestFetchMax
...
+                     || is_qnan(startRefValues[whichDestValue]))
+            {
+                // max(x, qNaN) = max(qNaN, x) = x,
+                // max(qNaN, qNaN) = qNaN,
+                if (is_qnan(startRefValues[whichDestValue / spec_vals.size()])
+                    && is_qnan(startRefValues[whichDestValue]))
+                    return !is_qnan(testValues[whichDestValue]);
+                else if (is_qnan(
+                             startRefValues[whichDestValue / spec_vals.size()]))
+                    return !std::isnan(testValues[whichDestValue])
```
svenvh: I find the logic a bit hard to follow as the comments don't seem to follow the code order.  If I'm not mistaken this is the `max(x, qNaN)` case?  So this returns false (pass) when the device returns a NaN instead of `x`, which seems wrong.

Same applies for the `max(qNaN, x)` case below I suppose.

test_conformance/c11_atomics/common.h
shajder: main merged

test_conformance/c11_atomics/test_atomics.cpp
shajder: corrected

test_conformance/c11_atomics/test_atomics.cpp
shajder: corrected

test_conformance/c11_atomics/test_atomics.cpp
shajder: corrected

test_conformance/c11_atomics/test_atomics.cpp
shajder: Corrected. Indeed it is hard to follow (mostly due to the number of cases to handle) and you are correct - the logic was inverted (also for min). Correction revealed other problems which had to be fixed with additional condition.

svenvh: Nice, looks better now.  Thanks for also fixing min!

svenvh (approved):

bashbaug:

PR #2397, opened 2025-05-22
fix correctly rounded behavior for math bruteforce tests

### LINKED ISSUE #2387: math bruteforce correctly rounded square root vs. non-correctly rounded test confusion
I'm seeing behavior with the math bruteforce test that seems incorrect, specifically when comparing the non-correctly rounded "sqrt" test to the correctly rounded "sqrt_cr" test.  In short, on a device that supports single-precision correctly rounded divides and square roots (meaning that `CL_FP_CORRECTLY_ROUNDED_DIVIDE_SQRT` is in the bitfield returned when querying `CL_DEVICE_SINGLE_FP_CONFIG`), it looks like the build option `-cl-fp32-correctly-rounded-divide-sqrt` is included when building **all** of the programs tested by the math bruteforce test.  This includes the non-correctly rounded "sqrt" test, so there is effectively no difference between the "sqrt" and the "sqrt_cr" test.  It is also including the build option on other tests, such as `sin`, which are unrelated to divide and square root, which is not necessarily wrong per se but which I also find surprising.

Shouldn't the `-cl-fp32-correctly-rounded-divide-sqrt` build option only be passed when building the "sqrt_cr" and "divide_cr" programs?

Full disclosure: There is another side effect of this issue, at least on our GPU implementation.  We currently use a different implementation of the **fp16** square root when we see the `-cl-fp32-correctly-rounded-divide-sqrt` build option.  This implementation meets the currently documented fp16 accuracy requirements, but the fp16 implementation without the `-cl-fp32-correctly-rounded-divide-sqrt` build option does not.  This means that the current test behavior is masking failures, or in other words, we may currently be passing the CTS only by chance.  If we fix this issue, we will need clarity whether our implementation must to use the correctly rounded (and slower) square root all of the time, or whether the fp16 square root accurate requirements should be relaxed (1 ULP is sufficient) like we did recently for "divide".

bashbaug:
fixes #2387 

Corrects the "correctly rounded" behavior for the math bruteforce tests.  Specifically:

* Only applies the `-cl-fp32-correctly-rounded-divide-sqrt` build option for the `divide_cr` and `sqrt_cr` tests.  The other tests do not receive this build option.  This means that there is a difference in the behavior of the `divide` and `divide_cr` tests and the `sqrt` and `sqrt_cr` tests, and the "correctly rounded" build option is not applied to the fp16 or fp64 tests.
* Removes the build option to toggle testing the correctly rounded divide and square root tests since it no longer needed.  Instead, the test names can be used to choose whether to test the correctly rounded functions or the non-correctly rounded functions.

Additionally:

* Relaxes the fp16 sqrt accuracy requirements to 1 ULP.  This is needed to pass this test on some of our devices.  This part is still under discussion, so I will keep this PR as a draft until it is settled.

### DIFF
--- a/test_conformance/math_brute_force/binary_operator_float.cpp
+++ b/test_conformance/math_brute_force/binary_operator_float.cpp
@@ -754,10 +754,12 @@ int TestFunc_Float_Float_Float_Operator(const Func *f, MTdata d,
         test_info.tinfo[i].d = MTdataHolder(genrand_int32(d));
     }
 
+    bool correctlyRounded = strcmp(f->name, "divide_cr") == 0;
+
     // Init the kernels
     BuildKernelInfo build_info{ test_info.threadCount, test_info.k,
                                 test_info.programs, f->nameInCode,
-                                relaxedMode };
+                                relaxedMode, correctlyRounded };
     if ((error = ThreadPool_Do(BuildKernelFn,
                                gMaxVectorSizeIndex - gMinVectorSizeIndex,
                                &build_info)))
--- a/test_conformance/math_brute_force/common.cpp
+++ b/test_conformance/math_brute_force/common.cpp
@@ -102,7 +102,7 @@ void EmitEnableExtension(std::ostringstream &kernel,
     if (needsFp16) kernel << "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n";
 }
 
-std::string GetBuildOptions(bool relaxed_mode)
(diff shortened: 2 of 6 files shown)


svenvh (approved):

lpavank (approved):
I reviewed the changes and tested locally on one of our devices (brutefoce suite with the knobs and time given) and we are OK with the changes at this time.

bashbaug:
Merging as discussed in the July 15th teleconference.

PR #2399, opened 2025-05-27
Added support for cl_ext_float_atomics in CBasicTestFetchMinSpecialFloats with atomic_half

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchMinSpecialFloats with support for atomic_half.


test_conformance/c11_atomics/common.h
```diff
@@ -111,44 +111,74 @@ union DoubleIntUnion {
...
+{
+    // Extract FP16 exponent and mantissa
+    uint16_t h_exp = (((cl_half)v) >> (CL_HALF_MANT_DIG - 1)) & 0x1F;
+    uint16_t h_mant = ((cl_half)v) & 0x3FF;
+
+    // NaN test
+    return (h_exp == 0x1F && h_mant != 0);
+}
+
+inline bool IsHalfInfinity(const cl_half v)
```
svenvh: Seems unused?

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3484,6 +3484,13 @@ class CBasicTestFetchMinSpecialFloats
             CBasicTestMemOrderScope<HostAtomicType,
                                     HostDataType>::OldValueCheck(false);
         }
+        else if constexpr (std::is_same_v<HostDataType, HOST_HALF>)
+        {
+            auto spec_vals = GetSpecialValues();
```
svenvh:
```suggestion
            const auto &spec_vals = GetSpecialValues();
```

In a few other places too.

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3539,6 +3546,35 @@ class CBasicTestFetchMinSpecialFloats
                 }
             }
         }
+        else if constexpr (std::is_same_v<HostDataType, HOST_HALF>)
+        {
+            if (special_values.empty())
+            {
+                special_values = {
+                    0xffff, 0x0000, 0x7c00, /*INFINITY*/
```
svenvh: Could it be that these are first cast to `float` instead of being directly constructed as halfs?  If so, we'd end up with regular float values of these integers, not the intended special values.

shajder: You are absolutely right, good catch. This came in with the `HostHalf` wrapper: a braced literal like `0x7c00` has type `int`, so it selects `HostHalf(int)`, which delegates to `HostHalf(cl_uint)` and runs `cl_half_from_float(static_cast<float>(value))`. The bit-pattern constructor `HostHalf(cl_half)` is never picked, so the value ends up interpreted as a float magnitude instead of raw half bits.

I'll correct it and, since the fix lives in a method common to all `CBasicTestFetch*SpecialFloats` classes, take the opportunity to introduce the `CBasicTestFetchSpecialFloats` base class that was planned anyway. The existing Add variant has the same latent issue, so consolidating into the base fixes it there too.

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -3734,13 +3701,37 @@ class CBasicTestFetchMinSpecialFloats
...
             HostAtomicType, HostDataType>::ExecuteSingleTest(deviceID, context,
                                                              queue);
     }
     cl_uint NumResults(cl_uint threadCount, cl_device_id deviceID) override
     {
-        if constexpr (std::is_same_v<HostDataType, HOST_FLOAT>)
+        if constexpr (
+            std::is_same_v<
+                HostDataType,
+                HOST_FLOAT> || std::is_same_v<HostDataType, HOST_HALF>)
```
marekkozl: There is pending bugfix at #2700 conflicting with this. Preferably we could merge smaller change #2700 first to maintain authors and rebase this one.

shajder: sure, we are waiting for @svenvh approval for this one anyway and #2700 seems ready to go

svenvh (approved):

PR #2400, opened 2025-05-28
Avoid some undefined behavior in test_bruteforce.

hvdijk:
* Ulp_Error*: ilogb(reference) - 1 may overflow if reference is zero.
* binary_i_double Test: DoubleFromUInt32's result is a cl_double and the attempt is to store it as a cl_double, but p was defined as a pointer to cl_ulong, resulting in an unintended implicit conversion that is not valid for out-of-range doubles.
* exp2, tanpi: ensure early exit for NaN.
* shift_right_sticky_128: avoid out-of-range shift if shift value is exactly 64.
* scalbn: e += n may overflow if n is large, move it after the check for large n.

### DIFF
--- a/test_common/harness/errorHelpers.cpp
+++ b/test_common/harness/errorHelpers.cpp
@@ -387,8 +387,7 @@ static float Ulp_Error_Half_Float(float test, double reference)
     }
 
     // reference is a normal power of two or a zero
-    int ulp_exp =
-        HALF_MANT_DIG - 1 - std::max(ilogb(reference) - 1, HALF_MIN_EXP - 1);
+    int ulp_exp = HALF_MANT_DIG - std::max(ilogb(reference), HALF_MIN_EXP);
 
     // Scale the exponent of the error
     return (float)scalbn(testVal - reference, ulp_exp);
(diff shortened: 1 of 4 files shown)


hvdijk:
Background: after updating our build infrastructure to build OpenCL CTS with a newer compiler version, I got this result:
```
ldexp...
144:           ldexp fp32     .
ERROR: ldexp: inf ulp error at {0x1.6694eep+20 (0x49b34a77), 2147483207}: *0x0p+0 (0x00000000) vs. inf (0x7f800000) at index: 12055
```
That is, our implementation was correctly returning infinity for `ldexp(0x1.6694eep+20, 2147483207)`, but the OpenCL CTS reference result was computed to be zero. When seeing that, I built OpenCL CTS with `clang -fsanitize=undefined` and addressed the issues that came up when running `test_brute_force`.

svenvh (approved):
LGTM, thanks for the fixes!

lakshmih:
Reviewing, could you hold on till end of week. Thanks

lakshmih (approved):

bashbaug:
Merging as discussed in the July 8th teleconference.

PR #2401, opened 2025-05-28
Allow building with cmake --compile-no-warning-as-error.

hvdijk:
In CMake 3.24+, there is built-in support for adding -Werror that does not require adding -Werror explicitly, and allows it to be downgraded to a warning if the user wants that. Use this, to account for warnings that have false positives.


hvdijk:
Ah, it's just because `test_conformance/vulkan` forcibly overrides all `CXXFLAGS`, including `-Werror`. Likewise, `test_conformance/common/vulkan_wrapper` keeps `-Werror` but then appends `-Wno-error`. Both seem dodgy, but if that's how it works, it's then consistent with that to also forcibly override `CMAKE_COMPILE_WARNING_AS_ERROR` the same way.

svenvh (approved):
LGTM, thanks!

> Both seem dodgy, but if that's how it works, it's then consistent with that to also forcibly override CMAKE_COMPILE_WARNING_AS_ERROR the same way.

I agree that these overrides are acceptable.  Ideally those CMakeLists wouldn't overwrite CXXFLAGS, but that's what it is currently.

bashbaug:
> > Both seem dodgy, but if that's how it works, it's then consistent with that to also forcibly override CMAKE_COMPILE_WARNING_AS_ERROR the same way.
> 
> I agree that these overrides are acceptable. Ideally those CMakeLists wouldn't overwrite CXXFLAGS, but that's what it is currently.

This seems like something we should fix.  Do we already have an issue so this doesn't (continue to) slip through the cracks?

bashbaug (approved):

svenvh:
> Do we already have an issue so this doesn't (continue to) slip through the cracks?

I've raised https://github.com/KhronosGroup/OpenCL-CTS/issues/2430 now.

bashbaug:
Merging as discussed in the July 1st teleconference (and offline).

PR #2402, opened 2025-05-29
Added initialization of buffer once imported in OpenCL for test_buffer_single_queue_fence test

sergiupopescu22:
For testing the interoperability between Vulkan and OpenCL, the test exports a Vulkan buffer and then OpenCL imports it. After importing, OpenCL applies a kernel that adds a constant value to the initial buffer (which is not initialized) and afterwards it expects that all the elements in the buffer to have that specific constant value (thus the test fails).

I added an initialization of the buffer, once it is imported into OpenCL.

### DIFF
--- a/test_conformance/vulkan/test_vulkan_interop_buffer.cpp
+++ b/test_conformance/vulkan/test_vulkan_interop_buffer.cpp
@@ -577,6 +577,25 @@ int run_test_with_one_queue(
             // bufferSizeList[i]
             global_work_size[0] = bufferSize;
 
+            err = clEnqueueAcquireExternalMemObjectsKHRptr(
+                cmd_queue1, vkBufferList.size(), buffers, 0, nullptr, nullptr);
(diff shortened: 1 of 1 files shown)


test_conformance/vulkan/test_vulkan_interop_buffer.cpp
```diff
@@ -708,6 +727,12 @@ int run_test_with_one_queue(
                                                + numBuffers);
             externalMemory.erase(externalMemory.begin(),
                                  externalMemory.begin() + numBuffers);
+
+            if (error_1)
+            {
+                clReleaseMemObject(error_1);
```
joselopeqti: `error_1` is already being released at the end of this test

sergiupopescu22: Thank you for pointing this out! I removed this part, only the fix for the initialization remains.

joselopeqti (dismissed):

kpet:
Closing as agreed in 2026/06/09 teleconference.

PR #2406, opened 2025-06-02
Added support for cl_ext_float_atomics in CBasicTestFetchMaxSpecialFloats with atomic_half

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchMaxSpecialFloats with support for atomic_half.


bashbaug (approved):
I didn't do a super detailed review, but I did look over the code, and I ran the tests and verified that additional "half" tests are tested and that the tests pass, so I don't see any merge blockers.

PR #2408, opened 2025-06-03
SVM buffer address equivalence tests

bashbaug:
Here are some tests to check how SVM pointers on the host relate to pointer values on the device when a buffer memory object is created from the SVM pointer.  See:

* https://github.com/KhronosGroup/OpenCL-Docs/issues/1369
* https://github.com/KhronosGroup/OpenCL-Docs/issues/1370

Some other interesting tests might be:

* [ ] Create a buffer from a pointer offset into the SVM allocation, rather than the base of the SVM allocation
* [ ] Create a sub-buffer from the buffer created from the SVM pointer

### DIFF
--- a/test_conformance/SVM/CMakeLists.txt
+++ b/test_conformance/SVM/CMakeLists.txt
@@ -4,6 +4,7 @@ set(${MODULE_NAME}_SOURCES
     main.cpp
     test_allocate_shared_buffer.cpp
     test_allocate_shared_buffer_negative.cpp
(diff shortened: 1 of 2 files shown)


bashbaug:
Discussed in the June 17th memory subgroup:

- More vendors have tried these tests.  Most tests are passing on most implementations, but some vendors are failing the test that creates a buffer memory object from an offset SVM allocation.  We need to decide if this is something we want to support.

PR #2409, opened 2025-06-05
tests for cl_khr_spirv_queries

bashbaug:
See: https://github.com/KhronosGroup/OpenCL-Docs/pull/1385


test_conformance/api/spirv_capability_deps.def
```diff
@@ -0,0 +1,312 @@
+//
```
kpet: If the SPIRV-Headers become a hard dependency, maybe we should just generate this file at build time?

karolherbst:
looks good, tested my implementation of `cl_khr_spirv_queries` against this PR and everything passes.

Though I think if `cl_khr_spirv_extended_debug_info` is supported you could also check if `OpenCL.DebugInfo.100` is.

bashbaug:
> Though I think if `cl_khr_spirv_extended_debug_info` is supported you could also check if `OpenCL.DebugInfo.100` is.

Good catch.  I added this, although I don't have a device to test it with, and according to [opencl.gpuinfo.org](https://opencl.gpuinfo.org/listextensions.php) there aren't any devices in the database that support it, either, hrm.

As an aside, the tests are currently (informationally) printing any non-required SPIR-V capabilities that are reported.  Should I do something similar for SPIR-V extended instruction sets and SPIR-V extensions also?  This might help to find other missing checks like this one... or it could just be noise, once usage of the SPIR-V queries becomes more widespread.

karolherbst:
> > Though I think if `cl_khr_spirv_extended_debug_info` is supported you could also check if `OpenCL.DebugInfo.100` is.
> 
> Good catch. I added this, although I don't have a device to test it with, and according to [opencl.gpuinfo.org](https://opencl.gpuinfo.org/listextensions.php) there aren't any devices in the database that support it, either, hrm.
> 

don't have tests here for that either.

> As an aside, the tests are currently (informationally) printing any non-required SPIR-V capabilities that are reported. Should I do something similar for SPIR-V extended instruction sets and SPIR-V extensions also? This might help to find other missing checks like this one... or it could just be noise, once usage of the SPIR-V queries becomes more widespread.

I think with this extension in mind it's fine to report more capabilities or extensions than necessary. They can also be part of vendor/ext extensions the tests here isn't aware off, so I'm pretty sure it's gonna cause some noise sooner or later.

test_conformance/api/test_spirv_queries.cpp
```diff
@@ -0,0 +1,764 @@
...
+    {
+        capabilities.push_back(spv::CapabilityPipeStorage);
+    }
+
+    // Required for OpenCL 2.2, or OpenCL 3.0 devices supporting sub-groups.
+    if (version == Version(2, 2)
+        || (version >= Version(3, 0) && deviceSubGroupsSupport == CL_TRUE))
+    {
+        capabilities.push_back(spv::CapabilitySubgroupDispatch);
+    }
```
kpet: Shouldn't these requirements only apply to devices that support SPIR-V 1.1?

bashbaug: Yes, good catch!  Fixed.

karolherbst:
need to add "cl_khr_spirv_queries" inside test_conformance/compiler/test_compiler_defines_for_extensions.cpp

otherwise `test_compiler compiler_defines_for_extensions` will fail: "FAIL: Extension cl_khr_spirv_queries is not in the list of approved Khronos extensions!"

bashbaug:
> need to add "cl_khr_spirv_queries" inside test_conformance/compiler/test_compiler_defines_for_extensions.cpp

Fixed.

I _really_ want to get that test fixed; it's kind of silly how we need to modify it every time a new KHR extension gets published.

test_conformance/api/test_spirv_queries.cpp
bashbaug: I think I'm going to leave the check for SPIR-V 1.1 as-is for now, but after thinking about this a bit more, shouldn't these capability really be supported whenever a device supports SPIR-V 1.1 **or newer**?  The capabilities were added in SPIR-V 1.1, but until they're removed in some future SPIR-V version, I expect they would be available even if a device only supported, say, SPIR-V 1.2.

This might also affect how we document these capabilities in the SPIR-V environment spec.  Maybe I should file an issue for this, specifically.

karolherbst:
> > need to add "cl_khr_spirv_queries" inside test_conformance/compiler/test_compiler_defines_for_extensions.cpp
> 
> Fixed.
> 
> I _really_ want to get that test fixed; it's kind of silly how we need to modify it every time a new KHR extension gets published.

yeah.. this list really should be generated

test_conformance/api/test_spirv_queries.cpp
kpet: Yeah, it should be 1.1 or newer because support for SPIR V 1.2 would in turn pull 1.1 capability requirements.

test_conformance/api/spirv_capability_deps.def
bashbaug: Done.  I initially didn't want to add a build dependency on python, but we already need python to build SPIR-V binaries for the `spirv_new` test, so perhaps this is OK?

Maybe this test should move to the `spriv_new` executable instead of `api`, since it's SPIR-V related?  Then the dependency on python will all be in one place?

Or, if we decide we really don't want to bother with generation at build time, we can just drop the last commit.

test_conformance/api/test_spirv_queries.cpp
bashbaug: I'd like to leave this as-is for now, is possible, in the interest of getting these changes in soon (we're already seeing some "failures" due to the missing changes to compiler_defines_for_extensions).  My rationale is:

* This _probably_ isn't an issue in practice currently, since most implementations that advertise support for newer versions of SPIR-V (e.g. 1.2 or newer) also advertise support for the older versions of SPIR-V (1.1)
* Even if an implementation only advertises support for the newer versions of SPIR-V, we'll just skip a few checks, and the tests won't "fail" unexpectedly.

I'd be happy to file an issue to fix this so it doesn't slip through the cracks.

test_conformance/api/spirv_capability_deps.def
bashbaug: > Maybe this test should move to the `spriv_new` executable instead of `api`, since it's SPIR-V related? Then the dependency on python will all be in one place?

I looked at moving this test to `spirv_new` instead of `api` and I think this would be relatively straightforward, but there is one side effect of moving the test to `spirv_new` we should consider: the `spirv_new` tests only run on a device that supports SPIR-V, as determined by `check_spirv_compilation_readiness()`.  Although it'd debatably be a little silly to do so, we do not currently require a device to support SPIR-V to support `cl_khr_spirv_queries`, and a device that does not support SPIR-V could return the empty set for all of the new SPIR-V queries and be conformant.  Is this a scenario we care about?

If so, we should not move the test to `spirv_new`.  If not, we should document the SPIR-V requirement, and perhaps add a basic "consistency check" to ensure a device advertising support for `cl_khr_spirv_queries` also supports SPIR-V.

karolherbst (approved):

test_conformance/api/spirv_capability_deps.def
bashbaug: @kpet @karolherbst any strong opinions?  Should we keep the test where it is currently, in `test_api`, or move it to `test_spirv_new`?  I believe this is the last major question to resolve before merging.

If we move the test to `test_spirv_new` I will add a very API consistency test that runs on all devices.  Thinking a bit more, this could take one of two forms:

1. It could check that if a device supports `cl_khr_spirv_queries` and does not support SPIR-V that the queries succeed, but are empty.  This is basically the current behavior.
2. Or, if we decide to require SPIR-V support if a device supports `cl_khr_spirv_queries`, it could check for this coupling.

ahesham-arm (approved):

kpet (approved):

PR #2411, opened 2025-06-05
add missing image support checks to a few SPIR-V tests

bashbaug:
The tests for new image operands should not run on devices that do not support images.


bashbaug:
Merging... obvious fix that only impacts SPIR-V.

PR #2413, opened 2025-06-06
cl_khr_external_memory_ahb test

dcrawleyqti:
Android AHardwareBuffer test. Initial testing for buffers and image.


test_conformance/extensions/cl_khr_external_memory_ahb/procs.h
```diff
@@ -0,0 +1,41 @@
+//
```
ahesham-arm: This file should not be needed. Please use the new macro for registring tests. `REGISTER_TEST` instead of `ADD_TEST` or `REGISTER_TEST_VERSION` instead of `ADD_TEST_VERSION`.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -0,0 +1,1521 @@
+//
+// Copyright (c) 2023 The Khronos Group Inc.
```
ahesham-arm: Incorrect date.

test_conformance/extensions/cl_khr_external_memory_ahb/main.cpp
```diff
@@ -0,0 +1,38 @@
+//
+// Copyright (c) 2023 The Khronos Group Inc.
```
ahesham-arm: Incorrect date.

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
```diff
@@ -0,0 +1,189 @@
+//
+// Created by joshkell on 8/18/24.
+//
+
+#include <cstring>
```
ahesham-arm: These headers include the C++ namespaced versions of types and functions but then use the non-namespaced one. This should be either `#include <string.h>` or the usage should be, e.g. `std::size_t` and `std::memset`.

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
```diff
@@ -0,0 +1,189 @@
+//
+// Created by joshkell on 8/18/24.
```
ahesham-arm: Missing copyright header.

test_conformance/extensions/cl_khr_external_memory_ahb/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(MODULE_NAME CL_KHR_EXTERNAL_MEMORY_AHB)
+
+set(${MODULE_NAME}_SOURCES
+         main.cpp
```
ahesham-arm: Incorrect alignment.

test_conformance/extensions/CMakeLists.txt
```diff
@@ -5,6 +5,9 @@
 add_subdirectory( cl_ext_cxx_for_opencl )
 add_subdirectory( cl_khr_command_buffer )
 add_subdirectory( cl_khr_dx9_media_sharing )
+if(ANDROID_HARDWARE_BUFFER_IS_SUPPORTED)
```
ahesham-arm: Where is this defined, or how is it detected?

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
```diff
@@ -0,0 +1,189 @@
...
+//
+
+#include <cstring>
+#include <cstdlib>
+#include "debug_ahb.h"
+
+AHardwareBuffer_UsageFlags flag_list[] = {
```
ahesham-arm: This can be made `constexpr`.

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
```diff
@@ -0,0 +1,189 @@
...
+        AHARDWAREBUFFER_USAGE_GPU_SAMPLED_IMAGE,
+        AHARDWAREBUFFER_USAGE_GPU_FRAMEBUFFER,
+        AHARDWAREBUFFER_USAGE_COMPOSER_OVERLAY,
+        AHARDWAREBUFFER_USAGE_PROTECTED_CONTENT,
+        AHARDWAREBUFFER_USAGE_VIDEO_ENCODE,
+        AHARDWAREBUFFER_USAGE_SENSOR_DIRECT_DATA,
+        AHARDWAREBUFFER_USAGE_GPU_DATA_BUFFER,
+        AHARDWAREBUFFER_USAGE_GPU_CUBE_MAP,
+        AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE,
+        AHARDWAREBUFFER_USAGE_FRONT_BUFFER, // This is not in older NDK 25
```
ahesham-arm:
```
// This is not in older NDK 25
````
Is there a way to detect the NDK version programatically and guard against the version at compile time?

File: test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp [hunk 10]
```diff
@@ -0,0 +1,189 @@
...
+};
+
+char * ahardwareBufferDecodeUsageFlagsToString(AHardwareBuffer_UsageFlags flags) {
```
ahesham-arm: Please use `std::string` instead.

test_conformance/extensions/cl_khr_external_memory_ahb/procs.h
dcrawleyqti: This file has been removed

test_conformance/extensions/cl_khr_external_memory_ahb/main.cpp
dcrawleyqti: Corrected

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
dcrawleyqti: Done

test_conformance/extensions/cl_khr_external_memory_ahb/CMakeLists.txt
dcrawleyqti: Corrected

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
dcrawleyqti: Added the header

test_conformance/extensions/CMakeLists.txt
dcrawleyqti: Intended to be similar to VULKAN_IS_SUPPORTED. Instead I used ANDROID_PLATFORM GREATER 28 to make sure all AHB functions used in the test were supported.

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
dcrawleyqti: Done

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
dcrawleyqti: This is now done at the CMakelists.txt level since most AHB functions used in the test also require a more recent version of the Android API

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
dcrawleyqti: Done

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
dcrawleyqti: Fixed

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -0,0 +1,1899 @@
...
+
+                size_t srcBytes = get_image_size(&imageInfo);
+                test_assert_error(srcBytes > 0, "Image cannot have zero size");
+
+                BufferOwningPtr<char> srcData;
+                generate_random_image_data(&imageInfo, srcData, seed);
+
+                void *hardware_buffer_data = nullptr;
+                ahb_result = AHardwareBuffer_lock(
+                    aHardwareBuffer, AHARDWAREBUFFER_USAGE_CPU_WRITE_OFTEN, -1,
```
Jeremy-Kemp, line 283: There is a mismatch here between `AHARDWAREBUFFER_USAGE_CPU_WRITE_OFTEN` and `usage` that was used to create the AHB.

Some of the flags in `test_usages` don't contain any CPU flags and so the call to `AHardwareBuffer_lock` can fail in those cases.

I think all of the variants in `test_usages` will need CPU flags set on them.

dcrawleyqti: Good catch! Added the CPU flags to all of the test_usages variants

test_conformance/extensions/cl_khr_external_memory_ahb/debug_ahb.cpp
```diff
@@ -0,0 +1,212 @@
...
+    AHARDWAREBUFFER_USAGE_PROTECTED_CONTENT,
+    AHARDWAREBUFFER_USAGE_VIDEO_ENCODE,
+    AHARDWAREBUFFER_USAGE_SENSOR_DIRECT_DATA,
+    AHARDWAREBUFFER_USAGE_GPU_DATA_BUFFER,
+    AHARDWAREBUFFER_USAGE_GPU_CUBE_MAP,
+    AHARDWAREBUFFER_USAGE_GPU_MIPMAP_COMPLETE,
+    AHARDWAREBUFFER_USAGE_FRONT_BUFFER,
+};
+
+std::string *
```
ahesham-arm: Why is this a pointer? What I meant when I said to use `std::string` is to use what C++ has to offer to working with strings in general, not just changing the return type of the function.

Here is my untested suggestion:
```
std::string aHardwareBufferDecodeUsageFlagsToString(AHardwareBuffer_UsageFlags flags)
{
    if (flags == 0)
    {
        return "UNKNOWN_FLAG";
    }

    std::vector<std::string> active_flags;
    for (const auto f : flag_list)
    {
        if ((f & flags) == f)
        {
            active_flags.push_back(ahardwareBufferUsageFlagToString(f));
        }
    }

    if (active_flags.empty())
    {
        return "UNKNOWN_FLAG";
    }

    return std::accumulate(active_flags.begin() + 1, active_flags.end(), active_flags.front(),
        [](std::string acc, const std::string& flag)
        {
            return std::move(acc) + "|" + flag;
        });
}
```

dcrawleyqti: Appreciate the suggestion!

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb_negative.cpp
```diff
@@ -0,0 +1,246 @@
...
+    if (!is_extension_available(
+            device, "cl_khr_external_memory_android_hardware_buffer"))
+    {
+        log_info("cl_khr_external_memory_android_hardware_buffer is not "
+                 "supported on this platform. "
+                 "Skipping test.\n");
+        return TEST_SKIPPED_ITSELF;
+    }
+
+    AHardwareBuffer_Desc aHardwareBufferDesc = { 0 };
```
ahesham-arm, line 43: This does not set `stride` (well, it sets it to zero). `stride` is ignored for `AHardwareBuffer_allocate()` but it is not clear to me if this would cause a problem with `AHardwareBuffer_isSupported()`.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -0,0 +1,1911 @@
...
+
+                // Populate AHB with random data
+                size_t pixelSize = get_pixel_size(&format.clImageFormat);
+                image_descriptor imageInfo = { 0 };
+                imageInfo.format = &format.clImageFormat;
+                imageInfo.type = format.clMemObjectType;
+                imageInfo.width = resolution.width;
+                imageInfo.height = resolution.height;
+                imageInfo.rowPitch = hardware_buffer_desc.stride * pixelSize;
+                test_assert_error(imageInfo.rowPitch
```
ahesham-arm, line 271: These early returns will leak `aHardwareBuffer` as `AHardwareBuffer_release()` is not called.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -0,0 +1,1911 @@
...
+                        unsigned pixel_index = row * imageInfo.width + col;
+                        switch (outputType)
+                        {
+                            case kFloat: {
+                                auto *cl_ptr = reinterpret_cast<cl_float4 *>(
+                                    ocl_verify_data.data());
+                                auto *ahb_ptr = reinterpret_cast<cl_float4 *>(
+                                    ahb_verify_data.data());
+
+                                if ((cl_ptr[pixel_index].s0
```
ahesham-arm, line 426: Do we expect these floating point comparisons to pass without any epsilons?

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -0,0 +1,1911 @@
...
+                AHardwareBuffer_release(aHardwareBuffer);
+                aHardwareBuffer = nullptr;
+            }
+        }
+    }
+
+    return TEST_PASS;
+}
+
+REGISTER_TEST(test_enqueue_copy_image_to_buffer)
```
ahesham-arm, line 988: These tests seem to duplicate a lot of setup boilerplate code, can this be extracted into a single static helper function?

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -0,0 +1,1911 @@
...
+                imageInfo.rowPitch = resolution.width * resolution.height
+                    * pixelSize; // Data is tightly packed
+                test_assert_error(imageInfo.rowPitch
+                                      >= pixelSize * imageInfo.width,
+                                  "Row pitch is smaller than width");
+
+                size_t origin[] = { 0, 0, 0 };
+                size_t region[] = { imageInfo.width, imageInfo.height, 1 };
+
+                auto verificationValue = static_cast<char *>(malloc(pixelSize));
```
ahesham-arm, line 1678: Please use a smart pointer instead. `std::unique_ptr` probably works for this case.

ahesham-arm (changes requested):

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb_negative.cpp
ahesham-arm: Checked the AOSP source code, this will not cause a problem. It is intended to be populated by calling `AHardwareBuffer_describe()`

ahesham-arm (approved):

PR #2414, opened 2025-06-09
Test releasing a command-buffer after submission but before execution has finished

EwanC:
Add cl_khr_command_buffer test that is it valid to release a command-buffer after it has been enqueued but before execution is finished.

This stresses the semantics from [clReleaseCommandBufferKHR](https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/clReleaseCommandBufferKHR.html#_description) that: "After the command_buffer reference count becomes zero **and has finished execution**, the command-buffer is deleted"


bashbaug (commented):
LGTM.  There are things we could do with user events that would ensure that execution has not finished, but this case is much more likely to happen in practice so I think it's the more important one to test.

lakshmih:
Reviewing and will comment by next week, thanks

bashbaug:
Merging offline with multiple approvals in place, as discussed in the July 1st teleconference.

PR #2415, opened 2025-06-10
Filter out CL_DEVICE_TYPE_CUSTOM from multi device testing in conform…

lakshmih:
…ance tests

Change-Id: I0561da359fc61cf6fad96b6cdad2e629b52df02b

### DIFF
--- a/test_conformance/multiple_device_context/test_multiple_devices.cpp
+++ b/test_conformance/multiple_device_context/test_multiple_devices.cpp
@@ -191,36 +191,50 @@ static int test_device_set(size_t deviceCount, size_t queueCount,
 REGISTER_TEST(two_devices)
 {
     cl_platform_id platform;
-    cl_device_id devices[2];
+    cl_device_id devices[MAX_DEVICES];
+    cl_device_id devicesFiltered[MAX_DEVICES];
     int err;
-    cl_uint numDevices;
+    cl_uint numDevices, numDevicesFiltered;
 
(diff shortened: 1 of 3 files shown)


test_conformance/multiple_device_context/test_multiple_devices.cpp
```diff
@@ -191,36 +191,50 @@ static int test_device_set(size_t deviceCount, size_t queueCount,
...
     err = clGetPlatformIDs(1, &platform, NULL);
     test_error( err, "Unable to get platform" );
 
     /* Get some devices */
-    err = clGetDeviceIDs(platform,  CL_DEVICE_TYPE_ALL, 2, devices, &numDevices );
-    test_error( err, "Unable to get 2 devices" );
+    err = clGetDeviceIDs(platform,  CL_DEVICE_TYPE_ALL, MAX_DEVICES, devices, &numDevices );
+    test_error( err, "Unable to get multiple devices" );
+
+    /* Filter out CL_DEVICE_TYPE_CUSTOM devices */
```
rjodinchr, line 206: That logic is copy/pasted below, we may have a function for it then.

bashbaug: Or, could both of these places use the "GetOpposingDevice" function already in the harness?  It seems to roughly do what we'd want this code to do also - to find another valid device to test with.

PR #2421, opened 2025-06-11
allow specifying CL_DEVICE_TYPE_ALL as the harness device type

bashbaug:
Currently, selecting a different device in a platform to test is rather cumbersome, for two reasons:

1. The default device type tested is the "default" device and there is at most one default device in a platform.  This means that, by itself, choosing any non-zero device index is by definition out-of-range:

```sh
$ CL_PLATFORM_INDEX=1 CL_DEVICE_INDEX=1 ./test_conformance/basic/test_basic 
Requesting Default device based on command line for platform index 1 and device index 1
device index out of range -- choosen_device_index (1) >= num_devices (1)
```

2. To choose a non-default device type you therefore need to explicitly specify another device type also, but "all" is not a valid device type in the harness.  This means that you need to know both the device type and the index of the device within that device type to choose the device to test.

```sh
$ CL_DEVICE_TYPE=all CL_PLATFORM_INDEX=1 CL_DEVICE_INDEX=1 ./test_conformance/basic/test_basic 
Unknown CL_DEVICE_TYPE env variable setting: all.
Aborting...
Aborted (core dumped)
```

This PR aims to fix (2), by allowing "all" as a device type.  In the future, we could consider making the default device type "all" vs. "default", which would fix (1) also, but that will likely need more discussion and should be done in a separate PR.

### DIFF
--- a/test_common/harness/testHarness.cpp
+++ b/test_common/harness/testHarness.cpp
@@ -197,7 +197,10 @@ int runTestHarnessWithCheck(int argc, const char *argv[], int testNum,
     if (env_mode != NULL)
     {
         based_on_env_var = 1;
-        if (strcmp(env_mode, "gpu") == 0
+        if (strcmp(env_mode, "all") == 0
+            || strcmp(env_mode, "CL_DEVICE_TYPE_ALL") == 0)
+            device_type = CL_DEVICE_TYPE_ALL;
+        else if (strcmp(env_mode, "gpu") == 0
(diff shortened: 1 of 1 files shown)


kpet (approved):
This change is fine but it reminded me of https://github.com/KhronosGroup/OpenCL-CTS/pull/1073. We should decide whether to retain environment variables or command line parameters to control device/platform selection.

lakshmih:
Need time to review this, thanks.

bashbaug:
Merging as discussed in the July 15th teleconference.

PR #2423, opened 2025-06-16
Fix program leaks in test_spir compile_and_link

444973455:
Fixed program leaks caused by missing clRelease calls


svenvh (approved):

bashbaug:
Merging as discussed in the July 1st teleconference (and offline).

PR #2424, opened 2025-06-16
Added support for cl_ext_float_atomics in CBasicTestFetchSubSpecialFloats with atomic_double

shajder:
Related to #2142, according to the work plan, extending CBasicTestFetchSubSpecialFloats with support for atomic_double.


test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -2161,7 +2161,10 @@ class CBasicTestFetchSubSpecialFloats
         : CBasicTestFetchSpecialFloats<HostAtomicType, HostDataType>(dataType,
                                                                      useSVM)
     {
-        if constexpr (std::is_same_v<HostDataType, HOST_FLOAT>)
+        if constexpr (
+            std::is_same_v<
+                HostDataType,
+                HOST_FLOAT> || std::is_same_v<HostDataType, HOST_DOUBLE>)
```
bashbaug, line 2167: Can this use instead:

```suggestion
        if constexpr (is_host_fp_v<HostDataType>)
```

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -2206,7 +2209,10 @@ class CBasicTestFetchSubSpecialFloats
                        HostDataType *startRefValues,
                        cl_uint whichDestValue) override
     {
-        if constexpr (std::is_same_v<HostDataType, HOST_FLOAT>)
+        if constexpr (
+            std::is_same_v<
+                HostDataType,
+                HOST_FLOAT> || std::is_same_v<HostDataType, HOST_DOUBLE>)
```
bashbaug, line 2215: Consider:

```suggestion
        if constexpr (is_host_fp_v<HostDataType>)
```

test_conformance/c11_atomics/test_atomics.cpp
shajder: Agreed, this pattern repeats in a several other locations and it comes from PRs layering on the same code over time. If you don't mind, I'd prefer to fix all of these together in a single housekeeping PR once everything here is merged. Is that okay with you?

bashbaug: I think you'll need to touch this line anyhow when https://github.com/KhronosGroup/OpenCL-CTS/pull/2417 gets merged, but I'm fine fixing it in a final cleanup PR if that's what you want to do.

bashbaug (dismissed):

bashbaug:
Merging (offline) as discussed in the September 15th teleconference.

bashbaug (approved):

PR #2426, opened 2025-06-26
Fix 'fpclassify: ambiguous call' in MSVC 2022

wenju-he:
Similar to #2219, we see "'fpclassify': ambiguous call" error in test_conformance\basic\test_fpmath.cpp.
This PR fixes the issue by adding utility function isnan_fp in testHarness.h.
Note IsHalfNaN in test_conformance/math_brute_force/utility.h isn't touched, to avoid change to this file.

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -37,11 +37,7 @@ double sRGBmap(float fc)
 {
     double c = (double)fc;
 
-#if !defined(_WIN32)
-    if (std::isnan(c)) c = 0.0;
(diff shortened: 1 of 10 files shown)


wenju-he:
Alternative fix to the error I see is to add constexpr to https://github.com/KhronosGroup/OpenCL-CTS/blob/9265cbb2c274/test_conformance/basic/test_fpmath.cpp#L104

The fix in this PR looks more general.

bashbaug (approved):

bashbaug (approved):

Issue #2427, opened 2025-06-27
[CTS] conversions CustomConversionsTest need skip fp_sat_xxx


wenhu1024:
follow https://registry.khronos.org/OpenCL/sdk/3.0/docs/man/html/ 

```
Conversions to floating-point type shall conform to IEEE-754 rounding rules. The _sat modifier may not be used for conversions to floating-point formats.
```

therefore, i think CustomConversionsTest also need skip fp_sat_xxx

PR #2428, opened 2025-06-27
conversions CustomConversionsTest skip fp_sat_xxx

wenhu1024:
related https://github.com/KhronosGroup/OpenCL-CTS/issues/2427

### DIFF
--- a/test_conformance/conversions/basic_test_conversions.cpp
+++ b/test_conformance/conversions/basic_test_conversions.cpp
@@ -448,6 +448,13 @@ cl_int CustomConversionsTest::Run()
             continue;
         }
 
+        // skip illegal saturated conversions to float type
+        if (kSaturated == sat
+            && (outType == kfloat || outType == kdouble || outType == khalf))
+        {
+            continue;
+        }
+
         // skip longs on embedded
         if (!gHasLong
             && (inType == klong || outType == klong || inType == kulong


test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -448,6 +448,12 @@ cl_int CustomConversionsTest::Run()
             continue;
         }
 
+        // skip illegal saturated conversions to float type
+        if (kSaturated == sat && (outType == kfloat || outType == kdouble))
```
shajder: Good catch, I think you should take into account khalf as well, thanks!

wenhu1024: OK,thanks!

shajder:
LGTM

bashbaug (commented):
I think this is fine, but when I run the conversions test suite I don't see any difference in the number of tests that are run.

I'm only running the "wimpy" tests, and only on scalars so the tests run faster, but I don't think this should matter?  For completeness my command line is:

`./test_conformance/conversions/test_conversions -w -1`

If the same tests are run before and after this change then what does it do exactly?  Thanks!

wenhu1024:
If run ./test_conversions fp_sat_xxx , the test result is always pass.

for example:
```
./test_conversions float_sat_int


conversions...
conversions passed
PASSED sub-test.
PASSED test.

```

PR #2429, opened 2025-06-30
Retain output memory objects for simultaneous command buffer tests

Nuullll:
Memory objects created in `EnqueueSimultaneousPass()` are used by kernels
that don't execute until the user event is signaled. Without retaining
these objects, they would be destroyed before the deferred kernel
execution occurs.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
@@ -337,6 +337,8 @@ struct SimultaneousMutableDispatchTest : public BasicMutableCommandBufferTest
                                                       * buffer_size_multiplier,
                                                   nullptr, &error);
         test_error(error, "clCreateBuffer failed");
+        // Retain new output memory object until the end of the test.
+        retained_output_buffers.push_back(new_out_mem);
 
         cl_mutable_dispatch_arg_khr arg_1{ 1, sizeof(new_out_mem),
                                            &new_out_mem };
@@ -428,6 +430,8 @@ struct SimultaneousMutableDispatchTest : public BasicMutableCommandBufferTest
 
     clKernelWrapper kernel_fill;
     clProgramWrapper program_fill;
+
+    std::vector<clMemWrapper> retained_output_buffers;
 
     const size_t test_global_work_size = 3 * sizeof(cl_int);
     const cl_int pattern_pri = 42;


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
```diff
@@ -377,6 +377,8 @@ struct SimultaneousMutableDispatchTest : public BasicMutableCommandBufferTest
                                     &pd.wait_events[2], nullptr);
         test_error(error, "clEnqueueReadBuffer failed");
 
+        // Retain new output memory object until the end of the test.
+        new_out_mems.push_back(std::move(new_out_mem));
```
EwanC: Slight readability preference for doing this straight after the `clCreateBuffer` call that creates `new_out_mem`, as it keeps code relating to the lifetime of the buffer in one place.

EwanC (approved):

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
bashbaug: This test is also rather complicated.  Maybe we should consider a different name for the vector, so it's clearer what it's doing?

bashbaug (dismissed):
Ooh, good catch.

I was initially thinking that the command buffer should be responsible for retaining the buffer (see https://github.com/KhronosGroup/OpenCL-Docs/pull/885), but then I remembered that this didn't apply to kernel arguments recorded into command buffers, so retaining the memory object "manually" is required in this case.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
Nuullll: Thanks for reviewing. I've addressed both comments:
1. Moved buffer lifetime management closer to creation: The `retained_output_buffers.push_back()` call is now placed immediately after the `clCreateBuffer()` call, keeping all code related to the buffer's lifetime management in one location.
 Note: I also removed `std::move()` since the buffer needs to remain valid for subsequent OpenCL API calls `clUpdateMutableCommandsKHR` and `clEnqueueReadBuffer`. The `clMemWrapper` copy constructor will properly handle the reference counting.

2. Improved variable naming: Renamed the vector from `new_out_mems` to `retained_output_buffers` to better clarify its purpose - retaining memory objects to prevent premature deallocation during test execution.

EwanC (approved):

bashbaug (approved):

bashbaug:

PR #2431, opened 2025-07-01
SVM tests where the size is zero

bashbaug:
Here are some tests to check how SVM APIs behave when the size is zero, see:

* https://github.com/KhronosGroup/OpenCL-Docs/issues/1403

These tests don't fail right now if some of the calls are unsuccessful.  This ensures that all sub-tests run on a given implementation.

### DIFF
--- a/test_conformance/SVM/CMakeLists.txt
+++ b/test_conformance/SVM/CMakeLists.txt
@@ -15,6 +15,7 @@ set(${MODULE_NAME}_SOURCES
     test_shared_address_space_fine_grain.cpp
     test_shared_address_space_fine_grain_buffers.cpp
     test_shared_sub_buffers.cpp
+    test_size_zero.cpp
     test_migrate.cpp
 )
 
--- a/test_conformance/SVM/test_size_zero.cpp
+++ b/test_conformance/SVM/test_size_zero.cpp
@@ -1,0 +1,191 @@
+//
+// Copyright (c) 2025 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "common.h"
+
+static int svm_size_zero_helper(cl_device_id device, cl_context context,
+                                cl_command_queue queue,
+                                cl_svm_mem_flags svmFlags)
+{
+    cl_int error = CL_SUCCESS;
+    cl_int value = 42;
+
+    clEventWrapper event;
+    cl_command_type cmdType = 0;
+
+    // Try allocating zero bytes of SVM.
+    // This should not crash.
+    void* svmZeroPtr =
+        clSVMAlloc(context, svmFlags, 0, svmFlags | CL_MEM_READ_WRITE);
+    log_info("      allocating zero bytes returned %p\n", svmZeroPtr);
+
+    // We should be able to free whatever we allocated.
+    clSVMFree(context, svmZeroPtr);
+
+    // Try freeing an explicit NULL pointer.
+    // This should not crash.
+    clSVMFree(context, nullptr);
+
+    clSVMWrapper svmPtr(context, sizeof(value), svmFlags | CL_MEM_READ_WRITE);
+    test_assert_error(svmPtr() != nullptr, "clSVMAlloc failed");
+
+    // Try filling zero bytes of the SVM pointer
+    event = nullptr; // Reset the event
+    error = clEnqueueSVMMemFill(queue, svmPtr(), &value, sizeof(value), 0, 0,
+                                nullptr, &event);
+    // test_error(error, "clEnqueueSVMMemFill with zero size failed");
+    log_info("      clEnqueueSVMMemFill with zero size returned %d\n", error);
+
+    // Check that the event for the zero-sized fill has the right command type
+    if (error == CL_SUCCESS)
+    {
+        error = clGetEventInfo(event, CL_EVENT_COMMAND_TYPE, sizeof(cmdType),
+                               &cmdType, nullptr);
+        test_error(error, "clGetEventInfo failed for CL_EVENT_COMMAND_TYPE");
+        // test_assert_error(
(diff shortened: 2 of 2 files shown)


bashbaug:
I'm going to add a couple of additional SVM-related "size is zero" tests:

* [ ] Calling clSetKernelExecInfo with CL_KERNEL_EXEC_INFO_SVM_PTRS with an empty set of SVM pointers.
    * This should definitely work, since this case is explicitly described in the spec.
* [ ] Calling clEnqueueSVMFree to asynchronously free an empty set of SVM pointers.

bashbaug:
Discussed in the July 8th teleconference.  One additional case to add:

* [ ] Calling clEnqueueSVMFree to asynchronously free a NULL pointer.

bashbaug:
> clEnqueueSVMMigrateMem with a NULL pointer and a size equal to zero.

Observation: All of the implementations I've tested consider this to be an error: `CL_INVALID_VALUE`.

bashbaug:
CC @karolherbst

test_conformance/SVM/test_corner_cases.cpp
```diff
@@ -0,0 +1,300 @@
...
+            cmdType, CL_COMMAND_SVM_FREE);
+    }
+
+    clSVMWrapper svmPtr(context, sizeof(value), svmFlags | CL_MEM_READ_WRITE);
+    test_assert_error(svmPtr() != nullptr, "clSVMAlloc failed");
+
+    // Try filling zero bytes of the SVM pointer
+    event = nullptr; // Reset the event
+    error = clEnqueueSVMMemFill(queue, svmPtr(), &value, sizeof(value), 0, 0,
+                                nullptr, &event);
```
karolherbst, line 72: I _think_ it's reasonable to assume that `size` being 0 should return an error. Not sure there are any valid uses cases for it and it probably always points to a programming error.

The spec currently states `CL_INVALID_VALUE if size is not a multiple of pattern_size` and I wonder if the intention was to also disallow 0 or not. Technically 0 is a multiple of anything and according to the spec an implementation will have to allow this behavior.

But similar arguments could be made for the other enqueue operations taking a size.

test_conformance/SVM/test_corner_cases.cpp
```diff
@@ -0,0 +1,300 @@
...
+        const void* svm_pointers[] = { nullptr };
+        const size_t sizes[] = { 0 };
+        event = nullptr; // Reset the event
+        error = clEnqueueSVMMigrateMem(queue, 1, svm_pointers, sizes,
+                                       CL_MIGRATE_MEM_OBJECT_CONTENT_UNDEFINED,
+                                       0, nullptr, &event);
+        // test_error(error, "clEnqueueSVMMigrateMem with NULL pointer failed");
+        log_info("      clEnqueueSVMMigrateMem with NULL pointer returned %s\n",
+                 IGetErrorString(error));
+    }
```
karolherbst, line 132: Yeah, that's a good one. The spec has a clear intention to allow this behavior: `CL_INVALID_VALUE if sizes[i] is non-zero range [svm_pointers[i], svm_pointers[i]+sizes[i]) is not contained within an existing clSVMAlloc allocation.`. So there is an error when the pointer is not part of an SVM allocation, but only for non-zero ranges. I think the wording could be a little clearer there, because I also missed that part, so it would be good to have a test for this specifically.

bashbaug: Hmm, `clEnqueueSVMMigrateMem` has a special-case for when the size is zero:

> Also, if sizes[i] is zero, then the entire allocation containing svm_pointer[i] is migrated.

Maybe this means that a `NULL` pointer shouldn't be valid, even when the size is zero?

test_conformance/SVM/test_corner_cases.cpp
bashbaug: Sorry for the delay, took a while to get back to this.

For unified SVM we decided that a size of zero is valid and this should not be an error, and we added tests for this scenario in https://github.com/KhronosGroup/OpenCL-CTS/pull/2436.

We don't _need_ to do the same for the non-extended SVM, but it'd be nice to have the same behavior.

I'd prefer not to revisit the unified SVM resolution, but I suppose that's an option, too.

test_conformance/SVM/test_corner_cases.cpp
bashbaug: We need to figure out what should happen for this case, both for standard SVM and for unified SVM.  As best I can tell:

1. All implementations I've tested are returning `CL_INVALID_VALUE` for this case, where the pointer is `NULL` and the size is zero.
2. We added a test for this case in https://github.com/KhronosGroup/OpenCL-CTS/pull/2436, but with a [comment ](https://github.com/KhronosGroup/OpenCL-CTS/pull/2436#issuecomment-3110356048) that "we're not sure if it's a valid test".

So:

1. We could keep this test as-is, which would align with the current unified SVM tests, but it would mean many shipping implementations are non-conformant.
2. We could remove this test or turn it into a negative test that verifies this scenario returns `CL_INVALID_VALUE`, but keep the test as-is for devices that support unified SVM.
3. We could remove this test and remove the test for unified SVM, or turn both tests into negative tests.

Note: It's slightly unfortunate that this is inconsistent behavior with the other "size equals zero" cases, but because this API explicitly has different behavior when the size is equal to zero, the inconsistency may be unavoidable.  Specifically, we [have](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueSVMMigrateMem):

> _svm_pointers_ is a pointer to an array of pointers. Each pointer in this array must be within an allocation produced by a call to [clSVMAlloc](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSVMAlloc).
> sizes is an array of sizes. [snip] if sizes[i] is zero, then the entire allocation containing svm_pointer[i] is migrated.

and the somewhat ambiguous error condition mentioned above:

> [CL_INVALID_VALUE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_VALUE) if sizes[i] is non-zero and the range [svm_pointers[i], svm_pointers[i]+sizes[i]) is not contained within an existing [clSVMAlloc](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clSVMAlloc) allocation.

bashbaug:
I've updated this PR to check for the unified SVM extension before running, even though these are SVM tests, as discussed in the November 4th memory subgroup.  I've also updated the tests to actually check for and report errors vs. logging the errors informationally, and I've updated the PR to target the unified SVM branch instead of the main branch.

All of the tests are passing with the unified SVM emulation layer.

I believe these tests are ready for review and merging now, so I am removing "draft" and marking the PR ready for review.

PR #2432, opened 2025-07-02
Add `cl_ext_immutable_memory_objects` tests writing to and from buffer

MichaelRizkalla-arm:
This change extends the test coverage for https://github.com/KhronosGroup/OpenCL-Docs/pull/1280

The change tests:
1. Writing to immutable buffers.
2. Writing to buffer/image from immutable buffers.
3. Reading from immutable buffers.

This change adds the following tests:
1. `test_negative_imagearraycopy`
2. `test_negative_imagearraycopy3d`
3. `test_immutable_bufferreadwriterect`
4. `test_immutable_arrayreadwrite`
5. `test_write_from_immutable_buffer_to_buffer`
6. `test_immutable_buffer_map_*`

and extends the following tests:
1. `test_arrayimagecopy3d`
2. `test_arrayimagecopy`
3. `test_imagearraycopy3d`
4. `test_imagearraycopy`
5. `test_buffer_copy`
6. `test_buffer_partial_copy`

### DIFF
--- a/test_conformance/basic/test_imagearraycopy.cpp
+++ b/test_conformance/basic/test_imagearraycopy.cpp
@@ -27,6 +27,82 @@ using test_function_t = int (*)(cl_device_id, cl_context, cl_command_queue,
                                 cl_mem_flags, cl_mem_flags, cl_mem_object_type,
                                 const cl_image_format *);
 
+static int test_negative_imagearraycopy_single_format(
+    cl_device_id device, cl_context context, cl_command_queue queue,
+    cl_mem_flags image_flags, cl_mem_flags buffer_flags,
+    cl_mem_object_type image_type, const cl_image_format *format)
+{
+    std::unique_ptr<cl_uchar, decltype(&free)> bufptr{ nullptr, free },
+        imgptr{ nullptr, free };
+    clMemWrapper image;
+    clMemWrapper buffer;
+    const size_t img_width = 512;
+    const size_t img_height = 512;
+    const size_t img_depth = (image_type == CL_MEM_OBJECT_IMAGE3D) ? 32 : 1;
+    size_t elem_size;
+    size_t buffer_size;
+    cl_int err;
+    RandomSeed seed(gRandomSeed);
+
+    const size_t origin[3] = { 0, 0, 0 },
+                 region[3] = { img_width, img_height, img_depth };
+
+    log_info("Testing %s %s\n",
+             GetChannelOrderName(format->image_channel_order),
+             GetChannelTypeName(format->image_channel_data_type));
+
+    elem_size = get_pixel_size(format);
+    buffer_size =
+        sizeof(cl_uchar) * elem_size * img_width * img_height * img_depth;
+
+    if (image_flags & CL_MEM_USE_HOST_PTR || image_flags & CL_MEM_COPY_HOST_PTR)
+    {
+        imgptr.reset(static_cast<cl_uchar *>(
+            create_random_data(kUChar, seed, buffer_size)));
+    }
+
+    bufptr.reset(
+        static_cast<cl_uchar *>(create_random_data(kUChar, seed, buffer_size)));
+
+    if (CL_MEM_OBJECT_IMAGE2D == image_type)
+    {
+        image = create_image_2d(context, image_flags, format, img_width,
+                                img_height, 0, imgptr.get(), &err);
+    }
+    else
+    {
+        image =
+            create_image_3d(context, image_flags, format, img_width, img_height,
+                            img_depth, 0, 0, imgptr.get(), &err);
+    }
+    test_error(err, "create_image_xd failed");
+
+    if (!(image_flags & CL_MEM_USE_HOST_PTR
+          || image_flags & CL_MEM_COPY_HOST_PTR))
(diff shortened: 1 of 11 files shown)


test_conformance/buffers/testBase.h
```diff
@@ -27,4 +27,8 @@ extern const char* flag_set_names[];
 
 #define NUM_FLAGS 5
 
+extern const cl_mem_flags immutable_flag_set[];
+extern const char* immutable_flag_set_names[];
+#define NUM_IMMUTABLE_FLAGS 3
```
bashbaug: I find this confusing.

* Why is `NUM_FLAGS` equal to 5 when there are 8 entries in the array?
* Why are there separate arrays for immutable flags that are used sometimes, but not every time?

Could we have a single array instead and filter based on the flags we care about testing?

Alternatively, could we have two distinct arrays, and iterate over both if we want to test both mutable and immutable flags?

test_conformance/buffers/test_buffer_copy.cpp
```diff
@@ -76,10 +77,20 @@ static int test_copy( cl_command_queue queue, cl_context context, int num_elemen
...
+         src_flag_id++)
     {
-        for (int dst_flag_id = 0; dst_flag_id < NUM_FLAGS; dst_flag_id++)
+        for (int dst_flag_id = 0; dst_flag_id < NUM_FLAGS + NUM_IMMUTABLE_FLAGS;
+             dst_flag_id++)
         {
+            if (((flag_set[src_flag_id] & CL_MEM_IMMUTABLE_EXT)
+                 || (flag_set[dst_flag_id] & CL_MEM_IMMUTABLE_EXT))
+                && !is_extension_available(device,
+                                           "cl_ext_immutable_memory_objects"))
```
bashbaug: Can we factor out this check?  `is_extension_available` requires allocating space for a string query and then parsing the queried string, so it isn't cheap.

test_conformance/basic/test_imagearraycopy.cpp
```diff
@@ -188,17 +264,57 @@ REGISTER_TEST(imagearraycopy)
...
-    return test_imagearraycommon(device, context, queue, CL_MEM_READ_WRITE,
-                                 CL_MEM_READ_WRITE, CL_MEM_OBJECT_IMAGE2D,
-                                 test_imagearraycopy_single_format);
+    int error = test_imagearraycommon(device, context, queue, CL_MEM_READ_WRITE,
+                                      CL_MEM_READ_WRITE, CL_MEM_OBJECT_IMAGE2D,
+                                      test_imagearraycopy_single_format);
+
+    if (is_extension_available(device, "cl_ext_immutable_memory_objects"))
+    {
+        error += test_imagearraycommon(
```
bashbaug: We should use `|=` here instead of `+=`.  Since we have:

```c++
typedef enum test_status
{
    TEST_PASS = 0,
    TEST_FAIL = 1,
    TEST_SKIP = 2,
    TEST_SKIPPED_ITSELF = -100,
} test_status;
```

If both tests fail with `+=` we'll end up with `TEST_SKIP`.

```suggestion
        error |= test_imagearraycommon(
```

This pattern is used in a few other places in this PR, also.

test_conformance/buffers/test_buffer_map.cpp
```diff
@@ -671,6 +671,96 @@ static int test_buffer_map_read( cl_device_id deviceID, cl_context context, cl_c
 
 }   // end test_buffer_map_read()
 
+int test_immutable_buffer_map(cl_device_id device, cl_context context,
+                              cl_command_queue queue, int num_elements,
+                              size_t size, char *type, int loops)
```
bashbaug:
```suggestion
                              size_t size, const char *type, int loops)
```

If this is OK, we can remove an explicit cast below, also.

test_conformance/buffers/test_buffer_write.cpp
```diff
@@ -1974,3 +1974,245 @@ REGISTER_TEST(buffer_write_async_ulong)
 
 }   // end test_buffer_ulong_write_array_async()
 
+
+int immutable_test_buffer_write(cl_device_id device, cl_context context,
+                                cl_command_queue queue, int num_elements,
+                                size_t size, char *type, int loops,
```
bashbaug:
```suggestion
                                size_t size, const char *type, int loops,
```

test_conformance/basic/test_imagearraycopy.cpp
MichaelRizkalla-arm: I've updated all the occurrences.

test_conformance/buffers/test_buffer_copy.cpp
MichaelRizkalla-arm: I moved all similar checks to a variable outside the loop.

test_conformance/buffers/testBase.h
MichaelRizkalla-arm: I followed the first option; a single array with all available flags to test. Updated `buffer` test suite existing tests to skip to new flags whenever needed.

lakshmih (approved):

PR #2434, opened 2025-07-03
Fix data race in mutable command buffer simultaneous execution test

Nuullll:
Prior to this change, both `clEnqueueReadBuffer` calls before and after updating the command buffer were writing to the same `output_buffer`, causing a data race condition and the first call's result to be overwritten. This commit introduces separate destination vectors (`output_buffer` and `updated_output_buffer`) for these operations and verifies both results independently to ensure test integrity.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
@@ -297,6 +297,7 @@ struct SimultaneousMutableDispatchTest : public BasicMutableCommandBufferTest
     {
         cl_int offset;
         std::vector<cl_int> output_buffer;
+        std::vector<cl_int> updated_output_buffer;
         // 0:user event, 1:offset-buffer fill event, 2:kernel done event
         clEventWrapper wait_events[3];
     };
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
```diff
@@ -329,7 +330,7 @@ struct SimultaneousMutableDispatchTest : public BasicMutableCommandBufferTest
         error = clEnqueueReadBuffer(work_queue, out_mem, CL_FALSE,
                                     pd.offset * sizeof(cl_int), data_size(),
                                     pd.output_buffer.data(), 1,
```
EwanC: We don't verify the contents of ` pd.output_buffer` from this call, before the second `clEnqueueReadBuffer` called that overwrites it, so this `clEnqueueReadBuffer` call doesn't add much testing value right now. Would it be worth adding a new std::vector member to `SimulPassData` so that both the `clEnqueueReadBuffer` outputs are stored and can verified afterwards? Otherwise, we may as well fix this race condition by removing this `clEnqueueReadBuffer` call.

Nuullll: Thanks! I've separated the destination vectors for these two calls.

EwanC (approved):
LGTM!

PR #2435, opened 2025-07-03
Use CL_KERNEL_WORK_GROUP_SIZE more often

karolherbst:
Drivers _may_ choose to advertise values for `CL_DEVICE_MAX_WORK_GROUP_SIZE` or `CL_DEVICE_MAX_WORK_ITEM_SIZES`  that kernels without a `reqd_work_group_size` are not able to be launched with.

The CTS should therefore make sure that the local_size passed to `clEnqueueNDRangeKernel` does not exceed `CL_KERNEL_WORK_GROUP_SIZE`

This fixes it up in two places I've noticed this not happening.

### DIFF
--- a/test_conformance/profiling/execute_multipass.cpp
+++ b/test_conformance/profiling/execute_multipass.cpp
@@ -106,28 +106,6 @@ static int run_kernel( cl_device_id device, cl_context context, cl_command_queue
     threads[1] = h;
     threads[2] = d;
 
-    err = clGetDeviceInfo(device, CL_DEVICE_MAX_WORK_ITEM_SIZES,
-                          3 * sizeof(size_t), (size_t *)localThreads, NULL);
-    if (err)
-    {
-        log_error("clGetDeviceInfo(CL_DEVICE_MAX_WORK_ITEM_SIZES) failed\n");
-        return -1;
-    }
-    err = clGetDeviceInfo(device, CL_DEVICE_MAX_WORK_GROUP_SIZE, sizeof(size_t),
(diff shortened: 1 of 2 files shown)


test_conformance/profiling/execute_multipass.cpp
```diff
@@ -161,6 +139,27 @@ static int run_kernel( cl_device_id device, cl_context context, cl_command_queue
...
+        log_error("clGetDeviceInfo(CL_KERNEL_WORK_GROUP_SIZE) failed\n");
+        return -1;
+    }
+    localThreads[0] =
+        std::min({ localThreads[0], threads[0], maxWorkgroupSize });
+    localThreads[1] = std::min(
+        { localThreads[1], threads[1], maxWorkgroupSize / localThreads[0] });
+    localThreads[2] =
+        std::min({ localThreads[2], threads[2],
+                   maxWorkgroupSize / (localThreads[0] * localThreads[1]) });
```
bashbaug, line 144: This is fine as-is, but if we end up doing this in a lot of places we probably ought to add a helper function in the harness for it.

Note: I did find `get_max_common_3D_work_group_size` and `get_max_allowed_work_group_size`, but neither appear to be a replacement for this code.

bashbaug (approved):
If we want to get more of these then we'll probably want to check at least all of the places where we're currently querying for `CL_DEVICE_MAX_WORK_ITEM_SIZES`.

bashbaug:
Discussed in the August 5th teleconference. OK to merge when the conflict is resolved.

PR #2437, opened 2025-07-07
Fix buffer size for mutable dispatch command buffer test

Nuullll:


bashbaug (approved):

bashbaug:

PR #2438, opened 2025-07-07
Test for mutable-dispatch original work-dim

EwanC:
New test following on from OpenCL-Docs Issues discussion https://github.com/KhronosGroup/OpenCL-Docs/issues/1390#issuecomment-3023818903

Noting that we have no test coverage for using the original value of `work_dim` during command-buffer update. All of our current CTS testing uses `0` for the `work_dim` to signify no update from the original value, however this test explicitly uses the original value.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_dim.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_dim.cpp
@@ -82,14 +82,10 @@ struct MutableDispatchWorkDim : public InfoMutableCommandBufferTest
                                         __global uint *dst1,
                                         __global uint *dst2)
             {
-                size_t gid0 = get_global_id(0);
-                dst0[gid0] = get_global_size(0);
-
-                size_t gid1 = get_global_id(1);
-                dst1[gid1] = get_global_size(1);
-
-                size_t gid2 = get_global_id(2);
-                dst2[gid2] = get_global_size(2);
+                size_t gid = get_global_linear_id();
+                dst0[gid] = get_global_size(0);
+                dst1[gid] = get_global_size(1);
+                dst2[gid] = get_global_size(2);
             })";
 
         cl_int error = create_single_kernel_helper(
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_dim.cpp
```diff
@@ -0,0 +1,223 @@
...
+            {
+                size_t gid0 = get_global_id(0);
+                dst0[gid0] = get_global_size(0);
+
+                size_t gid1 = get_global_id(1);
+                dst1[gid1] = get_global_size(1);
+
+                size_t gid2 = get_global_id(2);
+                dst2[gid2] = get_global_size(2);
+            })";
```
bashbaug: This probably won't matter in practice, but strictly speaking I think there is a data race here, since multiple work-items will share the same global ID in each dimension for a 3D dispatch.

Options:
1. Convince oursleves that this doesn't matter in practice and do nothing.
2. Switch to a 1D dispatch instead, so each global ID is unique.
3. Use a linear global ID instead.

bashbaug (commented):
This is a nice and clean test.  I have one small question, see below.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_dim.cpp
EwanC: ah that's a great spot, I'm not sure we have any mutable-dispatch tests that are more than 1D, so that's why I though having a 3D update test would have some bonus coverage.

I considered using linear global id, but since it's a OpenCL C 2.0 function I thought that might be an unnecessary constraint on when the test. I don't like the idea of adding code when we know it's a possible data race, so I'll update this patch to use that.

bashbaug: I don't think it's worth worrying about this, but if you did want to eliminate the OpenCL C 2.0 requirement imposed by `get_global_linear_id()` then you could compute the global linear ID by hand in the kernel.

bashbaug (approved):
LGTM!

bashbaug:

Issue #2279, opened 2025-02-16
Patchy coverage for `clEnqueueCopyBufferRect`


kpet:
- No coverage of the case where the provided pitches are zero (see https://github.com/kpet/clvk/pull/753)
- This command is only tested by `test_bufferreadwriterect` which does not seem to cover all possible region dimensionalities either. We should review what coverage is needed and extend the test accordingly.

shajder:
@kpet @bashbaug  In my evaluation of test coverage for the `test_bufferreadwriterect` function, I'm assuming that by randomizing the argument values with equal probability, we're likely to cover all necessary scenarios for the function’s dimensional input. This covers all cases of memory allocation sizes, specifically within the limits of (1..`CL_DEVICE_MAX_MEM_ALLOC_SIZE`) for 3D regions. Please correct me if I am wrong but I can't spot any missing coverage for dimensionalities.

Testing every possible region dimension is unnecessarily exhaustive. Instead, we can focus on key scenarios, such as reducing randomization by selecting special cases. For example, testing could be structured as follows:

- line stretching along the X-axis: (1..maxX, 1, 1),
- plane extending over the X and Y axes: (1..maxX, 1..maxY, 1),
- full 3D volume: (1..maxX, 1..maxY, 1..maxZ), and so on.

Would that be the intended result ?

PR #2440, opened 2025-07-08
Added sub-case for basic test where the provided pitches are zero

shajder:
Related to #2279


test_conformance/basic/test_bufferreadwriterect.cpp
```diff
@@ -506,17 +538,18 @@ static int test_bufferreadwriterect_impl(cl_device_id device,
...
                 log_info("%zu Copy %zu offset (%zu,%zu,%zu) -> %zu offset "
                          "(%zu,%zu,%zu) region (%zux%zux%zu = %zu)\n",
                          iter, src, soffset[0], soffset[1], soffset[2], dst,
                          doffset[0], doffset[1], doffset[2], sregion[0],
                          sregion[1], sregion[2],
                          sregion[0] * sregion[1] * sregion[2]);
-                if ((err = test_functions.copy(src, soffset, sregion, dst,
-                                               doffset, dregion)))
+                if ((err = copy_region(src, soffset, sregion, dst, doffset,
+                                       dregion)))
```
bashbaug: I don't understand this part of the change.  Shouldn't this continue to call `test_functions.copy` instead of `copy_region` directly?

Why are extra `{curly braces}` needed?

test_conformance/basic/test_bufferreadwriterect.cpp
```diff
@@ -88,78 +88,110 @@ void print_buffer(BufferType* buf, size_t w, size_t h, size_t d) {
...
 
-
-
 // This function invokes the CopyBufferRect CL command and then mirrors the operation on the host side verify buffers.
-int copy_region(size_t src, size_t soffset[3], size_t sregion[3], size_t dst, size_t doffset[3], size_t dregion[3]) {
-
+int copy_region(size_t src, size_t soffset[3], size_t sregion[3], size_t dst,
+                size_t doffset[3], size_t dregion[3])
+{
+    bool comp_pitch = (genrand_int32(mt) % 2) != 0 ? true : false;
```
bashbaug: Nitpick: I don't know what `comp_pitch` is supposed to mean.  Could we spell "comp" out completely to add clarity?

Additionally, with this changes we either always seem to compute all four pitches, or we pass zero for all four pitches.  Should we test different combinations of zero and non-zero pitches?

bashbaug (commented):
This adds additional coverage for `clEnqueueCopyBufferRect`.

Should we add additional coverage for `clEnqueueReadBufferRect` and `clEnqueueWriteBufferRect` also?  If this is something we want, it's fine to add it in a separate PR.

test_conformance/basic/test_bufferreadwriterect.cpp
shajder: Corrected

test_conformance/basic/test_bufferreadwriterect.cpp
shajder: Corrected

test_conformance/basic/test_bufferreadwriterect.cpp
```diff
@@ -88,78 +88,118 @@ void print_buffer(BufferType* buf, size_t w, size_t h, size_t d) {
...
+                 src_row_pitch, src_slice_pitch, dst_row_pitch, dst_slice_pitch,
+                 0, NULL, NULL))
             != CL_SUCCESS)
         {
-            CL_EXIT_ERROR(err, "clEnqueueCopyBufferRect failed between %u and %u",(unsigned)src,(unsigned)dst);
+            CL_EXIT_ERROR(err,
+                          "clEnqueueCopyBufferRect failed between %u and %u",
+                          (unsigned)src, (unsigned)dst);
         }
     }
```
joselopeqti, line 212: The calculation for `spitch`, `sslice`, `dpitch` and `dslice` in lines 209 to 213 has to be changed as well. The specs says:

>  If src_row_pitch is 0, src_row_pitch is computed as region[0].
>  If src_slice_pitch is 0, src_slice_pitch is computed as region[1] × src_row_pitch.
>  If dst_row_pitch is 0, dst_row_pitch is computed as region[0].
>  If dst_slice_pitch is 0, dst_slice_pitch is computed as region[1] × dst_row_pitch.

So when the row or slice pitches are 0, the region values should be used.

```suggestion
    size_t spitch = (src_row_pitch == 0) ? sregion[0] : src_row_pitch;
    size_t sslice = (src_slice_pitch == 0) ? sregion[1] * spitch : src_slice_pitch;    
    size_t dpitch = (dst_row_pitch == 0) ? dregion[0] : dst_row_pitch;
    size_t dslice = (dst_slice_pitch == 0) ? dregion[1] * dpitch : dst_slice_pitch;
```

joselopeqti (changes requested):

test_conformance/basic/test_bufferreadwriterect.cpp
shajder: FYI: @bashbaug @kpet 

> So when the row or slice pitches are 0, the region values should be used.

To me this sounds like a specification issue. If anything other than the actual source/destination buffer `width`/`height` is used for the row/slice pitches, we will no longer be copying a rectangular (or cubic) region. Instead, the copy will become skewed because each row will start at an incorrect offset. Am I missing something ?

shajder: @joselopeqti I've applied your correction with a minor caveat:

>    size_t dpitch = (dst_row_pitch == 0) ? **sregion**[0] : dst_row_pitch;
>    size_t dslice =
>        (dst_slice_pitch == 0) ? **sregion**[1] * dpitch : dst_slice_pitch;

shajder:
> Marcin will file a spec issue to track resolution.

@bashbaug  after double-checking the spec and the test, I've applied @joselopeqti's corrections. I believe the fallback of pitches to `region[N]`-related values when equal to zero may indeed be confusing for developers following the logic of our CTS test. On the other hand, a parameters of `clCreateBuffer` has no width, height, or depth - these are only conventional values in our test. In that sense, the fallback could legitimately be many things, and for `clEnqueueCopyBufferRect` the most reasonable choice is `region[N]`. I think we should leave it as is, without the issue.

test_conformance/basic/test_bufferreadwriterect.cpp
```diff
@@ -48,20 +52,19 @@ size_t height [TotalImages];
 size_t depth  [TotalImages];
 
 // cl buffer and host buffer.
-cl_mem buffer [TotalImages];
-BufferType* verify[TotalImages];
-BufferType* backing[TotalImages];
+clMemWrapper buffer[TotalImages];
```
joselopeqti: @shajder could you move the buffer array to the function scope?

shajder: > @shajder could you move the buffer array to the function scope?

@joselopeqti sure can do, how about moving it all to struct ? Seems a bit less invasive (no need to change functions definitions).

joselopeqti: I'm not sure about the struct, maybe keeping a pointer to the clMemWrappers in the namespace but declare them in the function scope?

shajder: Corrected

joselopeqti (approved):

test_conformance/basic/test_bufferreadwriterect.cpp
joselopeqti: Thanks!

bashbaug:

Issue #2276, opened 2025-02-14
conventions for `int` return values


bashbaug:
Throughout the CTS, many functions return `int`, but for different purposes.  Sometimes, the `int` is an OpenCL error code.  In other places, the `int` is the test status, such as whether the test passed, failed or was skipped.

It is very difficult to keep the different meanings of the `int` return value correct, and is is very common to unintentionally return an OpenCL error code instead of a test status.  Because both `CL_SUCCESS` and `TEST_PASS` both have the value (zero), this mistake does usually does not cause an issue in practice, though there are no guarantees this is the case now or in the future.

Is this something we want to fix?  I think this could take one of two forms:

1. Embrace that the test status and OpenCL errors can be intermixed and stop trying to differentiate between the two.
    * The value zero will always indicate that the test passed, regardless whether it is `CL_SUCCESS` or `TEST_PASS`.
    * If we switched `TEST_FAIL` to be a negative value, a negative value could always indicate that the test failed, regardless whether it is an OpenCL error code or `TEST_FAIL`.
    * If we switched `TEST_SKIPPED_ITSELF` to be a positive value, a positive value coudl always indicate that the test was skipped.
    * Note, if we did this, we could clean up parts of the harness; for example, we wouldn't need both `test_error` and `test_error_fail`.
2. Use a strongly typed test status instead, which would make it impossible to mix the two.

PR #2441, opened 2025-07-09
Testing Existing SVM APIs remaining APIs tests

JohnKesapidesARM:
Tests for the following APIs:
* clEnqueueSVMMemcpy
* clEnqueueSVMMemFill
* clEnqueueSVMMap/clEnqueueSVMUnMap
* clEnqueueSVMMigrateMem
* clEnqueueSVMMemFree
* clSetKernelArgSVMPointer
* clSetKernelExecInfo

Change-Id: I46ef6e7086dde83f9b1d4adc222637d1e7f7ee14


JohnKesapidesARM:
@bashbaug

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -165,12 +173,32 @@ template <typename T> class USVMWrapper {
...
+            test_error(err, "clCreateBuffer failed.");
+
+            err =
+                clSetKernelArg(kernel_CopyMemory, 1, sizeof(src_mem), &src_mem);
+            test_error(err, "clSetKernelArg failed.");
+
+            size_t gws{ count * sizeof(T) };
+            err = clEnqueueNDRangeKernel(queue, kernel_CopyMemory, 1, nullptr,
+                                         &gws, nullptr, 0, nullptr, nullptr);
+            test_error(err, "clEnqueueNDRangeKernel failed\n");
```
bashbaug: Could you explain the benefits of switching to a copy kernel for reading and writing SVM vs. calling clEnqueueSVMMemcpy?  It should be much more common to call clEnqueueSVMMemcpy to do these types of transfers in real application, so my gut feeling is that we should use clEnqueueSVMMemcpy for testing as much as possible and use a copy kernel as a special case only when it is needed for some reason.

test_conformance/SVM/common.h
```diff
@@ -85,5 +86,25 @@ extern cl_int create_cl_objects(cl_device_id device_from_harness, const char** p
...
 
+static inline cl_int check_event_type(cl_event event,
+                                      cl_command_type expectedCommandType)
+{
+    cl_command_type commandType;
+    cl_int error = clGetEventInfo(event, CL_EVENT_COMMAND_TYPE,
+                                  sizeof(cl_command_type), &commandType, NULL);
+    test_error(error, "clGetEventInfo failed");
+
+    return commandType == expectedCommandType ? CL_SUCCESS : TEST_FAIL;
```
bashbaug: This is a useful wrapper, but I don't think we should be mixing `CL_SUCCESS` and `TEST_FAIL` (see https://github.com/KhronosGroup/OpenCL-CTS/issues/2276).  Since it looks like this helper is returning an OpenCL error code consider returning `CL_INVALID_VALUE` instead of `TEST_FAIL`?

test_conformance/SVM/test_unified_svm_execinfo.cpp
```diff
@@ -0,0 +1,312 @@
...
+            }
+
+            struct d { global unsigned char* ptr; };
+            kernel void test_IndirectAccessWrite(global struct d* dst, const global unsigned char* src)
+            {
+                dst->ptr[get_global_id(0)] = src[get_global_id(0)];
+            }
+        )";
+
+        cl_program program;
```
bashbaug: It'd be better to use a `clProgramWrapper` here, then the program will be properly released even if an error occurs (also, then you wouldn't need the explicit release).

test_conformance/SVM/unified_svm_fixture.h
bashbaug: Note, this also means we'll compile a kernel everytime we instantiate a USVM wrapper, which will also negatively impact test execution time.

test_conformance/SVM/test_unified_svm_execinfo.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+    // Test reading from USM pointer indirectly using clSetKernelExecInfo.
+    // The test will perform a memcpy on the device.
+    cl_int test_svm_exec_info_read(USVMWrapper<cl_uchar> *mem)
+    {
+        cl_int err = CL_SUCCESS;
+
+        std::vector<cl_uchar> src_data(alloc_count, 0);
+
+        for (size_t it = 0; it < test_iterations; it++)
```
bashbaug: How much extra test coverage do we get executing this test 100 times vs. one time, or perhaps a handful of times?

Could we get away with executing the test once passing the base of the USVM allocation to clSetKernelExecInfo, and another time passing a pointer to the middle of the USVM allocation?

If we are going to execute the test many times, can we factor out the calls to clCreateBuffer, so we aren't creating and destroying the same sized buffers each iteration?

bashbaug (commented):
Hi @JohnKesapidesARM , I'll have some more comments later but I wanted to get these initial comments out ASAP.  In general though, these tests look really good, thank you!

test_conformance/SVM/test_unified_svm_free.cpp
```diff
@@ -0,0 +1,182 @@
...
+    cl_int test_SVMFree(std::vector<void *> &buffers)
+    {
+        cl_int err = CL_SUCCESS;
+
+        clEventWrapper event;
+
+        CallbackData data{ context };
+
+        err = clEnqueueSVMFree(queue, buffers.size(), buffers.data(),
+                               callback_svm_free, &data, 0, 0, &event);
```
bashbaug, line 91: This is testing a very specific usage of `clEnqueueSVMFree`, specifically where the SVM memory is freed with a callback.  Should we test without a callback, also?

Note that for devices that support system SVM the spec requires the callback to be used when the set of SVM pointers to be freed includes system allocations..

test_conformance/SVM/test_unified_svm_free.cpp
```diff
@@ -0,0 +1,182 @@
...
+                                   void *svm_pointers[], void *user_data)
+{
+    auto data = (CallbackData *)user_data;
+
+    data->svm_pointers.resize(num_svm_pointers, 0);
+
+    for (size_t i = 0; i < num_svm_pointers; ++i)
+    {
+        data->svm_pointers[i] = svm_pointers[i];
+        clSVMFree(data->context, svm_pointers[i]);
```
bashbaug: For devices that support system SVM, or in the context of unified SVM when the SVM capabilities includes `CL_SVM_CAPABILITY_SYSTEM_ALLOCATED_KHR`, I think we should be freeing the allocation with the appropriate system free function rather than `clSVMFree` if the SVM pointer was allocated with a system allocation function.

test_conformance/SVM/test_unified_svm_map_unmap.cpp
```diff
@@ -0,0 +1,136 @@
...
+    cl_int run() override
+    {
+        cl_int err;
+        cl_map_flags test_flags[] = { CL_MAP_READ, CL_MAP_WRITE,
+                                      CL_MAP_WRITE_INVALIDATE_REGION,
+                                      CL_MAP_READ | CL_MAP_WRITE };
+
+        cl_uint max_ti = static_cast<cl_uint>(deviceUSVMCaps.size());
+
+        for (cl_uint ti = 0; ti < max_ti; ti++)
```
bashbaug, line 73: This test should only execute for SVM types that have the capability `CL_SVM_CAPABILITY_HOST_MAP_KHR`.

test_conformance/SVM/unified_svm_fixture.h
JohnKesapidesARM: I run into an interesting corner with clEnqueueSVMMemcpy.
The short explanation is that there is no guarantee it will execute on the device.

Here is a bit more detail:

For normal SVM allocations: It is an issue when one of the arguments is a native host pointer.  It may not be possible to import/"use directly" the host pointer on the device, so vendors may fall back on the host to service the API call. 
This is usually opaque to the end user, and apart from coverage loss it's not that big of an issue.

but here is where it gets interesting though:
If we try to do clEnqueueSVMMemcpy between Device-SVM and a native host pointer. If we can't import the host pointer directly on the device then this would require as to fall back on the host but Device SVM is not accessible on the device, so there is no way to service the call.

test_conformance/SVM/common.h
JohnKesapidesARM: good point, I'll update it

test_conformance/SVM/test_unified_svm_execinfo.cpp
JohnKesapidesARM: 👍

test_conformance/SVM/test_unified_svm_free.cpp
JohnKesapidesARM: good point I'll support for that case

test_conformance/SVM/test_unified_svm_free.cpp
JohnKesapidesARM: I'll add coverage for that case also

test_conformance/SVM/test_unified_svm_execinfo.cpp
JohnKesapidesARM: yeah I guess if offset from the start works for one case it should work for all. I'll drop the loop and do start and middle

test_conformance/SVM/unified_svm_fixture.h
bashbaug: OK, I'm even more convinced that we should revert these changes.  If we need to adjust the specification to accomodate these types of implementations, I'm open to this discussion, but I think applications are generally going to expect `clEnqueueSVMMemcpy` to work even when the SVM pointer is not accessible on the host (specifically, when the `CL_SVM_CAPABILITY_HOST_READ_KHR` and/or `CL_SVM_CAPABILITY_HOST_WRITE_KHR` capabilities are not present).

Alternatively, we could keep the spec as-is and implementations would be required to find a way to make this work if they want to expose these specific capability combinations.  It's a bit outside of the scope of this PR, but I can think of several possibilities to do this, such as wraping the host pointer in a buffer and copying the memory via a kernel (as is done here, but within the implementation rather than in the applicaiton), or by copying twice - for example, first on the device to temporary memory that is accessible on the device and the host, and then on the host to the native host pointer.

bashbaug:
FYI, after I fixed the map/unmap test to only run on SVM types with the `HOST_MAP` capability (and a few bugs in the emulation layer itself) I'm passing all of these tests on most of the devices I've tested with the USVM emulation layer.  Cool!

test_conformance/SVM/unified_svm_fixture.h
JohnKesapidesARM: OK I agree this should be its own discussion, I'll loop in Kevin. Makes sense that for App developers the expectation is that clEnqueueSVMMemcpy would execute on the device, but atm there is no such requirement from the spec. Vendors might have implemented that on the host for other reasons as well (like simplicity or performance). 
Another concern here is that in the tests we use it to test/validate device operations. these vendors might pass the tests but miss correct test coverage. I'll revert the change, but I think we need clarity on this before we merge these into main

test_conformance/SVM/test_unified_svm_map_unmap.cpp
JohnKesapidesARM: will fix

JohnKesapidesARM:
updated :)

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -77,6 +77,8 @@ template <typename T> class USVMWrapper {
     cl_int allocate(const size_t count,
                     const std::vector<cl_svm_alloc_properties_khr> props_ = {})
     {
+        cl_int err;
```
bashbaug: Nitpick: Since we aren't using the error code in any new places in this function, I'd prefer to revert this change, both to reduce diffs and to keep it narrowly scoped.

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -165,12 +168,18 @@ template <typename T> class USVMWrapper {
...
+            test_error(err, "clFinish failed");
         }
         else if (caps & CL_SVM_CAPABILITY_DEVICE_WRITE_KHR)
         {
             err = clEnqueueSVMMemcpy(queue, CL_TRUE, data + offset, source,
                                      count * sizeof(T), 0, nullptr, nullptr);
             test_error(err, "clEnqueueSVMMemcpy failed");
+
+            err = clFinish(queue);
+            test_error(err, "clFinish failed");
```
bashbaug: Nitpick: Since the memcpy is blocking I don't think this `clFinish` is adding anything.

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -215,12 +224,18 @@ template <typename T> class USVMWrapper {
...
         }
         else if (caps & CL_SVM_CAPABILITY_DEVICE_READ_KHR)
         {
             err = clEnqueueSVMMemcpy(queue, CL_TRUE, dst, data + offset,
                                      count * sizeof(T), 0, nullptr, nullptr);
             test_error(err, "clEnqueueSVMMemcpy failed");
+
+            err = clFinish(queue);
+            test_error(err, "clFinish failed");
```
bashbaug: Nitpick: Same as above, since the memcpy is blocking I don't think this `clFinish` is adding anything.

test_conformance/SVM/test_unified_svm_setarg.cpp
```diff
@@ -0,0 +1,193 @@
...
+        }
+
+        const char *programString = R"(
+            kernel void test_kernel( const global char* src, global char* dst)
+            {
+                dst[get_global_id(0)] = src[get_global_id(0)];
+            }
+        )";
+
+        cl_program program;
```
bashbaug: Nitpick: If we used a `clProgramWrapper` instead then we would not need to explicitly release the program object.

bashbaug:
Yup, all LGTM, thanks!

bashbaug:
Merging (into the unified SVM test branch) as discussed in the July 22nd memory subgroup.

PR #2443, opened 2025-07-13
test_vulkan: fix build flags and warnings

kpet:
- Do not override parent-provided value for CMAKE_CXX_FLAGS
- Stop building with -fpermissive (not required/bad practice)
- Delete unused variables
- Remove unnecessary casts


bashbaug (approved):
Merging offline, as discussed in the July 15th memory subgroup.

Issue #2282, opened 2025-02-18
Check that `clSetKernelArg` called with `arg_size` of 0 for an argument in the `local` address space returns `CL_INVALID_ARG_SIZE`


kpet:
We could potentially widen this issue to cover all negative testing for `clSetKernelArg`

Work plan:

`test_api`

- [ ] `CL_INVALID_KERNEL` - kernel is not a valid kernel object
- [ ] `CL_INVALID_ARG_INDEX` - arg_index is not a valid argument index
- [x] `CL_INVALID_ARG_VALUE` - image declared with the read_only qualifier and arg_value refers to an image object created with cl_mem_flags of `CL_MEM_WRITE_ONLY`, PR #2445
- [x] `CL_INVALID_ARG_VALUE` - image argument is declared with the write_only qualifier and arg_value refers to an image object created with cl_mem_flags of `CL_MEM_READ_ONLY`, PR #2445
- [ ] `CL_INVALID_MEM_OBJECT` - argument declared to be a memory object when the specified arg_value is not a valid memory object
- [ ] `CL_INVALID_SAMPLER` - type sampler_t when the specified arg_value is not a valid sampler object
- [ ] `CL_INVALID_DEVICE_QUEUE` - type queue_t when the specified arg_value is not a valid device queue object
- [ ] `CL_INVALID_ARG_SIZE` - arg_size does not match the size of the data type for an argument that is not a memory object
- [ ] `CL_INVALID_ARG_SIZE` - argument is a memory object and arg_size != sizeof(cl_mem)
- [x] `CL_INVALID_ARG_SIZE` - arg_size is zero and the argument is declared with the local qualifier, PR #2449
- [ ] `CL_INVALID_ARG_SIZE` - argument is a sampler and arg_size != sizeof(cl_sampler)
- [ ] `CL_INVALID_MEM_OBJECT` - argument declared to be a depth image, depth image array, multi-sample image, multi-sample image array, multi-sample depth image, or a multi-sample depth image array when the specified arg_value does not follow the rules described above for a depth memory object or memory array object argument

shajder:
After reviewing CTS source code I identified below missing nevative coverage for clSetKernelArg:

`test_api`

- [ ] `CL_INVALID_KERNEL` - kernel is not a valid kernel object
- [ ] `CL_INVALID_ARG_INDEX` - arg_index is not a valid argument index
[...]

PR #2445, opened 2025-07-15
Added test to verify negative result of clSetKernelArg with CL_INVALID_ARG_VALUE

shajder:
Related to #2282, according to work plan with mismatched read/write only qualifier for image_t objects

### DIFF
--- a/test_conformance/api/test_kernels.cpp
+++ b/test_conformance/api/test_kernels.cpp
@@ -87,24 +87,19 @@ const char *sample_two_kernel_program[] = {
 "\n"
 "}\n" };
 
-const char *sample_read_only_image_test_kernel = {
-    "__kernel void read_only_image_test(__write_only image2d_t img, __global "
-    "uint4 *dst)\n"
-    "{\n"
-    "    write_imageui(img, (int2)(get_global_id(0), get_global_id(1)), "
-    "dst[0]);\n"
-    "}\n"
-};
+const char *sample_read_only_image_test_kernel = R"(
+    __kernel void read_only_image_test(__write_only image2d_t img, __global uint4 *src)
+    {
+        write_imageui(img, (int2)(get_global_id(0), get_global_id(1)), src[0]);
+    }
+)";
 
-const char *sample_write_only_image_test_kernel = {
-    "__kernel void write_only_image_test(__read_only image2d_t src, __global "
-    "uint4 *dst)\n"
-    "{\n"
-    "    dst[0]=read_imageui(src, (int2)(get_global_id(0), "
-    "get_global_id(1)));\n"
-    "}\n"
-};
-
+const char *sample_write_only_image_test_kernel = R"(
+    __kernel void write_only_image_test(__read_only image2d_t src, __global uint4 *dst)
+    {
+        dst[0]=read_imageui(src, (int2)(get_global_id(0), get_global_id(1)));
+    }
+)";
 
 REGISTER_TEST(get_kernel_info)
 {
@@ -727,11 +722,11 @@ REGISTER_TEST(negative_set_read_write_image_arg)
 {
     cl_int error = CL_SUCCESS;
     clProgramWrapper program;
-    clKernelWrapper kernels[2];
-    clMemWrapper images[2], buffer;
+    clKernelWrapper write_image_kernel, read_image_kernel;
+    clMemWrapper write_only_image, read_only_image;
     const char *test_kernels[2] = { sample_read_only_image_test_kernel,
                                     sample_write_only_image_test_kernel };
-    constexpr cl_image_format formats = { CL_RGBA, CL_UNSIGNED_INT8 };
+    constexpr cl_image_format format = { CL_RGBA, CL_UNSIGNED_INT8 };
     const int size_dim = 128;
 
     // Setup the test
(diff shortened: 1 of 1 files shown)


test_conformance/api/test_kernels.cpp
```diff
@@ -87,6 +87,24 @@ const char *sample_two_kernel_program[] = {
 "\n"
 "}\n" };
 
+const char *sample_read_only_image_test_kernel = {
```
bashbaug: We should use C++ raw string literals for new code.

For an example, see: https://github.com/KhronosGroup/OpenCL-CTS/blob/d918b4165763a55d831a37efc8aea35eb5a09fdc/test_conformance/basic/test_writeimage.cpp#L32-L34

test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +722,65 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
     return TEST_PASS;
 }
+
+REGISTER_TEST(negative_set_read_write_image_arg)
+{
+    cl_int error = CL_SUCCESS;
+    clProgramWrapper program;
+    clKernelWrapper kernels[2];
+    clMemWrapper images[2], buffer;
+    const char *test_kernels[2] = { sample_read_only_image_test_kernel,
```
bashbaug: We never loop over these arrays, so can we use more descriptive names instead?  I'm thinking of things like:

write_image_kernel, read_image_kernel, write_only_image, read_only_image

test_conformance/api/test_kernels.cpp
```diff
@@ -87,6 +87,24 @@ const char *sample_two_kernel_program[] = {
 "\n"
 "}\n" };
 
+const char *sample_read_only_image_test_kernel = {
+    "__kernel void read_only_image_test(__write_only image2d_t img, __global "
+    "uint4 *dst)\n"
```
bashbaug: `dst` is never written to, so this should probably be `src`?

```suggestion
    "uint4 *src)\n"
```

test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +722,65 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+    images[1] = create_image_2d(context, CL_MEM_WRITE_ONLY, &formats, size_dim,
+                                size_dim, 0, nullptr, &error);
+    test_error(error, "create_image_2d failed");
+
+    error = clSetKernelArg(kernels[0], 1, sizeof(buffer), &buffer);
+    test_error(error, "clSetKernelArg failed");
+
+    // Run the test
+    error = clSetKernelArg(kernels[0], 0, sizeof(images[0]), &images[0]);
+    test_failure_error_ret(error, CL_INVALID_ARG_VALUE,
```
bashbaug, line 757: I'm not going to lose a ton of sleep over this, but it seems odd to me that we need to use `test_failure_error_ret` here vs. `test_failure_error`.

test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +722,65 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+
+REGISTER_TEST(negative_set_read_write_image_arg)
+{
+    cl_int error = CL_SUCCESS;
+    clProgramWrapper program;
+    clKernelWrapper kernels[2];
+    clMemWrapper images[2], buffer;
+    const char *test_kernels[2] = { sample_read_only_image_test_kernel,
+                                    sample_write_only_image_test_kernel };
+    constexpr cl_image_format formats = { CL_RGBA, CL_UNSIGNED_INT8 };
```
bashbaug: Nitpick: This is a single format, not multiple formats.

```suggestion
    constexpr cl_image_format format = { CL_RGBA, CL_UNSIGNED_INT8 };
```

test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +722,65 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+    images[0] = create_image_2d(context, CL_MEM_READ_ONLY, &formats, size_dim,
+                                size_dim, 0, nullptr, &error);
+    test_error(error, "create_image_2d failed");
+
+    images[1] = create_image_2d(context, CL_MEM_WRITE_ONLY, &formats, size_dim,
+                                size_dim, 0, nullptr, &error);
+    test_error(error, "create_image_2d failed");
+
+    error = clSetKernelArg(kernels[0], 1, sizeof(buffer), &buffer);
+    test_error(error, "clSetKernelArg failed");
```
bashbaug: Since we aren't running these kernels, strictly speaking we don't need to worry about any non-image kernel arguments.  This means that we don't need to create this buffer, either.

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp
shajder: Similarly, one small advantage is that we can pass `TEST_FAIL` implicitly.

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp
shajder: Done. All your remarks apply to the source test `negative_set_immutable_memory_to_writeable_kernel_arg`, and probably to more tests in this [source file](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/api/test_kernels.cpp). It would likely be best to perform such a refactoring all at once.

test_conformance/api/test_kernels.cpp
bashbaug: Yeah, I suspect that the `test_failure_error` macro should return `TEST_FAIL` when the comparison fails, rather than `errCode != expectedErrCode`, but that's going to be a much bigger change.  I'll just resolve this comment.

test_conformance/api/test_kernels.cpp
bashbaug: I see.  That's too bad, but at least the new code added by this PR is "good"!

bashbaug (approved):

bashbaug:

hvdijk:
(Accidentally reported on the wrong PR first, sorry.) When an implementation doesn't support images, this can fail due to `read_imageui` and `write_ui` not being available. I have opened #2510 for this.

PR #2448, opened 2025-07-21
Added test to verify negative result of clSetKernelArg with CL_INVALID_ARG_SIZE

shajder:
Related to #2282, according to work plan related to CL_INVALID_ARG_SIZE result


test_conformance/api/test_kernels.cpp
```diff
@@ -87,6 +88,130 @@ const char *sample_two_kernel_program[] = {
...
+    {
+        iterate_type(std::get<Out>(t));
+        for_each_elem<Out + 1, Tp...>(t);
+    }
+
+    template <typename TestType> void test_invalid_arg(ExplicitType type)
+    {
+        unsigned int sizes[] = { 1, 2, 4, 8, 16 };
+        std::vector<char> buf;
+        for (unsigned i = 0; i < ARRAY_SIZE(sizes); i++)
```
EwanC: I like that this class is using some pretty advanced C++, contrasted this a C array needing a macro to work out it's size 😁 Can we use a std::array for sizes instead with a range based for loop.

test_conformance/api/test_kernels.cpp
```diff
@@ -87,6 +88,130 @@ const char *sample_two_kernel_program[] = {
...
+            clProgramWrapper program;
+            clKernelWrapper kernel;
+
+            size_t destStride = get_explicit_type_size(type) * sizes[i];
+
+            std::ostringstream vecNameStr;
+            vecNameStr << get_explicit_type_name(type);
+            if (sizes[i] != 1) vecNameStr << sizes[i];
+
+            std::stringstream sstr;
```
EwanC: I like string streams, but this instance looks like it could just as easily be:
```c++
std::string pragma_str;
if (type == kDouble)
    pragma_str = "#pragma OPENCL EXTENSION cl_khr_fp64 : enable\n";
...
```

test_conformance/api/test_kernels.cpp
```diff
@@ -87,6 +88,130 @@ const char *sample_two_kernel_program[] = {
...
+            if (type == kDouble)
+                sstr << "#pragma OPENCL EXTENSION cl_khr_fp64 : enable\n";
+
+            if (type == kHalf)
+                sstr << "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n";
+
+            std::string program_source = str_sprintf(
+                std::string(sample_arg_size_test_kernel), sstr.str().c_str(),
+                vecNameStr.str().c_str(), vecNameStr.str().c_str());
```
EwanC: Would define a variable for `vecNameStr.str().c_str()` rather than duplicating the expression for each arg.

test_conformance/api/test_kernels.cpp
```diff
@@ -87,6 +88,130 @@ const char *sample_two_kernel_program[] = {
...
+
+            const char *ptr = program_source.c_str();
+            cl_int error = create_single_kernel_helper(
+                context, &program, &kernel, 1, &ptr, "arg_size_test");
+
+            if (error != CL_SUCCESS)
+                throw std::runtime_error("Unable to build test program\n");
+
+            buf.resize(destStride);
```
EwanC: Do we actually need to resize a buffer each iteration, can we just define a type of the max possible size to use for all the iterations, e.g. `cl_ulong16`, and use that as the `clSetKernelArg` pointer.

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp
shajder: hmm - not sure if I am reading you right. Wouldn't that be a hassle to define 11 x 5 names? Then, to pick it up in each iuteration we would need some kind of `std::map<ExplicitType, std::vector<std::string>>`, right ?

test_conformance/api/test_kernels.cpp
shajder: Corrected, thanks!

test_conformance/api/test_kernels.cpp
EwanC: I meaning more like define a variable on line 175 `auto vec_name_cstr = vecNameStr.str().c_str()` and use it for both the arguments here.

shajder: I get it, thanks - `vecNameStr.str()` might be a temporary object so let's stick to additional `std::string` variable. Corrected

EwanC (dismissed):

EwanC (approved):

bashbaug (approved):

bashbaug:
@outofcontrol the CLA bot seems stuck here - any ideas?

lakshmih:
Still reviewing this, could you wait on merging? Thanks.

test_conformance/api/test_kernels.cpp
```diff
@@ -109,6 +111,113 @@ const char *sample_write_only_image_test_kernel = R"(
...
+        else if (type == kHalf && !fp16_supported)
+            return true;
+        else if (strchr(get_explicit_type_name(type), ' ') != 0)
+            return true;
+        return false;
+    }
+
+    void test_invalid_arg(ExplicitType type)
+    {
+        printf("\nTesting explicit type %d\n", type);
```
lakshmih: Need to remove debug print..

test_conformance/api/test_kernels.cpp
```diff
@@ -109,6 +111,113 @@ const char *sample_write_only_image_test_kernel = R"(
...
+    %s
+    __kernel void arg_size_test(%s src, __global %s *dst)
+    {
+        dst[0]=src;
+    }
+)";
+
+namespace {
+
+struct ArgSizeTypesIterator
```
lakshmih: Do we still need this iterator type?

test_conformance/api/test_kernels.cpp
shajder: removed

test_conformance/api/test_kernels.cpp
shajder: Moved to one block. Side note: Removing the template approach in favor of `ExplicitType` iteration can be seen as a downgrade. The main justification for this change is to resolve compile-time confusion between `cl_ushort` and `cl_half`, which requires additional code and reduces readability.

bashbaug:
Merging as agreed in the September 16th teleconference.

PR #2450, opened 2025-07-22
Added test to verify negative result of clSetKernelArg with CL_INVALID_ARG_SIZE and memory object argument

shajder:


bashbaug (dismissed):

test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +711,42 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+    test_error(error,
+               "Unable to get mem_obj_size_test kernel for built program");
+
+    std::vector<cl_uchar> mem_data(256, 0);
+    clMemWrapper buffer = clCreateBuffer(
+        context, CL_MEM_USE_HOST_PTR, mem_data.size(), mem_data.data(), &error);
+    test_error(error, "clCreateBuffer failed");
+
+    // Run the test - CL_INVALID_ARG_SIZE
+    error = clSetKernelArg(mem_obj_arg_kernel, 0, sizeof(buffer) * 2, &buffer);
```
EwanC: There's a bit of a gap in my C++ knowledge about the interaction of the `sizeof` opeator and user defined conversion operator, do you know or have you checked `sizeof(buffer)` calls `operator T() const { return object; }` then doing `sizeof()` on the returned object? It might be simpler to replace `sizeof(buffer)` with `sizeof(cl_mem)` to remove the distraction of this possibility.

EwanC (dismissed):

test_conformance/api/test_kernels.cpp
shajder: Good find, thanks - corrected!

PR #2451, opened 2025-07-22
Rename Version major and minor getters

joselopeqti:
Both these functions cause a conflict when using an old version of the GNU C Library with the header file sys/sysmacros.h where major() and minor() are defined as a macro

### DIFF
--- a/test_common/harness/testHarness.h
+++ b/test_common/harness/testHarness.h
@@ -28,8 +28,8 @@ public:
     Version(): m_major(0), m_minor(0) {}
 
     Version(cl_uint major, cl_uint minor): m_major(major), m_minor(minor) {}
-    int version_major() const { return m_major; }
-    int version_minor() const { return m_minor; }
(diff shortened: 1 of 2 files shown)


test_common/harness/testHarness.h
```diff
@@ -28,8 +28,8 @@ class Version {
     Version(): m_major(0), m_minor(0) {}
 
     Version(cl_uint major, cl_uint minor): m_major(major), m_minor(minor) {}
-    int major() const { return m_major; }
-    int minor() const { return m_minor; }
+    int version_major() const { return m_major; }
+    int version_minor() const { return m_minor; }
```
EwanC: Nit, but would probably go for `get_major()`/`get_minor()` since `version` feels redundant as it's implied already on the class.

EwanC (dismissed):

bashbaug:
Merging as discussed in the August 26th teleconference.

PR #2454, opened 2025-07-24
Fix test_spir not checking for the required extension

ahesham-arm:
`clGetDeviceInfo` should fail with `CL_INVALID_VALUE` when queried for `CL_DEVICE_SPIR_VERSIONS` on devices that do not claim to support the extension that provides it, `cl_khr_spir`.

Following this change, the test is skipped instead of failing on devices that do not support `cl_khr_spir`.

### DIFF
--- a/test_conformance/spir/main.cpp
+++ b/test_conformance/spir/main.cpp
@@ -6911,12 +6911,13 @@ int main (int argc, const char* argv[])
         cl_device_id device = get_platform_device(device_type, choosen_device_index, choosen_platform_index);
         printDeviceHeader(device);
 
+        REQUIRE_EXTENSION("cl_khr_spir");
+
         std::vector<Version> versions;
         get_spir_version(device, versions);
 
-        if (!is_extension_available(device, "cl_khr_spir")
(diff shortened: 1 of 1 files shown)


test_conformance/spir/main.cpp
```diff
@@ -6911,6 +6911,8 @@ int main (int argc, const char* argv[])
         cl_device_id device = get_platform_device(device_type, choosen_device_index, choosen_platform_index);
         printDeviceHeader(device);
 
+        REQUIRE_EXTENSION("cl_khr_spir");
```
bashbaug, line 6914: It's not hurting anything, but there is another check for `cl_khr_spir` five lines below that is now redundant and can be removed.

bashbaug (dismissed):

test_conformance/spir/main.cpp
ahesham-arm: Thanks for spotting that, removed the other check.

bashbaug (approved):

bashbaug:
Straightforward change and obvious fix, merging immediately.

PR #2456, opened 2025-07-25
Added test to verify negative result of clSetKernelArg with CL_INVALID_MEM_OBJECT

shajder:

### DIFF
--- a/test_conformance/api/test_kernels.cpp
+++ b/test_conformance/api/test_kernels.cpp
@@ -87,6 +87,13 @@ const char *sample_two_kernel_program[] = {
 "\n"
 "}\n" };
 
+const char *sample_mem_obj_size_test_kernel = R"(
+    __kernel void mem_obj_size_test(__global int *src, __global int *dst)
+    {
+        size_t  tid = get_global_id(0);
+        dst[tid] = src[tid];
+    }
+)";
 
 REGISTER_TEST(get_kernel_info)
 {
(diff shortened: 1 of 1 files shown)


test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +711,36 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+    test_error(error,
+               "Unable to get mem_obj_size_test kernel for built program");
+
+    std::vector<cl_uchar> mem_data(256, 0);
+    clMemWrapper buffer = clCreateBuffer(
+        context, CL_MEM_USE_HOST_PTR, mem_data.size(), mem_data.data(), &error);
+    test_error(error, "clCreateBuffer failed");
+
+    // Run the test - CL_INVALID_MEM_OBJECT
+    error = clSetKernelArg(mem_obj_arg_kernel, 0, sizeof(buffer), nullptr);
```
bashbaug, line 737: I don't think this is an error:

> If the argument is declared to be a pointer of a built-in scalar or vector type, or a user defined structure type in the global or constant address space, the memory object specified as argument value must be a buffer object (or NULL).

shajder: yep, I've noticed it fails and I started to wonder what does that mean:

> specified arg_value is not a valid memory object

I've experimented with pointers and constants and the result was either crash or no error. I will appreciate for your suggestions, thanks.

bashbaug: Given the discussion here: https://github.com/KhronosGroup/OpenCL-Docs/issues/482#issuecomment-795685410

The only cases the CTS should be checking for are when the object is valid and when the object is `NULL`.

Since the case where the object is `NULL` is not an error, I'm not sure what else we can do to check for this `CL_INVALID_MEM_OBJECT` error case that is not implementation defined.

shajder: understood, I am closing this one!

PR #2457, opened 2025-07-28
Update '-list' option

rjodinchr:
'-list' option is used to print all sub-tests. But some test do not support it at all. And all test do not display it the same way, making it quite complicated for external tools to extract them.

That CL clean the usage so that tests:
- Print the sub-tests list with either '-list' (to prevent breaking legacy usage) or '--list' (to match other options)
- Do not print anything else when the option is used

### DIFF
--- a/test_common/harness/testHarness.cpp
+++ b/test_common/harness/testHarness.cpp
@@ -22,6 +22,7 @@
 #include <cassert>
 #include <deque>
 #include <mutex>
+#include <set>
 #include <stdexcept>
 #include <thread>
 #include <vector>
@@ -172,12 +173,11 @@ void version_expected_info(const char *test_name, const char *api_name,
 
 static void list_tests(int testNum, test_definition testList[])
 {
-    std::vector<std::string> names;
+    std::set<std::string> names;
     for (int i = 0; i < testNum; i++)
     {
-        names.emplace_back(testList[i].name);
+        names.insert(testList[i].name);
     }
-    std::sort(names.begin(), names.end());
     for (const auto &name : names)
     {
         log_info("\t%s\n", name.c_str());


test_common/harness/testHarness.cpp
```diff
@@ -169,6 +169,14 @@ void version_expected_info(const char *test_name, const char *api_name,
...
              test_name, api_name, expected_version, api_name, device_version);
 }
+
+static void list_tests(int testNum, test_definition testList[])
+{
+    for (int i = 0; i < testNum; i++)
+    {
+        log_info("\t%s\n", testList[i].name);
+    }
+}
```
ahesham-arm: You don't need to pass arguments to this function, the tests are already part of `test_registry` and are in a container. Additionally, I think it would be better to have the output of this function be an alphabetically sorted list to guarantee stability regardless of the order of adding the tests, and it's just easier to skim through a sorted list visually if you're operating the test binaries yourself.

Here is my suggestion:
```suggestion
static void list_tests()
{
    auto& test_list = test_registry::getInstance().definitions();
    std::vector<test_definition> sorted_list(test_list.size());
    std::partial_sort_copy(test_list.begin(), test_list.end(), sorted_list.begin(), sorted_list.end());
    
    for (const auto& test : sorted_list)
    {
        log_info("\t%s\n", test.name);
    }
}

```

rjodinchr: I agree with the sorting part.

> You don't need to pass arguments to this function, the tests are already part of test_registry and are in a container.

I am not sure I agree. `runTestHarnessWithCheck` is called with `testNum` and `testList` as parameters. Maybe it should not and just get those from the container, but as long as it is the case, I think it's better to rely on those for `list_tests` as it is inside `runTestHarnessWithCheck`.

An example might be that something could be filtering the test earlier in the flow. That would ask the question whether this print is at the right place (but that's what `--help` would also print), but I do believe we don't want to ignore what has been passed to `runTestHarnessWithCheck` and just use that.

ahesham-arm: sure, we can unify how parameters are passed to the harness in later work

test_common/harness/testHarness.cpp
```diff
@@ -169,6 +169,20 @@ void version_expected_info(const char *test_name, const char *api_name,
              "reports %s version %s)\n",
              test_name, api_name, expected_version, api_name, device_version);
 }
+
+static void list_tests(int testNum, test_definition testList[])
+{
+    std::vector<std::string> names;
```
ahesham-arm: If the point is sorting, you can use a sorted container for this, e.g. `std::set<std::string>`

test_common/harness/testHarness.cpp
```diff
@@ -169,6 +169,20 @@ void version_expected_info(const char *test_name, const char *api_name,
...
+    for (int i = 0; i < testNum; i++)
+    {
+        names.emplace_back(testList[i].name);
+    }
+    std::sort(names.begin(), names.end());
+    for (auto name : names)
```
ahesham-arm: This creates a copy, I would use `for (const auto& name : names)` instead.

test_conformance/spir/main.cpp
```diff
@@ -6771,10 +6779,12 @@ static int ParseCommandLine (int argc, const char *argv[],
...
         log_info( "\tno-unzip\t\tDo not extract test files from Zip; use existing.\n" );
 
-        for( unsigned int i = 0; i < (sizeof(spir_suites) / sizeof(sub_suite)); i++ )
-        {
-            log_info( "\t\t%s\n", spir_suites[i].name );
-        }
+        ListTests();
+        return 0;
+    }
+    else if (!strcmp(argv[1], "--list") || !strcmp(argv[1], "-list"))
```
hvdijk, line 6785: This doesn't check `argc > 1`, so when `test_spir` is called without arguments, this calls `strcmp(NULL, "--list")` and (typically) segfaults.

rjodinchr: Are you making a fix for it, or should I take care of it?

hvdijk: I can submit a PR later today, or if you're able to submit one sooner than that, even better.

rjodinchr: Please make a PR. I will review it and will make sure it gets merge in during tomorrow's OpenCL weekly.

hvdijk: Was able to get to it quicker than I thought, opened #2509.

PR #2458, opened 2025-07-29
Added test to verify negative result of clSetKernelArg with CL_INVALID_ARG_INDEX

shajder:


test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +704,28 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+
+    // Setup the test
+    error = create_single_kernel_helper(context, &program, nullptr, 1,
+                                        sample_single_test_kernel, nullptr);
+    test_error(error, "Unable to build test program");
+
+    sampler_arg_kernel = clCreateKernel(program, "sample_test", &error);
+    test_error(error, "Unable to get sample_test kernel for built program");
+
+    // Run the test - CL_INVALID_ARG_INDEX
```
EwanC: This comment doesn't give the reader much extra info as-is, it would add value if instead the comment said that the "sample_test" test kernel has 2 arguments, so `2` is out of range as arguments are indexed from zero.

test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +704,28 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+    error = clSetKernelArg(sampler_arg_kernel, 2, sizeof(cl_sampler), nullptr);
```
EwanC: `cl_sampler` types aren't used anywhere else in this file, can we use `sizeof(cl_mem)` to avoid a reader having to work if there's a reason we've started using `cl_sampler` for this particular test.

test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +704,28 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
 
     return TEST_PASS;
 }
+
+REGISTER_TEST(negative_invalid_arg_index)
+{
+    cl_int error = CL_SUCCESS;
+    clProgramWrapper program;
+    clKernelWrapper sampler_arg_kernel;
```
EwanC: The kernel names at the top of the file are "sample" rather than "sampler", a "sampler" is a separate concept to what we're trying to test here, so would prefer we didn't complicate the test by using that terminology. Would just call this variable `kernel`.

EwanC (approved):
Some nitpicks but LGTM

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp
shajder: Done :)

bashbaug (approved):
Merging as discussed in the August 12th teleconference.

PR #2460, opened 2025-07-30
Added test to verify negative result of clSetKernelArg with CL_INVALID_KERNEL

shajder:


test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +704,22 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
 }
+
+REGISTER_TEST(negative_invalid_kernel)
+{
+    cl_int error = CL_SUCCESS;
+    clKernelWrapper sampler_arg_kernel;
+
+    clMemWrapper mem = clCreateBuffer(context, CL_MEM_READ_ONLY,
+                                      sizeof(cl_float), NULL, &error);
+
```
EwanC, line 737: Could stick a `test_error(error, "clCreateBuffer failed");`  check in here before doing the `clSetKernelArg`

test_conformance/api/test_kernels.cpp
```diff
@@ -704,3 +704,22 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
 
     return TEST_PASS;
 }
+
+REGISTER_TEST(negative_invalid_kernel)
+{
+    cl_int error = CL_SUCCESS;
+    clKernelWrapper sampler_arg_kernel;
```
EwanC: we're not initializing this with anything to do with samplers, so would go with a simpler variable name like `kernel`

test_conformance/api/test_kernels.cpp
shajder: thanks

test_conformance/api/test_kernels.cpp
shajder: done

EwanC (approved):

EwanC (dismissed):

PR #2461, opened 2025-07-30
Fix warnings treated as errors in Vulkan tests.

ahesham-arm:
Unused variables and mismatched sign comparisons.


kpet (approved):
Trivial warning fixing. Merging.

PR #2462, opened 2025-07-30
Added test to verify negative result of clSetKernelArg with CL_INVALID_DEVICE_QUEUE

shajder:


test_conformance/api/test_kernels.cpp
```diff
@@ -718,6 +730,44 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
 
+REGISTER_TEST(negative_invalid_arg_queue)
+{
+    OpenCLCFeatures features;
+    get_device_cl_c_features(device, features);
+
+    const Version clc_version = get_device_latest_cl_c_version(device);
+    if ((clc_version < Version(2, 0))
+        || !features.supports__opencl_c_device_enqueue)
+        return TEST_SKIPPED_ITSELF;
```
bashbaug, line 778: I don't think these checks are quite right.  Consider the case where the device is an OpenCL 2.0 device, so the latest OpenCL C version is 2.0, but no features are supported, because features were not added until OpenCL 3.0.  In this case, I think the test will be skipped, though it should run.

I think the checks should be:

1. If the OpenCL C version is less than 2.0, then skip the test.
2. If the OpenCL C version is exactly 2.0, then run the test.
    * Depending how we set things up, we probably don't need to check for this specifically.
4. The OpenCL C version is 3.0 or newer, then run the test if the device enqueue feature is supported, otherwise skip the test.

test_conformance/api/test_kernels.cpp
```diff
@@ -718,6 +730,44 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+                                        &sample_queue_test_kernel, "queue_test",
+                                        build_opts.c_str());
+    test_error(error, "Unable to build test program");
+
+    queue_arg_kernel = clCreateKernel(program, "queue_test", &error);
+    test_error(error, "Unable to get queue_test kernel for built program");
+
+    // Run the test - CL_INVALID_DEVICE_QUEUE
+    error = clSetKernelArg(queue_arg_kernel, 0, sizeof(cl_command_queue),
+                           queue_arg);
```
bashbaug: This is a reasonable test and we shouldn't remove it, but it's not _that_ interesting because it's passing `nullptr` as the command queue (due to the behavior of the default constructor for clCommandQueueWrapper).  Aside: If we really wanted to, we could pass `nullptr` explicitly instead, because the wrapper really isn't buying us anything here.

It might be interesting to add a test that passes the host command queue from the harness as well, which should also be identified as an invalid device queue.

test_conformance/api/test_kernels.cpp
```diff
@@ -754,6 +766,54 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+    clCommandQueueWrapper queue_arg;
+
+    // Setup the test
+    error = create_single_kernel_helper(context, &program, nullptr, 1,
+                                        &sample_queue_test_kernel, "queue_test",
+                                        build_opts.c_str());
+    test_error(error, "Unable to build test program");
+
+    queue_arg_kernel = clCreateKernel(program, "queue_test", &error);
+    test_error(error, "Unable to get queue_test kernel for built program");
```
bashbaug: This doesn't work - we either need the kernel name to be `nullptr`, or the kernel handle must not be `nullptr`.

Since I can't think of a reason not to do so, I'd recommend just using the harness to create the kernel also, and pass a non-`nullptr` kernel handle.

```suggestion
    // Setup the test
    error = create_single_kernel_helper(context, &program, &queue_arg_kernel, 1,
                                        &sample_queue_test_kernel, "queue_test",
                                        build_opts.c_str());
    test_error(error, "Unable to create test kernel");
```

rjodinchr (approved):

lakshmih (approved):

test_conformance/api/test_kernels.cpp
```diff
@@ -754,6 +766,51 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+    OpenCLCFeatures features;
+    get_device_cl_c_features(device, features);
+
+    const Version clc_version = get_device_latest_cl_c_version(device);
+    if (clc_version < Version(2, 0)
+        || (clc_version >= Version(3, 0)
+            && !features.supports__opencl_c_device_enqueue))
+        return TEST_SKIPPED_ITSELF;
+
+    std::string build_opts = "-cl-std=CL" + clc_version.to_string();
```
bashbaug, line 780: This is fine, but note that the harness will already compile with the highest supported OpenCL C version unless another OpenCL C version is explicitly requested via passed-in build options.  This means that we don't need to setup these `build_opts` explicitly or pass any build options to the harness.

test_conformance/api/test_kernels.cpp
```diff
@@ -754,6 +766,51 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+
+    // Setup the test
+    error = create_single_kernel_helper(context, &program, &queue_arg_kernel, 1,
+                                        &sample_queue_test_kernel, "queue_test",
+                                        build_opts.c_str());
+    test_error(error, "Unable to create test kernel");
+
+    // Run the test - CL_INVALID_DEVICE_QUEUE
+    error = clSetKernelArg(queue_arg_kernel, 0, sizeof(cl_command_queue),
+                           &queue_arg);
```
bashbaug, line 795: Nitpick: if we just passed in `nullptr` directly here, we wouldn't need the `queue_arg` wrapper.


```suggestion
    error = clSetKernelArg(queue_arg_kernel, 0, sizeof(cl_command_queue),
                           nullptr);
```

bashbaug (approved):

PR #2470, opened 2025-08-01
Added test to verify negative result of clSetKernelArg with CL_INVALID_MEM_OBJECT and image object argument

shajder:

[run-test: test_api negative_set_kernel_arg_invalid_image]


test_conformance/api/test_kernels.cpp
```diff
@@ -754,6 +765,55 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+                        el.first.c_str(), el.second.c_str(), el.second.c_str());
+
+        const char *ptr = program_source.c_str();
+        cl_int error = create_single_kernel_helper(context, &program, &kernel,
+                                                   1, &ptr, "arg_image_test");
+        test_error(error, "Unable to build test program");
+
+        kernel = clCreateKernel(program, "arg_image_test", &error);
+        test_error(error,
+                   "Unable to get arg_image_test kernel for built program");
```
rjodinchr: `create_single_kernel_helper` creates the kernel, no need for `clCreateKernel` I believe.

rjodinchr (changes requested):
Can you rebase on ToT, and add a CI run?

test_conformance/api/test_kernels.cpp
shajder: Duplication removed.

shajder:
> Can you rebase on ToT, and add a CI run?

CI test with POCL added, it fails with `CL_SUCCESS` which looks like a bug in POCL implementation. I went through my desktop implementations:

-Intel both GPU and CPU implementation fails with `CL_INVALID_ARG_VALUE`. As per spec:
> [CL_INVALID_ARG_VALUE](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_INVALID_ARG_VALUE)
> -if the argument is an image declared with the read_only qualifier and arg_value points to an image object created with the memory flag [CL_MEM_WRITE_ONLY](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_MEM_WRITE_ONLY)
> -if the argument is an image declared with the write_only qualifier and arg_value points to an image object created with the memory flag [CL_MEM_READ_ONLY](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_MEM_READ_ONLY) or [CL_MEM_IMMUTABLE_EXT](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_MEM_IMMUTABLE_EXT)

I can't see any possibility to meet above conditions.

-Nvidia passes with expected `CL_INVALID_MEM_OBJECT`

rjodinchr:
That's fine, it is also good to know that the test is catching an implementation bug, it is the goal. Just update `ci/pocl/golden.json` with `fail` instead of `pass` for that test.

shajder:
> That's fine, it is also good to know that the test is catching an implementation bug, it is the goal. Just update `ci/pocl/golden.json` with `fail` instead of `pass` for that test.

Just thinking at loud, does marking it "fail" actually prevent this from reaching anyone on the POCL side?

rjodinchr:
> Just thinking at loud, does marking it "fail" actually prevent this from reaching anyone on the POCL side?

No, it does not notify anyone in the pocl community. Feel free to open a bug on their side. Otherwise, they will notice it once that patch gets merged and they update the CTS.

Note that we are using pocl 7.1, so even if they fix it upstream, the status will not change until we uprev the version of pocl used here.

rjodinchr (approved):

PR #2472, opened 2025-08-05
Added test to verify negative result of clSetKernelArg with CL_INVALID_MEM_OBJECT and MSAA image object argument

shajder:

[run-test: test_api negative_set_kernel_arg_invalid_image_msaa]


test_conformance/api/test_kernels.cpp
```diff
@@ -718,6 +729,62 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+REGISTER_TEST(negative_set_kernel_arg_invalid_image_msaa)
+{
+    PASSIVE_REQUIRE_IMAGE_SUPPORT(device)
+
+    if (!is_extension_available(device, "cl_khr_gl_msaa_sharing"))
+    {
+        log_info("Test not run because 'cl_khr_gl_msaa_sharing' extension is "
+                 "not supported by the tested device\n");
+        return TEST_SKIPPED_ITSELF;
+    }
```
bashbaug:
```suggestion
    REQUIRE_EXTENSION("cl_khr_gl_msaa_sharing");
```

test_conformance/api/test_kernels.cpp
```diff
@@ -718,6 +729,62 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+
+    cl_int error = CL_SUCCESS;
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+
+    std::vector<std::pair<std::string, std::string>> image_types = {
+        { "image2d_msaa_t", "2" },
+        { "image2d_array_msaa_t", "4" },
+        { "image2d_msaa_depth_t", "4" },
+        { "image2d_array_msaa_depth_t", "4" }
```
bashbaug, line 804: I'm not convinced that all of these cases should return an error.

I think `image2d_msaa_t` should work.  Why isn't the 2D image created below trivially an MSAA image, just one with a single sample?

It's _probably_ reasonable to expect that `image2d_array_msaa_t` and `image2d_array_msaa_depth_t` will fail, since the image that is created is a 2D image and not a 2D image array.  Do we have similar testing for non-MSAA images?

I think that `image2d_msaa_depth_t` is an especially grey area, since the image is a 2D image, just not one with a `CL_DEPTH` image channel order.  Do we expect implementations to check the channel order?

Note that the error description in `clSetKernelArg` was clarified recently in https://github.com/KhronosGroup/OpenCL-Docs/pull/1493.  The text in the currently published specs is:

> If the argument is a multi-sample 2D depth image, the arg_value entry must be a pointer to a multisample depth image object.

This is somewhat ambiguous though - what is considered to be "a multisample depth image object"?  Also, note that the spec does not require the image to be a 2D image, which seems like an oversight.

The updated text in the GitHub spec sources but not yet in a published specification is:

> If the kernel argument is a 2D MSAA depth image, then the image memory object must be of image type [CL_MEM_OBJECT_IMAGE2D](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#CL_MEM_OBJECT_IMAGE2D) and image channel order [CL_DEPTH](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#CL_DEPTH).

This is more precise, but maybe we got the `CL_DEPTH` image channel order part wrong.

shajder: > I think image2d_msaa_t should work. Why isn't the 2D image created below trivially an MSAA image, just one with a single sample?

While mathematically a single-sample image is a subset of MSAA, I believe `image2d_msaa_t` and `image2d_t` are treated as mutually exclusive types by the API. Specifically, we cannot pass an `image2d_msaa_t` to a `read_imageX` function that uses a sampler, nor can we use an `image2d_t` with the sampler-less read functions. It’s a similar situation to `OpenGL`, where you cannot use a `sampler2D` for a texture created with `glTexImage2DMultisample` (and vice-versa with `sampler2DMS`).

> If the kernel argument is a 2D MSAA depth image, then the image memory object must be of image type [CL_MEM_OBJECT_IMAGE2D](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#CL_MEM_OBJECT_IMAGE2D) and image channel order [CL_DEPTH](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#CL_DEPTH).

My interpretation relies on the "must" keyword here. It implies that if the kernel expects `image2d_msaa_depth_t`, the implementation is required to verify that the passed object actually has the `CL_DEPTH` channel order. If a user passes a non-depth image, that fails to meet the `CL_DEPTH` channel order requirement, which is why we expect `clSetKernelArg` to return an error. Does that make sense?

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp
```diff
@@ -754,6 +765,57 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+    REQUIRE_EXTENSION("cl_khr_gl_msaa_sharing");
+
+    cl_int error = CL_SUCCESS;
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+
+    std::vector<std::pair<std::string, std::string>> image_types = {
+        { "image2d_msaa_t", "2" },
+        { "image2d_array_msaa_t", "4" },
+        { "image2d_msaa_depth_t", "4" },
```
rjodinchr:
```suggestion
        { "image2d_msaa_depth_t", "2" },
```
```c
float4 read_imagef(image2d_msaa_t image, int2 coord, int sample)
```
https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#built-in-image-sampler-less-read-functions

test_conformance/api/test_kernels.cpp
```diff
@@ -754,6 +765,57 @@ REGISTER_TEST(negative_set_immutable_memory_to_writeable_kernel_arg)
...
+        { "image2d_array_msaa_depth_t", "4" }
+    };
+
+    constexpr cl_image_format format = { CL_RGBA, CL_UNSIGNED_INT8 };
+    constexpr size_t size_dim = 32;
+    clMemWrapper image =
+        create_image_2d(context, CL_MEM_READ_ONLY, &format, size_dim, size_dim,
+                        0, nullptr, &error);
+    test_error(error, "clCreateBuffer failed");
```
rjodinchr:
```suggestion
    test_error(error, "create_image_2d failed");
```

test_conformance/api/test_kernels.cpp
```diff
@@ -89,6 +91,15 @@ const char *sample_two_kernel_program[] = {
 "\n"
 "}\n" };
 
+const char *sample_arg_image_msaa_test_kernel = R"(
+    #pragma OPENCL EXTENSION cl_khr_gl_msaa_sharing : enable
+    __kernel void arg_image_test(%s src, __global float4 *dst)
+    {
+        int%s coord = (int%s)(get_global_id(0));
+        dst[0]=read_imagef(src, coord, 0);
```
rjodinchr, line 99:
```
float read_imagef(image2d_msaa_depth_t image, int2 coord, int sample)
float read_imagef(image2d_array_msaa_depth_t image, int4 coord, int sample)
```
https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#built-in-image-sampler-less-read-functions

Assigning `float` (from `read_imagef`) to `float4` (`dst[0]`) feels wrong.

rjodinchr (changes requested):
Could you rebase on ToT?
Can you also add a CI run please?

test_conformance/api/test_kernels.cpp
shajder: Implicit conversions are allowed, float to float4 shouldn't be a problem - the other way around is not allowed, as per spec: https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#implicit-conversions

test_conformance/api/test_kernels.cpp

test_conformance/api/test_kernels.cpp

shajder:
> Could you rebase on ToT? Can you also add a CI run please?

Done. POCL skips, as for my local implementations, only Intel GPU supports `cl_khr_gl_msaa_sharing` but fails with `CL_SUCCESS` so I can't fully test this PR.

rjodinchr (dismissed):

lakshmih (dismissed):

rjodinchr (approved):

rjodinchr:
Merging as discussed during the July 14 teleconference.

PR #2474, opened 2025-08-07
Fix compiler warnings in test_vulkan

ahesham-arm:
Use the correct print format specifiers for `uint64_t` and fix loop index type to address sign warning.


kpet (approved):
Trivial change. Merging.

PR #2476, opened 2025-08-07
Update the GitHub workflow for Android

ahesham-arm:
Use the latest stable LTS version of the NDK (r28c) and add `ANDROID_PLATFORM` to the CMake configuration options, set to 29, so that Android Hardware Buffer tests can be compiled.


.github/workflows/presubmit.yml
```diff
@@ -107,7 +109,8 @@ jobs:
         if: ${{ matrix.arch == 'android-arm' || matrix.arch == 'android-aarch64' }}
         shell: bash
         run: |
-          echo "CMAKE_CONFIG_ARGS_ANDROID=-DCMAKE_ANDROID_ARCH_ABI=${ANDROID_ARCH_ABI}" >> $GITHUB_ENV
+          export CMAKE_CONFIG_ARGS_ANDROID="-DCMAKE_ANDROID_ARCH_ABI=${ANDROID_ARCH_ABI} -DANDROID_PLATFORM=${ANDROID_PLATFORM}"
```
kpet: Do you need the export?

ahesham-arm: No, removed.

Issue #2473, opened 2025-08-06
Test updates for command-buffer simultaneous-use rewok


EwanC:
CTS changes require for cl_khr_command_buffer PR https://github.com/KhronosGroup/OpenCL-Docs/pull/1411:
* Remove test for pending state query
* Either rework existing simulataneous use tests so that it tests the new definition, or delete them and create new more suitable tests without tech-debt from old definition. Without update it might be hard to write a test that avoids data races from the possibility of concurrent executions.
* Add tests for pipelined submission of a command-buffer not created with simultaneous-use. Ideally stressing indriect dependencies as well as direct ones:
  * in-order queue to express depdencies
  * out-of-order queue with event dependencies to express depenencies 
  * barrier to express for dependencies
* Add cl_khr_command_buffer_mutable_dispatch tests for updating and enqueueing pipelined submissions of a non-simultaneous use command-buffer with depdencies between each enqueue, and only do a blocking wait at the end and verify final result.
* Add cl_khr_command_buffer_mutable_dispatch tests for a simultaneous-use command-buffer, where two invocations are scheduled such that they can run concurrently, but the second invocation is updated such that it uses different inputs/outputs to avoid race conditions in the kernel.

EwanC:
Closing this as it was addressed by merged PR https://github.com/KhronosGroup/OpenCL-CTS/pull/2477

PR #2477, opened 2025-08-08
CTS updates for reworked simultaneous use

EwanC:
Actions test plan from https://github.com/KhronosGroup/OpenCL-CTS/issues/2473 to update CTS tests to reflect changes from cl_khr_command_buffer PR https://github.com/KhronosGroup/OpenCL-Docs/pull/1411

* Adds new test in`command_buffer_pipelined_enqueue.cpp` for multiple enqueues without blocking in-between, but serialized execution.
* Removed test for `CL_COMMAND_BUFFER_STATE_PENDING_KHR` state query.
* Remove negative test for `clEnqueueCommandBuffer` pending state error.
* Simplify `cl_khr_command_buffer` tests that stress simultaneous-use by testing multiple serialized enqueues of the same command-buffer, which doesn't now require the device imultaneous-use capability
* Remove simultaneous-use command-buffer creation in base class to off, and require tests do it themselves if they require it.
* Rewrite mutable dispatch simultaneous test to test updating both pipelined enqueues, and updating the new definition of simultaneous-use

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_pipelined_enqueue.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_pipelined_enqueue.cpp
@@ -280,6 +280,9 @@ struct BarrierPipelined : public CommandBufferPipelined
             0, nullptr, out_of_order_command_buffer, 0, nullptr, nullptr);
         test_error(error, "clEnqueueCommandBufferKHR failed");
 
+        error = clEnqueueBarrier(out_of_order_queue);
+        test_error(error, "clEnqueueBarrier failed");
+
         std::vector<cl_int> output_data(num_elements);
         error = clEnqueueReadBuffer(out_of_order_queue, out_mem, CL_TRUE, 0,
                                     data_size(), output_data.data(), 0, nullptr,


test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
```diff
@@ -27,10 +27,7 @@ BasicCommandBufferTest::BasicCommandBufferTest(cl_device_id device,
     : CommandBufferTestBase(device), context(context), queue(nullptr),
       num_elements(0), simultaneous_use_support(false),
       out_of_order_support(false), queue_out_of_order_support(false),
-      // try to use simultaneous path by default
-      simultaneous_use_requested(true),
-      // due to simultaneous cases extend buffer size
-      buffer_size_multiplier(1), command_buffer(this)
+      buffer_size_multiplier(1) /* extends buffer */, command_buffer(this)
```
bashbaug: Minor: I don't understand what "extends buffer" is referring to here.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_pipelined_enqueue.cpp
EwanC, line 1: This file is worth paying attention to in review, it adds new tests that cover cases that could have been considered simultaneous-use in the previous definition, but now are not.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_simultaneous.cpp
EwanC, line 1: This file is worth paying attention to in review, it tests updating a command-buffer which may still be executing as there are no blocking waits between enqueues.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_pipelined_enqueue.cpp
```diff
@@ -0,0 +1,307 @@
...
+    InOrderPipelined(cl_device_id device, cl_context context,
+                     cl_command_queue queue)
+        : CommandBufferPipelined(device, context, queue)
+    {}
+
+    cl_int Run() override
+    {
+        cl_int error = RecordCommandBuffer(command_buffer);
+        test_error(error, "RecordCommandBuffer failed");
+
```
AhmedAmraniAkdi, line 134: Hello, I believe the buffer needs to be initialised to 0, otherwise it will use whatever data it had and the test will fail because it is expecting 43.

Same for pipeline_barrier_deps and pipeline_event_deps.

EwanC: Thanks for catching that, I've pushed a commit that zero initialized the buffer in some setup code shared by all the pipeline tests.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_profiling.cpp
```diff
@@ -356,19 +279,13 @@ struct CommandBufferSubstituteQueueProfiling : public BasicCommandBufferTest
...
+REGISTER_TEST(profiling)
 {
-    return MakeAndRunTest<CommandBufferProfiling<true>>(device, context, queue,
-                                                        num_elements);
+    return MakeAndRunTest<CommandBufferProfiling>(device, context, queue,
+                                                  num_elements);
 }
 
-REGISTER_TEST(substitute_queue_profiling)
+REGISTER_TEST(profiling_substitue_queue)
```
bashbaug:
```suggestion
REGISTER_TEST(profiling_substitute_queue)
```

bashbaug (dismissed):
I've run these tests with the command buffer emulation layer and I'm not seeing any unexpected failures, although it would be most interesting for somebody to run these tests on an implementation that does not support simultaneous use.

test_conformance/extensions/cl_khr_command_buffer/basic_command_buffer.cpp
EwanC: I've pushed a commit that tries to document this member properly in the class definition, rather than doing a abbreviated description here.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_pipelined_enqueue.cpp
```diff
@@ -0,0 +1,318 @@
...
+        error = clEnqueueBarrier(out_of_order_queue);
+        test_error(error, "clEnqueueBarrier failed");
+
+        error = clEnqueueCommandBufferKHR(
+            0, nullptr, out_of_order_command_buffer, 0, nullptr, nullptr);
+        test_error(error, "clEnqueueCommandBufferKHR failed");
+
+        std::vector<cl_int> output_data(num_elements);
+        error = clEnqueueReadBuffer(out_of_order_queue, out_mem, CL_TRUE, 0,
```
AhmedAmraniAkdi, line 287: Hello Ewan, sorry for not catching this earlier, but I believe there needs to be a barrier before the clEnqueueReadBuffer, as there is no guarantee it will wait until all work is done (out of order queue).

EwanC: Nice spot, I've pushed a commit adding a barrier in for that missing synchronization.

PR #2478, opened 2025-08-11
Extended support for random-coord region operations with clEnqueueMapImage

### LINKED ISSUE #2348: CTS lacks coverage for using non-zero origins for clEnqueueMapImage on image types other than 2D
test_enqueue_map_image in basic does this for 2D images but there is no testing for other image types.

shajder:
Fixes #2348 according to issue description

[run-test: test_basic enqueue_map_image]


test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,151 +218,290 @@ REGISTER_TEST(enqueue_map_buffer)
...
+void random_region_coords(size_t *offset, size_t *region,
+                          const cl_mem_object_type image_type,
+                          const size_t &imageSize, const MTdataHolder &d)
+{
+    offset[0] = (size_t)random_in_range(0, (int)imageSize - 1, d);
+    region[0] = (size_t)random_in_range(1, (int)(imageSize - offset[0] - 1), d);
+    offset[1] = (size_t)random_in_range(0, (int)imageSize - 1, d);
+    region[1] = (size_t)random_in_range(1, (int)(imageSize - offset[1] - 1), d);
+    offset[2] = (size_t)random_in_range(0, (int)imageSize - 1, d);
+    region[2] = (size_t)random_in_range(1, (int)(imageSize - offset[2] - 1), d);
```
rjodinchr: `offset[x]` can be equal to `imageSize - 1`? If so , it would mean calling `random_in_range(1, 0, d)`, which does not feel right.


```suggestion
    offset[0] = (size_t)random_in_range(0, (int)imageSize - 1, d);
    region[0] = (size_t)random_in_range(1, (int)(imageSize - offset[0]), d);
    offset[1] = (size_t)random_in_range(0, (int)imageSize - 1, d);
    region[1] = (size_t)random_in_range(1, (int)(imageSize - offset[1]), d);
    offset[2] = (size_t)random_in_range(0, (int)imageSize - 1, d);
    region[2] = (size_t)random_in_range(1, (int)(imageSize - offset[2]), d);
```

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,151 +218,290 @@ REGISTER_TEST(enqueue_map_buffer)
...
+                                size_t x_off = offset[0] + x;
+                                size_t y_off = offset[1] + y;
+                                size_t z_off = offset[2] + z;
+                                size_t ref_loc =
+                                    (z_off * image_size * image_size
+                                     + y_off * image_size + x_off);
+                                cl_uchar *pixel =
+                                    &mappedRegion
+                                        [z * slicePitch + y * rowPitch
+                                         + x * pixel_size]; // is this correct ?
```
rjodinchr: We should figure out if it is correct or not, and either fix it or remove the comment

test_conformance/basic/test_enqueue_map.cpp
shajder: Corrected as requested, FYI [the code](https://github.com/KhronosGroup/OpenCL-CTS/blob/e2580bded220c6ac9786a1008b479aa96d32d359/test_conformance/basic/test_enqueue_map.cpp#L233-L238) was operational in the test before my changes:

```
            offset[0] = (size_t)random_in_range(0, (int)imageSize - 1, d);
            region[0] =
                (size_t)random_in_range(1, (int)(imageSize - offset[0] - 1), d);
            offset[1] = (size_t)random_in_range(0, (int)imageSize - 1, d);
            region[1] =
                (size_t)random_in_range(1, (int)(imageSize - offset[1] - 1), d);
```

test_conformance/basic/test_enqueue_map.cpp
shajder: Thanks, I put it ~year ago and can't recall what was the reason. I just debug it with intel implementation, looks correct, test passes.

rjodinchr (commented):
I would appreciate it if you could create some helper functions to make the code easier to read and maintain. The current implementation has too many indentation levels in a single function.
Also, could you enable a run in GitHub CI: https://github.com/KhronosGroup/OpenCL-CTS#running-targeted-ci-tests-on-pull-requests ?
Thank you

rjodinchr:
You need to add `test_basic` in `ci/pocl/golden.json` with the expected result.

shajder:
> You need to add `test_basic` in `ci/pocl/golden.json` with the expected result.

Thanks, nice feature!

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,78 +218,234 @@ REGISTER_TEST(enqueue_map_buffer)
...
+            region[1] = region[2] = 1;
+            break;
+        }
+        case CL_MEM_OBJECT_IMAGE2D_ARRAY:
+        case CL_MEM_OBJECT_IMAGE2D: {
+            offset[2] = 0;
+            region[2] = 1;
+            break;
+        }
+    }
```
rjodinchr: I'm not sure about that, let me know if I'm mistaken.
`CL_MEM_OBJECT_IMAGE1D_ARRAY` is a collection of 1D images, right? Shouldn't we test the second dimension then?
The same goes for `CL_MEM_OBJECT_IMAGE2D_ARRAY`.

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,78 +218,234 @@ REGISTER_TEST(enqueue_map_buffer)
...
+            img_region[0] = image_size;
+            break;
+        }
+        case CL_MEM_OBJECT_IMAGE2D: {
+            size_t image_size = std::min(
+                size_t(256),
+                (size_t)std::sqrt(max_image_data_size / max_bytes_per_pixel));
+            memObject = create_image_2d(context, mem_flag, &fmt, image_size,
+                                        image_size, 0, (void *)data, &error);
+            img_region[0] = img_region[1] = image_size;
```
marekkozl, line 246: Can we have consistency in logging behavior after image creation fails i.e. use of `test_error`

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,78 +218,234 @@ REGISTER_TEST(enqueue_map_buffer)
...
+                {
+                    memcpy(referenceData.data(), hostPtrData.data(),
+                           max_image_data_size);
+
+                    const size_t pixel_size = get_pixel_size(&fmt);
+
+                    size_t img_region[3] = { 1, 1, 1 };
+                    error = create_image_type(memObject, context, image_type,
+                                              mem_flag, fmt, img_region,
+                                              hostPtrData.data());
```
marekkozl: You are iterating over `flag_set` with `cl_mem_flags mem_flag = flag_set[src_flag_id];`
Use of `hostPtrData.data()` will always produce you a pointer so it implies `hasHostPtr` flag set be always there.

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,78 +218,234 @@ REGISTER_TEST(enqueue_map_buffer)
...
+        case CL_MEM_OBJECT_IMAGE2D_ARRAY: {
+            size_t image_size = std::min(
+                size_t(256),
+                (size_t)std::cbrt(max_image_data_size / max_bytes_per_pixel));
+            memObject = create_image_2d_array(
+                context, mem_flag, &fmt, image_size, image_size, image_size, 0,
+                0, (void *)data, &error);
+            img_region[0] = img_region[1] = img_region[2] = image_size;
+            break;
+        }
```
marekkozl, line 278: add default handling - if there is no supported enum - SUCCESS will be returned

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,78 +218,234 @@ REGISTER_TEST(enqueue_map_buffer)
...
+        return TEST_PASS;
+    }
 
+    cl_int run_tests_for_image_type(const size_t *img_region,
+                                    const size_t pixel_size,
+                                    const cl_mem_object_type &image_type,
+                                    const bool is_immutable_image)
+    {
+        cl_int error = CL_SUCCESS;
+        size_t image_size = img_region[0];
```
marekkozl: nitpick: a future change that gives width/height/depth independent sizes will require more refactoring

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,78 +218,234 @@ REGISTER_TEST(enqueue_map_buffer)
...
+        for (int i = 0; i < num_test_per_image_type; i++)
```
marekkozl: signed/unsigned comparison

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,78 +218,234 @@ REGISTER_TEST(enqueue_map_buffer)
...
+void random_region_coords(size_t *offset, size_t *region,
+                          const cl_mem_object_type image_type,
+                          const size_t &imageSize, const MTdataHolder &d)
+{
+    offset[0] = (size_t)random_in_range(0, (int)imageSize - 1, d);
+    region[0] = (size_t)random_in_range(1, (int)(imageSize - offset[0]), d);
+    offset[1] = (size_t)random_in_range(0, (int)imageSize - 1, d);
+    region[1] = (size_t)random_in_range(1, (int)(imageSize - offset[1]), d);
+    offset[2] = (size_t)random_in_range(0, (int)imageSize - 1, d);
+    region[2] = (size_t)random_in_range(1, (int)(imageSize - offset[2]), d);
```
marekkozl: previously it was dropping the region by 1, but seems it can go to the imageSize boundary - so it is a fix, right?

test_conformance/basic/test_enqueue_map.cpp
shajder: Corrected. The case seemed debatable at first (separate array layers felt like separate images) but the spec explicitly allows mapping a region spanning multiple layers.

test_conformance/basic/test_enqueue_map.cpp
shajder: Done

test_conformance/basic/test_enqueue_map.cpp
shajder: Corrected

test_conformance/basic/test_enqueue_map.cpp
shajder: What should be returned instead? The list of supported types isn't expanding fast, so an explicit default case felt like overkill, added a log statement and an early-out to skip the test instead.

test_conformance/basic/test_enqueue_map.cpp
shajder: Done

test_conformance/basic/test_enqueue_map.cpp
shajder: Indeed, this is a fix, nothing critical though. The test didn't crash previously, it just could theoretically have hit a zero-sized region in that edge case.

test_conformance/basic/test_enqueue_map.cpp
rjodinchr: `IMAGE1D_ARRAY` & `IMAGE2D` are doing the same thing

shajder: Cases merged

test_conformance/basic/test_enqueue_map.cpp
rjodinchr: That doesn't seem sufficient to me.

Currently, reaching the default case triggers a `log_info` and skips the test due to the `null` `memObject`, which is unexpected behavior. This should be treated as an error indicating a flaw in the test itself, so an assertion or similar safeguard is necessary.

Furthermore, if `create_image` fails, two redundant `test_error` messages are generated (one should be enough if it has all the information).

shajder: The behaviour was already handled and logged, so I wouldn't call it "unexpected", but I've pushed an explicit fail instead: logs the unsupported type and returns `TEST_FAIL` rather than silently skipping - as requested.

As for the "redundant" test_error: it's not just a message macro, it also acts as an error handler, which is the standard pattern throughout CTS, right?

rjodinchr: Having the whole `basic` suite pass while this subtest is skipped because of an internal error is definitely not what an OpenCL-CTS user expects. As such, I would consider that unexpected behavior.
I'm not saying it would happen with that PR, but we never know how that code will evolve, so we had better be safe now than sorry later on.

I understand `test_error` is not just a message macro; it returns. I would still argue that we should not have two of them. I would be in favor of removing the one here (in `create_image_type`) and adding more information (the `image_type`) to the other one.
Note that `return error` at the end of `create_image_type` does not make much sense, as we either return before that statement or return `CL_SUCCESS`. Removing `test_error` would bring it back to a logical state that I believe is easier to read, understand, and maintain.

shajder: Pushed. Worth noting this request does deviate from the CTS convention of failing at the point of error rather than deferring to the caller, not a problem at this scale, but flagging it so it's not read as the default going forward.

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -259,64 +473,87 @@ REGISTER_TEST(enqueue_map_image)
...
-                    value = spin - value;
-                    referenceData[((offset[1] + y) * imageSize + offset[0]) * 4
-                                  + x] = value;
+                    for (size_t x = 0; x < region[0]; x++)
+                    {
+                        size_t x_off = offset[0] + x;
+                        size_t y_off = offset[1] + y;
+                        size_t z_off = offset[2] + z;
+                        size_t ref_loc = (z_off * image_size * image_size
+                                          + y_off * image_size + x_off);
```
rjodinchr: How sure are we that the image width is equal to the image height?
Should this be:
```suggestion
                        size_t ref_loc = (z_off * img_region[0] * img_region[1]
                                          + y_off * image_size + x_off);
```
At least to make sure that whatever happens next, we get the proper dimension size to compute that offset?

test_conformance/basic/test_enqueue_map.cpp
```diff
@@ -169,78 +218,243 @@ REGISTER_TEST(enqueue_map_buffer)
...
+            region[2] = 1;
+            break;
+        }
+        case CL_MEM_OBJECT_IMAGE1D: {
+            offset[1] = offset[2] = 0;
+            region[1] = region[2] = 1;
+            break;
+        }
+    }
+}
```
rjodinchr, line 299: I think we could simplify that function by changing its prototype. Something like that:

```suggestion
void random_region_coords(size_t *offset, size_t *region,
                          const size_t *img_region, const MTdataHolder &d)
{
    offset[0] = (size_t)random_in_range(0, (int)img_region[0] - 1, d);
    region[0] = (size_t)random_in_range(1, (int)(img_region[0] - offset[0]), d);
    offset[1] = (size_t)random_in_range(0, (int)img_region[1] - 1, d);
    region[1] = (size_t)random_in_range(1, (int)(img_region[1] - offset[1]), d);
    offset[2] = (size_t)random_in_range(0, (int)img_region[2] - 1, d);
    region[2] = (size_t)random_in_range(1, (int)(img_region[2] - offset[2]), d);
}
```
I think `img_region[x]` is always at least `1` looking at how we are creating it in the code.

Note that this suggestion would also require to update the caller.

rjodinchr: Note that with that proposed change, I believe the switch on `image_type` is not needed anymore.

test_conformance/basic/test_enqueue_map.cpp
shajder: Testing images are equally sized indeed, this idea comes from [original test](https://github.com/KhronosGroup/OpenCL-CTS/blob/e2580bded220c6ac9786a1008b479aa96d32d359/test_conformance/basic/test_enqueue_map.cpp#L212-L225). I corrected suggested and few more similar locations. It doesn't affect image size, it is only a precautionary measure.

joselopeqti (changes requested):
Would it be possible to apply the following patch? It fixes 2 different issues we have observed:

- clGetSupportedImageFormats  needs an access flag when used with non-immutable allocations, so CL_MEM_READ_WRITE is added for this case.
- The `y_pitch` used for mapping should be the slicePitch in the case of 1D array images

[fix.patch](https://github.com/user-attachments/files/30003748/fix.patch)

shajder:
> Would it be possible to apply the following patch? It fixes 2 different issues we have observed:
> 
> * clGetSupportedImageFormats  needs an access flag when used with non-immutable allocations, so CL_MEM_READ_WRITE is added for this case.
> * The `y_pitch` used for mapping should be the slicePitch in the case of 1D array images
> 
> [fix.patch](https://github.com/user-attachments/files/30003748/fix.patch)

Thanks, I spotted the `clGetSupportedImageFormats` issue earlier today and drew the same conclusion (CL_MEM_READ_WRITE). As for y_pitch, I will take a look now.

shajder:
> * The `y_pitch` used for mapping should be the slicePitch in the case of 1D array images

Done

test_conformance/basic/test_enqueue_map.cpp
shajder: Corrected

joselopeqti:
> Thanks, I spotted the clGetSupportedImageFormats issue earlier today and drew the same conclusion (CL_MEM_READ_WRITE). As for y_pitch, I will take a look now.

Could you apply it like in the patch file? `CL_MEM_IMMUTABLE_EXT` should be enough for querying and would not need `CL_MEM_READ_WRITE`

shajder:
> > Thanks, I spotted the clGetSupportedImageFormats issue earlier today and drew the same conclusion (CL_MEM_READ_WRITE). As for y_pitch, I will take a look now.
> 
> Could you apply it like in the patch file? `CL_MEM_IMMUTABLE_EXT` should be enough for querying and would not need `CL_MEM_READ_WRITE`

Could you point me to the section of the `ext_immutable_memory_objects_EXT` extension (or core spec) that documents `CL_MEM_IMMUTABLE_EXT` as a substitute for an access qualifier when calling `clGetSupportedImageFormats`? Per the [flags description](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#image-format-mapping), only `CL_MEM_READ_WRITE`, `CL_MEM_READ_ONLY`, `CL_MEM_WRITE_ONLY`, or `CL_MEM_KERNEL_READ_AND_WRITE` are given meaning there, `CL_MEM_IMMUTABLE_EXT` describes post-creation mutability, not kernel access, and elsewhere it's always combined with an explicit access qualifier rather than replacing one. Without a spec reference for the exception, I'd rather apply `CL_MEM_READ_WRITE` unconditionally to all `flag_set` entries, immutable included, it costs nothing and removes the ambiguity.

joselopeqti:
In the description for flags in [clGetSupportedImageFormats](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#clGetSupportedImageFormats):

> flags may also be [CL_MEM_IMMUTABLE_EXT](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_MEM_IMMUTABLE_EXT) to query image formats that may be used to create immutable images

shajder:
> In the description for flags in [clGetSupportedImageFormats](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#clGetSupportedImageFormats):
> 
> > flags may also be [CL_MEM_IMMUTABLE_EXT](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_MEM_IMMUTABLE_EXT) to query image formats that may be used to create immutable images


It fails on my nvidia implementation:

> ERROR: clGetSupportedImageFormats failed to return supported formats! (CL_INVALID_VALUE from .../test_enqueue_map.cpp:367)

`clGetSupportedImageFormats` returns `CL_INVALID_VALUE` unless passed flags in exact range `CL_MEM_READ_WRITE/READ_ONLY/WRITE_ONLY/KERNEL_READ_AND_WRITE/IMMUTABLE_EXT`. Do you want to apply your exact patch or can we stick to this:

```
                cl_mem_flags query_flag = is_immutable_image
                    ? CL_MEM_IMMUTABLE_EXT
                    :  CL_MEM_READ_WRITE;

```

joselopeqti:
That code is fine for me

joselopeqti (approved):

rjodinchr (approved):

bashbaug:

PR #2479, opened 2025-08-12
Update D3D10 and D3D11 interop tests

ahesham-arm:
Both tests depend on a very old DirectX SDK (August 2009) and expect it to be extracted to an {NV_TOOLS} environment variable. They additionally require definining {ARCH} as a CMake configuration option, which is not needed.

Update both projects to use DirectX libraries provided by the Windows SDK and drop the unneeded configuration options.


nikhiljnv (approved):
LGTM.

Issue #2480, opened 2025-08-13
add tests for cl_khr_external_semaphore_dx_fence


bashbaug:
We re-added the cl_khr_external_semaphore_dx_fence extension for the July 2025 spec release, see: https://github.com/KhronosGroup/OpenCL-Docs/pull/1359

We do not currently have any tests for this extension, though.

CC @lakshmih

PR #2482, opened 2025-08-14
Add tests for cl_khr_external_semaphore_dx_fence

joselopeqti:
- Add tests to import DirectX 12 fences as semaphores.
- Add tests to export semaphores as DirectX 12 fences.
- Add queries tests.
- Add negative tests.

Fixes: https://github.com/KhronosGroup/OpenCL-CTS/issues/2480


test_conformance/common/directx_wrapper/directx_wrapper.cpp
```diff
@@ -0,0 +1,95 @@
...
+    if (FAILED(hr))
+    {
+        throw std::runtime_error(
+            "Failed to create DirectX 12 command allocator");
+    }
+}
+
+DirectXWrapper::~DirectXWrapper()
+{
+    if (dx_command_allocator)
```
ahesham-arm: If these are `ComPtr` you would not need to manage their lifetime manually.

nikhiljnv:
Any reason we can not add dx_fence as just another handle type to cl_khr_external_semaphore tests and need to create separate one?

nikhiljnv (approved):
LGTM.

PR #2483, opened 2025-08-14
Extension version checking

ahesham-arm:
This is a proposal to improve how we check for CL extensions.

The goal is to centralise how we check for extensions and the extension version. The change ensures that, unless otherwise explicitly stated by the test, the device is queried to check for the latest version (as specified in the CL headers) of an extension.

If we agree to this change I can then have follow-up work that uses the new macros everywhere, instead of manual checks.

The pull request is split into commits for easier reviewing, I suggest you review each commit on its own, instead of the pull request as a whole.

### DIFF
--- a/test_common/harness/testHarness.h
+++ b/test_common/harness/testHarness.h
@@ -150,16 +150,21 @@ template <typename T> T *register_test(const char *name, Version version)
 
 #define REGISTER_TEST(name) REGISTER_TEST_VERSION(name, Version(1, 2))
 
-#define REQUIRE_EXTENSION(name)                                                \
+#define REQUIRE_EXTENSION(ext)                                                 \
     do                                                                         \
     {                                                                          \
-        if (!is_extension_available(device, name))                             \
+        if (!is_extension_available(device, ext##_EXTENSION_NAME,              \
+                                    ext##_EXTENSION_VERSION))                  \
         {                                                                      \
-            log_info(name                                                      \
+            log_info(ext##_EXTENSION_NAME                                      \
                      " is not supported on this device. Skipping test.\n");    \
             return TEST_SKIPPED_ITSELF;                                        \
         }                                                                      \
     } while (0)
+
+#define HAS_EXTENSION(ext)                                                     \
+    is_extension_available(device, ext##_EXTENSION_NAME,                       \
+                           ext##_EXTENSION_VERSION)
 
 extern int gFailCount;
 extern int gTestCount;
--- a/test_conformance/api/negative_enqueue_map_image.cpp
+++ b/test_conformance/api/negative_enqueue_map_image.cpp
@@ -29,7 +29,7 @@ REGISTER_TEST(negative_enqueue_map_image)
 {
     constexpr size_t image_dim = 32;
 
-    REQUIRE_EXTENSION("cl_ext_immutable_memory_objects");
+    REQUIRE_EXTENSION(CL_EXT_IMMUTABLE_MEMORY_OBJECTS);
 
     static constexpr cl_mem_flags mem_flags[]{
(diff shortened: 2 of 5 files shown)


test_common/harness/deviceInfo.h
```diff
@@ -26,12 +27,28 @@ std::string get_device_info_string(cl_device_id device,
...
- * exception if the extension is not supported by the device. */
-cl_version get_extension_version(cl_device_id device,
-                                 const char *extensionName);
+/**
+ * @brief Check if an extension is supported by an OpenCL device.
+ *
+ * In addition to checking that the extension is in the list of extensions
+ * supported by the device, the function will check the extension version.
+ * To guarantee compatibility with any breaking changes, the function
+ * succeeds only on an exact version match.
```
bashbaug, line 38: I think I understand the rationale for checking for an exact version match, but I also think this is likely to cause confusion and problems in the future.  Couple of options to consider:

* Do nothing. :smile:
* Rename the function so it's more obvious that it is checking for the exact version, and leave the door open for additional overloads in the future - `is_exact_extension_version_available`?
* Add more version checking smarts, maybe something like checking for an exact version for major versions less than 1, or a matching major version and a minor version equal to or greater?

Note, we may want to ignore mismatching "patch" versions regardless.

test_conformance/api/negative_enqueue_map_image.cpp
```diff
@@ -29,7 +29,7 @@ REGISTER_TEST(negative_enqueue_map_image)
 {
     constexpr size_t image_dim = 32;
 
-    REQUIRE_EXTENSION("cl_ext_immutable_memory_objects");
+    REQUIRE_EXTENSION(CL_EXT_IMMUTABLE_MEMORY_OBJECTS);
```
bashbaug, line 32: This is clever, but I wonder if it's a bit too clever, since things like intellisense linking and even grep'ing for these symbols won't work.  Should we just type out the extension name and extension version instead?

```c++
REQUIRE_EXTENSION(
    CL_EXT_IMMUTABLE_MEMORY_OBJECTS_EXTENSION_NAME,
    CL_EXT_IMMUTABLE_MEMORY_OBJECTS_VERSION);
```

test_common/harness/testHarness.h
```diff
@@ -150,17 +150,22 @@ template <typename T> T *register_test(const char *name, Version version)
 
 #define REGISTER_TEST(name) REGISTER_TEST_VERSION(name, Version(1, 2))
 
-#define REQUIRE_EXTENSION(name)                                                \
+#define REQUIRE_EXTENSION(ext)                                                 \
     do                                                                         \
     {                                                                          \
-        if (!is_extension_available(device, name))                             \
+        if (!is_extension_available(device, ext##_EXTENSION_NAME,              \
+                                    ext##_EXTENSION_VERSION))                  \
```
bashbaug, line 157: This may be OK, but so we go in with eyes open - I think this means that the tests (that use this macro at least) will only run for implementations that support  the extension version in the headers, and only that version.

Do we want to have the ability to test other extension versions also?  For example, I could envision tests for an extension that have one codepath for version 1.X and another codepath for version 2.X, kind of like the tests for integer dot product, but perhaps with a breaking change between 1.x and 2.X.  For usages like this one I don't think we'd be able to use this REQUIRE_EXTENSION macro the way it is currently defined.

test_common/harness/deviceInfo.h
EwanC: I don't have anything to add, but this thread made me think of this CTS Issue we still have  https://github.com/KhronosGroup/OpenCL-CTS/issues/2152 open about how the versions of provisional extensions is tested. It would be nice if we could resolve that issue as part of this PR.

bashbaug:
Removing "focused review" for now while discussions are ongoing.

PR #2493, opened 2025-08-18
Fix channel count for RGBx

lakshmih:
RGBx is defined as a 4 channel format where the 4th channel is ignored (5.3.1.1)

### DIFF
--- a/test_common/harness/imageHelpers.cpp
+++ b/test_common/harness/imageHelpers.cpp
@@ -137,10 +137,10 @@ uint32_t get_channel_order_channel_count(cl_channel_order order)
         case CL_RGx: return 2;
 
         case CL_RGB:
-        case CL_sRGB:
-        case CL_sRGBx: return 3;
+        case CL_sRGB: return 3;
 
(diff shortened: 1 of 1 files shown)


test_common/harness/imageHelpers.cpp
```diff
@@ -137,10 +137,10 @@ uint32_t get_channel_order_channel_count(cl_channel_order order)
         case CL_RGx: return 2;
 
         case CL_RGB:
-        case CL_RGBx:
         case CL_sRGB:
         case CL_sRGBx: return 3;
 
+        case CL_RGBx:
```
bashbaug, line 142: I can definitely see places where it is wrong to return "3" vs. "4" for `CL_RGBx`, but have you checked that there are no places where we will check the fourth "ignored" channel if we return "4" instead of "3"?

I think we should also keep `CL_RGBx` and `CL_sRGBx` in sync, so if we do end up moving `CL_RGBx`, we should move `CL_sRGBx` also.

PR #2494, opened 2025-08-19
Add negative tests for context API functions

MichaelRizkalla-arm:
This PR adds negative tests for the following API functions:

- clCreateContext
- clCreateContextFromType
- clRetainContext
- clReleaseContext
- clSetContextDestructorCallback

Also, define a new macro `test_object_failure_ret` to test for an expected error code and the returned object is `NULL` as a result of a failure.

### DIFF
--- a/test_conformance/api/negative_context.cpp
+++ b/test_conformance/api/negative_context.cpp
@@ -1,0 +1,290 @@
+//
+// Copyright (c) 2025 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+#include "testBase.h"
+
+/* Negative Tests for clCreateContext */
+REGISTER_TEST(negative_create_context)
+{
+    cl_context_properties props[3] = {
+        CL_CONTEXT_PLATFORM, reinterpret_cast<cl_context_properties>(nullptr), 0
+    };
+    cl_int err = 0;
+    cl_context ctx = clCreateContext(props, 1, &device, nullptr, nullptr, &err);
+    test_object_failure_ret(
+        ctx, err, CL_INVALID_PLATFORM,
+        "clCreateContext should return CL_INVALID_PLATFORM when:\"an invalid "
+        "platform object is used with the CL_CONTEXT_PLATFORM property\" using "
+        "a nullptr",
+        TEST_FAIL);
+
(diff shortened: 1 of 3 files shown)


test_conformance/api/negative_context.cpp
```diff
@@ -0,0 +1,300 @@
...
+        props[1] = reinterpret_cast<cl_context_properties>(nullptr);
+        ctx = clCreateContext(props, 1, &device, nullptr, nullptr, &err);
+        test_object_failure_ret(
+            ctx, err, CL_INVALID_PROPERTY,
+            "clCreateContext should return CL_INVALID_PROPERTY when: \"context "
+            "property name in properties is not a supported property name\"",
+            TEST_FAIL);
+
+        cl_context_properties invalid_value{ -1 };
+        props[0] = CL_CONTEXT_INTEROP_USER_SYNC;
```
EwanC, line 46: we could enter this control flow if `get_device_cl_version(device)` returns `Version(1,1)` but `CL_CONTEXT_INTEROP_USER_SYNC` isn't added to the spec until version 1.2

test_conformance/api/negative_context.cpp
```diff
@@ -0,0 +1,300 @@
...
+
+    cl_context_info INVALID_PARAM_VALUE = 0;
+    err = clGetContextInfo(context, INVALID_PARAM_VALUE, 0, nullptr, nullptr);
+    test_failure_error_ret(
+        err, CL_INVALID_VALUE,
+        "clGetContextInfo should return CL_INVALID_VALUE when: \"param_name is "
+        "not one of the supported values\"",
+        TEST_FAIL);
+
+    err = clGetContextInfo(context, CL_CONTEXT_REFERENCE_COUNT, 0, &param_value,
```
EwanC: using a non-zero value, e.g `sizeof(cl_uint) - 1` would be a marginally stronger test.

test_conformance/api/negative_context.cpp
```diff
@@ -0,0 +1,300 @@
...
+            "clCreateContext should return CL_INVALID_PROPERTY when: \"the "
+            "value specified for a supported property name is not valid\"",
+            TEST_FAIL);
+
+        props[0] = CL_DEVICE_TYPE;
+        props[1] = 0;
+        ctx = clCreateContext(props, 1, &device, nullptr, nullptr, &err);
+        test_object_failure_ret(
+            ctx, err, CL_INVALID_PROPERTY,
+            "clCreateContext should return CL_INVALID_PROPERTY when: \"the "
```
EwanC: Does this provide extra coverage that the first check for this case on line 41 doesn't?

test_conformance/api/negative_context.cpp
MichaelRizkalla-arm: I updated the branch condition to `>= Version(1, 2)`.

test_conformance/api/negative_context.cpp
MichaelRizkalla-arm: Done

test_conformance/api/negative_context.cpp
MichaelRizkalla-arm: No, it's been removed.

kpet:
Merging as agreed in 2026/01/13 teleconference.

PR #2495, opened 2025-08-19
Implement negative tests for cl_device_id API functions

MichaelRizkalla-arm:


test_common/harness/parseParameters.cpp
```diff
@@ -50,6 +50,12 @@ void helpInfo()
     --num-worker-threads <num>
         Select parallel execution with the specified number of worker threads.
 
+    --invalid-object-scenarios=<option_1>,<option_2>....
```
EwanC, line 61: Would be nice to stick in the help message what the default value is, i.e nullptr only.

MichaelRizkalla-arm: I modified the help message to indicate that `nullptr` is the default value.

bashbaug: Could we also please add help text so it is very clear that `valid_object_wrong_type` is NOT required for OpenCL conformance, because it is testing implementation-defined behavior?  From the spec:

> An OpenCL implementation must check for a NULL object to determine if an object is valid. The behavior for all other invalid objects is implementation-defined.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,534 @@
...
+    {
+        log_info("Platform has every Device Type... Skipping Test\n");
+    }
+    else
+    {
+        test_failure_error_ret(
+            err, CL_DEVICE_NOT_FOUND,
+            "clGetDeviceIDs should return CL_DEVICE_NOT_FOUND when: \"no "
+            "OpenCL devices that matched device_type were found\"",
+            TEST_FAIL);
```
bashbaug, line 127: Does it make sense to move this check into the loop, so we unconditionally check all device types, and we don't stop with the first device type that returns an error?

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,534 @@
...
+            "\"host_timestamp or device_timestamp is NULL\" using nullptr for "
+            "host_timestamp ",
+            TEST_FAIL);
+    }
+    else
+    {
+        log_info("Platform does not Support Timers\n");
+        log_info("Skipping CL_INVALID_VALUE tests\n");
+
+        err = clGetDeviceAndHostTimer(device, device_timestamp, host_timestamp);
```
bashbaug, line 193: This is fine and this test should execute quickly, but mentioning for completeness: this is duplicating the "consistency_device_and_host_timer" test.  Same with `clGetHostTimer`, below.

test_conformance/api/negative_device.cpp
```diff
@@ -0,0 +1,534 @@
...
+                               "a valid device\"",
+                               TEST_FAIL);
+    }
+
+    cl_platform_id platform = getPlatformFromDevice(device);
+    // Initialise timer_resolution to a Non-0 value as CL2.1/2 devices must
+    // support timer synchronisation
+    cl_ulong timer_resolution = 1;
+    auto device_version = get_device_cl_version(device);
+    if (device_version >= Version(2, 1))
```
lakshmih: This isnt needed here and elsewhere in the file (ln 155) since it should be handled by . REGISTER_TEST_VERSION(<name>, Version(2, 1))

test_common/harness/testHarness.h
```diff
@@ -257,6 +258,41 @@ extern std::string get_platform_info_string(cl_platform_id platform,
                                             cl_platform_info param_name);
 extern bool is_platform_extension_available(cl_platform_id platform,
                                             const char *extensionName);
+extern std::vector<cl_device_id> get_all_available_devices();
+extern bool different_device_available(cl_device_id deviceID,
+                                       cl_device_id *different_device);
```
lakshmih: Are these functions (get_all_available_devices and different_device_available) defined?

test_common/harness/testHarness.h
```diff
@@ -257,6 +258,41 @@ extern std::string get_platform_info_string(cl_platform_id platform,
...
+enum InvalidObject
+{
+    Nullptr = 1 << 1,
```
lakshmih: Should this be 1 << 0?

test_common/harness/parseParameters.cpp
MichaelRizkalla-arm: I've added a note to specify that the option is not required.

test_conformance/api/negative_device.cpp
MichaelRizkalla-arm: I removed the check and a similar one in the following test.

test_common/harness/testHarness.h
MichaelRizkalla-arm: No, I removed the declarations.

PR #2497, opened 2025-08-20
Fix basic progvar_prog_scope_init and progvar_prog_scope_uninit

AhmedAmraniAkdi:
The test was mapping a memory object using CL_MAP_READ but then writing to its host_ptr using memset.

### DIFF
--- a/test_conformance/basic/test_progvar.cpp
+++ b/test_conformance/basic/test_progvar.cpp
@@ -1264,7 +1264,7 @@ static int l_write_read_for_type(cl_device_id device, cl_context context,
             }
 
             cl_uchar* read_ptr = (cl_uchar*)clEnqueueMapBuffer(
-                queue, read_mem, CL_TRUE, CL_MAP_READ, 0, read_data_size, 0, 0,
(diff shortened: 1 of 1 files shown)


kpet:
FYI @JohnKesapidesARM

test_conformance/basic/test_progvar.cpp
```diff
@@ -1264,7 +1264,7 @@ static int l_write_read_for_type(cl_device_id device, cl_context context,
             }
 
             cl_uchar* read_ptr = (cl_uchar*)clEnqueueMapBuffer(
-                queue, read_mem, CL_TRUE, CL_MAP_READ, 0, read_data_size, 0, 0,
+                queue, read_mem, CL_TRUE, CL_MAP_WRITE, 0, read_data_size, 0, 0,
```
bashbaug, line 1267: Since it looks like we're overwriting the entire mapped region I think we could use [CL_MAP_WRITE_INVALIDATE_REGION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MAP_WRITE_INVALIDATE_REGION) instead, which might be a bit more efficient, but I doubt this will matter much in practice.

bashbaug (approved):

PR #2498, opened 2025-08-20
mutable_command_info: remove unnecessary Skip function

rjodinchr:
The check implementated by that Skip function is already implemented in 'InfoMutableCommandBufferTest::Skip()'.

Also this is trying to get the extension_version before checking whether the extension is supported, leading to false negative for device not supporting the extension.


rjodinchr:
@bashbaug @pj87, please take a look

PR #2499, opened 2025-08-21
Extra mutable dispatch coverage

EwanC:
Adds tests to cover points 2 & 3 from the questions asked about cl_khr_command_buffer_mutable_dispatch in https://github.com/KhronosGroup/OpenCL-Docs/issues/1437

* New test for point 2 from issue, `mutable_dispatch_updates_persist`, testing multiple enqueues of a command-buffer after update, and that the updated argument persists for all of them.
* New test for point 3 pseudocode from issue in test `mutable_dispatch_set_kernel_arg`


test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_update_state.cpp
```diff
@@ -0,0 +1,345 @@
...
+                data[tid] += value;
+            })";
+
+        cl_int error = create_single_kernel_helper(
+            context, &program, &kernel, 1, &add_kernel, "add_kernel");
+        test_error(error, "Creating kernel failed");
+        return CL_SUCCESS;
+    }
+
+    cl_int Run() override
```
bashbaug: I don't think this version of "Run" from the base class ever gets called - does it?

If not, it can be removed.

EwanC: oops, I forgot to delete this from the base class which I did a refactor. Removed the the method now, thanks for catching.

bashbaug (approved):

PR #2502, opened 2025-08-26
Implement negative tests for clEnqueueMarkerWithWaitList

MichaelRizkalla-arm:
This change adds negative tests for `clEnqueueMarkerWithWaitList`

### DIFF
--- a/test_conformance/api/negative_enqueue_marker.cpp
+++ b/test_conformance/api/negative_enqueue_marker.cpp
@@ -1,0 +1,97 @@
+//
+// Copyright (c) 2025 The Khronos Group Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "testBase.h"
+#include "harness/typeWrappers.h"
+
+REGISTER_TEST(negative_enqueue_marker_with_wait_list)
+{
+    cl_platform_id platform = getPlatformFromDevice(device);
+    cl_context_properties props[3] = {
+        CL_CONTEXT_PLATFORM, reinterpret_cast<cl_context_properties>(platform),
+        0
+    };
+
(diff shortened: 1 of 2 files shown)


test_conformance/api/negative_enqueue_marker.cpp
```diff
@@ -0,0 +1,96 @@
...
+        "clEnqueueMarkerWithWaitList should return CL_INVALID_EVENT_WAIT_LIST "
+        "when: \"num_events_in_wait_list is 0 but event_wait_list is not "
+        "NULL\"",
+        TEST_FAIL);
+    test_assert_error(ret_event == nullptr,
+                      "if clEnqueueMarkerWithWaitList failed, no ret_event "
+                      "should be created");
+
+    err = clEnqueueMarkerWithWaitList(
+        queue, 1, reinterpret_cast<cl_event*>(&context), &ret_event);
```
EwanC: Not sure implementations are expected to do an invalid check the event object is valid, passed nullptr. See https://github.com/KhronosGroup/OpenCL-Docs/issues/482#issuecomment-795685410

bashbaug: I agree with Ewan that this case should not be tested.

MichaelRizkalla-arm: Thanks for pointing this out!
I've updated the test case with a wait list that contains a `nullptr` event.

test_conformance/api/negative_enqueue_marker.cpp
```diff
@@ -0,0 +1,97 @@
...
+        clCreateContext(props, 1, &device, nullptr, nullptr, &err);
+    test_error(err, "clCreateContext failed");
+
+    cl_event ret_event = nullptr;
+
+    err = clEnqueueMarkerWithWaitList(nullptr, 0, nullptr, &ret_event);
+    test_failure_error_ret(
+        err, CL_INVALID_COMMAND_QUEUE,
+        "clEnqueueMarkerWithWaitList should return CL_INVALID_COMMAND_QUEUE "
+        "when: \"context is not a valid context\" using a nullptr",
```
EwanC: nitpick - think this should be talking about valid command-queues not valid contexts

bashbaug: In case it's helpful, here is a specific suggestion:

```suggestion
        "when: \"command_queue is not a valid host command-queue\" using a nullptr",
```

MichaelRizkalla-arm: I've updated the error message.

EwanC (approved):

bashbaug:
Merging as discussed over email.

PR #2503, opened 2025-08-26
c11_atomics: Fix cl_uint --> cl_half conversion on host

Nuullll:


Nuullll:
> Hi, I'm having a hard time connecting the dots together. Can you please provide a test command line that I can run that demonstrates the problem? Thanks!

Sorry for the late response - I was distracted by other work.

The issue can be reproduced by running:
```
./OpenCL-CTS/build/test_conformance/c11_atomics/test_c11_atomics svm_atomic_store
```
with Intel OpenCL CPU implementation (which has fp16 support enabled).

## Root cause
In the OpenCL kernel:
```opencl
__kernel void test_atomic_kernel(uint threadCount, uint numDestItems, volatile __global atomic_half *destMemory, __global half *oldValues)
{
  uint  tid = get_global_id(0);
  atomic_store_explicit(&destMemory[tid], tid, memory_order_relaxed, memory_scope_all_devices);
}
```
When `uint tid` is stored as a half value, the OpenCL device properly converts the integer to IEEE 754 half-precision floating point format. However, in the host reference code, the original implementation simply cast `cl_uint` directly to `cl_half`:
```cpp
host_atomic_store(&destMemory[tid], (HostDataType)tid, MemoryOrder());
```
`cl_half` is defined as `uint16_t` in OpenCL Header, a direct cast from `cl_uint` to `cl_half` just truncates the integer bits, rather than performing proper float-to-half conversion.

bashbaug:
Ah, gotcha, thank you for the explanation!

FWIW, it was the static cast to a `float` in the snippet below that looked questionable (note: v is a `uint`)::

```c++
            // For half types, convert from float to proper half-precision bit
            // pattern
            return cl_half_from_float(static_cast<float>(v), gHalfRoundingMode);
```

But, it looks like this is matching what the kernel is doing (convert the `uint` to a `half`), so it's correct.

Interestingly, there does not appear to be a problem with our GPU device, which does not support SVM atomics but does support fp16 atomic load and store.  I am running:

```sh
./test_conformance/c11_atomics/test_c11_atomics atomic_store
```

I'll add "focused review" and we'll see if we can get this merged next week.

test_conformance/c11_atomics/common.h
```diff
@@ -452,6 +452,22 @@ class CBasicTest : CTest {
...
+            // For half types, convert from float to proper half-precision bit
+            // pattern
+            return cl_half_from_float(static_cast<float>(v), gHalfRoundingMode);
+        }
+        else
+        {
+            // For non-half types, direct cast is fine
+            return static_cast<HostDataType>(v);
+        }
+    }
```
shajder: Thanks, this is great find, and I see your point. The stored value is indeed not aligned with `ExpectedValue` method. 

However, while the correction is necessary, the location you proposed is not consistent with other host-atomics modifications. Please take a look at [this](https://github.com/KhronosGroup/OpenCL-CTS/blob/e641de99a5c6f671c63e490acd5da0e492a4438f/test_conformance/c11_atomics/host_atomics.h#L97-L132) example. Your correction should be placed around [here](https://github.com/KhronosGroup/OpenCL-CTS/blob/e641de99a5c6f671c63e490acd5da0e492a4438f/test_conformance/c11_atomics/host_atomics.h#L172-L182). Could you please move your changes to `host_atomics.h` ? If you think you will not be able to make the update within a week or two please let me know so I will take over.

Nuullll:
> Interestingly, there does not appear to be a problem with our GPU device, which does not support SVM atomics but does support fp16 atomic load and store. I am running:
> 
> ```shell
> ./test_conformance/c11_atomics/test_c11_atomics atomic_store
> ```

@bashbaug Thanks for pointing this out. `atomic_store` passes for both CPU and GPU, because by default it does not use `HostFunction` for verification:
https://github.com/KhronosGroup/OpenCL-CTS/blob/e641de99a5c6f671c63e490acd5da0e492a4438f/test_conformance/c11_atomics/common.h#L327-L333
https://github.com/KhronosGroup/OpenCL-CTS/blob/e641de99a5c6f671c63e490acd5da0e492a4438f/test_conformance/c11_atomics/common.h#L1141
https://github.com/KhronosGroup/OpenCL-CTS/blob/e641de99a5c6f671c63e490acd5da0e492a4438f/test_conformance/c11_atomics/common.h#L1426-L1428
https://github.com/KhronosGroup/OpenCL-CTS/blob/e641de99a5c6f671c63e490acd5da0e492a4438f/test_conformance/c11_atomics/common.h#L146-L155
https://github.com/KhronosGroup/OpenCL-CTS/blob/e641de99a5c6f671c63e490acd5da0e492a4438f/test_conformance/c11_atomics/test_atomics.cpp#L87-L92

The same issue can be exposed on CPU if we force using host threads for verification:
```
./test_conformance/c11_atomics/test_c11_atomics atomic_store -host
```

test_conformance/c11_atomics/common.h
Nuullll: Thanks for the comment.
However, I don't think we should place the correction in the body of `host_atomic_store` or `host_atomic_exchange` -- the error happens when we cast (truncate) a `cl_uint` value to a `cl_half` (which is effectively `uint16_t`), which correlates to the C-style cast `(HostDataType)tid` used **before** calling `host_atomic_*`.

shajder: Correct me if I am mistaken at any point. The `cl_uint tid` is cast to `HostDataType` which should be fine as long as `tid` doesn't overflow. Since it should remain within the range 0-512, I don't see any issue so far. Then, in `host_atomic_exchange` the correction below should perform exactly the same operation as your current implementation in `ConvertToHostDataType` (which, by the way, is not unrolled at build time):

```
template <typename AtomicType, typename CorrespondingType>
CorrespondingType host_atomic_exchange(volatile AtomicType *a, CorrespondingType c,
                                       TExplicitMemoryOrderType order)
{
    CorrespondingType ct=c;
    if constexpr (std::is_same_v<AtomicType, HOST_ATOMIC_HALF>)
        ct = cl_half_from_float(static_cast<float>(c), gHalfRoundingMode);

#if defined( _MSC_VER ) || (defined( __INTEL_COMPILER ) && defined(WIN32))
    if (sizeof(CorrespondingType) == 2)
        return InterlockedExchange16(reinterpret_cast<volatile SHORT *>(a), ct);
    else
        return InterlockedExchange(reinterpret_cast<volatile LONG *>(a), ct);
#elif defined(__GNUC__)
    return __sync_lock_test_and_set(a, ct);
#else
  log_info("Host function not implemented: atomic_exchange\n");
  return 0;
#endif
}
```

Am I missing something obvious ? Please let me know if that works, I can't test that on my machine by my own.

Nuullll: Thank you for the feedback. I've reconsidered the approach and believe placing the conversion inside `host_atomic_exchange` has a semantic issue: by the time the function receives the parameter `c`, the C-style cast `(HostDataType)tid` has already corrupted the value through integer truncation, although we would be able to **recover** the right data as long as `tid` does not overflow, it's not guaranteed if we further extend the test.

Instead, I propose a more robust solution: defining a `HostHalf` wrapper class that ensures proper conversion at construction time:

```cpp
class HostHalf {
public:
    HostHalf(cl_uint value = 0): value_(cl_half_from_float(static_cast<float>(value), gHalfRoundingMode)) {}
    operator cl_half() const { return value_; }

    friend std::ostream &operator<<(std::ostream &os, const HostHalf &hh) {
        os << cl_half_to_float(hh.value_);
        return os;
    }

private:
    cl_half value_;
};


// ...

// #define HOST_HALF               cl_half
#define HOST_HALF               HostHalf
```

so that it automatically handles all cast sites through conversion operators.

Would this approach be acceptable? I can provide an updated patch if you agree with this direction.

shajder: > placing the conversion inside `host_atomic_exchange` has a semantic issue: by the time the function receives the parameter `c`, the C-style cast `(HostDataType)tid` has already corrupted the value through integer truncation

Please notice one important discrepancy in your approach - you are assuming that casting from `cl_int` to `cl_half` may result in integer truncation, meanwhile converting `cl_int` to `cl_half` should work as expected. Sadly, maximum value of positive half-float value in the IEEE 754 standard is even smaller than `uint16_t` (a.k.a. `cl_half`). In other words, if your `cl_int` value is too large for `uint16_t` you will also overflow when converting via `cl_half_from_float` procedure. That is why we need control over the logic underlying the test to avoid exceeding the maximum range.

As I mentioned, I am not able to test the code on my machine so I will appreciate your answer. Could you temporarly apply my suggestion so we could disscuse it in the thread ? If we will spot any problems with it then we could apply different solution.

Nuullll: Thank you for the clarification. You're absolutely right that with the current test implementation where `tid` is bounded (0-1024), both approaches would produce identical results since the values stay within the representable range of half-float.

I've tested your suggested approach by adding the conversion in `host_atomic_exchange` and can confirm it works correctly for the current tests. Here's what I found:

**Your approach works because:**
- Current `tid` values (0-1024) are all representable in half-float (max finite value ~65504)
- After the cast `(cl_half)tid`, we get the raw bit pattern (e.g., 5 --> `0x0005`)
- Then `cl_half_from_float(static_cast<float>(0x0005), mode)` interprets `0x0005` as float `5.0` and produces `0x4500` (correct half-float for 5.0)

**The semantic difference I was concerned about:** When values exceed `uint16_t` range:

```cpp
cl_uint x = 1 << 16;  // 65536, 0x10000
// Cast approach:
cl_half cast = static_cast<cl_half>(x);           // 0x0000 (truncation)
cl_half_from_float(static_cast<float>(cast), mode); // 0x0000 (0.0)

// Direct conversion approach:
cl_half_from_float(static_cast<float>(x), mode);   // 0x7C00 (+Infinity)
```

The intention might be exchanging a large value (which naturally overflows to Infinity) with the atomic pointer, while the truncated zero is unexpected.

However, given that:
1. The current tests are designed with bounded `tid` values
2. Your approach is consistent with existing patterns in the codebase

I'm fine with proceeding with your suggested approach. Alternatively, if there's value in making the code more robust against future extensions (where `tid` or other input might exceed current bounds), we could add a comment or assertion documenting the value range assumption, or we could still consider the wrapper class approach. Please let me know your preference.

shajder: Thank you, Adding comments or assertions to handle cases exceeding half-float bounds seems like a good approach. I assume this should be done before each `host_atomic_store`/`host_atomic_exchange` call. Would you be able to implement those changes?

test_conformance/c11_atomics/common.h
```diff
@@ -452,6 +452,22 @@ class CBasicTest : CTest {
...
+    // Helper function to convert cl_uint to appropriate atomic value
+    HostDataType ConvertToHostDataType(cl_uint v)
+    {
+        if (DataType()._type == TYPE_ATOMIC_HALF)
+        {
+            // For half types, convert from float to proper half-precision bit
+            // pattern
+            return cl_half_from_float(static_cast<float>(v), gHalfRoundingMode);
+        }
+        else
```
silverclaw: This else is redundant here, since the previous block ended with a terminating statement.
Thus the following indentation is also redundant.

Nuullll: Thanks! Nice catch.
I'll update after finalizing discussion with shajder

test_conformance/c11_atomics/common.h
Nuullll: > Thank you, Adding comments or assertions to handle cases exceeding half-float bounds seems like a good approach. I assume this should be done before each `host_atomic_store`/`host_atomic_exchange` call. Would you be able to implement those changes?

I just realized that: if we need to add comments/assertions before each `host_atomic_store`/`host_atomic_exchange`, then why don't we fix the lossy cast problem using current approach or a wrapper class?

shajder: > if we need to add comments/assertions before each `host_atomic_store`/`host_atomic_exchange`

We actually do not need that, it’s extra caution.

Just to be on the same page, I’ll recap. I understand the fix is necessary. The discussion above refers to a minor issue: the location of the fix. I’m fine making it more robust, but your solution should be comprehensive. Specifically, the wrapper class should handle other host atomic operations like `host_atomic_fetch_add`, `host_atomic_fetch_sub` and so on. Does that make sense?

test_conformance/c11_atomics/host_atomics.h
```diff
@@ -175,7 +243,7 @@ CorrespondingType host_atomic_exchange(volatile AtomicType *a, CorrespondingType
     else
         return InterlockedExchange(reinterpret_cast<volatile LONG *>(a), c);
```
shajder: Some of Windows CI/CD checks failed here, below correction works for me (build only)

```
    if constexpr (sizeof(CorrespondingType) == 2)
    {
        return InterlockedExchange16(
            reinterpret_cast<volatile SHORT*>(a),
            static_cast<SHORT>( static_cast<std::uint16_t>(c) )
        );
    }
    else
    {
        return InterlockedExchange(
            reinterpret_cast<volatile LONG*>(a),
            static_cast<LONG>( static_cast<std::uint32_t>(c) )
        );
    }

```

shajder (dismissed):

bashbaug:
@Nuullll would you mind resolving the merge conflicts?  I took a look and they don't look too bad, but they aren't quite something I'm comfortable resolving via the web UI.  Thanks!

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1736,105 +1596,36 @@ class CBasicTestFetchSub
...
+                sums.push_back(subtract(ref_vals.begin(), ref_vals.end()));
+                sums.push_back(subtract(ref_vals.rbegin(), ref_vals.rend()));
 
                 std::sort(sums.begin(), sums.end());
+                assert(std::all_of(sums.begin(), sums.end(),
+                                   [](const HostDataType &val) {
+                                       return std::isfinite(
+                                           static_cast<double>(val));
+                                   })
+                       && "Infinite subtraction value detected!");
```
Nuullll, line 1623: This assertion indeed helped capture the `kahan_sub` implementation bug in #2368. See the fix here: https://github.com/KhronosGroup/OpenCL-CTS/pull/2503/files#diff-aac1c2081764e51ba19f3e93970d6de6c44ca0b530316e76acddfa4bb4a1829cR1533

test_conformance/c11_atomics/test_atomics.cpp
```diff
@@ -1659,7 +1530,7 @@ template <> double kahan_sub<double>(const std::vector<double> &nums)
     {
         double y = num - compensation;
         double t = sum - y;
-        compensation = (t - sum) - y;
```
Nuullll, line 1662: `(t-sum) == -y`, so `compensation == -2y` here, causing `sum` to explode and overflow to infinity.

Nuullll:
@shajder @bashbaug I've resolved the conflict and fixed a minor issue of #2368 , please take another look. Thanks!

lakshmih (dismissed):

test_conformance/c11_atomics/test_atomics.cpp
shajder: Thanks, good catch but this is still not quite correct. It should be:

```
double y = - num - compensation;
double t = sum + y;
compensation = (t - sum) - y;
```

PR #2504, opened 2025-08-26
remove checks for the command-buffer pending state

bashbaug:
This is a very small subset of the changes in #2477 to get things building again, since the command-buffer pending state is no longer in the spec or headers.


ahesham-arm (approved):
Needed to get the CTS compiling again after https://github.com/KhronosGroup/OpenCL-Headers/pull/289

PR #2505, opened 2025-08-27
Implement negative tests for cl_command_queue functions

MichaelRizkalla-arm:
This change adds negative tests to `cl_command_queue`-related APIs:
- clCreateCommandQueue
- clCreateCommandQueueWithProperties
- clSetDefaultDeviceCommandQueue
- clRetainCommandQueue
- clReleaseCommandQueue
- clGetCommandQueueInfo
- clSetCommandQueueProperty

### DIFF
--- a/test_conformance/api/negative_queue.cpp
+++ b/test_conformance/api/negative_queue.cpp
@@ -16,92 +16,184 @@
 #include "testBase.h"
 #include "harness/typeWrappers.h"
 
+#include <vector>
+
 REGISTER_TEST(negative_create_command_queue)
 {
-    cl_command_queue_properties device_props = 0;
-    cl_int error = clGetDeviceInfo(device, CL_DEVICE_QUEUE_PROPERTIES,
-                                   sizeof(device_props), &device_props, NULL);
-    test_error(error, "clGetDeviceInfo for CL_DEVICE_QUEUE_PROPERTIES failed");
+    cl_int err = 0;
+    clCreateCommandQueue(nullptr, device, 0, &err);
+    test_failure_error_ret(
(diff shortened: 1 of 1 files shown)


test_conformance/api/negative_queue.cpp
```diff
@@ -16,92 +16,236 @@
...
     {
-        // Early return as we can't check correct error is returned for
-        // unsupported property.
-        return TEST_PASS;
+        clCreateCommandQueue(context, different_device, 0, &err);
+        test_failure_error_ret(
+            err, CL_INVALID_DEVICE,
+            "clCreateCommandQueue should return CL_INVALID_DEVICE when: "
+            "\"device is not associated with context\"",
+            TEST_FAIL);
```
bashbaug: Our implementations currently pass these tests, but I belive this is testing implementation-defined behavior according to the agreement in https://github.com/KhronosGroup/OpenCL-Docs/issues/482#issuecomment-795685410, since this is passing a non-`NULL` device object and expecting the error code `CL_INVALID_DEVICE`.

Same goes for the other "different_device_available" tests.

Options to consider:
1. Remove the tests, since they're testing implementation-defined behavior.
2. Keep the tests, but only log an error code other than `CL_INVALID_DEVICE` and do not fail the test.
3. Revisit the requirements for an implementation in this case.

MichaelRizkalla-arm: I have modified the test to retrieve a valid OpenCL device using `GetOpposingDevice` that is not associated with `context`.

bashbaug (approved):
Thanks!  LGTM.

lakshmih (approved):

PR #2506, opened 2025-08-27
c11_atomics: Filter invalid memory orders for 2 order instructions

alan-baker:
* OpenCL C (and SPIR-V) require that the failure memory order is not stronger than the success memory order.

Also see Khronos internal memory model issue #181

CC @bashbaug

### DIFF
--- a/test_conformance/c11_atomics/common.h
+++ b/test_conformance/c11_atomics/common.h
@@ -691,12 +691,25 @@ public:
                                            cl_context context,
                                            cl_command_queue queue)
     {
-        // Comparator for success and failure memory orders.
-        // Failure must not be stronger than success.
-        const auto compareOrderStrength = [](TExplicitMemoryOrderType success,
-                                             TExplicitMemoryOrderType failure) {
-            if (success == MEMORY_ORDER_RELEASE)
-                return failure == MEMORY_ORDER_RELAXED;
+        // Comparator for orders and scopes.
+        const auto checkValidity = [](TExplicitMemoryOrderType success,
+                                      TExplicitMemoryOrderType failure,
+                                      TExplicitMemoryScopeType scope) {
+            // Both memory order arguments must be set (or neither).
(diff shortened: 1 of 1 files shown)


test_conformance/c11_atomics/common.h
```diff
@@ -691,6 +691,15 @@ class CBasicTestMemOrder2Scope
...
                                            cl_command_queue queue)
     {
+        // Comparator for success and failure memory orders.
+        // Failure must not be stronger than success.
+        const auto compareOrderStrength = [](TExplicitMemoryOrderType success,
+                                             TExplicitMemoryOrderType failure) {
+            if (success == MEMORY_ORDER_RELEASE)
+                return failure == MEMORY_ORDER_RELAXED;
+            return failure <= success;
+        };
```
bashbaug: I do think we need to add some filtering here, but the conditions that are checked do not appear to be the conditions described in the spec.  Specifically, the spec says:

> The failure argument shall not be memory_order_release nor memory_order_acq_rel. The failure argument shall be no stronger than the success argument.

Couple of other suggestions:

* Somewhat bike-sheddy, but could we call this something like `checkValidity` instead of `compareOrderStrength`, since the filtering will need to take more than just the strength into account?
* Could we fold some of the prior checks into the same function, for added clarity?
    * "both memory order arguments must be set (or none)"
    * "memory scope without memory order is not allowed"

Note, I know we have an ongoing discussion happening on the internal issue...

alan-baker: I think combining the checks makes sense. I'll do that once we resolve the internal discussion.

alan-baker: I've updated this to combine the checks and with the resolution of the internal issue.

bashbaug (approved):
LGTM, but it'd be good to get a few more eyes on this before merging.

Note: There is a lot in our review queue right now, but we'll try to get this in ASAP.

joselopeqti (approved):

PR #2509, opened 2025-09-01
Fix -list option in test_spir

hvdijk:
This handles the case where test_spir is called without arguments.


bashbaug:
Merging as agreed in the September 2nd teleconference.

PR #2510, opened 2025-09-01
Fix test_api negative_set_read_write_image_arg.

hvdijk:
When the implementation does not have image support, the functions used are not required to be defined and the test should be skipped.


kpet (approved):

bashbaug (approved):
Merging as agreed in the September 2nd teleconference.

PR #2511, opened 2025-09-03
Fix buffer overflow in MutableDispatchWorkDim test

Nuullll:
The test had critical buffer overflow issues:

1. Buffer size was calculated incorrectly: used update_elements (4) instead
   of total work items. For 3D kernels, this meant allocating 16 bytes
   when 64*4=256 bytes were needed for the updated 4x4x4 grid.

2. Original 2x2x2 grid writes 8 elements (32 bytes) but buffer was only
   16 bytes, causing overflow on first execution.

3. Updated 4x4x4 grid writes 64 elements (256 bytes) with massive
   overflow into adjacent memory.

4. Verify function only checked one dimension instead of total elements
   in the 3D grid.

Fixed by:
- Calculating total work items as product of all dimensions
- Using update_total_elements (64) for buffer allocation
- Updating Verify calls to check correct number of elements
- Adding constants for original_total_elements and update_total_elements


EwanC (approved):
Thanks for the fix

PR #2512, opened 2025-09-03
Modified kernel code to correspond to the Image declared format (CL_UNSIGNED_INT8) 

sergiupopescu22:
**For mutable_dispatch_image_1d_arguments & mutable_dispatch_image_2d_arguments:**
As the images are created using CL_UNSIGNED_INT8, the kernel does not use correct instructions, as they are designed for signed variable. This fix consists of modifying the kernel code to use unsigned instructions and auxiliary variables .

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_enqueue.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_enqueue.cpp
@@ -429,9 +429,6 @@ struct EnqueueCommandBufferEventWaitListNullOrEventsNull
                                "CL_INVALID_EVENT_WAIT_LIST",
                                TEST_FAIL);
 
-        error = clSetUserEventStatus(event, CL_COMPLETE);
-        test_error(error, "Unable to set user event to complete");
-
         return CL_SUCCESS;
     }
 };


test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_enqueue.cpp
```diff
@@ -429,6 +429,9 @@ struct EnqueueCommandBufferEventWaitListNullOrEventsNull
                                "CL_INVALID_EVENT_WAIT_LIST",
                                TEST_FAIL);
 
+        error = clSetUserEventStatus(event, CL_COMPLETE);
+        test_error(error, "Unable to set user event to complete");
```
bashbaug: I'm not following why the user event needs to be set to "complete" before it is released and destroyed.  Can you say more?

AhmedAmraniAkdi: Hi Ben, this is from the clReleaseEvent specification:

"Developers should be careful when releasing their last reference count on events created by [clCreateUserEvent](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateUserEvent) that have not yet been set to status of [CL_COMPLETE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_COMPLETE) or an error."

I understand from this that an user event needs to be set to CL_COMPLETE before releasing its last reference.

bashbaug: The spec text referenced above is purely a caution that if you enqueue a command dependent on the user event, and you have not set the event to "complete" before it is released and destroyed, then it is impossible to satisfy the dependency and the command will never execute.  It does not describe a requirement to set the event to "complete" before releasing and destroying it.

In this specific case, we're enqueueing a command that appears to be dependent on the user event, but the command is ill-formed and we fail the test if the command does not return an error.  Therefore, there are no commands dependent on the user event, and it should be fine to release and destroy the event before setting it to "complete".

AhmedAmraniAkdi: I understand, thanks Ben, we will remove this change from the pull request.

PR #2513, opened 2025-09-03
Make subgroup tests run on devices without non uniform work groups

karolherbst:
This allows implementors to support `cl_khr_subgroup_ballot` and `cl_khr_subgroup_non_uniform_vote` without requiring non uniform workgroup support.

### DIFF
--- a/test_conformance/subgroups/test_subgroup_ballot.cpp
+++ b/test_conformance/subgroups/test_subgroup_ballot.cpp
@@ -896,8 +896,21 @@ REGISTER_TEST(subgroup_functions_ballot)
         return TEST_SKIPPED_ITSELF;
     }
 
-    constexpr size_t global_work_size = 170;
-    constexpr size_t local_work_size = 64;
+    int error = 0;
+
+    // Non-uniform work-groups are an optional feature from 3.0 onward.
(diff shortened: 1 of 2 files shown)


test_conformance/subgroups/test_subgroup_ballot.cpp
```diff
@@ -896,8 +896,21 @@ REGISTER_TEST(subgroup_functions_ballot)
...
+    if (get_device_cl_version(device) >= Version(3, 0))
+    {
+        error = clGetDeviceInfo(
+            device, CL_DEVICE_NON_UNIFORM_WORK_GROUP_SUPPORT, sizeof(cl_bool),
+            &device_supports_non_uniform_wg, nullptr);
+        test_error(error, "clGetDeviceInfo failed");
+    }
+
+    const size_t global_work_size = device_supports_non_uniform_wg ? 170 : 192;
+    const size_t local_work_size = 64;
```
bashbaug, line 912: Probably not worth worrying about, but we could consider something like the following if we wanted to make this a bit more resilient to future changes:

```suggestion
    const size_t local_work_size = 64;
    const size_t global_work_size = device_supports_non_uniform_wg ? 170 : local_work_size * 3;
```

Issue #2501, opened 2025-08-25
Enqueueing a kernel with a reqd_work_group_size with a NULL local_work_size


bashbaug:
We don't seem to have any coverage for the case where a kernel has a required work-group size:

```c++
__attribute__((reqd_work_group_size(2, 3, 4)))
kernel void test(...) { ... }
```

And, the ND-range when enqueueing the kernel passes `NULL` as the local work-group size:

```c++
clEnqueueNDRangeKernel(
    queue, kernel, work_dim, global_work_offset, global_work_size, NULL,
    num_events_in_wait_list, event_wait_list, event);
```

In this case, I believe the implementation should use the required work-group size as the work-group size, though the spec isn't as clear about this as I'd like.  We may want to clarify the spec, also.

The non-uniform work-groups test suite has some provisions for testing a required work-group size and a local work-group size equal to `NULL`, but this specific combination does not appear to be tested.

The API test for "kernel_required_group_size" executes a kernel, but passes an explicit local work-group size.

The API test for "kernel_attributes" only compiles kernels and never executes them.

Testing notes:
* We should test cases where the passed-in `work_dim` is 1, 2, and 3.
    * If the passed-in `work_dim` is 2 then the third component (Z) of the required work-group size must be 1.
    * If the passed-in `work_dim` is 1 then both the second and third components (Y and Z) of the required work-group size must be 1.
* We should test both the command-queue path (`clEnqueueNDRangeKernel`) and the command-buffer path (`clCommandNDRangeKernelKHR`).
* We should also test that the required work-group size is returned by `clGetKernelSuggestedLocalWorkSizeKHR`, for implementations that support `cl_khr_suggested_local_work_size`.

bashbaug:
Discussed in the August 26th teleconference.  We agree that this should be valid, and hence we should have testing for this scenario.

Added mobica-backlog.

PR #2514, opened 2025-09-08
add test for a NULL local work size and a required work-group size

bashbaug:
see #2501 

This tests the following scenarios:

1. Execute a kernel with a required work-group size, passing `NULL` as the local work size.
2. Query the suggested work-group size for a kernel with a required work-group size.

### DIFF
--- a/test_conformance/api/test_kernel_attributes.cpp
+++ b/test_conformance/api/test_kernel_attributes.cpp
@@ -440,6 +440,7 @@ REGISTER_TEST(null_required_work_group_size)
                          device_max_work_item_sizes[1],
                          device_max_work_item_sizes[2],
                          device_max_work_group_size);
+                continue;
             }
 
             const cl_int zero = 0;


test_conformance/api/test_kernel_attributes.cpp
```diff
@@ -336,3 +336,160 @@ REGISTER_TEST(kernel_attributes)
     }
     return success ? TEST_PASS : TEST_FAIL;
 }
+
+REGISTER_TEST(null_required_work_group_size)
+{
+    cl_int error = CL_SUCCESS;
+
+    bool checkSuggestedLocalWorkSize = false;
```
EwanC: Bit of a nitpick, but think the surrounding variable naming style tends to snake_case (same comment for a couple other variables later on). Also not sure if we need a specific variable for this, or can  just check  `clGetKernelSuggestedLocalWorkSizeKHR` isn't `nullptr` later on.

bashbaug: Hmm, good point.  I'll just check if the function pointer is `nullptr`.

I'll see what I can do about snake_case vs. camelCase also.  The CTS seems to use both, but snake_case is definitely more pervasive in this file.  Thanks!

EwanC (dismissed):

PR #2515, opened 2025-09-09
Corrected bufferreadwriterect test where provided pitches are zero for read and write ops

shajder:
Related to #2279


rjodinchr (approved):

bashbaug:
Discussed in the July 21st teleconference.  OK to merge offline with a Qualcomm approval.

karolherbst:
Hit a regression with this: https://github.com/KhronosGroup/OpenCL-CTS/issues/2765

bashbaug:
Should we revert this PR until the regression is addressed?

karolherbst:
> Should we revert this PR until the regression is addressed?

maybe.. but I've also noticed other implementations (pocl and intel) to not throw an error, so I'm not 100% sure it's not an issue in the implementation, but it sure looks like an issue in the CTS so far.

PR #2519, opened 2025-09-11
Fix path to spv file in test_compiler.

hvdijk:
In #1220, test_compiler was changed to load write_kernel.spv* from a file using a relative path
test_conformance/compiler/spirv_bin/write_kernel.spv*. This works when test_compiler is run from the build directory, but does not work when it is run from the test_conformance directory, which is how run_conformance.py expects to run it. Adjust the path accordingly.

### DIFF
--- a/test_conformance/compiler/test_unload_platform_compiler.cpp
+++ b/test_conformance/compiler/test_unload_platform_compiler.cpp
@@ -32,8 +32,8 @@ const std::string slash = "\\";
 #else
 const std::string slash = "/";
 #endif
-std::string compilerSpvBinaries = "test_conformance" + slash + "compiler"
-    + slash + "spirv_bin" + slash + "write_kernel.spv";
+std::string compilerSpvBinaries =
+    "compiler" + slash + "spirv_bin" + slash + "write_kernel.spv";
 
 const std::string spvExt = ".spv";
 


test_conformance/compiler/test_unload_platform_compiler.cpp
```diff
@@ -32,8 +32,8 @@ const std::string slash = "\\";
 #else
 const std::string slash = "/";
 #endif
-std::string compilerSpvBinaries = "test_conformance" + slash + "compiler"
-    + slash + "spirv_bin" + slash + "write_kernel.spv";
+std::string compilerSpvBinaries =
```
ahesham-arm, line 35: While you're right that this will work if tests are invoked using the Python script, it will break if run from the root of the build directory, or really any other directory.

I think the better solution here is to define the path relative to the test binary, regardless of where the working directory is. I can create a patch to fix this.

hvdijk: That would be fine too, or do it the way that `test_spirv_new` does it where it accepts a command-line argument that specifies the location, but the CSV files set the option so that it works from the `test_conformance` directory.

ahesham-arm: I have created #2523 which uses your suggestion. Please have a look and let me know if we should close this PR in favour of it.

I plan on cleaning up the code and have this functionality be part of the harness so we can share it across `spirv_new` and `compiler`, but this address the immediate issue.

hvdijk: I can't test it right now to verify it works as intended, but at a quick glance that looks very nice, thanks.

hvdijk:
Closing in favour of #2523

PR #2521, opened 2025-09-12
Remove hard dependency on SPIR-V Headers

joselopeqti:
In https://github.com/KhronosGroup/OpenCL-CTS/pull/2409 a hard dependency on SPIR-V Headers was added. This change removes the hard dependency on them and adds a TEST_FAIL in test_conformance/api/test_spirv_queries.cpp when the SPIR-V Headers are not provided but the `cl_khr_spirv_queries` extension is available.

### DIFF
--- a/test_conformance/api/test_spirv_queries.cpp
+++ b/test_conformance/api/test_spirv_queries.cpp
@@ -770,7 +770,8 @@ REGISTER_TEST(spirv_query_dependencies)
 
 #else
 
-REGISTER_TEST(spirv_query_extension_check_absence_of_files) {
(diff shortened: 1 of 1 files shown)


joselopeqti:
I agree that it is not that bad to clone them. But in my opinion, we should try to keep the base dependencies to a minimum and just add them to the tests that require them, like it is done for other tests.

Couldn't the SPIR-V test just be added as its own test?

bashbaug:
Could you have a look at https://github.com/KhronosGroup/OpenCL-CTS/pull/2527, which moves in-depth testing (and hence the SPIR-V header dependency) from `test_api` to `test_spirv_new`, and see if this works for you?

PR #2523, opened 2025-09-15
Fix file paths in test_compiler

ahesham-arm:
Add a command line argument to override the SPIR-V file paths if needed, similar to `test_spirv_new`. Set the default path to the one required by `run_conformance.py`, which assumes the current working directory to be `<build_dir>/test_conformance`.

### DIFF
--- a/test_conformance/compiler/main.cpp
+++ b/test_conformance/compiler/main.cpp
@@ -13,10 +13,62 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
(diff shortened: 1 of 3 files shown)


bashbaug (approved):
LGTM

I'd eventually like to move this functionality to the harness, but it's fine to do that as a separate step, and it may be easier to let some other refactoring land first (e.g. #2527).

coldav:
Could we get this fix could merged please?

PR #2524, opened 2025-09-15
Unified SVM Memcpy and check_for_common_memory_type

### LINKED ISSUE #2517: Unified SVM Memcpy and check_for_common_memory_type
Found while extending the unified SVM tests for system SVM scenarios, and discussed in the September 9th memory subgroup: The current unified SVM memcpy tests include the function `check_for_common_memory_type`:

https://github.com/KhronosGroup/OpenCL-CTS/blob/b63ef2d8f6b30b00c0b0da4b534ecc2d03aa6b72/test_conformance/SVM/test_unified_svm_mem_cpy.cpp#L213

The function appears to be filtering combinations of SVM that are tested, based on the SVM capabilities.

https://github.com/KhronosGroup/OpenCL-CTS/blob/b63ef2d8f6b30b00c0b0da4b534ecc2d03aa6b72/test_conformance/SVM/test_unified_svm_mem_cpy.cpp#L166-L175

Is the filtering correct?  Specifically, which SVM capabilities should be required for a memcpy source, and which capabilities should be required for a memcpy destination?

From the discussion here https://github.com/KhronosGroup/OpenCL-Docs/pull/1282#discussion_r1913880258, perhaps we only need to check whether the memcpy source supports reading from the device, and whether the memcpy destination supports writing on the device?

CC: @JohnKesapidesARM

JohnKesapidesARM:
* Refactor check_for_common_memory_type to only check for device access.
* rename check_for_common_memory_type to caps_compatibility_check
* Update test_unified_svm_mem_cpy and test_unified_svm_mem_fill.

fixes #2517

### DIFF
--- a/test_conformance/SVM/test_unified_svm_mem_cpy.cpp
+++ b/test_conformance/SVM/test_unified_svm_mem_cpy.cpp
@@ -167,7 +167,7 @@ struct UnifiedSVMOPs : UnifiedSVMBase
         {
             for (cl_uint dst_ti = 0; dst_ti < max_ti; dst_ti++)
             {
-                if (check_for_common_memory_type(src_ti, dst_ti))
+                if (caps_compatibility_check(src_ti, dst_ti))
                 {
                     log_info(
                         "   testing clEnqueueSVMMemcpy() SVM type %u -> SVM "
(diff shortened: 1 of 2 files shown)


JohnKesapidesARM:
@bashbaug  fix for common memory type

Issue #1364, opened 2021-12-01
missing coverage for floating-point types in work group scan and reduce tests


bashbaug:
I'm creating this issue to track possible improvements to the work group scan and reduce tests:

1. There is a lot of code duplication between tests.  Each operation is in its own test file but there is a LOT of commonality between them.  Can we at least condense to a single file for all of the reduction tests and all of the scan tests?
2. Within each file there is a lot of code duplicate for each type.  We should be able to condense to a single function.
3. There appears to be missing test coverage for the `float`, `double`, and `half` types.

bashbaug:
Changing title.  The code duplication described by (1) and (2) was fixed by #1401 but the missing test coverage for the floating-point types (3) still remains.

See also https://github.com/KhronosGroup/OpenCL-Docs/issues/833.

shajder:
@bashbaug can we close this issue or is there anything more to add here? Thanks

PR #2525, opened 2025-09-15
Added floating point types support for work_group_reduce tests

shajder:
Related to #1364

### DIFF
--- a/test_conformance/workgroups/test_wg_scan_reduce.cpp
+++ b/test_conformance/workgroups/test_wg_scan_reduce.cpp
@@ -21,6 +21,10 @@
 
 #include "testBase.h"
 
+cl_half_rounding_mode gHalfRoundingMode = CL_HALF_RTE;
+constexpr cl_half g_half_min = 0xfbff;
+constexpr cl_half g_half_max = 0x7bff;
+
 static std::string make_kernel_string(const std::string &type,
                                       const std::string &kernelName,
                                       const std::string &func)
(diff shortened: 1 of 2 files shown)


test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -64,41 +68,88 @@ template <> struct TestTypeInfo<cl_ulong>
...
+
 template <typename T> struct Add
 {
     using Type = T;
     static constexpr const char *opName = "add";
     static constexpr T identityValue = 0;
-    static T combine(T a, T b) { return a + b; }
+    static T combine(T a, T b)
+    {
+        if (std::is_same_v<T, cl_half>)
```
bashbaug: I personally would prefer to use template specialization on the `cl_half` type here vs. checks on `std::is_same_v`, though this is a matter of personal preference.

If we're going to use `std::is_same_v` though, can we at least use `if constexpr`?

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -112,29 +163,133 @@ template <typename C> struct Reduce
...
-                             i + j);
-                    return -1;
+                    if (fabs(result - outptr[i + j]) > max_err)
+                    {
+                        log_info("%s_%s: Error at %zu\n", testName, testOpName,
+                                 i + j);
+                        return -1;
+                    }
+                }
+                else if (std::is_same_v<Type, cl_half>)
```
bashbaug: Does it make sense to refactor this into a `check_result` function or similar, especially if it could be reused for the work-group scans?

If not, this is another place where we should use `if constexpr`.

```suggestion
                else if constexpr (std::is_same_v<Type, cl_half>)
```

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -112,29 +163,133 @@ template <typename C> struct Reduce
...
+                double min_range = -999.0;
+                for (size_t j = 0; j < max_wg_size; j++)
+                    ref_vals[j] = get_random_float(min_range, max_range, d);
+
+                for (size_t i = 0; i < (size_t)n_elems; i += max_wg_size)
+                {
+                    size_t work_group_size = std::min(max_wg_size, n_elems - i);
+                    memcpy(&inptr[i], ref_vals.data(),
+                           sizeof(Type) * work_group_size);
+                }
```
bashbaug, line 266: I believe this means we'll be using the same input data for all work-groups.  While that's not inherently bad, it does mean we're doing redundant testing by running multiple work-groups.

Some options to consider:
1. Do nothing, these tests don't run for very long so a little redundant testing is fine.
2. Test with different random data for each work-group instead, but verify against the worst-case `max_err`.
3. Only test a single work-group.

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -112,29 +163,133 @@ template <typename C> struct Reduce
...
+        if constexpr (std::is_floating_point_v<
+                          Type> || std::is_same_v<Type, cl_half>)
+        {
+            std::vector<Type> ref_vals(max_wg_size, 0);
+            if (std::is_same_v<Type, cl_half>)
+            {
+                // to prevent overflow limit range of randomization
+                float max_range = 99.0;
+                float min_range = -99.0;
+                // generate reference values for one work group
```
bashbaug, line 224: Do we want at least a couple of work-groups to test special values, things like NaNs, infinities, maybe +/-0, etc?

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: `cl_half` specialization added

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: > Does it make sense to refactor this into a check_result...

Make sense, `check_result` added

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: > 2. Test with different random data for each work-group instead, but verify against the worst-case max_err.

Please keep in mind that the `max_err` may be different for other random sets. That's why I propagated the same data across other workgroups to avoid such computations during the verification phase. Since it is not a performance load and the code is consistent with integral types perhaps it may stay as is.

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: Sure, do you think we can go for new test just to keep special cases separated ? I presume it may be more readable.

bashbaug: I realized today while looking at an unrelated issue that we never define any specific NaN / infinity / -0 vs. +0 behavior for these work-group scan and reduction functions.  Since we don't define any specific behavior for these special values, I'm fine testing with this limited range, for now at least.

test_conformance/workgroups/test_wg_scan_reduce.cpp
bashbaug: I'm leaning towards "do nothing", but I don't think this is consistent with the integral types.  As far as I can tell:

* This code, for the floating-point types, generates a set of random data for the "max work-group size", then stamps out the same data for each work-group, until each of the "n_elems" has a value.
* The code below, for the integral types, generates a random value for each of the "n_elems", which means that each work-group will have a different set of random data.

Is this correct?

shajder: yes, your description is correct. In other words kernels and CL calls stays the same for integral and floating-point types but generation and verification part is not consistent indeed.

bashbaug:
Related to #2529 and #2533.

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -122,19 +207,100 @@ template <typename C> struct Reduce
         }
         return 0;
     }
+
+    static void generate_reference_values(Type *inptr, size_t n_elems,
+                                          size_t max_wg_size, Type &max_err = 0)
```
bashbaug: Slight nitpick: This is generating **input** values (and the max error for those input values), not **reference** values.  Might be best to rename this function to avoid confusion.

bashbaug (dismissed):

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: Done

bashbaug (approved):

PR #1401, opened 2022-03-14
refactor work group scan and reduction tests

bashbaug:
This PR refactors the work-group scan and reduction tests to reduce duplication and make it easier to add support for other scan and reduction operators or types in the future.

PR #2529, opened 2025-09-19
Added floating point types support for work_group_scan_inclusive tests

shajder:
Related to #1364


bashbaug:
Related to https://github.com/KhronosGroup/OpenCL-CTS/pull/2533.

rjodinchr (commented):
I think that PR could be refactored a bit to avoid too much code duplication

shajder:
> I think that PR could be refactored a bit to avoid too much code duplication

@rjodinchr thanks for the feedback. Please note the [comment](https://github.com/KhronosGroup/OpenCL-CTS/issues/1364#issuecomment-1242360025) from the parent issue, duplication issues were resolved in #1401. That being said, I can confirm there is room for unification, however I would prefer to add a separate PR once all related branches are merged. How does that sound to you?

rjodinchr:
> > I think that PR could be refactored a bit to avoid too much code duplication
> 
> @rjodinchr thanks for the feedback. Please note the [comment](https://github.com/KhronosGroup/OpenCL-CTS/issues/1364#issuecomment-1242360025) from the parent issue, duplication issues were resolved in #1401. That being said, I can confirm there is room for unification, however I would prefer to add a separate PR once all related branches are merged. How does that sound to you?

I think the fp16/fp64 sections that are duplicated 3 times (for Min/Max/Add), should be refactored in a single function (with a template maybe). I don't see a reason to wait for that.

shajder:
> I think the fp16/fp64 sections that are duplicated 3 times (for Min/Max/Add), should be refactored in a single function (with a template maybe). I don't see a reason to wait for that.

As far as I understand you refer to the code which was the subject of a request in #2525; please see that thread for the context and how it was resolved.

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -620,6 +716,37 @@ REGISTER_TEST_VERSION(work_group_scan_inclusive_add, Version(2, 0))
...
+
+    result |= run_test<ScanInclusive<Add<cl_float>>>(device, context, queue,
+                                                     num_elements);
+
+    if (is_extension_available(device, "cl_khr_fp64"))
+    {
+        result |= run_test<ScanInclusive<Add<cl_double>>>(device, context,
+                                                          queue, num_elements);
+    }
+
```
rjodinchr, line 752: duplicated

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -640,6 +767,37 @@ REGISTER_TEST_VERSION(work_group_scan_inclusive_max, Version(2, 0))
...
+    result |= run_test<ScanInclusive<Max<cl_float>>>(device, context, queue,
+                                                     num_elements);
+
+    if (is_extension_available(device, "cl_khr_fp64"))
+    {
+        result |= run_test<ScanInclusive<Max<cl_double>>>(device, context,
+                                                          queue, num_elements);
+    }
+
```
rjodinchr: duplicated

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -660,6 +818,37 @@ REGISTER_TEST_VERSION(work_group_scan_inclusive_min, Version(2, 0))
...
+    result |= run_test<ScanInclusive<Min<cl_float>>>(device, context, queue,
+                                                     num_elements);
+
+    if (is_extension_available(device, "cl_khr_fp64"))
+    {
+        result |= run_test<ScanInclusive<Min<cl_double>>>(device, context,
+                                                          queue, num_elements);
+    }
+
```
rjodinchr: duplicated

rjodinchr (commented):
I refer to the 3 comments "duplicated" added below.

Those sections look the same with different templates for Min/Max/Add. I don't see why we should duplicate that code 3 times here while we could just have it in one place.

This is independent of previous/future PR, if I'm not mistaken.

shajder:
> This is independent of previous/future PR, if I'm not mistaken.

corrected

rjodinchr (approved):
Thank you for the change.
LGTM.

I think in a future PR we could also refactor inclusive and reduce, right? they are also the same and could also be part of the same function if we add another template argument to it?

shajder:
> Thank you for the change. LGTM.
> 
> I think in a future PR we could also refactor inclusive and reduce, right? they are also the same and could also be part of the same function if we add another template argument to it?

Okay

PR #2532, opened 2025-09-23
Use a more genuine kernel for the command buffer device enqueue negative test

AhmedAmraniAkdi:
The current one has no arguments and does not enqueue any work.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
@@ -398,16 +398,17 @@ struct CommandNDRangeKernelWithKernelEnqueueCall : public BasicCommandBufferTest
 
         const char* kernel_str =
             R"(
-__kernel void enqueue_call_func() {
-  }
-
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp
```diff
@@ -398,16 +398,17 @@ struct CommandNDRangeKernelWithKernelEnqueueCall : public BasicCommandBufferTest
...
+            __kernel void enqueue_call_func(__global int* out_mem) {
+            out_mem[get_global_id(0)] = 0x1234;
+              }
+            
+            __kernel void enqueue_call_kernel(__global int* out_mem) {
+            queue_t def_q = get_default_queue();
+            ndrange_t ndrange = ndrange_1D(1);
+            enqueue_kernel(def_q, CLK_ENQUEUE_FLAGS_WAIT_KERNEL, ndrange,
+                               ^{enqueue_call_func(out_mem);});
+              }
```
EwanC, line 410: Minor: Since we're touching the kernels we could improve the indentation, right now the body of the function isn't indented and the closing braces don't align to anything.

PR #2533, opened 2025-09-24
Added floating point types support for work_group_scan_exclusive tests

shajder:
Related to #1364


test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -22,8 +22,8 @@
 #include "testBase.h"
 
 cl_half_rounding_mode gHalfRoundingMode = CL_HALF_RTE;
-constexpr cl_half g_half_min = 0xfbff;
-constexpr cl_half g_half_max = 0x7bff;
+constexpr cl_half g_half_min = 0xFC00;
+constexpr cl_half g_half_max = 0x7C00;
```
rjodinchr, line 26: Why do do not keep the maximum finite values?

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -458,10 +483,79 @@ template <typename C> struct ScanExclusive
 
     static void generate_input_values(Type *inptr, size_t n_elems,
                                       size_t max_wg_size,
-                                      const Type *max_err = nullptr)
+                                      Type *const max_err = nullptr)
```
rjodinchr, line 486: I understand why we want to add a const, but no need to remove the one already there, right?

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -458,10 +483,79 @@ template <typename C> struct ScanExclusive
...
+            if constexpr (std::is_same_v<Type, cl_half>)
+            {
+                // to prevent overflow limit range of randomization
+                float max_range = 99.0;
+                float min_range = -99.0;
+                // generate reference values for one work group
+                for (size_t j = 0; j < max_wg_size; j++)
+                    ref_vals[j] = cl_half_from_float(
+                        get_random_float(min_range, max_range, d),
+                        gHalfRoundingMode);
```
rjodinchr, line 502: Why do we need to select in that specific range? we do not have any range for integers.

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -458,10 +483,79 @@ template <typename C> struct ScanExclusive
...
+
+                // populate reference data across all work groups
+                for (size_t i = 0; i < (size_t)n_elems; i += max_wg_size)
+                {
+                    size_t wg_size = std::min(max_wg_size, n_elems - i);
+                    memcpy(&inptr[i], ref_vals.data(), sizeof(Type) * wg_size);
+                }
```
rjodinchr, line 509: Why do we duplicate the data? we generate random values for integers for each workgroup it seems.

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: Integer add wraps deterministically and is order-independent, so reference and device match even on overflow, no range needed. FP isn't: scan order is unspecified, and once a partial sum saturates to positive or negative INF the value is lost and INF (or INF-INF -> NaN) can appear at different positions on device vs reference, causing false mismatches. The range just keeps every partial sum finite.

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: The two consts aren't the same: the old const Type * protects the pointee, but max_err is now written to (it's an out-param), so that const has to go; the new Type *const only locks the pointer, not the data.

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: sorry, I've implemented that sep 2025, I don't recall the details right now and I will have to dive deeper. I will get back later.

shajder: As per spec for [work_group_scan_exclusive](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#work-group-functions):

> The exclusive scan operation takes a binary associative operator op with an identity I and n (where n is the size of the work-group) elements [a0, a1, …​ an-1] and returns [I, a0, (a0 op a1), …​ (a0 op a1 op …​ op an-2)]. If op = add, the identity I is 0. If op = min, the identity I is INT_MAX, UINT_MAX, LONG_MAX, ULONG_MAX, for int, uint, long, ulong types and is +INF for floating-point types. Similarly if op = max, the identity I is INT_MIN, 0, LONG_MIN, 0 and -INF. For the example above, the exclusive scan add operation on the ordered set [3 1 7 0 4 1 6 3] would return [0 3 4 11 11 15 16 22].

In other words identity value (`I`) for `work_group_scan_exclusive` (`+INF` for min, `-INF` for max, on `half`/`float`/`double`) will be put at the first element of output buffer. Thus, we need to take this into account at the host side during verification of the results. Keep in mind identity value has no significant meaning for the reduce and inclusive-scan variants.

test_conformance/workgroups/test_wg_scan_reduce.cpp
shajder: Yes, it's intentional: `ref_vals` holds one work group's data replicated to every group. Integer `add` is exact, so each group can use independent random values. FP add isn't, we bound the association/rounding error, and that per-position `max_err` is computed once from a single group; it's only valid for every group if they all hold the same data, which is why we replicate instead of randomizing per group. Strictly this is only needed for `add`; `min`/`max` are exact and could use independent data, but splitting them out costs extra code and readability for no real gain.

rjodinchr (approved):

kpet:

test_conformance/workgroups/test_wg_scan_reduce.cpp
```diff
@@ -458,10 +483,79 @@ template <typename C> struct ScanExclusive
...
+                {
+                    // compute maximal summation error
+                    float s = std::abs(cl_half_to_float(ref_vals[0]));
+                    for (size_t i = 1; i < (size_t)n_elems; i++)
+                    {
+                        max_err[i] = cl_half_from_float(
+                            std::abs((max_wg_size - 1) * CL_HALF_EPSILON * s),
+                            gHalfRoundingMode);
+                        s += std::abs(cl_half_to_float(ref_vals[i]));
+                    }
```
rjodinchr, line 521: `ref_vals` contains `max_wg_size` elements. Iterating over `n_elems` seems like a bug. I'm seeing a crash with this code. I've not completely identified it yet.

shajder: Confirmed, the same problem refers to 3 other locations. Do you want me to create PR with the fix?

rjodinchr: Yes please 🙏

shajder: [PR](https://github.com/KhronosGroup/OpenCL-CTS/pull/2726) added, thanks

PR #2534, opened 2025-09-25
One last file path fix for test_compiler.

hvdijk:
Although run_conformance.py runs from test_conformance, it will change the current working directory before launching each individual test. Adjust the path accordingly.


bashbaug (dismissed):

bashbaug:
I think I resolved the conflict from #2535 properly, but if @hvdijk (or @dneto0) could verify that'd be super - thanks!

PR #2535, opened 2025-09-26
Add helper to wrap result of std::filesystem::path::u8string()

dneto0:
In C++17 the return type of std::filesystem::path::u8string() is std::string, but in C++20 the return type changed to std::u8string.

Add a helper to copy a std::u8string to a std::string, to be used when a std::string is required. This fixes the build for C++20.


dneto0:
This fixes the build for me when std::u8string is available.

It conflicts with #2534.  If 2534 lands first, then delete the obvious conflict.

test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp
```diff
@@ -229,9 +229,8 @@ struct MutableDispatchWorkGroups : public BasicMutableCommandBufferTest
             CL_STRUCTURE_TYPE_MUTABLE_DISPATCH_CONFIG_KHR
         };
         const void *configs[1] = { &dispatch_config };
-        error =
-            clUpdateMutableCommandsKHR(single_command_buffer, num_configs,
-                                       config_types, configs);
+        error = clUpdateMutableCommandsKHR(single_command_buffer, num_configs,
+                                           config_types, configs);
```
rjodinchr: This is just changing the format, I don't think that should be part of that PR.

rjodinchr (approved):
LGTM once `test_conformance/extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/mutable_command_work_groups.cpp` is reverted

bashbaug (approved):
Thank you for fixing this!

3a2l:
I understand that this changed under C++20 but I don't understand why the code is being compiled with that standard when we have
```
set(CMAKE_CXX_STANDARD 17)
```
in the CMake file. Thanks for fixing this regardless.

rjodinchr:
We had that issue in the Google infrastructure where we are not using CMake to build the CTS.

PR #2537, opened 2025-10-06
Added coverage of using multiple kernels created from the same source program with different builds

### LINKED ISSUE #2164: No explicit coverage of using multiple kernels created from the same source program with different builds
OpenCL programs can be built multiple times, with different options that can affect the behaviour of the kernels. We do not, AFAICT,  test this explicitly in the CTS.

Suggested test:
- Create two buffers
- Create a program from source (the source contains a single kernel that writes a value defined by a preprocessor macro into a buffer)
- Build the program, setting the macro to one value
- Create a kernel from the program
- Enqueue the kernel, with buffer 1 as a parameter
- Build the program again, setting the macro to a different value
- Create a kernel from the program
- Enqueue the kernel, with buffer 2 as a parameter
- Call clFinish on the queue
- Check that both buffers contain different values and the right value for each kernel

Also consider adding a variant using clCompileProgram/clLinkProgram.

shajder:
Fixes #2164 according to issue description

### DIFF
--- a/test_conformance/compiler/test_compile.cpp
+++ b/test_conformance/compiler/test_compile.cpp
@@ -159,6 +159,14 @@ const char *link_static_function_access = // use with compile_static_function
     "extern int foo(int, int);\n"
     "int access_foo() { int blah = foo(3, 4); return blah + 5; }\n";
 
+const char *multi_build_test_kernel = R"(
+__kernel void test_kernel(__global int *dst)
+{
+    int tid = get_global_id(0);
+    dst[tid] = BUILD_OPT_VAL;
+}
+)";
+
 static int test_large_single_compile(cl_context context, cl_device_id deviceID,
                                      unsigned int numLines)
 {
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_compile.cpp
```diff
@@ -3933,3 +3941,96 @@ REGISTER_TEST(compile_and_link_status_options_log)
 
     return 0;
 }
+
+REGISTER_TEST(multiple_build_program)
+{
+    int error = CL_SUCCESS;
```
EwanC:
```suggestion
    cl_int error = CL_SUCCESS;
```

test_conformance/compiler/test_compile.cpp
```diff
@@ -3933,3 +3941,96 @@ REGISTER_TEST(compile_and_link_status_options_log)
...
+    size_t threads[1] = { 1 };
```
EwanC: I don't think `threads` needs to be an array type for a single element. Also the term `threads` is a bit overloaded with host side threading, would prefer a name like `total_work_items` or `global_size`, etc

test_conformance/compiler/test_compile.cpp
```diff
@@ -3933,3 +3941,96 @@ REGISTER_TEST(compile_and_link_status_options_log)
...
+    clProgramWrapper program = nullptr;
```
EwanC: We could just define `program` on line 3953, rather than initializing it to nullptr first

test_conformance/compiler/test_compile.cpp
```diff
@@ -3933,3 +3941,96 @@ REGISTER_TEST(compile_and_link_status_options_log)
...
+    error = clEnqueueReadBuffer(queue, out_stream_1, true, 0,
+                                sizeof(cl_int) * threads[0], test_values.data(),
+                                0, NULL, NULL);
+    test_error(error, "clEnqueueReadBuffer failed");
+
+    for (size_t i = 0; i < test_values.size(); i++)
+    {
+        if (test_values[i] != 2)
+        {
+            log_error("Unexpected test value %d for kernel0 at pos %zu.\n",
```
EwanC:
```suggestion
            log_error("Unexpected test value %d for kernel1 at pos %zu.\n",
```

test_conformance/compiler/test_compile.cpp
```diff
@@ -3933,3 +3941,95 @@ REGISTER_TEST(compile_and_link_status_options_log)
 
     return 0;
 }
+
+REGISTER_TEST(multiple_build_program)
+{
+    cl_int error = CL_SUCCESS;
+    size_t threads = 0;
+
+    threads = std::min(num_elements, 512);
```
EwanC:
```suggestion
    size_t threads = std::min(num_elements, 512);
```

shajder: done

test_conformance/compiler/test_compile.cpp
```diff
@@ -3933,3 +3941,93 @@ REGISTER_TEST(compile_and_link_status_options_log)
 
     return 0;
 }
+
+REGISTER_TEST(multiple_build_program)
+{
+    cl_int error = CL_SUCCESS;
+    size_t threads = std::min(num_elements, 512);
```
bashbaug: Nitpick: "threads" is a massively overloaded term, and what's the significance of 512?

Could this just use whatever is passed as `num_elements`?  Is testing a single value sufficient (vs. 512)?
If we're going to rename the variable (vs. reusing `num_elements`), can it be "const"?

shajder: Done

bashbaug (dismissed):

test_conformance/compiler/test_compile.cpp
```diff
@@ -3933,3 +3941,93 @@ REGISTER_TEST(compile_and_link_status_options_log)
 
     return 0;
 }
+
+REGISTER_TEST(multiple_build_program)
+{
+    cl_int error = CL_SUCCESS;
+    const size_t num_threads = num_elements;
```
bashbaug, line 3956: Actually, I do have a minor review comment still - can we get rid of `num_threads` completely now and just use `num_elements`?

Also, did you see the prior review comment about skipping this test in non-online compiler mode?

shajder: > Actually, I do have a minor review comment still - can we get rid of `num_threads` completely now and just use `num_elements`?

please notice `num_threads` is uses as argument for `clEnqueueNDRangeKernel` which expects `size_t`* so I will have to cast `int`* to `size_t`*, right?

bashbaug: Ah, gotcha, works for me!

bashbaug (approved):

PR #2539, opened 2025-10-07
Add AHardwareBufferWrapper

ahesham-arm:
Add a wrapper around AHB for proper resource deallocation and refactor existing tests to use the wrapper.

### DIFF
--- a/test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
+++ b/test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
@@ -134,19 +134,15 @@ REGISTER_TEST(test_images)
 
                 CHECK_AHARDWARE_BUFFER_SUPPORT(aHardwareBufferDesc, format);
 
-                AHardwareBuffer *aHardwareBuffer = nullptr;
-                int ahb_result = AHardwareBuffer_allocate(&aHardwareBufferDesc,
-                                                          &aHardwareBuffer);
-                if (ahb_result != 0)
-                {
-                    log_error("AHardwareBuffer_allocate failed with code %d\n",
-                              ahb_result);
-                    return TEST_FAIL;
-                }
+                AHardwareBufferWrapper aHardwareBuffer(&aHardwareBufferDesc);
+                log_info(
+                    "Testing %s\n",
+                    ahardwareBufferFormatToString(format.aHardwareBufferFormat)
+                        .c_str());
 
-                const cl_mem_properties props[] = {
+                cl_mem_properties props[] = {
                     CL_EXTERNAL_MEMORY_HANDLE_ANDROID_HARDWARE_BUFFER_KHR,
-                    reinterpret_cast<cl_mem_properties>(aHardwareBuffer), 0
+                    aHardwareBuffer.get_props(), 0
                 };
 
                 cl_mem image = clCreateImageWithProperties(
@@ -181,8 +177,6 @@ REGISTER_TEST(test_images)
 
                 test_error(clReleaseMemObject(image),
                            "Failed to release image");
-                AHardwareBuffer_release(aHardwareBuffer);
-                aHardwareBuffer = nullptr;
             }
         }
     }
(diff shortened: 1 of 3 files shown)


test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -1883,28 +1835,178 @@ REGISTER_TEST(test_blob)
...
+                  "CL_INVALID_IMAGE_DESCRIPTOR but returned %s: "
+                  "CL_INVALID_IMAGE_DESCRIPTOR if image_desc is not NULL when "
+                  "using cl_khr_external_memory_android_hardware_buffer.\n",
+                  IGetErrorString(err));
+        return TEST_FAIL;
+    }
+
+    mem = clCreateImageWithProperties(context, props_buffer, CL_MEM_READ_WRITE,
+                                      nullptr, nullptr, nullptr, &err);
+    if (CL_IMAGE_FORMAT_NOT_SUPPORTED != err)
```
joselopeqti: One question, should this be or not be allowed? 

I believe the spec is not clear for this case. It is clear when creating a buffer, that it requires the AHardwareBuffer to be `AHARDWAREBUFFER_FORMAT_BLOB`. But when creating images, it says that it returns `CL_IMAGE_FORMAT_NOT_SUPPORTED`:
> if properties includes an AHardwareBuffer external memory handle and the AHardwareBuffer format is not supported.

So, is it really `AHARDWARE_BUFFER_FORMAT_BLOB` an unsupported image format? Because for example, blob could be interpreted as an image of height 1 and width buffer_size, with a cl_image_format of `CR_R, CL_UNSIGNED_INT8`.

kpet: I don't think having an implicit mapping of `AHARDWARE_BUFFER_FORMAT_BLOB` to a specific image format makes sense. An AHB with `AHARDWARE_BUFFER_FORMAT_BLOB` format could hold anything and this mapping is bound to be non-sensical in many cases.

Having said that I agree with you that there is potentially room for clarifying or changing the specification. Two possible changes that I see:
1. We explicitly say that creating an image from an AHB with `AHARDWARE_BUFFER_FORMAT_BLOB` format is not supported. This would avoid any confusion in the future.
2. We add support for this which would likely mean requiring both a `cl_image_format` and `cl_image_descriptor` from the application. I can see this path being useful as a shorthand to creating a buffer from an AHB and then an image fro that buffer. On the flip side, it would mean a special case and more API surface.

joselopeqti: Ok, thank you for the clarification. I agree, in my opinion 1 would be the best, simpler and consistent with what the extension says for creating buffers. 2 would imply adding a special behavior just for this case.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -1883,28 +1835,178 @@ REGISTER_TEST(test_blob)
...
-        AHardwareBuffer_release(aHardwareBuffer);
-        aHardwareBuffer = nullptr;
+    }
+
+    return TEST_PASS;
+}
+
+// Confirm correct error codes are returned when passing invalid arguments
+// accompanied by passing a AHB in cl_mem_properties
+REGISTER_TEST(negative)
```
joselopeqti: This test could be moved to `test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb_negative.cpp`

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
ahesham-arm: OpenCL specification updated.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
ahesham-arm: Why?

joselopeqti: Other negative tests are there

ahesham-arm: Done. Had to fix the line endings, so I did that in a separate commit.

nikhiljnv:
Merging as discussed in memory subgroup on October 28th 2025.

PR #2542, opened 2025-10-13
Added an in-order queue variant for the semaphores_import_export_fd test

### LINKED ISSUE #2213: switch external_semaphores_import_export_fd to in-order queues
See discussion: https://github.com/KhronosGroup/OpenCL-CTS/pull/2110#discussion_r1887957529

We should switch this test to in-order queues (instead of out-of-order queues) so it can run on more devices.

shajder:
Fixes #2213 according to mentioned discussion

### DIFF
--- a/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
+++ b/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
@@ -114,35 +114,18 @@ REGISTER_TEST_VERSION(external_semaphores_import_export_fd, Version(1, 2))
         test_error(err,
                    "clGetDeviceInfo for CL_DEVICE_QUEUE_PROPERTIES failed");
 
-        if ((device_props & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE) == 0)
+        if ((device_props & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE) != 0)
         {
-            log_info(
-                "Queue property CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE not "
-                "supported. Skipping test.\n");
-            return TEST_SKIPPED_ITSELF;
-        }
-        // Create ooo queue
-        clCommandQueueWrapper test_queue = clCreateCommandQueue(
-            context, device, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &err);
-        test_error(err, "Could not create command queue");
+            // Create ooo queue
+            clCommandQueueWrapper test_queue = clCreateCommandQueue(
+                context, device, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &err);
+            test_error(err, "Could not create command queue");
 
-        cl_int status = doTest(device, context, test_queue);
-        if (status != TEST_PASS && status != TEST_SKIPPED_ITSELF)
-        {
-            total_status = TEST_FAIL;
-        }
-    }
-
-    // test external semaphore sync fd with in-order queue
-    {
-        clCommandQueueWrapper test_queue =
-            clCreateCommandQueue(context, device, 0, &err);
-        test_error(err, "Could not create command queue");
-
-        cl_int status = doTest(device, context, test_queue);
-        if (status != TEST_PASS && status != TEST_SKIPPED_ITSELF)
-        {
-            total_status = TEST_FAIL;
+            cl_int status = doTest(device, context, test_queue);
+            if (status != TEST_PASS && status != TEST_SKIPPED_ITSELF)
+            {
+                total_status = TEST_FAIL;
+            }
         }
     }
 


joselopeqti (approved):

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
bashbaug, line 1: This is OK as-is, but it'd be nice to do a bit of refactoring to improve maintainability and perhaps improve test execution speed:

1. The checks whether the required extensions are available should be moved out of the `doTest` function since they're invariant.  May as well update them to use the new `REQUIRE_EXTENSION` macro while we're at it.
2. Consider using the in-order queue from the harness vs. creating another new queue.  We'll want to pass in the test queue if we do this vs. creating it in `doTest`.
3. Especially if we do (2), consider moving the checks for out-of-order queue availability and creation of the out-of-order queue out of `doTest` also.

shajder: @bashbaug before I start refactoring there is one more concern I wanted to raise. I've noticed that the test uses only `cl_khr_external_semaphore_sync_fd`. Shouldn't we launch similar (the same?) test for `cl_khr_external_semaphore_opaque_fd` as well ?

bashbaug: > Shouldn't we launch similar (the same?) test for `cl_khr_external_semaphore_opaque_fd` as well ?

Probably... but let's not try to boil the ocean.  Here's my suggestion:

1. Keep the test specific to sync FD for now and get this PR cleaned up and merged.
2. Try to expand the test to opaque FDs as well.  Since they're both FDs, maybe this is fairly trivial?  Do this as a follow-on PR.
3. See if it makes sense to expand the test logic to other semaphore types as well (e.g. cl_khr_external_semaphore_win32), either in the same test or as a different test.  If it does make sense, do it as another follow-on PR,

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
```diff
@@ -142,16 +100,60 @@ template <bool inorder> cl_int doTest(cl_device_id device, cl_context context)
...
+                              sizeof(device_props), &device_props, NULL);
+        test_error(err,
+                   "clGetDeviceInfo for CL_DEVICE_QUEUE_PROPERTIES failed");
+
+        if ((device_props & CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE) == 0)
+        {
+            log_info(
+                "Queue property CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE not "
+                "supported. Skipping test.\n");
+            return TEST_SKIPPED_ITSELF;
```
bashbaug: We shouldn't return `TEST_SKIPPED_ITSELF` here because this will skip testing the in-order queue case.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
```diff
@@ -142,16 +100,60 @@ template <bool inorder> cl_int doTest(cl_device_id device, cl_context context)
...
+        cl_int status = doTest(device, context, test_queue);
+        if (status != TEST_PASS && status != TEST_SKIPPED_ITSELF)
+        {
+            total_status = TEST_FAIL;
+        }
     }
-    status = doTest<true>(device, context);
-    if (status != TEST_PASS && status != TEST_SKIPPED_ITSELF)
+
+    // test external semaphore sync fd with in-order queue
```
bashbaug: Do we need to test this case separately, or is it sufficient to just test with the in-order queue from the harness?

shajder: hm, maybe I missed your point, you wrote earlier:

> Consider using the in-order queue from the harness vs. creating another new queue. We'll want to pass in the test queue if we do this vs. creating it in doTest

thus, there are two similar in-order tests. Do you think it is exaggeration ?

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
shajder: > * Try to expand the test to opaque FDs as well.  Since they're both FDs, maybe this is fairly trivial?  Do this as a follow-on PR.
> * See if it makes sense to expand the test logic to other semaphore types as well (e.g. cl_khr_external_semaphore_win32), either in the same test or as a different test.  If it does make sense, do it as another follow-on PR,

@bashbaug I just created [draft-PR](https://github.com/KhronosGroup/OpenCL-CTS/pull/2568), to be frank I am not entirely sure will that work. I will appreciate your expertise. Thanks

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
bashbaug: Right, I think we only need to test with the queue from the harness, which is an in-order queue.  This will save needing to create another in-order queue, which should be fairly inexpensive, but why bother if we can just use the in-order queue from the harness?

This means that we only need to test the case below using the queue from the harness (currently on Line 149), and do not need to test this case separately that creates a new in-order queue.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
shajder: Corrected

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_sync_fd.cpp
shajder: Done

PR #2543, opened 2025-10-14
c11_atomics: Fix verification loop

gwawiork:
Looks like bug. @shajder Could you look at this fix? Thanks


shajder:
> Looks like bug. @shajder Could you look at this fix? Thanks

thanks @gwawiork

PR #2548, opened 2025-10-16
Fix release call in AHB image read test

xinjin01:
Ensure clEnqueueReleaseExternalMemObjectsKHR targets imported_image instead of the non-external opencl_image, matching the prior acquire call.


joselopeqti (approved):

paulfradgley (approved):
LGTM

PR #2550, opened 2025-10-20
Specify memory scope and memory order for the atomic operations in generic_address_space generic_atomics_variant generic_atomics_invariant

AhmedAmraniAkdi:
Use the explicit version of the atomic_load/store and atomic_fetch_add with memory order relaxed and memory scope workgroup to allow devices that only support the minimum CL_DEVICE_ATOMIC_MEMORY_CAPABILITIES which are (CL_DEVICE_ATOMIC_ORDER_RELAXED | CL_DEVICE_ATOMIC_SCOPE_WORK_GROUP) to run the tests.

The test should only require the relaxed ordering and memory scope workgroup anyway.

From the specificiation:

"The non-explicit atomic_store function requires support for OpenCL C 2.0, or OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device features"

"The non-explicit atomic_load function requires support for OpenCL C 2.0 or OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device features."

"The non-explicit atomic_fetch_key functions require support for OpenCL C 2.0, or OpenCL C 3.0 or newer and both the __opencl_c_atomic_order_seq_cst and __opencl_c_atomic_scope_device features."

### DIFF
--- a/test_conformance/generic_address_space/atomic_tests.cpp
+++ b/test_conformance/generic_address_space/atomic_tests.cpp
@@ -35,7 +35,7 @@ kernel void testKernel(global atomic_int* globalPtr, local atomic_int* localPtr)
     int wgid = get_group_id(0);
     int wgsize = get_local_size(0);
 
-    if (tid == 0) atomic_store(localPtr, 0);
+    if (tid == 0) atomic_store_explicit(localPtr, 0, memory_order_relaxed, memory_scope_work_group);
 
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
Merging as discussed in the October 28th teleconference.

PR #2552, opened 2025-10-20
Remove spir half vloada_half tests as this builtin does not exist in OpenCL

AhmedAmraniAkdi:
The test was using vloada_half which does not exist for scalars.

Also removed the files test.vloada_half_*.* from half.zip.

For test.vloada_half3_global, changed the OpenCL kernel to use vload_half instead of vloada_half.

Build failures will return a proper failure now, before, the test was passing in this case.

More Info: https://github.com/KhronosGroup/OpenCL-Docs/issues/648

### DIFF
--- a/test_conformance/spir/main.cpp
+++ b/test_conformance/spir/main.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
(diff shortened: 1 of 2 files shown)


bashbaug:
Merging as discussed in the October 21st teleconference.

PR #2553, opened 2025-10-21
image_from_small_buffer_negative expecting incorrect error code

paulfradgley:
CL_INVALID_MEM_OBJECT is not a valid error return from clCreateImage.

Result should either be CL_INVALID_IMAGE_SIZE or CL_INVALID_IMAGE_FORMAT_DESCRIPTOR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,12 @@ if(USE_CL_EXPERIMENTAL)
   add_definitions(-DCL_EXPERIMENTAL)
 endif(USE_CL_EXPERIMENTAL)
 
+option(SANITIZER_THREAD "Build with the thread sanitiser" OFF)
+if (SANITIZER_THREAD)
+    add_compile_options(-fsanitize=thread)
+    add_link_options(-fsanitize=thread)
+endif (SANITIZER_THREAD)
+
 #-----------------------------------------------------------
 # Default Configurable Test Set
 #-----------------------------------------------------------


paulfradgley:
[CL_INVALID_IMAGE_FORMAT_DESCRIPTOR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_IMAGE_FORMAT_DESCRIPTOR)
if image_format is NULL
if values specified in image_format are not valid
if properties includes an AHardwareBuffer external memory handle and image_format is not NULL
**if an image is created from a buffer and the row pitch, or slice pitch, if the [cl_ext_image_from_buffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#cl_ext_image_from_buffer) extension is supported, or base address alignment do not follow the rules described for creating an image from a buffer**

[CL_INVALID_IMAGE_DESCRIPTOR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_IMAGE_DESCRIPTOR)
if properties includes an AHardwareBuffer external memory handle and image_desc is not NULL

[CL_INVALID_IMAGE_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_IMAGE_SIZE)
if the image dimensions specified in image_desc are not valid or exceed the maximum image dimensions described in the [Device Queries](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#device-queries-table) table for all devices in context
**if the [cl_ext_image_from_buffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#cl_ext_image_from_buffer) extension is supported and an image is created from a buffer and the buffer passed in mem_object is too small to be used as a data store for the image, e.g. if its size is smaller than the value returned for [CL_IMAGE_REQUIREMENTS_SIZE_EXT](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_IMAGE_REQUIREMENTS_SIZE_EXT) for the parameters used to create the image**

test_conformance/images/kernel_read_write/test_cl_ext_image_from_buffer.cpp
```diff
@@ -629,7 +629,7 @@ int image_from_small_buffer_negative(cl_device_id device, cl_context context,
 
                 clCreateImage(context, flag, &format, &image_desc, nullptr,
                               &err);
-                test_failure_error(err, CL_INVALID_MEM_OBJECT,
+                test_failure_error(err, CL_INVALID_IMAGE_FORMAT_DESCRIPTOR,
```
bashbaug: Should we accept `CL_IMAGE_SIZE` here also, or fix the spec to disambiguate which error code should be returned?

FWIW, if I understand this scenario correctly and I had to pick, I'd probably return `CL_IMAGE_SIZE` due to:

> an image is created from a buffer and the buffer passed in mem_object is too small

PR #2554, opened 2025-10-21
Added a test for semaphore payloads that are ignored

### LINKED ISSUE #2384: add a test for semaphore payloads that are ignored
The spec currently says for the _sema_payload_list_ that may be passed to [clEnqueueWaitSemaphoresKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueWaitSemaphoresKHR) and [clEnqueueSignalSemaphoresKHR](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueSignalSemaphoresKHR):

> sema_payload_list points to the list of values of type cl_semaphore_payload_khr containing valid semaphore payload values to wait on. This can be set to NULL or will be ignored when all semaphores in the list of sema_objects are of type [CL_SEMAPHORE_TYPE_BINARY_KHR](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_API.html#CL_SEMAPHORE_TYPE_BINARY_KHR).

Even if this wording changes slightly - see https://github.com/KhronosGroup/OpenCL-Docs/pull/1362#discussion_r2070804827 - it looks like we're currently always passing NULL as our semaphore payload list.  We should add some targeted testing to ensure that a non-NULL semaphore payload list is properly accepted and that it is ignored.

shajder:
Fixes #2384 according to issue description

### DIFF
--- a/test_conformance/extensions/cl_khr_semaphore/test_semaphores_payload.cpp
+++ b/test_conformance/extensions/cl_khr_semaphore/test_semaphores_payload.cpp
@@ -20,9 +20,11 @@ namespace {
 
 struct PayloadSemaphore : public SemaphoreTestBase
 {
+    clSemaphoreWrapper sema_sec;
+
     PayloadSemaphore(cl_device_id device, cl_context context,
                      cl_command_queue queue, cl_int nelems)
-        : SemaphoreTestBase(device, context, queue, nelems)
+        : SemaphoreTestBase(device, context, queue, nelems), sema_sec(this)
     {}
 
     cl_int Run() override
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_semaphore/test_semaphores_payload.cpp
```diff
@@ -0,0 +1,83 @@
...
+        : SemaphoreTestBase(device, context, queue, nelems)
+    {}
+
+    cl_int Run() override
+    {
+        cl_int err = CL_SUCCESS;
+        // Create ooo queue
+        clCommandQueueWrapper queue = clCreateCommandQueue(
+            context, device, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &err);
+        test_error(err, "Could not create command queue");
```
bashbaug: Let's not use an out-of-order queue here.  It will restrict which devices this test will run on - or at least, it should, the checks for out-of-order queue support are missing.  The events ordering the signal and the wait are missing, too.)

Note, we don't need to create an in-order queue to run on, and we can use the in-order queue passed in from the framework.

It's a little silly to signal and wait on the semaphore from the same in-order queue, but I believe it's valid.

shajder: Corrected

test_conformance/extensions/cl_khr_semaphore/test_semaphores_payload.cpp
```diff
@@ -0,0 +1,79 @@
...
+            static_cast<cl_semaphore_properties_khr>(
+                CL_SEMAPHORE_TYPE_BINARY_KHR),
+            0
+        };
+        semaphore =
+            clCreateSemaphoreWithPropertiesKHR(context, sema_props, &err);
+        test_error(err, "Could not create semaphore");
+
+
+        clSemaphoreWrapper sema_sec = nullptr;
```
joselopeqti: Could you move `sema_sec` outside of the run function, so it is a member of the struct?
Also, `sema_sec` needs to be initialized in the constructor to get the reference to the struct, like this:
```
    PayloadSemaphore(cl_device_id device, cl_context context,
                     cl_command_queue queue, cl_int nelems)
        : SemaphoreTestBase(device, context, queue, nelems),
        sema_sec(this)
    {}
```

test_conformance/extensions/cl_khr_semaphore/test_semaphores_payload.cpp
```diff
@@ -0,0 +1,79 @@
...
+        cl_semaphore_payload_khr payload_list[] = { 1, 2 };
+        cl_semaphore_khr semaphores[2] = { semaphore, sema_sec };
+
+        // Signal semaphore
+        err = clEnqueueSignalSemaphoresKHR(queue, 2, semaphores, payload_list,
+                                           0, nullptr, nullptr);
+        test_error(err, "Could not signal semaphore");
+
+        // Wait semaphore
+        err = clEnqueueWaitSemaphoresKHR(queue, 2, semaphores, payload_list, 0,
```
joselopeqti: Can we use different values for the payload in the wait than the ones used in the signal? This would ensure that implementations ignore the payload

test_conformance/extensions/cl_khr_semaphore/test_semaphores_payload.cpp
shajder: Done

test_conformance/extensions/cl_khr_semaphore/test_semaphores_payload.cpp
shajder: Done

PR #2556, opened 2025-10-24
Simplify initialization of validation data in buffers buffer_fill

joselopeqti:
Instead of using three different `for` loops to fill the validation data, sets the whole buffer to 0 and only writes to section that should be filled.

### DIFF
--- a/test_conformance/buffers/test_buffer_fill.cpp
+++ b/test_conformance/buffers/test_buffer_fill.cpp
@@ -1,6 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
-// 
+//
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


test_conformance/buffers/test_buffer_fill.cpp
```diff
@@ -921,17 +921,15 @@ REGISTER_TEST(buffer_fill_int)
         log_info( "Testing random fill from offset %d for %d elements: \n", (int)offset_elements, (int)fill_elements );
 
         for ( i = 0; i < 5; i++ ){
+            volatile size_t j = 0;
```
bashbaug, line 1057: Why is `volatile` needed here?

EwanC: If we're changing the `j` declaration, i'd prefer to define `j` in each loops instead of using a variable that gets reused across multiple loops, i.e prefer `for(size_t j = ...`  That's more readable in my opinion.

PR #2559, opened 2025-10-28
Added test to verify process of building a kernel with long name

### LINKED ISSUE #2422: Test the maximum kernel name and SPIR-V entry point name length
- The limit for SPIR-V seems to be the universal limit for the size of a literal string: 65536
- I could not find a limit for OpenCL C after a quick search

shajder:
Closes #2422 according to issue description

### DIFF
--- a/test_conformance/compiler/test_build_helpers.cpp
+++ b/test_conformance/compiler/test_build_helpers.cpp
@@ -117,13 +117,6 @@ __kernel void sample_test_D(__global float *src, __global int *dst)
 #endif
 )";
 
-const char *sample_name_size_test_kernel = R"(
-    __kernel void %s(int src, __global int *dst)
-    {
-        dst[0]=src;
-    }
-)";
-
 REGISTER_TEST(load_program_source)
 {
     int error;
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1009,3 +1017,127 @@ REGISTER_TEST(get_program_build_info)
...
+        clKernelWrapper kernel =
+            clCreateKernel(linkedProgram, name.c_str(), &error);
+        test_error(error, "clCreateKernel after link failed");
+    }
+#endif
+
+    return TEST_PASS;
+}
+
+REGISTER_TEST(kernel_name_size)
```
bashbaug, line 1109: This is another test we should probably skip if we're not in online compiler mode (gCompilationMode == kOnline), since it unconditionally calls `clCreateProgramWithSource`.

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1009,3 +1017,127 @@ REGISTER_TEST(get_program_build_info)
...
+{
+    cl_int status = TEST_PASS;
+    for (cl_uint len = 32; len < 2048; len *= 2)
```
bashbaug: Check: do you want to run the 2K case, or stop at 1K?  Either way, it might be clearer to use `<=` here.

```suggestion
    for (cl_uint len = 32; len <= 2048; len *= 2)
```

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1009,3 +1017,127 @@ REGISTER_TEST(get_program_build_info)
...
+    {
+        status = test_kernel_name_len(context, device, len);
```
bashbaug: Nitpick: We can move the declaration here (and remove the declaration above):

```suggestion
        cl_int status = test_kernel_name_len(context, device, len);
```

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1009,3 +1017,127 @@ REGISTER_TEST(get_program_build_info)
...
+        if (name != std::string(kernel_name.data()))
+        {
+            log_error("Kernel name mismatch! expected=%s got=%s\n",
+                      name.c_str(), kernel_name.data());
+            return TEST_FAIL;
+        }
+    }
+
+#if defined(CL_VERSION_1_2)
+    {
```
bashbaug: The CTS should always be built with the latest headers, so we shouldn't need to check for this define.  Do you intend to check whether the device supports OpenCL 1.2 here?

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -116,6 +117,13 @@ __kernel void sample_test_D(__global float *src, __global int *dst)
 #endif
 )";
 
+const char *sample_name_size_test_kernel = R"(
```
bashbaug: Consider declaring this as a local variable in the test function itself - saves scrolling to find it.  Or, if you choose to keep it as a global variable, consider declaring it "static".

bashbaug (dismissed):

test_conformance/compiler/test_build_helpers.cpp
shajder: Thanks! Now that you mentioned the compilation mode, I realized we should also cover the SPIR-V compilation mode as well. The infrastructure for this is already implemented through `create_single_kernel_helper` and the `--compilation-mode` flag, so I’ve rearranged part of the test to use that path.

test_conformance/compiler/test_build_helpers.cpp

test_conformance/compiler/test_build_helpers.cpp

test_conformance/compiler/test_build_helpers.cpp

test_conformance/compiler/test_build_helpers.cpp

test_conformance/compiler/test_build_helpers.cpp
bashbaug: This is missing a runtime check now for whether the device supports OpenCL 1.2, but I'd be REALLY surprised to see an OpenCL 1.0 or 1.1 device running CTS anymore, so I'm just going to resolve this comment and we can cross the pre-1.2 bridge if we come to it.

PR #2560, opened 2025-10-28
Add ThreadSanitizer option

gorazd-sumkovski-arm:
Add the `SANITIZER_THREAD` CMake option which will add the `-fsanitize=thread` flag to the compiler and linker when enabled.


bashbaug (approved):
Merging as discussed in the November 4th teleconference.

PR #2561, opened 2025-10-28
Add multi device and negative tests for cl_ext_buffer_device_address

gtrebuchet-arm:
The tests checks that invalid parameters for clSetKernelArgDevicePointerEXT are reported successfully and ensure that a kernel can access a buffer from their respective device address on each device in a multi device context.

### DIFF
--- a/test_conformance/extensions/cl_ext_buffer_device_address/buffer_device_address.cpp
+++ b/test_conformance/extensions/cl_ext_buffer_device_address/buffer_device_address.cpp
@@ -14,7 +14,10 @@
 //
 
 #include "harness/typeWrappers.h"
+#include "harness/extensionHelpers.h"
 #include <cinttypes>
+#include <vector>
+#include <string>
 
 #define BUF_SIZE 1024
 #define BUF_SIZE_STR "1024"
@@ -421,6 +424,8 @@ private:
 
 REGISTER_TEST(private_address)
 {
+    REQUIRE_EXTENSION("cl_ext_buffer_device_address");
+
     BufferDeviceAddressTest test_fixture = BufferDeviceAddressTest(
         device, context, queue, CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT);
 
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_ext_buffer_device_address/buffer_device_address.cpp
```diff
@@ -435,3 +440,161 @@ REGISTER_TEST(private_address)
...
+        ctx, props, CL_MEM_READ_WRITE, 16, nullptr, &error);
+    std::vector<cl_mem_device_address_ext> addresses(numDevices);
+    error =
+        clGetMemObjectInfo(buffer, CL_MEM_DEVICE_ADDRESS_EXT,
+                           sizeof(cl_mem_device_address_ext) * addresses.size(),
+                           addresses.data(), nullptr);
+    test_error_ret(error, "clGetMemObjectInfo failed\n", TEST_FAIL);
+    for (cl_mem_device_address_ext a : addresses)
+    {
+        if (a == 0) test_fail("Unexpected null cl_mem_device_address_ext\n");
```
pjaaskel: Is it not allowed to have 0 as a buffer address? Of course this is theoretical, but is it explicitly not allowed?

gtrebuchet-arm: You are right, I don't see anything preventing that in the spec.
I'd argue thats a useful check anyway. That seems so wrong that I don't expect any sensible implementation to do that.

lakshmih:
Qualcomm is reviewing this and will post comments. Thanks

test_conformance/extensions/cl_ext_buffer_device_address/buffer_device_address.cpp
pjaaskel: I'd just remove the check since you will check against the kernel-queried values later anyhow?

gtrebuchet-arm: Yea, realized after answering, I agree. I'll fix it.

test_conformance/extensions/cl_ext_buffer_device_address/buffer_device_address.cpp
```diff
@@ -435,3 +440,157 @@ REGISTER_TEST(private_address)
...
+    {
+        queues[i] = clCreateCommandQueue(ctx, devices[i], 0, &error);
+        test_error_ret(error, "Unable to create command queue\n", TEST_FAIL);
+    }
+    static std::string source = R"(
+        void kernel test_device_address(
+            global ulong* ptr,
+            global ulong* ptr2)
+        {
+            *ptr = as_ulong(ptr2);
```
lakshmih: as_ulong cannot accept a pointer as argument so this kernel might need to be rewritten.

gtrebuchet-arm: Thanks for pointing that out. Replaced with a pattern.

test_conformance/extensions/cl_ext_buffer_device_address/buffer_device_address.cpp
```diff
@@ -435,3 +440,172 @@ REGISTER_TEST(private_address)
...
+    test_error(error, "Unable to create test kernel");
+
+    error = clSetKernelArgDevicePointerEXT(nullptr, 0, 0);
+    test_failure_error_ret(
+        error, CL_INVALID_KERNEL,
+        "clSetKernelArgDevicePointerEXT should return CL_INVALID_KERNEL when: "
+        "\"kernel is not a valid kernel object\"",
+        TEST_FAIL);
+
+    error = clSetKernelArgDevicePointerEXT(kernel, 1, 0x15465);
```
lakshmih, line 596: As per the spec "The device pointer can only be used for arguments that are declared to be a pointer to global memory allocated with clCreateBufferWithProperties with the CL_MEM_DEVICE_PRIVATE_ADDRESS_EXT property" 
So this is indeed an error condition but the spec does not mention the error to be returned in this case. This needs to be discussed and clarified in the spec.

kpet: https://github.com/KhronosGroup/OpenCL-Docs/pull/1492

franz:
LGTM (once that spec clarification is worked out)

Issue #2563, opened 2025-10-31
'-cl-uniform-work-group-size' is not tested anywhere


rjodinchr:
While that option might be complicated to test, we should at least make sure that the compiler does not return an error when the option is passed.

rjodinchr:
Fixed by #2564

PR #2564, opened 2025-10-31
Better support for -cl-uniform-work-group-size

rjodinchr:
Add support for build options available only after a certain version:
- -cl-uniform-work-group-size after 2.0
- -cl-no-subgroup-ifp after 2.1

Add specific test for cl-uniform-work-group-size
- Check that test can be executed when work group size is uniform.
- Check that test returns the proper error code when work group size is not uniform.

Ref #2563

### DIFF
--- a/test_conformance/compiler/test_build_options.cpp
+++ b/test_conformance/compiler/test_build_options.cpp
@@ -45,13 +45,7 @@ const char *include_test_kernel[] = {
 "\n"
 "}\n" };
 
-const char *options_test_kernel[] = {
-    "__kernel void sample_test(__global float *src, __global int *dst)\n"
-    "{\n"
-    "    size_t tid = get_global_id(0);\n"
-    "    dst[tid] = (int)src[tid];\n"
-    "}\n"
-};
+const char *options_test_kernel[] = { "__kernel void sample_test() {}\n" };
 
 std::array optimization_options{
     std::pair{ "-cl-single-precision-constant", Version(1, 0) },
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_build_options.cpp
```diff
@@ -93,31 +90,42 @@ REGISTER_TEST(options_build_optimizations)
     int error;
     cl_build_status status;
 
-    for(size_t i = 0; i < sizeof(optimization_options) / (sizeof(char*)); i++) {
+    Version version = get_device_cl_version(device);
 
+    for (size_t i = 0; i < sizeof(optimization_options) / (sizeof(char *)); i++)
```
lakshmih: for loop @ line 95 computes size incorrectly now that optimization_options is an array of std::pair. Makes test segfault:

lakshmih: Suggest switching to a modern std::array to avoid any similar issue in the future, so:

std::array<std::pair<const char *, Version>, 12> optimization_options = {{
    { "-cl-single-precision-constant", Version(1, 0) },
    { "-cl-denorms-are-zero", Version(1, 0) },
    { "-cl-opt-disable", Version(1, 0) },
    { "-cl-mad-enable", Version(1, 0) },
    { "-cl-no-signed-zeros", Version(1, 0) },
    { "-cl-unsafe-math-optimizations", Version(1, 0) },
    { "-cl-finite-math-only", Version(1, 0) },
    { "-cl-fast-relaxed-math", Version(1, 0) },
    { "-w", Version(1, 0) },
    { "-Werror", Version(1, 0) },
    { "-cl-uniform-work-group-size", Version(2, 0) },
    { "-cl-no-subgroup-ifp", Version(2, 1) },
}}; 
and for loop can be simplified to:

for (size_t i = 0; i < optimization_options.size(); i++)

rjodinchr: Done.
As we are using C++17, I use a more compact declaration of the array.

rjodinchr: Windows is not happy about the compact declaration...

rjodinchr: I was just missing the `#include <array>` 😅

test_conformance/compiler/test_build_options.cpp
```diff
@@ -42,27 +45,23 @@ const char *include_test_kernel[] = {
...
+    std::pair{ "-cl-denorms-are-zero", Version(1, 0) },
+    std::pair{ "-cl-opt-disable", Version(1, 0) },
+    std::pair{ "-cl-mad-enable", Version(1, 0) },
+    std::pair{ "-cl-no-signed-zeros", Version(1, 0) },
+    std::pair{ "-cl-unsafe-math-optimizations", Version(1, 0) },
+    std::pair{ "-cl-finite-math-only", Version(1, 0) },
+    std::pair{ "-cl-fast-relaxed-math", Version(1, 0) },
+    std::pair{ "-w", Version(1, 0) },
+    std::pair{ "-Werror", Version(1, 0) },
+    std::pair{ "-cl-uniform-work-group-size", Version(2, 0) },
```
ahesham-arm, line 61: This compiler option requires `-cl-std=CL2.0`. From the specification:
```
Note: This option is missing before version 2.0.
```

rjodinchr: Do you intend to provide a patch to fix it? Or should I take care of it?

ahesham-arm: I'm afraid I don't know enough about how these tests are structured to be able to write a good enough patch for it. If you can please do, that would be great, thank you!

rjodinchr: https://github.com/KhronosGroup/OpenCL-CTS/pull/2595

PR #2565, opened 2025-10-31
fix api/negative_create_command_queue

rjodinchr:
When trying to trigger CL_INVALID_VALUE by passing invalid properties, we are using `-1` right now. But properties is a bitfield. Thus `-1` represent everything.
On devices not supporting a valid
property (`CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE` example), `clCreateCommandQueue` can return the error code
`CL_INVALID_QUEUE_PROPERTIES`.


rjodinchr:
Closing in favour of a driver patch

PR #2568, opened 2025-11-03
Extended external semaphore import/export tests to cover more handle types

shajder:
Due to discussion from #2542

### DIFF
--- a/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_handle_types.cpp
+++ b/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_handle_types.cpp
@@ -69,7 +69,7 @@ cl_int doTest(cl_device_id device, cl_context context, cl_command_queue queue,
     cl_semaphore_properties_khr sema_2_props[] = {
         static_cast<cl_semaphore_properties_khr>(CL_SEMAPHORE_TYPE_KHR),
         static_cast<cl_semaphore_properties_khr>(CL_SEMAPHORE_TYPE_BINARY_KHR),
-        handle_type, static_cast<cl_semaphore_properties_khr>(handle), 0
+        handle_type, (cl_semaphore_properties_khr)handle, 0
     };
 
     cl_semaphore_khr sema_2 =


test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore_handle_types.cpp
```diff
@@ -0,0 +1,178 @@
...
+REGISTER_TEST_VERSION(external_semaphores_import_export_fd, Version(1, 2))
+{
+    REQUIRE_EXTENSION("cl_khr_external_semaphore");
+
+    std::vector<cl_semaphore_properties_khr> handle_types = {
+        CL_SEMAPHORE_HANDLE_SYNC_FD_KHR, CL_SEMAPHORE_HANDLE_OPAQUE_FD_KHR,
+#ifdef _WIN32
+        CL_SEMAPHORE_HANDLE_D3D12_FENCE_KHR,
+        CL_SEMAPHORE_HANDLE_OPAQUE_WIN32_KHR
+#endif
```
bashbaug, line 113: I don't think this will work out-of-the-box for these win32 handle types because they are NTHANDLE (void*) handle types, which are different than the fd (int) handle types expected by `doTest`.

Does this work for the OPAQUE_FD handle type, though?

If it does, let's get a PR merged with it first, then we can look to expand to the win32 handle types, if it makes sense.

shajder: > Does this work for the OPAQUE_FD handle type, though?

(FYI @nikhiljnv) sadly, none of the CTS tests related to OPAQUE_FD handles works for me - my nvidia driver claims to support `cl_khr_external_semaphore_opaque_fd` but `clCreateSemaphoreWithPropertiesKHR` fails with `CL_INVALID_PROPERTY` every now and then. 

That being said, I think this code should work. If you take a look at [this piece](https://github.com/KhronosGroup/OpenCL-CTS/blob/ba991c01523ea11a0ab7d27e43736bbe7e256f0d/test_conformance/extensions/cl_khr_semaphore/test_semaphores_negative_create.cpp#L469-L496) - logic behind is very similar

> they are NTHANDLE (void*) handle types, which are different than the fd (int) handle types expected by doTest

I will try to verify with windows machine, still it will have to work with `clGetSemaphoreHandleForTypeKHR`, right ? I will apply the fix from other CTS test to follow `vulkan_wrapper.cpp` scheme like below:

```
    HANDLE handle;
...
    vkGetSemaphoreWin32HandleKHR(m_device, &vkSemaphoreGetWin32HandleInfoKHR,
                                 &handle);
...
    err = clGetSemaphoreHandleForTypeKHR(sema_1, device, handle_type,
                                         sizeof(handle), &handle, &handle_size);

```

PR #2569, opened 2025-11-03
Add AHB lifetime test

ahesham-arm:
Add lifetime test for AHardwareBuffer in which for both CL Buffers and CL Images the following steps are taken
 - Create AHB
 - Create mem object from the AHB
 - Release the AHB
 - Read and write from and to the mem object
 - Verify the reads and write have happened sucessfully

The CL implementation should maintain a reference count to the AHB since the AHB must not be deallocated for the test to pass.

### DIFF
--- a/test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
+++ b/test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
@@ -96,6 +96,23 @@ static const char *diff_images_kernel_source = {
         })"
 };
 
+static const char *lifetime_kernel_source = {
+    R"(
+            __kernel void increment_buffer(global uchar* buffer)
+            {
+                int tid = get_global_id(0);
+                buffer[tid] ++;
+            }
+
+            __kernel void set_image_color(write_only image2d_t ahb_image, float4 set_color)
+            {
+                int tidX = get_global_id(0);
+                int tidY = get_global_id(1);
+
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -1857,3 +1874,289 @@ REGISTER_TEST(blob)
...
+        AHardwareBufferWrapper aHardwareBuffer(&aHardwareBufferDesc);
+
+        const cl_mem_properties props[] = {
+            CL_EXTERNAL_MEMORY_HANDLE_ANDROID_HARDWARE_BUFFER_KHR,
+            aHardwareBuffer.get_props(),
+            0,
+        };
+
+        imported_buffer = clCreateBufferWithProperties(
+            context, props, CL_MEM_READ_WRITE, buffer_size, nullptr, &err);
```
joselopeqti: This call to `clCreateBufferWithProperties` should return `CL_INVALID_BUFFER_SIZE` if you pass `buffer_size`.

> [CL_INVALID_BUFFER_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_BUFFER_SIZE)
if size is non-zero and properties includes an AHardwareBuffer external memory handle

ahesham-arm: You're right, updated to pass `0` instead.

PR #2576, opened 2025-11-11
fix the null required work group size test

### LINKED ISSUE #2575: potential out of spec behavior required in api reqd_work_group_size test
The test will try to enqueue a 2x1x1 (work_dim = 1) kernel launch with a `reqd_work_group_size(2,3,1)` kernel and I'm not sure that's something we want to be supported at all.

`reqd_work_group_size` main purpose is to allow runtimes to compile more optimized kernels, e.g. they could inline `get_local_size(...)` calls unconditionally. While one could argue, that calling `get_local_size(1)` is 1 for work_dim = 1, it does prevent optimizations like the above and therefore limits the purpose of `reqd_work_group_size` here. Therefore I don't think we want to test this at all and make it a strict error in the CTS, if it isn't already.

This also means that runtimes will have to treat `reqd_work_group_size` like `work_group_size_hint`, because runtimes _can't_ assume that the given work group size is indeed what's always being used.

I don't know if the spec strictly disallows this, but with "[CL_INVALID_WORK_GROUP_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_INVALID_WORK_GROUP_SIZE) if local_work_size is specified and does not match the required work-group size for kernel in the program source." one could argue that 2x1x1 does not match 2x3x1 even with `work_dim = 1`.

bashbaug:
A 3D required work-group size is always valid, but a 1D or 2D required work-group size is only valid when the work-group size in those dimensions is equal to one.

fixes #2575

### DIFF
--- a/test_conformance/api/test_kernel_attributes.cpp
+++ b/test_conformance/api/test_kernel_attributes.cpp
@@ -385,14 +385,14 @@ REGISTER_TEST(null_required_work_group_size)
 
     struct KernelAttribInfo
     {
-        std::string str;
-        cl_uint max_dim;
+        cl_int wgs[3];
+        cl_uint min_dim;
     };
 
     std::vector<KernelAttribInfo> attribs;
-    attribs.push_back({ "__attribute__((reqd_work_group_size(2,1,1)))", 1 });
-    attribs.push_back({ "__attribute__((reqd_work_group_size(2,3,1)))", 2 });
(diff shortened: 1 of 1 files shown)


karolherbst:
Thanks, it fixes the issue I'm running into! I just had to do a bit of bug fixing myself (returning the required size in `clGetKernelSuggestedLocalWorkSizeKHR`)

lakshmih:
@joselopeqti  is reviewing this.

PR #2577, opened 2025-11-12
Add tests for cloned kernel via clCloneKernel

MichaelRizkalla-arm:


test_conformance/api/test_clone_kernel.cpp
```diff
@@ -52,37 +53,52 @@ const char *clone_kernel_test_double[] =
...
-"    outbuf[6] = buf[0];\n"
-"}\n"
-"\n"
-"__kernel void buf_write_kernel(__global int* buf, int write_val)\n"
-"{\n"
-"    buf[0] = write_val;\n"
-"}\n"
-
- };
+const char* clone_kernel_test_kernel[] = {
```
EwanC, line 54: Can we make this a C++ raw string literal to make it more readable

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -223,16 +227,128 @@ int test_double_arg_clone(cl_device_id device, cl_context context,
...
-    if (abs(((cl_double*)pbufRes)[2] - d) > 0.0000001)
+    test_assert_error(abs(((cl_double*)pbufRes)[2] - d) <= 0.0000001,
+                      "clCloneKernel test failed.");
+
+    return TEST_PASS;
+}
+
+int test_args_enqueue_helper(cl_context context, cl_command_queue queue,
+                             cl_kernel* srcKernel, cl_int* value,
+                             cl_mem* bufOut)
```
EwanC: Don't think we need to `srcKernel`, `value`, and `bufOut` as pointers if their only usage is as inputs to the function that don't get checked for nullptr before being dereferenced.

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -295,9 +411,10 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
     // cl_mem
     clMemWrapper buf, bufOut;
 
-    char* pbuf = new char[BUF_SIZE];
-    char* pbufRes = new char[BUF_SIZE];
-    buf = clCreateBuffer(context, CL_MEM_READ_WRITE | CL_MEM_USE_HOST_PTR, BUF_SIZE, pbuf, &error);
+    std::unique_ptr<char[]> pbuf = std::make_unique<char[]>(BUF_SIZE);
+    std::unique_ptr<char[]> pbufRes = std::make_unique<char[]>(BUF_SIZE);
```
EwanC: Do we really need to use heap allocation here or can we just define these variables as arrays, and the variables scope will keep the stack memory alive for the same lifetime?

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -346,56 +463,462 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
...
+        error = test_double_arg_clone(device, context, queue, num_elements,
+                                      pbufRes.get(), bufOut);
+        test_error(error, "double arg clone test failed.");
+    }
+
+    return TEST_PASS;
+}
+
+int test_exec_enqueue_helper(cl_context context, cl_command_queue queue,
+                             BufPtr* pBuf, cl_int* svmPtr_Kernel,
```
EwanC: `pBuf` looks unused int this function

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -346,56 +463,462 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
...
+                             cl_kernel* srcKernel, cl_int* value)
```
EwanC: `srcKernel` and `value` don't need to be passed as pointers (same for function below)

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -346,56 +463,462 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
...
+                              || svmPtr_srcKernel_1 != NULL
+                              || svmPtr_cloneKernel_2 != NULL,
+                          "clSVMAlloc returned NULL");
+
+        // srcKernel, set args
+        if (test_svm_exec_info_helper(context, queue, pBuf, svmPtr_srcKernel,
+                                      &srcKernel, &intargs[0])
+            != TEST_PASS)
+        {
+            test_fail("test_svm_exec_info_helper failed for srcKernel.\n");
```
EwanC, line 633: are we leaking the SVM allocated memory when we hit the `test_fail` case in these `if` statements by returning early without calling `clSVMFree`

MichaelRizkalla-arm: I've added a struct that will clean-up the allocations.

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -324,78 +453,471 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
...
+    test_error(error, "clFinish failed");
+
+    error = clSetKernelArg(srcKernel, 0, sizeof(cl_int), &value);
+    test_error(error, "clSetKernelArg failed");
+    error = clSetKernelArgSVMPointer(srcKernel, 1, pBuf);
+    test_error(error, "clSetKernelArgSVMPointer failed");
+
+    error = clFinish(queue);
+    test_error(error, "clFinish failed");
+
```
EwanC: Do we need this `clFinish` call if we've already got one a few lines above before setting the arguments again?

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -324,78 +453,471 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
...
+{
+    cl_int error;
+
+    error = clSetKernelArgSVMPointer(srcKernel, 0, svmPtr_Kernel);
+    test_error(error, "clSetKernelArgSVMPointer failed");
+    error = clSetKernelArg(srcKernel, 1, sizeof(cl_int), &value);
+    test_error(error, "clSetKernelArg failed");
+
+    error = clFinish(queue);
+    test_error(error, "clFinish failed");
```
EwanC: why do we need a clFinish() call here? a comment would be useful if it is required.

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -324,78 +453,471 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
...
+
+        return TEST_PASS;
+    }
+    else
+    {
+        return TEST_SKIPPED_ITSELF;
+    }
+}
+
+REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
```
EwanC: Slight preference for splitting this into individual tests prefixed with `clone_kernel_*` then you run run them all by doing `./test_conformance/api/test_api clone_kernel*` or individually if you want do debug.

EwanC (dismissed):
I haven't had a chance to run this code, but LGTM with some minor comments

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -324,78 +453,471 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
...
+
+    error = clEnqueueSVMUnmap(queue, pBuf, 0, NULL, NULL);
+    test_error(error, "clEnqueueSVMUnmap failed");
+    error = clFinish(queue);
+    test_error(error, "clFinish failed");
+
+    error = clSetKernelArg(srcKernel, 0, sizeof(cl_int), &value);
+    test_error(error, "clSetKernelArg failed");
+    error = clSetKernelArgSVMPointer(srcKernel, 1, pBuf);
+    test_error(error, "clSetKernelArgSVMPointer failed");
```
bashbaug, line 574: This is called `test_svm_exec_info_helper` but I don't see any place where `clSetKernelExecInfo` is called.

Since we're passing `svmPtr_Kernel` indirectly, I think we need something like:


```suggestion
    error = clSetKernelArgSVMPointer(srcKernel, 1, pBuf);
    test_error(error, "clSetKernelArgSVMPointer failed");

    error = clSetKernelExecInfo(srcKernel, CL_KERNEL_EXEC_INFO_SVM_PTRS,
                                sizeof(svmPtr_Kernel), &svmPtr_Kernel);
    test_error(error, "clSetKernelExecInfo failed");
```

test_conformance/api/test_clone_kernel.cpp
```diff
@@ -324,78 +453,471 @@ REGISTER_TEST_VERSION(clone_kernel, Version(2, 1))
...
+        != TEST_PASS)
+    {
+        test_fail("test_exec_enqueue_helper failed.\n");
     }
 
-    if (((float*)pbufRes)[2] != farg)
+    return TEST_PASS;
+}
+
+template <size_t N> struct SVMCleanup
```
bashbaug: If we used the `clSVMWrapper` class more aggressively, could we get rid of `SVMCleanup`?

If `clSVMWrapper` is too rudimentary, could we enhance it?

bashbaug (changes requested):

bashbaug:
Removing "focused review" until the comments above are addressed.

test_conformance/api/test_clone_kernel.cpp
MichaelRizkalla-arm: It's redundant, I've removed it.

test_conformance/api/test_clone_kernel.cpp
MichaelRizkalla-arm: I removed it.

test_conformance/api/test_clone_kernel.cpp
MichaelRizkalla-arm: I've split the test into five different tests with prefix `clone_kernel_`.

test_conformance/api/test_clone_kernel.cpp
MichaelRizkalla-arm: Done.

test_conformance/api/test_clone_kernel.cpp
MichaelRizkalla-arm: I've updated `clSVMWrapper` to cover equality, and used it instead of `SVMCleanup`.

bashbaug (approved):

bashbaug:
Latest version LGTM - thanks!  @lakshmih is this one you'd like to review?

bashbaug:
Discussed in the May 26th teleconference.  OK to merge offline with one additional review (@lakshmih).

lakshmih (approved):

rjodinchr:
This commit broke some tests when running with `clvk` which is an OpenCL3.0 implementation with a OpenCL-C1.2 compiler.
Here is the fix for it: https://github.com/KhronosGroup/OpenCL-CTS/pull/2712
Please take a look, thank you

PR #2579, opened 2025-11-13
fixes a warning and a typo in the mutable dispatch test

bashbaug:
Fixes a warning in the mutable dispatch test with some compilers:

```
3>C:\git\OpenCL-CTS\test_conformance\extensions\cl_khr_command_buffer\cl_khr_command_buffer_mutable_dispatch\mutable_command_basic.h(82,16): warning C4805: '==': unsafe mix of type 'int' and type 'bool' in operation
```

Also fixes a misspelled variable name while we're at it.


EwanC (approved):
LGTM to address the immediate issue, but it's odd that `is_extension_available()` returns an `int` rather than a `bool` given the implementation returns a `bool` that gets casted.

PR #2583, opened 2025-11-18
Added test to verify reqd_work_group_size attribute with cl_khr_command_buffer extension

### LINKED ISSUE #2501: Enqueueing a kernel with a reqd_work_group_size with a NULL local_work_size

shajder:
Fixes #2501 according to issue description


test_conformance/extensions/cl_khr_command_buffer/command_buffer_kernel_attributes.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+        return CL_SUCCESS;
+    }
+
+    cl_int Run() override
+    {
+        cl_uint device_max_dim = 0;
+        cl_int error =
+            clGetDeviceInfo(device, CL_DEVICE_MAX_WORK_ITEM_DIMENSIONS,
+                            sizeof(device_max_dim), &device_max_dim, nullptr);
```
bashbaug, line 76: I'm OK with this as-is, but FWIW I don't quite understand the division of tasks between "Setup" and "Run".  Would it make more sense to move some parts of "Setup" - in particular creating the destination buffer - to "Run" here instead?  Or, should these device queries move out of "Run" and into "Setup" since they're invariant?

I personally would find the test easier to understand if it didn't override "Setup" at all and did everything in "Run" instead.

bashbaug (dismissed):

bashbaug (approved):
Re-approving after updating to resolve merge conflict.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_kernel_attributes.cpp
```diff
@@ -0,0 +1,234 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+#include "basic_command_buffer.h"
+
+#include <vector>
```
EwanC: I'd guess we can omit this and `vector.h` will be brought in transitively from `testHarness.h`

test_conformance/extensions/cl_khr_command_buffer/command_buffer_kernel_attributes.cpp
EwanC: I'd personally find having having `Run()` focus on the `attribs` loop the most readable layout. So move the info queries to `Setup()` and set member variables, and turn ` const std::string body_str ` into a static class member

EwanC (approved):
Couple of pedantic comments, but happy to merge as-is

test_conformance/extensions/cl_khr_command_buffer/command_buffer_kernel_attributes.cpp
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_kernel_attributes.cpp
shajder: Done

PR #2584, opened 2025-11-18
Registration for the profiling_timebase test has been changed: version 2.1 is now required.

### LINKED ISSUE #2582: checking the OpenCL version in the profiling_timebase test
In the profiling_timebase test case, there is a check of clGetDeviceAndHostTimer support using clGetPlatformInfoa.
The test is skipped if the OpenCL version on the platform is >= 3.0 and the result of clGetDeviceAndHostTimer is 0.
But clGetDeviceAndHostTimer is not available in OpenCL versions less than 2.1. In the case of OpenCL with versions lower than 2.1, there will be an attempt to use the test. Do I understand correctly that in the case of an OpenCL version less than 2.1, you need to skip the test case?

funch0za:
This fix is necessary to install the minimum version of OpenCL in the profiling_timebase test. This test uses clGetDeviceAndHostTimer, available only since version 2.1.
fix #2582

### DIFF
--- a/test_conformance/profiling/profiling_timebase.cpp
+++ b/test_conformance/profiling/profiling_timebase.cpp
@@ -18,7 +18,7 @@
 
 const char *kernelCode = "__kernel void kernel_empty(){}";
 
-REGISTER_TEST(profiling_timebase)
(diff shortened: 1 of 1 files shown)


bashbaug:
Merging as discussed over email.

PR #2592, opened 2025-12-04
Add AHB sub buffer test

ahesham-arm:
Add test for clCreateSubBuffer using a buffer created from a AHardwareBuffer


test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -50,6 +59,29 @@ ahb_image_size_table test_sizes[] = {
     { 64, 64 }, { 128, 128 }, { 256, 256 }, { 512, 512 }
 };
 
+uint32_t test_buffer_sizes[] = { 2, 8, 32, 128, 512, 2048, 16384, 65536 };
+
+ahb_usage_table test_buffer_usages[] = {
```
joselopeqti, line 78: I believe all these flags require `AHARDWAREBUFFER_USAGE_CPU_WRITE_OFTEN` as the test uses `AHardwareBuffer_lock` to write the initial data to the buffer

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -2160,3 +2192,128 @@ REGISTER_TEST(lifetime_image)
...
+                 "Skipping test.\n");
+        return TEST_SKIPPED_ITSELF;
+    }
+
+    AHardwareBuffer_Desc aHardwareBufferDesc = { 0 };
+    aHardwareBufferDesc.format = AHARDWAREBUFFER_FORMAT_BLOB;
+    for (auto usage : test_buffer_usages)
+    {
+        // Filter out usage flags that are not readable on device
+        if (!isAHBUsageReadable(usage.usageFlags))
```
joselopeqti: `isAHBUsageReadable` needs to be changed to check for more flags. Right now, it only checks for `AHARDWAREBUFFER_USAGE_GPU_SAMPLED_IMAGE` which is not present in any of the flags specified in `test_buffer_usages`, so the test is actually being skipped.
`AHARDWAREBUFFER_USAGE_GPU_DATA_BUFFER` and `AHARDWAREBUFFER_USAGE_SENSOR_DIRECT_DATA` should be added.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
ahesham-arm: Shouldn't `AHARDWAREBUFFER_USAGE_CPU_WRITE_RARELY` be a better fit here, if we're only locking once for the initial data?

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
ahesham-arm: You're right, I updated the helper function and added two more for checking host related flags. Updated the test code to use the new helpers.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
joselopeqti: Yes, you're right both `AHARDWAREBUFFER_USAGE_CPU_WRITE_RARELY ` and `AHARDWAREBUFFER_USAGE_CPU_READ_RARELY `

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -2160,3 +2210,129 @@ REGISTER_TEST(lifetime_image)
...
+                if (host_buffer[i] != host_sub_buffer[i])
+                {
+                    log_error(
+                        "At position i=%zu expected value %u but got %u\n", i,
+                        host_buffer[i], host_sub_buffer[i]);
+                    return TEST_FAIL;
+                }
+            }
+
+            err = clFinish(queue);
```
joselopeqti: This `clFinish` is not necessary as `clEnqueueReadBuffer` is blocking

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
ahesham-arm: Updated the flags to include `AHARDWAREBUFFER_USAGE_CPU_WRITE_RARELY` and changed the call to `AHardwareBuffer_lock` to use `AHARDWAREBUFFER_USAGE_CPU_WRITE_RARELY` as well as it is only used for initialising the buffer data.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
ahesham-arm: Removed.

PR #2593, opened 2025-12-05
Added tests validating clEnqueueSVMMemcpy when source and destination SVM ranges overlap

### LINKED ISSUE #2372: add a negative test for an overlapping clEnqueueSVMMemcpy
Found as part of the unified SVM discussions: We do not currently have a negative test for an overlapping clEnqueueSVMMemcpy.

We should test:

* A case where the end of the source memory partially overlaps the beginning of the destination memory.
* A case where the beginning of the source memory partially overlaps the end of the destination memory.
* A case where the source memory completely overlaps the destination memory.

Each of these cases should return `CL_MEM_COPY_OVERLAP`, as per the spec.

shajder:
Fixes #2372 according to issue description

### DIFF
--- a/test_conformance/SVM/test_negative_memcpy.cpp
+++ b/test_conformance/SVM/test_negative_memcpy.cpp
@@ -15,7 +15,7 @@
 //
 #include "common.h"
 
-REGISTER_TEST(svm_negative_dst_overlap_src)
+REGISTER_TEST(svm_negative_memcpy_overlap)
 {
     cl_int err = CL_SUCCESS;
 
@@ -25,7 +25,7 @@ REGISTER_TEST(svm_negative_dst_overlap_src)
 
     if (err != CL_SUCCESS || caps == 0)
     {
-        log_error("svm_negative_dst_overlap_src: SVM is not supported on this "
+        log_error("svm_negative_memcpy_overlap: SVM is not supported on this "
                   "device\n");
         return TEST_SKIP;
     }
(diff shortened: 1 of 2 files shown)


test_conformance/SVM/test_negative_memcpy.cpp
```diff
@@ -0,0 +1,180 @@
...
+                            &num_devices, 0);
+    test_error(err, "svm_negative_dst_overlap_src: create_cl_objects failed");
+
+    context = contextWrapper;
+    queue = queues[0];
+
+    const size_t dataSize = 32;
+
+    char *data_buf =
+        (char *)clSVMAlloc(contextWrapper, CL_MEM_READ_WRITE, dataSize * 2, 0);
```
bashbaug: Recommend using the `clSVMWrapper` here instead, which automatically handles freeing the SVM allocation.

test_conformance/SVM/test_negative_memcpy.cpp
```diff
@@ -0,0 +1,180 @@
...
+    if (!data_buf)
+    {
+        log_error(
+            "svm_negative_dst_overlap_src: invalid result for clSVMAlloc\n");
+        return TEST_FAIL;
+    }
```
bashbaug: Consider:

```suggestion
    test_assert_error(data_buf != nullptr, "clSVMAlloc failed");
```

bashbaug (commented):
Have you considered testing all three scenarios in one test instead?

This would let us reuse the same context / command queue / etc, even the same SVM allocation if we choose, which means the tests will run faster and the code will be shorter.

test_conformance/SVM/test_negative_memcpy.cpp
shajder: Done

test_conformance/SVM/test_negative_memcpy.cpp
shajder: Done

lakshmih (approved):

PR #2594, opened 2025-12-05
Tighten AHB buffer row pitch in test_cl_khr_external_memory_ahb test

xinjin01:
What was wrong:
enqueue_copy_buffer_to_image (and the related write/fill tests) mis-set imageInfo.rowPitch to width*height*pixelSize. Because get_image_size multiplies row pitch by height, this wrongly calculates the intended buffer size.

How it’s fixed: set rowPitch to the true per-line pitch (width*pixelSize) for all buffer-backed image cases so the calculated sizes match the actual data layout and stay within the expected memory footprint.


nikhiljnv (approved):
Approving based on Jose's approval.

nikhiljnv:
Merging as discussed in memory subgroup on December 9th 2025.

PR #2595, opened 2025-12-05
fix compiler/options_build_optimizations

rjodinchr:
We need to define -cl-std when compiling with option not in 1.0.

### DIFF
--- a/test_conformance/compiler/test_build_options.cpp
+++ b/test_conformance/compiler/test_build_options.cpp
@@ -13,8 +13,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#include "kernelHelpers.h"
 #include "testBase.h"
+#include "harness/kernelHelpers.h"
 #include "harness/os_helpers.h"
 #include "harness/testHarness.h"
 


ahesham-arm (approved):

svenvh (dismissed):

test_conformance/compiler/test_build_options.cpp
```diff
@@ -431,7 +434,7 @@ REGISTER_TEST(options_uniform_work_group_size)
     {
         return TEST_SKIPPED_ITSELF;
     }
-    const char *options = "-cl-uniform-work-group-size";
+    const char *options = "-cl-std=CL2.0 -cl-uniform-work-group-size";
```
bashbaug: Since OpenCL 3.0 devices do not necessarily need to support OpenCL C 2.0, I don't think it's generally safe to specify `-cl-std=CL2.0` unconditionally.  The safest thing to do would be to query the OpenCL C version(s) supported by the device and pick the latest.

Note that the harness already does this, unless we specify a different `-cl-std` version instead.  So, we could either explicitly pull similar logic in here instead, or rely on the harness to add the highest supported OpenCL C version implicitly.

https://github.com/KhronosGroup/OpenCL-CTS/blob/bd167754d9690a23d0210d04a9913de386ba31b8/test_common/harness/kernelHelpers.cpp#L774-L778

rjodinchr: Sounds good to me, let me update that PR

PR #2598, opened 2025-12-10
Fix test_non_uniform_work_group on devices that do no support it

ahesham-arm:
On devices that do not support non-uniform workgroups, the test was using incorrect round-up values for the global work group sizes. The global work group sizes were round up such that they are multiples of the local sizes, but those local sizes were later adjusted, in the call to `prepareDevice`, causing the test to (correctly) fail with the error code `CL_INVALID_WORK_GROUP_SIZE`, because the global work group sizes are no longer multiples of the local ones.

Example log before the fix, running `non_uniform_2d_basic`:
```
Rounding globalSize[0] = 42967 -> 43243
Rounding globalSize[1] = 13 -> 14
T E S T  P A R A M E T E R S :
    Number of dimensions:    2
    Global work group size:    {43243, 14}
    Local work group size:    {256, 1}
```
Note how the adjusted global size (43243) is not a multiple of the adjusted local size (256).

Example log after the fix, running the same test:
```
Rounding _globalSize[0] = 42967 -> 43008
Rounding _globalSize[1] = 13 -> 14
T E S T  P A R A M E T E R S :
    Number of dimensions:    2
    Global work group size:    {43008, 14}
    Local work group size:    {256, 1}
```
Adjusted global size is now a multiple of the local size.

### DIFF
--- a/test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
+++ b/test_conformance/non_uniform_work_group/TestNonUniformWorkGroup.cpp
@@ -261,50 +261,6 @@ TestNonUniformWorkGroup::TestNonUniformWorkGroup(
         return;
     }
 
-    // For OpenCL-3.0 support for non-uniform workgroups is optional, it's still
-    // useful to run these tests since we can verify the behavior of the
-    // get_enqueued_local_size() builtin for uniform workgroups, so we round up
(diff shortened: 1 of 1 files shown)


rjodinchr (approved):

lakshmih (approved):

bashbaug:

PR #2606, opened 2026-01-20
Fix test_vulkan_interop_buffer validation errors for missing vkUpdateDescriptorSets

shajder:
Fixes vulkan validation layer error:

Vulkan validation layer: Validation Error: [ VUID-vkCmdDispatch-None-08114 ] Object 0: handle = 0xb9181f0000000029, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x30b6e267 | vkCmdDispatch(): the descriptor VkDescriptorSet 0xb9181f0000000029[] [Set 0, Binding 1, Index 1, variable "bufferPtrList"] is being used in dispatch but has never been updated via vkUpdateDescriptorSets() or a similar call. The Vulkan spec states: Descriptors in each bound descriptor set, specified via vkCmdBindDescriptorSets, must be valid as described by descriptor validity if they are statically used by the VkPipeline bound to the pipeline bind point used by this command and the bound VkPipeline was not created with VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT (https://vulkan.lunarg.com/doc/view/1.4.304.0/windows/1.4-extensions/vkspec.html#VUID-vkCmdDispatch-None-08114)

### DIFF
--- a/test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
+++ b/test_conformance/common/vulkan_wrapper/vulkan_wrapper.cpp
@@ -1071,7 +1071,8 @@ VulkanComputePipeline::VulkanComputePipeline(
 
 VulkanComputePipeline::VulkanComputePipeline(
     const VulkanDevice &device, const VulkanPipelineLayout &pipelineLayout,
-    const VulkanShaderModule &shaderModule, const std::string &entryFuncName)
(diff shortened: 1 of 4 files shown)


dcrawleyqti:
Can also confirm this fixes the 08114 validation error for the buffer subtest. Qualcomm approves this change.

PR #2607, opened 2026-01-26
Add AHB buffer map/unmap test

ahesham-arm:
Add tests for clEnqueueMapBuffer and clEnqueueUnmapMemObject with a buffer created using a AHardwareBuffer

### DIFF
--- a/test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
+++ b/test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 
+#include "harness/deviceInfo.h"
 #include "harness/compat.h"
 #include "harness/kernelHelpers.h"
 #include "harness/imageHelpers.h"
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -2336,3 +2337,108 @@ REGISTER_TEST(sub_buffer)
...
+            {
+                log_error("AHardwareBuffer_unlock failed with code %d\n",
+                          ahb_result);
+                return TEST_FAIL;
+            }
+
+            void *mapped_ptr =
+                clEnqueueMapBuffer(queue, buffer, CL_BLOCKING, CL_MAP_READ, 0,
+                                   buffer_size, 0, nullptr, nullptr, &err);
+            test_error(err, "clEnqueueMapBuffer failed");
```
joselopeqti, line 2484: I just noticed that `clEnqueueAcquireExternalMemObjectsKHR` and `clEnqueueReleaseExternalMemObjectsKHR` are missing in this test and in `lifetime_buffer`, `lifetime_image` and `sub_buffer`.
They need to be called to comply with the spec:

> Applications must acquire the memory objects that are created using external handles before they can be used by any OpenCL commands queued to a command-queue

joselopeqti (changes requested):

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
```diff
@@ -2336,3 +2337,108 @@ REGISTER_TEST(sub_buffer)
...
+
+            for (size_t i = 0; i < buffer_size; ++i)
+            {
+                if (host_buffer[i] != mapped_data[i])
+                {
+                    log_error(
+                        "At position i=%zu expected value %u but got %u\n", i,
+                        host_buffer[i], mapped_data[i]);
+                    return TEST_FAIL;
+                }
```
rjodinchr, line 2498: `buffer` which is a `clMemWrapper` would be released while still mapped here. I think we should unmap before returning.

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
ahesham-arm: Done, thanks!

test_conformance/extensions/cl_khr_external_memory_ahb/test_ahb.cpp
ahesham-arm: Done, thanks!

rjodinchr (approved):

joselopeqti (approved):

PR #2608, opened 2026-01-27
Skip options_uniform_work_group_size test in SPIR-V mode

svenvh:
Currently there is no way to represent the `-cl-uniform-work-group-size` option in the SPIR-V binary, so skip the test in SPIR-V mode.

See also https://github.com/KhronosGroup/OpenCL-Docs/pull/370 .

### DIFF
--- a/test_conformance/compiler/test_build_options.cpp
+++ b/test_conformance/compiler/test_build_options.cpp
@@ -16,6 +16,7 @@
 #include "testBase.h"
 #include "harness/kernelHelpers.h"
 #include "harness/os_helpers.h"
+#include "harness/parseParameters.h"
 #include "harness/testHarness.h"
 
 #include <array>
@@ -432,6 +433,11 @@ REGISTER_TEST(options_denorm_cache)
 REGISTER_TEST(options_uniform_work_group_size)
 {
     if (get_device_cl_version(device) < Version(2, 0))
+    {
+        return TEST_SKIPPED_ITSELF;
+    }
+    // There is currently no way to represent this option in SPIR-V.
+    if (gCompilationMode == kSpir_v)
     {
         return TEST_SKIPPED_ITSELF;
     }


test_conformance/compiler/test_build_options.cpp
```diff
@@ -435,6 +436,11 @@ REGISTER_TEST(options_uniform_work_group_size)
     {
         return TEST_SKIPPED_ITSELF;
     }
+    // There is currently no way to represent this option in SPIR-V.
+    if (gCompilationMode == kSpir_v)
+    {
+        return TEST_SKIPPED_ITSELF;
```
bashbaug, line 442: I'm not quite following the rationale for this change - can you provide some more detail?

You're right that there is no way to "represent this option in SPIR-V", but there is no need to, because it is passed as a build option to `clBuildProgram` (or to `clCompileProgram`).  Therefore, I believe this case should still be tested when compiling via SPIR-V.

Note that this case is explicitly described in the spec in the description for [clEnqueueNDRangeKernel](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueNDRangeKernel):

> To support non-uniform work-groups:
> ...
> 2. The program object associated with kernel must support non-uniform work-groups. Specifically, this means:
> * ...
> * If the program was created with [clCreateProgramWithIL](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateProgramWithIL) or [clCreateProgramWithBinary](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateProgramWithBinary), the program must be compiled or built without the -cl-uniform-work-group-size build options.

svenvh: > You're right that there is no way to "represent this option in SPIR-V", but there is no need to, because it is passed as a build option to clBuildProgram (or to clCompileProgram)

Thanks so much for catching that!  Turns out I was led astray by an implementation bug, I will abandon this PR.

PR #2609, opened 2026-01-28
vectors: improve error messages

svenvh:
Clarify that the first number is the work-item.

Print hexadecimal numbers with a `0x` prefix.


bashbaug (approved):
LGTM.  Since this is only updating error messages, I am going to merge this immediately.

PR #2612, opened 2026-02-05
Fix test_vulkan_interop_image validation errors for missing vkUpdateDescriptorSets

### LINKED ISSUE #2487: test_vulkan: VVL error VUID-vkCmdDispatch-None-08114
> vkCmdDispatch(): the descriptor [VkDescriptorSet 0x80000000008, Set 0, Binding 1, Index 1, variable "image2DList"] is being used in dispatch but has never been updated via vkUpdateDescriptorSets() or a similar call.
The Vulkan spec states: Descriptors in each bound descriptor set, specified via vkCmdBindDescriptorSets, must be valid as described by descriptor validity if they are statically used by the VkPipeline bound to the pipeline bind point used by this command and the bound VkPipeline was not created with VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT (https://docs.vulkan.org/spec/latest/chapters/dispatch.html#VUID-vkCmdDispatch-None-08114)

shajder:
Fixes #2487

@nikhiljnv @karolherbst  I wasn’t able to fully validate the test due to missing support for simultaneous import/export semaphores. I’d really appreciate your review of this PR, thanks!


shajder:
> Our testing found that we need the attached change ...

applied

lakshmih (approved):

neiltrevett:
Merged as discussed in call on 5May26

PR #2616, opened 2026-02-17
Fix sync_fd imported semaphore undefined behaviour

ahesham-arm:
The following tests create an OpenCL semaphore using (fd == -1) then call `clEnqueueSignalSemaphoresKHR` on that semaphore. (fd == -1) refers to an object that has already signaled and enqueueing a signal command on it will lead to undefined behavior.

Quoting OpenCL specification:
```
The special value -1 for fd is treated like a valid sync file descriptor
referring to an object that has already signaled.
```
And
```
Signaling the same binary semaphore twice without an interleaving wait
may lead to undefined behavior.
```

- external_semaphores_simple_1
- external_semaphores_reuse
- external_semaphores_cross_queues_ooo
- external_semaphores_cross_queues_io
- external_semaphores_cross_queues_io2

This commit changes the tests to avoid signaling an already signaled semaphore and correctly re-import the semaphore's fd after enqueueing a wait successfully.

### DIFF
--- a/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
+++ b/test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
@@ -425,9 +425,14 @@ REGISTER_TEST_VERSION(external_semaphores_simple_1, Version(1, 2))
 
         // Signal semaphore
         clEventWrapper signal_event;
-        err = clEnqueueSignalSemaphoresKHR(queue, 1, &sema_ext.getCLSemaphore(),
-                                           nullptr, 0, nullptr, &signal_event);
-        test_error(err, "Could not signal semaphore");
+        if (vkExternalSemaphoreHandleType
+            != VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SYNC_FD)
+        {
+            err = clEnqueueSignalSemaphoresKHR(
+                queue, 1, &sema_ext.getCLSemaphore(), nullptr, 0, nullptr,
+                &signal_event);
+            test_error(err, "Could not signal semaphore");
+        }
 
         // Wait semaphore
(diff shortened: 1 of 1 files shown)


test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
joselopeqti, line 536: I believe `clEnqueueWaitSemaphoresKHR` should depend on the _task_events_ when using `VULKAN_EXTERNAL_SEMAPHORE_HANDLE_TYPE_SYNC_FD` as handle type because the semaphore is already signaled when the wait is executed so there is nothing waiting on them.
Maybe it would be better is the `clFinish` at the end is replaced by a `clWaitForEvents` waiting on the last _wait_event_. I think that would make the check that all events were completed more useful.

ahesham-arm: Hello, while I don't disagree with your suggestions, I don't think they should block this change as they are more of an improvement, than a fix to testing undefined behaviour, which is what I am trying to focus on here.

I am happy to apply your suggestions as a follow up PR if you agree that these changes do indeed fix external semaphores UB.

joselopeqti: Ok, I'm fine with applying the suggestions later as I agree that this PR fixes UB

joselopeqti (approved):

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
ahesham-arm: Thanks a lot, resolving this conversation.

bashbaug:
Merging as discussed in the March 3rd teleconference.

PR #2621, opened 2026-02-24
add test for loading and unloading ICDs (cl_khr_icd_unloadable)

bashbaug:
This PR adds a CTS test for the cl_khr_icd_unloadable extension, see:

* https://github.com/KhronosGroup/OpenCL-Docs/pull/1350
* https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/268

The CTS test follows a similar methodology as the ICD loader unit test:

1. The test is primarily implemented in a dynamic library "plugin".
    * The "plugin" links with the harness (and hence the OpenCL ICD loader) and makes OpenCL API calls.
    * The "plugin" currently runs a single test, which creates a kernel, executes it, checks the results, and cleans up.
2. The test executable is responsible for dynamically loading the "plugin" and invoking it to execute the test.
    * The test executable does NOT link with the harness or the OpenCL ICD loader and does not directly make any OpenCL API calls.
    * If the test executable is able to load the plugin, run the test, and unload the plugin N times (where N is currently 5), the overall test is considered to pass, and the test executable returns "success".  Otherwise, the test executable returns "failure".

### DIFF
--- a/test_conformance/extensions/cl_khr_icd_unloadable/main.cpp
+++ b/test_conformance/extensions/cl_khr_icd_unloadable/main.cpp
@@ -72,5 +72,6 @@ int main(int argc, const char *argv[])
         ClosePlugin(plugin);
     }
 
+    log_info("All testing complete.\n");
     return result;
 }


bashbaug:
For now, I am mostly interested in the high-level approach.  Is this something we would consider appropriate for a CTS test?  Is there a lighter-weight way to test these changes?

Of course, if the high-level approach looks OK, I would be interested in detailed comments also.  Thanks!

PR #2622, opened 2026-02-24
basic arraycopy fix

AhmedAmraniAkdi:
For the second subtest, map streams[0] (created with input_ptr as  CL_MEM_USE_HOST_PTR]) before modifying its host_ptr.

Set err to error_count at the end of the test, otherwise if the second subtest fails, it will print the last one failed too.

Rest of changes is formatting.

[run-test: test_basic arraycopy]


lakshmih (approved):

lakshmih (commented):
It isn't clear what additional coverage this is adding or why so would like that clarification

lakshmih (changes requested):

rjodinchr (changes requested):
Please do not format lines you are not changing, this is making the diff very complicated to read.

I agree with the err part at the end.

But I don't see any reason to add the map/unmap. Is it failing on your device?

AhmedAmraniAkdi:
Hello, I reverted the formatting changes and formatted only the affected lines.

> But I don't see any reason to add the map/unmap. Is it failing on your device?

> From looking at the change streams[0] (the cl_mem object) is never used again for the rest of the function, so there's no point in trying to synchronize between the host and device when the device memory object is never used again. The error_count accumulation at the end is useful.

streams[0] can be used again through input_ptr in the clEnqueueWriteBuffer call:
```
// Copy the array up from host ptr
        err = clEnqueueWriteBuffer(queue, streams[2], CL_TRUE, 0, sizeof(cl_uint)*num_elements, input_ptr, 0, NULL, NULL);
        test_error(err, "clEnqueueWriteBuffer failed");
```
input_ptr is owned by streams[0] and streams[0] is not released by that point. I believe we need to synchronise host and device, otherwise it is ambiguous what the test is expecting to be written to streams[2].

An alternative would be to allocate a new pointer to be used for clEnqueueWriteBuffer.

rjodinchr:
Instead of mapping/unmapping `streams[0]`, we should release it early right after the first subtest since it is not reused. This resolves the conflict on input_ptr in subsequent subtests more cleanly I believe.

AhmedAmraniAkdi:
Hi, I changed the fix as suggested, now we release streams[0] after the first subtest.

rjodinchr:
Could you add `[run-test: test_basic arraycopy]` to the PR description, or one of the commit messages, and update `ci/pocl/golden.json` accordingly?

rjodinchr (dismissed):

AhmedAmraniAkdi:
I've added the [run-test: test_basic arraycopy] tag to the description, I'll update the golden.json file once the test passes.

rjodinchr (approved):

lakshmih (approved):

bashbaug:
Merging as discussed in the August 4th teleconference.

PR #2624, opened 2026-02-27
minor bugfixes &  improvements

franz:


test_conformance/api/test_api_consistency.cpp
```diff
@@ -941,9 +1019,8 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
         // Returns CL_INVALID_OPERATION if no devices in context support
         // Intermediate Language Programs.
 
-        cl_uint bogus = 0xDEADBEEF;
-        clProgramWrapper ilProgram =
-            clCreateProgramWithIL(context, &bogus, sizeof(bogus), &error);
+        clProgramWrapper ilProgram = clCreateProgramWithIL(
+            context, &ExampleSpirv64, ExampleSpirv64Len, &error);
```
bashbaug: Can you explain the rationale for this change?

We only get here in the case where the device does not support intermediate language programs, so it shouldn't what we pass as the incoming IL.

I'd prefer not to embed binary SPIR-V blobs into CTS tests since they're very difficult to maintain.

franz: The rationale is that this tests makes an API call that triggers two separate error conditions in the same call, with different return values: CL_INVALID_OPERATION (for devices not supporting SPIRV) and CL_INVALID_VALUE (for invalid SPIRV input). But then it accepts only one returned error. 

At least in PoCL, the per-device SPIRV support checks are separate from input SPIR-V validity check, which happens to run first (any only once per API call). I can move the validity check to after per-device SPIRV support checks, which would make this change unnecessary, but the CTS test would remain ambiguous. If you prefer this solution, i can remove this commit from the PR.

bashbaug: Gotcha, makes sense, thanks.

I did notice the possible CL_INVALID_VALUE return value also, but note that it should be returned (**emphasis** mine):

> if the length-byte block of memory pointed to by il does not contain well-formed intermediate language input **that can be consumed by the OpenCL runtime**.

In other words, even if the SPIR-V is well-formed, if no devices in the context support SPIR-V, then it cannot "be consumed by the OpenCL runtime" and I believe it would be appropriate to return CL_INVALID_VALUE.

Therefore, maybe the best thing to do is to accept both error codes - CL_INVALID_OPERATION and CL_INVALID_VALUE?

franz: Yes, i have changed the commit to drop the embedded SPIRV and instead accept  both error values.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -944,6 +945,13 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
         cl_uint bogus = 0xDEADBEEF;
         clProgramWrapper ilProgram =
             clCreateProgramWithIL(context, &bogus, sizeof(bogus), &error);
+
+        // the clCreateProg API call above is called with invalid SPIRV input.
+        // if the implementation tests for SPIR-V validity before testing device
+        // support, it may return CL_INVALID_VALUE; accept that error too
+        if (error == CL_INVALID_VALUE)
+          error = CL_INVALID_OPERATION;
```
bashbaug: Nitpick: I think this is a little too clever and it means that the error message below is slightly misleading.  Suggest doing one of these two, instead:

1. Check both error conditions, and if neither is returned, fail the test using `test_fail`.  I found a couple of other places in the CTS where we are doing something like this.  Example:

    https://github.com/KhronosGroup/OpenCL-CTS/blob/65064216143344acda679c66a6992ddac9aca1ed/test_conformance/api/negative_queue.cpp#L182-L185

2. Check both possible error conditions using `test_assert_error`.  This is a bit more compact, but I don't see anywhere else where we are doing this already in the CTS.

I'd suggest moving the comment describing why CL_INVALID_VALUE is also accepted up a few lines, also, so it's next to the description for CL_INVALID_OPERATION, though it's pretty close where it is currently.

test_conformance/api/test_kernel_arg_info.cpp
```diff
@@ -542,33 +542,38 @@ size_t get_param_size(const std::string& arg_type, cl_device_id device,
     }
 
     size_t ret(0);
+    /* "signed" and "unsigned" type names are valid (int is implied) */
+    if (arg_type.find("signed") != std::string::npos)
```
bashbaug, line 546: Is this guaranteed to be valid?  Neither `signed` nor `unsigned` (without the `int`) are listed in https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#built-in-scalar-data-types.

test_conformance/images/clFillImage/test_fill_generic.cpp
```diff
@@ -302,6 +302,7 @@ int test_fill_image_generic( cl_context context, cl_command_queue queue, image_d
 
     // Unmap the image.
     error = clEnqueueUnmapMemObject(queue, image, mapped, 0, NULL, NULL);
+    error = error | clFinish(queue);
```
bashbaug: Nitpick, nice-to-have: I'd prefer to do this separately vs. combining the error code for the unmap and finish.

For bonus points, we could switch to `test_error` vs. testing the error code specifically.  Note that `test_error` will already print the file, line, and error code as a string.

test_conformance/run_conformance.py
```diff
@@ -265,13 +268,13 @@ def run_tests(tests):
         log_file.flush()
 
         print("")
-        if result != 0:
+        if result == 0 or result == 2 or result == -100:
+            log_file.write("     (" + get_time() + ")     Test " + test_name + " passed in " + str(run_time) + "s\n")
```
bashbaug: Interested in feedback from others in the working group: Should we log test skips separately?  Or is it OK to log these the same as a passing test?

test_conformance/api/test_kernel_arg_info.cpp
franz: OK, i didn't check the list of the builtin scalar data types; however, in the code of the test, there is "signed"  and "unsigned" (both without `int`) in the `std::vector` returned from method `generate_all_type_arguments(cl_device_id device)` and then the code calls `get_param_size` on each element of that vector.

I don't remember exactly, but IIRC the problem was that the `get_param_size` returned 0 for both these "implied int" types, and as a result `total_param_size` was miscalculated by 8 bytes, which then triggered an error because the actual total size of arguments was over limit, when used with SPIR-V compilation mode. IIRC Clang doesn't really care about PoCL's argument-related limits, so it will accept larger actual arguments, but the SPIRV translation path isn't so forgiving. I could try to recreate the error & get more details if you'd like.

test_conformance/images/clFillImage/test_fill_generic.cpp
franz: fixed

test_conformance/api/test_api_consistency.cpp
franz: fixed using 1st way

test_conformance/run_conformance.py
franz: I realized it's currently inconsistent, so I updated it print skipping separately from passing in both places where it's printed.

franz: FWIW, the main motivation for this commit was to fix #2643 , but since PR #2646 already fixes that bug and is approved, this commit is optional.

PR #2627, opened 2026-03-03
Fix left shift operator undefined behavior in integer_ops

joselopeqti:
As per the C++ Programming Reference page n0: 151, section: 7.6.7 Shift operators 
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/n5001.pdf

> The operands shall be prvalues of integral or unscoped enumeration type and integral promotions are performed. The type of the result is that of the promoted left operand. The behavior is undefined if the right operand is negative, or greater than or equal to the width of the promoted left operand.

### DIFF
--- a/test_conformance/integer_ops/test_extended_bit_ops_insert.cpp
+++ b/test_conformance/integer_ops/test_extended_bit_ops_insert.cpp
@@ -35,12 +35,12 @@ cpu_bit_insert(T tbase, T tinsert, cl_uint offset, cl_uint count)
     cl_ulong base = static_cast<cl_ulong>(tbase);
     cl_ulong insert = static_cast<cl_ulong>(tinsert);
 
-    cl_ulong mask = 0;
+    cl_ulong result = base;
     if (offset < 64)
-        mask = (count < 64) ? ((1ULL << count) - 1) << offset : ~0ULL;
-    cl_ulong shifted = 0;
-    if (offset < 64) shifted = (insert << offset);
-    cl_ulong result = (shifted & mask) | (base & ~mask);
+    {
+        cl_ulong mask = (count < 64) ? ((1ULL << count) - 1) << offset : ~0ULL;
+        result = ((insert << offset) & mask) | (base & ~mask);
+    }
 
     return static_cast<typename std::make_unsigned<T>::type>(result);
 }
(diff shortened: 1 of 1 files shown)


svenvh (dismissed):

test_conformance/integer_ops/test_extended_bit_ops_insert.cpp
```diff
@@ -35,8 +35,12 @@ cpu_bit_insert(T tbase, T tinsert, cl_uint offset, cl_uint count)
...
     cl_ulong insert = static_cast<cl_ulong>(tinsert);
 
-    cl_ulong mask = (count < 64) ? ((1ULL << count) - 1) << offset : ~0ULL;
-    cl_ulong result = ((insert << offset) & mask) | (base & ~mask);
+    cl_ulong mask = 0;
+    if (offset < 64)
+        mask = (count < 64) ? ((1ULL << count) - 1) << offset : ~0ULL;
+    cl_ulong shifted = 0;
+    if (offset < 64) shifted = (insert << offset);
+    cl_ulong result = (shifted & mask) | (base & ~mask);
```
bashbaug: Could we do something like this?  A bit more compact and easier to understand (IMHO):

```suggestion
    cl_ulong result = base;
    if (offset < 64)
    {
        cl_ulong mask = (count < 64) ? ((1ULL << count) - 1) << offset : ~0ULL;
        result = ((insert << offset) & mask) | (base & ~mask);
    }
```

I also considered this version with just the ternary operator also, but I think the version above with the single if statement is easier to understand.

```c++
    cl_ulong mask =
        (count < 64) && (offset < 64) ? ((1ULL << count) - 1) << offset : ~0ULL;
    cl_ulong result =
        (offset < 64) ? ((insert << offset) & mask) | (base & ~mask) : base;
```

bashbaug (commented):
Darn, I remember trying hard to avoid undefined behavior when we first did these tests (https://github.com/KhronosGroup/OpenCL-CTS/pull/1232), but I guess we missed this case.

Out of curiosity, how did you find this?

bashbaug (approved):

rjodinchr (approved):

svenvh:
> Not in this PR, but should we have an option to build with the undefined behavior sanitizer enabled?

Sorry for the late followup, but yes, I think that's valuable.  I've opened https://github.com/KhronosGroup/OpenCL-CTS/pull/2672 for this.

PR #2630, opened 2026-03-05
Added test to verify API consistency when linker or compiler unavailable

### LINKED ISSUE #955: Missing compiler optionality consistency tests
The CTS doesn't have consistency tests to check that when `CL_DEVICE_{COMPILER,LINKER}_AVAILABLE` return `CL_FALSE` `clBuildProgram`, `clCompileProgram` and `clLinkProgram` return `CL_{COMPILER,LINKER}_NOT_AVAILABLE` when appropriate.

shajder:
Fixes #955

@kpet @bashbaug I've noticed procedure `clBuildProgram` does not return `CL_LINKER_NOT_AVAILABLE`, only `CL_COMPILER_NOT_AVAILABLE`. Is this intentional ? What if compiler is available but linker isn't ? Should it still return `CL_COMPILER_NOT_AVAILABLE` ? Thanks

### DIFF
--- a/test_conformance/api/test_api_consistency.cpp
+++ b/test_conformance/api/test_api_consistency.cpp
@@ -921,7 +921,6 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
                           "support cl_khr_il_program");
 
         // Test setup:
-
         error = create_single_kernel_helper(context, &program, &kernel, 1,
                                             &test_kernel, "test");
         test_error(error, "Unable to create test kernel");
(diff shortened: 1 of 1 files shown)


test_conformance/api/test_api_consistency.cpp
```diff
@@ -961,6 +960,127 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
...
+        cl_int err =
+            clGetDeviceInfo(device, CL_DEVICE_COMPILER_AVAILABLE,
+                            sizeof(compilerAvail), &compilerAvail, nullptr);
+        test_error_fail(err, "clGetDeviceInfo failed");
+
+        if (!compilerAvail)
+        {
+            cl_uint specConst = 42;
+            error = clSetProgramSpecializationConstant(
+                program, 0, sizeof(specConst), &specConst);
```
bashbaug: Can you describe what this case is testing in more detail?

If we get here there's something in the string returned by `CL_DEVICE_IL_VERSION` or the array returned by `CL_DEVICE_ILS_WITH_VERSION`, so it would be odd to return that a compiler is unavailable.  If anything, it seems like we should test that a compiler **is** available in this case, otherwise how would you compile the programs created from a supported IL?

Regardless, we haven't created `program` for the call to `clSetProgramSpecializationConstant`, so I think we would need to accept `CL_INVALID_PROGRAM` or a bunch of other error conditions, not just `CL_COMPILER_NOT_AVAILABLE`.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -961,6 +960,127 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
...
+                "CL_COMPILER_NOT_AVAILABLE");
+        }
+    }
+
+    return TEST_PASS;
+}
+
+REGISTER_TEST(consistency_missing_compiler)
+{
+    cl_bool compilerAvail = true;
```
bashbaug: Nitpick - same type of code occurs in a few other places, also:

```suggestion
    cl_bool compilerAvail = CL_TRUE;
```

test_conformance/api/test_api_consistency.cpp
```diff
@@ -961,6 +960,127 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
...
+
+    cl_bool linkerAvail = true;
+    err = clGetDeviceInfo(device, CL_DEVICE_LINKER_AVAILABLE,
+                          sizeof(linkerAvail), &linkerAvail, nullptr);
+    test_error_fail(err, "clGetDeviceInfo failed");
+
+    if (linkerAvail || !compilerAvail)
+    {
+        log_info("Can't perform linker consistency check\n");
+        return TEST_SKIPPED_ITSELF;
```
bashbaug: This should be an error.  Both the full profile and the embedded profile say for [CL_DEVICE_LINKER_AVAILABLE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_LINKER_AVAILABLE):

> This must be [CL_TRUE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_TRUE) if [CL_DEVICE_COMPILER_AVAILABLE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_COMPILER_AVAILABLE) is [CL_TRUE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_TRUE).

I think it would be fine to skip this test if CL_DEVICE_COMPILER_AVAILABLE is CL_FALSE, though.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -961,6 +960,127 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
...
+            size_t tid = get_global_id(0);
+            dst[tid] = (int)src[tid];
+        }
+    )";
+
+    size_t line_length = strlen(sample_kernel);
+
+    /* New OpenCL API only has one entry point, so go ahead and just try it */
+    clProgramWrapper program = clCreateProgramWithSource(
+        context, 1, &sample_kernel, &line_length, &err);
```
bashbaug: Nitpicks:

1. This seems like a weird and confusing (and copy-pasted) comment.  Can we just remove it?
2. Since `sample_kernel` is NUL-terminated, we can just pass `nullptr` for `lengths`, vs. calling `strlen` explicitly.

test_conformance/api/test_api_consistency.cpp
shajder: Hmm, thanks for pointing that out. I tried to create a scenario where `clSetProgramSpecializationConstant` returns `CL_COMPILER_NOT_AVAILABLE`, which currently isn’t covered by the `CTS`. However, you’re right that the program should first be created with `clCreateProgramWithIL`.

I can’t find an appropriate test that covers the scenario (`CL_COMPILER_NOT_AVAILABLE` + `clCreateProgramWithIL`), so this will probably need to be handled as a separate issue/PR.

test_conformance/api/test_api_consistency.cpp
shajder: Done

test_conformance/api/test_api_consistency.cpp
shajder: Removed test for `CL_DEVICE_COMPILER_AVAILABLE`

test_conformance/api/test_api_consistency.cpp
shajder: Done

test_conformance/api/test_api_consistency.cpp
```diff
@@ -965,6 +964,97 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
...
+    err = clBuildProgram(program, 1, &device, NULL, NULL, NULL);
+    test_failure_error_ret(err, CL_COMPILER_NOT_AVAILABLE,
+                           "clBuildProgram should return "
+                           "CL_COMPILER_NOT_AVAILABLE",
+                           TEST_FAIL);
+
+    return TEST_PASS;
+}
+
+REGISTER_TEST(consistency_missing_linker)
```
rjodinchr, line 1011: `CL_DEVICE_LINKER_AVAILABLE == CL_FALSE && CL_DEVICE_COMPILER_AVAILABLE == CL_TRUE` is an illegal state according to the specification.

> CL_DEVICE_LINKER_AVAILABLE: This must be CL_TRUE for devices supporting the full
profile or if CL_DEVICE_COMPILER_AVAILABLE is CL_TRUE.

> CL_DEVICE_LINKER_AVAILABLE: This can be CL_FALSE for devices supporting the
embedded profile, but must be CL_TRUE if
CL_DEVICE_COMPILER_AVAILABLE is CL_TRUE.

So, I'm not sure what we are testing here.

rjodinchr (changes requested):

test_conformance/api/test_api_consistency.cpp
shajder: Corrected

rjodinchr: I think you went too far with that correction.
We are not checking that a call to clLinkProgram actually returns CL_LINKER_NOT_AVAILABLE.

The issue previous was mostly the call to clCompileProgram which could not succeed. But I do believe we should keep the call to clLinkProgram. It is expected to check first whether it can link, before looking at the program (which cannot be linked because not compiled).

shajder: hmm, wouldn't that trigger `CL_INVALID_OPERATION` (per spec: _if the compilation or build of a program executable for any of the devices listed in device_list by a previous call to clCompileProgram or clBuildProgram for program has not completed_), which sits right next to `CL_LINKER_NOT_AVAILABLE` in the error list with no prioritization between them. A conformant implementation may return either, so asserting `CL_LINKER_NOT_AVAILABLE` specifically may be debatable.

rjodinchr: I agree this is debatable, and we should talk about it during the weekly teleconference with everyone.

I do think it makes sense to first check whether any link could be performed (if not return CL_LINKER_NOT_AVAILABLE), then check whether the function arguments are valid.

shajder: We may allow both errors as valid result for now. I can't spot any error prioritization in the OpenCL spec, adding it here would set a precedent.

rjodinchr: I also think there are no error prioritization in the OpenCL spec, but I do believe they are some already in the OpenCL-CTS.
 
For example, in `negative_invalid_arg_index`, we call clSetKernelArg with an invalid index and an invalid arg value. But we check for `CL_INVALID_ARG_INDEX`.

rjodinchr: But allowing for both errors would make sense to me. That would make it more flexible for the implementers.

test_conformance/api/test_api_consistency.cpp
```diff
@@ -965,6 +964,106 @@ REGISTER_TEST_VERSION(consistency_il_programs, Version(3, 0))
...
+                  "CL_DEVICE_COMPILER_AVAILABLE is CL_TRUE.\n");
+        return TEST_FAIL;
+    }
+
+    if (!gIsEmbedded)
+    {
+        log_error("CL_DEVICE_LINKER_AVAILABLE must be CL_TRUE for devices "
+                  "supporting the full profile.\n");
+        return TEST_FAIL;
+    }
```
lakshmih: Suggest moving this up to ~line 1023 and collating with other linker availability checks. Otherwise looks good

lakshmih (changes requested):

rjodinchr (approved):

test_conformance/api/test_api_consistency.cpp
shajder: Thanks, done.

joselopeqti (approved):

Issue #594, opened 2020-02-13
Validate values returned by image device queries when images are not supported


kpet:
See https://github.com/KhronosGroup/OpenCL-Docs/pull/217.

PR #2632, opened 2026-03-09
Added support for querying max read/write image args when the device does not support images

shajder:
Related to #594

additional housekeeping:

-replaced raw pointers with std::vectors
-replaced OpenCL C style types with wrappers
-corrected consistency of returned values

[run-test: test_api consistency_read_write_images]
[run-test: test_api min_max_image_array_size]
[run-test: test_api min_max_read_image_args]
[run-test: test_api min_max_write_image_args]


test_conformance/api/test_api_min_max.cpp
```diff
@@ -992,61 +1025,80 @@ REGISTER_TEST(min_max_image_array_size)
...
+        /* Get the max image array width */
+        error = clGetDeviceInfo(device, CL_DEVICE_IMAGE_MAX_ARRAY_SIZE,
+                                sizeof(maxDimension), &maxDimension, NULL);
+        test_error(error, "Unable to get max image array size from device");
+
+        test_failure_error_ret(
+            maxDimension, 0,
+            "Missing image support but CL_DEVICE_IMAGE_MAX_ARRAY_SIZE  query "
+            "did not return 0",
+            TEST_FAIL);
```
rjodinchr: Shouldn't we return TEST_SKIPPED_ITSELF here?

test_conformance/api/test_api_min_max.cpp
```diff
@@ -341,48 +355,53 @@ REGISTER_TEST(min_max_read_image_args)
...
 
     /* Create a program with that many read args */
-    programSrc = (char *)malloc(strlen(sample_read_image_kernel_pattern[0])
-                                + (strlen(readArgPattern) + 6) * (maxReadImages)
-                                + strlen(sample_read_image_kernel_pattern[1])
-                                + 1 + 40240);
-
-    strcpy(programSrc, sample_read_image_kernel_pattern[0]);
-    strcat(programSrc, "read_only image2d_t srcimg0");
+    std::vector<char> programSrc(
```
rjodinchr: Why not using `std::string`?

test_conformance/api/test_api_min_max.cpp
```diff
@@ -490,42 +521,46 @@ REGISTER_TEST(min_max_write_image_args)
     }
 
     /* Create a program with that many write args + 1 */
-    programSrc = (char *)malloc(
+    std::vector<char> programSrc(
```
rjodinchr: `std::string`?

test_conformance/api/test_api_min_max.cpp
shajder: Corrected

test_conformance/api/test_api_min_max.cpp
shajder: Corrected

test_conformance/api/test_api_min_max.cpp
shajder: Corrected

test_conformance/api/test_api_consistency.cpp
```diff
@@ -625,6 +625,16 @@ REGISTER_TEST_VERSION(consistency_read_write_images, Version(3, 0))
                "Unable to query "
                "CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS");
 
+    if (checkForImageSupport(device))
+    {
+        test_failure_error_ret(maxReadWriteImageArgs, 0,
```
rjodinchr: `test_assert_error_ret`?

test_conformance/api/test_api_min_max.cpp
```diff
@@ -271,24 +271,37 @@ REGISTER_TEST(min_max_read_image_args)
...
 
-    PASSIVE_REQUIRE_IMAGE_SUPPORT(device)
-    image_format_desc.image_channel_order = CL_RGBA;
-    image_format_desc.image_channel_data_type = CL_FLOAT;
+    if (checkForImageSupport(device))
+    {
+        /* Get the max read image arg count */
+        error = clGetDeviceInfo(device, CL_DEVICE_MAX_READ_IMAGE_ARGS,
+                                sizeof(maxReadImages), &maxReadImages, NULL);
+        test_error(error, "Unable to get max read image arg count from device");
```
rjodinchr: Redundant call with the one after (line 307). Merge them and move them before the current if statement

test_conformance/api/test_api_min_max.cpp
```diff
@@ -407,70 +420,81 @@ REGISTER_TEST(min_max_read_image_args)
...
+    cl_int eventStatus;
     cl_uint minRequiredWriteImages = gIsEmbedded ? 1 : 8;
 
+    if (checkForImageSupport(device))
+    {
+        /* Get the max write image arg count */
+        error = clGetDeviceInfo(device, CL_DEVICE_MAX_WRITE_IMAGE_ARGS,
+                                sizeof(maxWriteImages), &maxWriteImages, NULL);
+        test_error(error,
+                   "Unable to get max write image arg count from device");
```
rjodinchr: Redundant call

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1053,73 +1069,80 @@ REGISTER_TEST(min_max_image_array_size)
     cl_ulong maxAllocSize;
     size_t minRequiredDimension = gIsEmbedded ? 256 : 2048;
 
-    if (checkForImageSupport(device))
+    if (!checkForImageSupport(device))
```
rjodinchr: I don't like that. This is making the diff much bigger than what it really is.
I don't see a real benefit from it as well.

I would prefer to keep the early exit here, and avoid the else. That should make the diff much smaller.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -341,48 +354,47 @@ REGISTER_TEST(min_max_read_image_args)
...
     test_error(error, "Failed to create the program and kernel.");
-    free(programSrc);
 
     result = clCreateBuffer(context, CL_MEM_READ_WRITE, sizeof(cl_float), NULL,
                             &error);
     test_error(error, "clCreateBufer failed");
 
     /* Create some I/O streams */
-    streams = new clMemWrapper[maxReadImages + 1];
+    std::vector<clMemWrapper> streams(0);
```
rjodinchr: Is there a reason for that explicit `0`?

```suggestion
    std::vector<clMemWrapper> streams;
```

test_conformance/api/test_api_min_max.cpp
```diff
@@ -490,42 +514,36 @@ REGISTER_TEST(min_max_write_image_args)
...
         create_single_kernel_helper(context, &program, &kernel, 1,
-                                    (const char **)&programSrc, "sample_test");
+                                    (const char **)&prog_data, "sample_test");
     test_error(error, "Failed to create the program and kernel.");
-    free(programSrc);
-
 
     /* Create some I/O streams */
-    streams = new clMemWrapper[maxWriteImages + 1];
+    std::vector<clMemWrapper> streams(0);
```
```suggestion
    std::vector<clMemWrapper> streams;
```

test_conformance/api/test_api_consistency.cpp
shajder: Corrected

test_conformance/api/test_api_min_max.cpp
shajder: Code duplication removed

test_conformance/api/test_api_min_max.cpp
shajder: done

test_conformance/api/test_api_min_max.cpp
shajder: done

test_conformance/api/test_api_min_max.cpp
shajder: Corrected along with few more fixes

test_conformance/api/test_api_min_max.cpp
shajder: done

rjodinchr (dismissed):

test_conformance/api/test_api_consistency.cpp
```diff
@@ -625,8 +625,17 @@ REGISTER_TEST_VERSION(consistency_read_write_images, Version(3, 0))
...
-    // clGetSupportedImageFormats, passing
-    // CL_MEM_KERNEL_READ_AND_WRITE
+    if (checkForImageSupport(device))
+    {
+        test_assert_error_ret(maxReadWriteImageArgs == 0,
+                              "Missing image support but "
+                              "CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS query did "
+                              "not return 0",
+                              TEST_FAIL);
+        return TEST_SKIPPED_ITSELF;
```
bashbaug: Couple of comments here:

1. I don't think it's correct to skip the test here, even if images aren't supported.  `clGetSupportedImageFormats` should still work even if no devices in the context support images by returning that no image formats are supported.  In other words, there are no defined error conditions for `clGetSupportedImageFormats` when no devices in the context support images.
2. This test is currently gated by OpenCL 3.0 since it is currently testing OpenCL 3.0 API consistency for read-write images.  I don't think it's a terrible place for this check, but if we're going to keep it here, we ought to do some minor refactoring so it runs on OpenCL 2.x devices, also.  Alternatively, maybe it makes more sense to move this check to some other test?  I'm a bit surprised there is no `test_api min_max_read_write_image_args`.

shajder: > 1. I don't think it's correct to skip the test here, even if images aren't supported.  `clGetSupportedImageFormats` should still work even if no devices in the context support images by returning that no image formats are supported.  In other words, there are no defined error conditions for `clGetSupportedImageFormats` when no devices in the context support images.

`clGetSupportedImageFormats` has no defined error condition for the no-image-support case, so technically call shouldn't cause trouble. Skip removed.

That being said, is the assertion below:

```
    if (maxReadWriteImageArgs == 0)
    {
        test_assert_error(
            totalReadWriteImageFormats == 0,
            "CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS returned 0 "
            "but clGetSupportedImageFormats(CL_MEM_KERNEL_READ_AND_WRITE) "
            "returned a non-empty set");
    }

```

 grounded in the spec? 5.3.2 does not say the returned set has to be empty in that case.

shajder: > 2\. This test is currently gated by OpenCL 3.0 since it is currently testing OpenCL 3.0 API consistency for read-write images.  I don't think it's a terrible place for this check, but if we're going to keep it here, we ought to do some minor refactoring so it runs on OpenCL 2.x devices, also.  Alternatively, maybe it makes more sense to move this check to some other test?  I'm a bit surprised there is no `test_api min_max_read_write_image_args`.

Are you thinking of a new test (new issue) covering that query, and dropping the check from here?

bashbaug: > does not say the returned set has to be empty in that case.

Hmm, well we do have in the current spec:

> [clGetSupportedImageFormats](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#clGetSupportedImageFormats) returns a union of image formats supported by all devices in the context.

It seems hard to support an image format without supporting images, but I don't mind saying that the returned set must be empty explicitly if it'd be helpful.  We could just add a sentence like "If no devices in the context support images, the number of supported image formats will be zero", or similar.  I think this is what we are currently testing (or plan to test), so this wouldn't require any new tests.

> Are you thinking of a new test (new issue) covering that query, and dropping the check from here?

I think we have several paths forward:

1. Drop the check from here and file an issue to add the missing check for read-write images in a more appropriate place.
2. Drop the check from here and add the missing check to some other test in this PR, which could be an entirely new test.
3. Keep the check here, but refactor the test slightly so it can run on 2.x devices in addition to 3.0 devices.

I lean toward (1) or (2) personally.

bashbaug: Here's a PR that says explicitly that clGetSupportedImageFormats must return that zero image formats are supported when no devices in the context support images: https://github.com/KhronosGroup/OpenCL-Docs/pull/1606

shajder: > I lean toward (1) or (2) personally.

Dropped the check here. Rather than adding the new test to this PR, I added it in a separate one (#2785) with reference to #594, since this PR looks close to done.

Also fixed the minimum for `CL_DEVICE_MAX_WRITE_IMAGE_ARGS`, which per tables 63 and 64 in the 3.1.1 spec is 64 for the full profile and 8 for embedded. The test was checking against 8 and 1.

bashbaug: > Also fixed the minimum for `CL_DEVICE_MAX_WRITE_IMAGE_ARGS`, which per tables 63 and 64 in the 3.1.1 spec is 64 for the full profile and 8 for embedded. The test was checking against 8 and 1.

I think we should make this change in a separate PR, since it looks like a number of currently conformant implementations are returning a value smaller than 64 today, see the [gpuinfo.org database](https://opencl.gpuinfo.org/displaydeviceinfo.php?name=CL_DEVICE_MAX_WRITE_IMAGE_ARGS).

Note that the minimum value was 8 for an OpenCL 1.2 implementation, then 64 for an OpenCL 2.x implementation, which was carried forward for OpenCL 3.x implementations.

shajder: > I think we should make this change in a separate PR

Correction moved to #2787.

Out of curiosity, should the change of minimum value between versions be noted in the spec? The table only gives the current value, while the spec spells out per-version requirements elsewhere quite often.

bashbaug (approved):

bashbaug:

PR #2634, opened 2026-03-11
Fix CTS api sub test failure for negative_get_command_queue_info

jujiang-del:
To fix a test failure in api sub-test "negative_get_command_queue_info".
The test code gets clGetDeviceInfo for the property CL_DEVICE_QUEUE_ON_DEVICE_PROPERTIES, when it create command queue for the device, "CL_QUEUE_ON_DEVICE" should be the correct input, instead of CL_QUEUE_PROPERTIES.

### DIFF
--- a/test_conformance/api/negative_queue.cpp
+++ b/test_conformance/api/negative_queue.cpp
@@ -435,7 +435,7 @@ REGISTER_TEST(negative_get_command_queue_info)
     if (device_supports_on_device_queue(device))
     {
         const cl_queue_properties properties[] = {
-            CL_QUEUE_PROPERTIES, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, 0
+            CL_QUEUE_ON_DEVICE, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, 0
         };
         cl_int err = CL_INVALID_VALUE;
         clCommandQueueWrapper cmd_queue = clCreateCommandQueueWithProperties(
--- a/test_conformance/api/negative_queue.cpp
+++ b/test_conformance/api/negative_queue.cpp
@@ -438,10 +438,10 @@ REGISTER_TEST(negative_get_command_queue_info)
         err = clGetCommandQueueInfo(queue, CL_QUEUE_SIZE,
                                     sizeof(queue_size), &queue_size, nullptr);
         test_failure_error_ret(err, CL_INVALID_COMMAND_QUEUE,
-                                "clGetCommandQueueInfo should return "
-                                "CL_INVALID_COMMAND_QUEUE when: \"command_queue "
-                                "is not a valid command-queue for param_name\"",
-                                TEST_FAIL);
+                               "clGetCommandQueueInfo should return "
+                               "CL_INVALID_COMMAND_QUEUE when: \"command_queue "
+                               "is not a valid command-queue for param_name\"",
+                               TEST_FAIL);
     }
 
     constexpr cl_command_queue_info invalid_param = -1;
--- a/test_conformance/api/negative_queue.cpp
+++ b/test_conformance/api/negative_queue.cpp
@@ -438,7 +438,7 @@ REGISTER_TEST(negative_get_command_queue_info)
             CL_QUEUE_PROPERTIES,
             CL_QUEUE_ON_DEVICE | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE,
             0
-            };
+        };
         cl_int err = CL_INVALID_VALUE;
         clCommandQueueWrapper cmd_queue = clCreateCommandQueueWithProperties(
             context, device, properties, &err);
--- a/test_conformance/api/negative_queue.cpp
+++ b/test_conformance/api/negative_queue.cpp
@@ -434,22 +434,14 @@ REGISTER_TEST(negative_get_command_queue_info)
 
     if (device_supports_on_device_queue(device))
     {
-        const cl_queue_properties properties[] = {
-            CL_QUEUE_PROPERTIES,
-            CL_QUEUE_ON_DEVICE | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, 0
-        };
-        cl_int err = CL_INVALID_VALUE;
-        clCommandQueueWrapper cmd_queue = clCreateCommandQueueWithProperties(
-            context, device, properties, &err);
-        test_error(err, "clCreateCommandQueueWithProperties");
         cl_uint queue_size = -1;
-        err = clGetCommandQueueInfo(cmd_queue, CL_QUEUE_SIZE,
+        err = clGetCommandQueueInfo(queue, CL_QUEUE_SIZE,
                                     sizeof(queue_size), &queue_size, nullptr);
         test_failure_error_ret(err, CL_INVALID_COMMAND_QUEUE,
-                               "clGetCommandQueueInfo should return "
-                               "CL_INVALID_COMMAND_QUEUE when: \"command_queue "
-                               "is not a valid command-queue for param_name\"",
-                               TEST_FAIL);
+                                "clGetCommandQueueInfo should return "
+                                "CL_INVALID_COMMAND_QUEUE when: \"command_queue "
+                                "is not a valid command-queue for param_name\"",
+                                TEST_FAIL);
     }
 
     constexpr cl_command_queue_info invalid_param = -1;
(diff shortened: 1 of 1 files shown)


test_conformance/api/negative_queue.cpp
bashbaug: I don't think this is a correct change.  Note that this is a negative test, so it is checking that the proper error code is returned.  Specifically, it is testing that `CL_INVALID_COMMAND_QUEUE` is returned when querying `CL_QUEUE_SIZE` for a command-queue that is **not** a device command-queue, therefore the command-queue should **not** be created with the `CL_QUEUE_ON_DEVICE` property.

IMHO there are some improvements we could make to this test - for example, we could run this scenario on all OpenCL 2.x or newer device, not just those that support device command-queues - but I do think the test is fundamentally correct.  Please let me know if you still believe otherwise - thanks1

bashbaug (changes requested):

test_conformance/api/negative_queue.cpp
jujiang-del: If keep the original CL_QUEUE_PROPERTIES, the API clCreateCommandQueueWithProperties will fail and won't proceed to the following calls.

bashbaug: No, the call with `CL_QUEUE_PROPERTIES` should succeed: It's creating a command queue with an array of properties, and the only property that is being passed is the `CL_QUEUE_PROPERTIES` bitfield with value `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE`.

In other words, this is the equivalent of this call without using the array of properties:

```c++
        clCommandQueueWrapper cmd_queue = clCreateCommandQueue(
            context, device, CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, &err);
```

edit: fixed copy-paste error

I ran this test on our CPU device (which supports device queues) and it behaved as expected:

```
>>>> clCreateCommandQueueWithProperties: context = 0x5cea69aec7c8, device = 12th Gen Intel(R) Core(TM) i9-12900K (CL_DEVICE_TYPE_CPU) (0x5cea69c1dbe8), properties = [ CL_QUEUE_PROPERTIES = CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE ]
<<<< clCreateCommandQueueWithProperties: returned 0x5cea6b7a8098 -> CL_SUCCESS
>>>> clGetCommandQueueInfo: command_queue = 0x5cea6b7a8098, param_name = CL_QUEUE_SIZE (00001094)
<<<< clGetCommandQueueInfo -> CL_INVALID_COMMAND_QUEUE
```

jujiang-del: The test code queries clGetDeviceInfo for the property CL_DEVICE_QUEUE_ON_DEVICE_PROPERTIES. When creating a command queue for the device, the correct flag to use is CL_QUEUE_ON_DEVICE. The runtime must validate that the device actually supports this capability. This requires carefully checking the value returned by clGetDeviceInfo.

When CL_QUEUE_PROPERTIES is incorrectly passed to clCreateCommandQueue, the call fails, and the subsequent negative test clGetCommandQueueInfo does not execute.

Additionally, several negative API test cases validate CL_QUEUE_PROPERTIES with the flag CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE, including:

negative_create_command_queue

negative_create_command_queue_with_properties

After fully investigating and verifying the behavior, including applying the CTS patch and updating the OpenCL runtime validation, all 109 out of 109 tests now pass.

bashbaug: I don't think I'm following - sorry.  Is the issue that your implementation does not support out-of-order **host** command queues, so the call to create the command queue is failing?  If so, that's plausible, we've had other similar problems recently.

This seems like a different issue / fix than the one in this PR, though.

jujiang-del: Looking into the test codes, the call clGetDeviceInfo gets the property CL_DEVICE_QUEUE_ON_DEVICE_PROPERTIES, when it create command queue for the device, "CL_QUEUE_ON_DEVICE" should be the correct input, instead of CL_QUEUE_PROPERTIES.

bashbaug: > when it create command queue for the device, "CL_QUEUE_ON_DEVICE" should be the correct input, instead of CL_QUEUE_PROPERTIES.

No, this is not the case.  Please have a look at the [table of properties for clCreateCommandQueueWithProperties](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#queue-properties-table).

The valid property names are in the left-most column of the table, which includes `CL_QUEUE_PROPERTIES`.  The valid property values are described in the right-most column of the table.  Note that `CL_QUEUE_ON_DEVICE` is a valid bitfield bit, but not a valid property name, therefore the change currently in this PR should produce a `CL_INVALID_VALUE` error due to the invalid property name rather than successfully create a command queue.

Is the issue that the device you are testing does not support an out-of-order host queue?  If so, we can fix this one of two ways.  Either:

1. We could check for out-of-order host queue support, and skip this scenario if the device does not support out-of-order host queues.
2. We could create (or perhaps just use) some other type of queue that is supported to test this scenario vs. an out-of-order host queue.

jujiang-del: Thanks for the review suggestion. Updated the PR. please review again.

lakshmih (dismissed):

test_conformance/api/negative_queue.cpp
bashbaug: This is still not right, sorry.

The test is now creating a command-queue with the `QUEUE_ON_DEVICE` property and querying `CL_QUEUE_SIZE` for this command-queue, but this should no longer be an error, because querying `CL_QUEUE_SIZE` for a device command-queue is valid.

We really need this to be a **host** command-queue, **without** the `CL_QUEUE_ON_DEVICE` property, to generate the expected error.

```
>>>> clCreateCommandQueueWithProperties: context = 0x638535bb1e08, device = 12th Gen Intel(R) Core(TM) i7-1270P (CL_DEVICE_TYPE_CPU) (0x638535bab258), properties = [ CL_QUEUE_PROPERTIES = CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE | CL_QUEUE_ON_DEVICE ]
<<<< clCreateCommandQueueWithProperties: returned 0x638538219468 -> CL_SUCCESS
>>>> clGetCommandQueueInfo: command_queue = 0x638538219468, param_name = CL_QUEUE_SIZE (00001094)
<<<< clGetCommandQueueInfo -> CL_SUCCESS
ERROR: clGetCommandQueueInfo should return CL_INVALID_COMMAND_QUEUE when: "command_queue is not a valid command-queue for param_name"! (Got CL_SUCCESS, expected CL_INVALID_COMMAND_QUEUE from /home/bashbaug/git/OpenCL-CTS/test_conformance/api/negative_queue.cpp:448)
```

bashbaug (changes requested):

test_conformance/api/negative_queue.cpp
bashbaug, line 445: How about if we replace this entire code block with:

```c++
    cl_uint queue_size = -1;
    err = clGetCommandQueueInfo(queue, CL_QUEUE_SIZE,
                                sizeof(queue_size), &queue_size, nullptr);
    test_failure_error_ret(err, CL_INVALID_COMMAND_QUEUE,
                            "clGetCommandQueueInfo should return "
                            "CL_INVALID_COMMAND_QUEUE when: \"command_queue "
                            "is not a valid command-queue for param_name\"",
                            TEST_FAIL);
```

Since `queue` is a host command-queue, this should generate the expected error, and then we don't even need to bother creating another command-queue.

jujiang-del: Thanks. Verified and updated as suggested.

test_conformance/api/negative_queue.cpp
```diff
@@ -434,16 +434,9 @@ REGISTER_TEST(negative_get_command_queue_info)
 
     if (device_supports_on_device_queue(device))
```
bashbaug, line 435: Strictly speaking, I think we could perform this check on any device supporting OpenCL 2.0 or newer, not just those that support on-device queues, which would give us a little more coverage.

bashbaug (approved):
Thanks, LGTM!

I left one comment below which would give us a little more coverage, but I would be fine merging this PR in its current state.

lakshmih (approved):

bashbaug:

PR #2638, opened 2026-03-13
Add command line flags to override test_compiler include directories

olestrohm-arm:
This adds flags `--include-test-directory` and
`--second-include-test-directory` flags to test_compiler to allow the user to specify their locations rather than relying on the test to infer their location based on the path of the test binary.

### DIFF
--- a/test_conformance/compiler/main.cpp
+++ b/test_conformance/compiler/main.cpp
@@ -18,9 +18,16 @@
 
 #include "harness/testHarness.h"
 #include "harness/stringHelpers.h"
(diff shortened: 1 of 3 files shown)


rjodinchr (commented):
I'm not against that change, but I would go even further: we could create the directories & files at runtime under `std::filesystem::temp_directory_path()` so that we don't have to install anything.
What do you think?

lakshmih (approved):

neiltrevett:

PR #2640, opened 2026-03-17
Added support for querying max image samplers when the device does not support images

shajder:
Related to #594
additional housekeeping:

-replaced raw pointers with std::vectors
-replaced OpenCL C style types with wrappers
-corrected consistency of returned values

[run-test: test_api min_max_samplers]


test_conformance/api/test_api_min_max.cpp
```diff
@@ -1443,37 +1455,37 @@ REGISTER_TEST(min_max_samplers)
...
+        std::strncat(programSrc.data(), samplerLine, programSrc.size());
     }
-    strcat(programSrc, sample_sampler_kernel_pattern[4]);
-
+    std::strncat(programSrc.data(), sample_sampler_kernel_pattern[4],
+                 programSrc.size());
 
     error =
         create_single_kernel_helper(context, &program, &kernel, 1,
                                     (const char **)&programSrc, "sample_test");
```
```suggestion
const char *prog_data = programSrc.data();
error =
create_single_kernel_helper(context, &program, &kernel, 1,
(const char **)&prog_data, "sample_test");
```
Casting `&programSrc` to `const char **` feels like an bug.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1443,37 +1455,37 @@ REGISTER_TEST(min_max_samplers)
     }
 
     /* Create a kernel to test with */
-    programSrc = (char *)malloc(
+    std::vector<char> programSrc(
```
rjodinchr: `std::string`?

test_conformance/api/test_api_min_max.cpp
shajder: Thanks, good find! Corrected.

test_conformance/api/test_api_min_max.cpp
shajder: I know contiguous storage has been guaranteed since C++11, so `std::string` would work fine here. I picked `std::vector<char>` for consistency with how kernel sources are assembled elsewhere in these tests, following earlier review guidance in this repo. Happy to switch if you prefer `std::string`.

rjodinchr: This is just building a string, using `std::string` would make more sense to me. It would also simplify the code after it (removing the use of `strncat` & `strlen`) I believe.

shajder: Corrected

test_conformance/api/test_api_min_max.cpp
```diff
@@ -19,6 +19,7 @@
 #include <ctype.h>
 #include <string.h>
 #include <cinttypes>
+#include <cstring>
```
rjodinchr: `string.h` & `cstring` are very similar, aren't they? Should we have only 1 of them?

test_conformance/api/test_api_min_max.cpp
```diff
@@ -1401,30 +1402,35 @@ REGISTER_TEST(min_max_samplers)
...
                             &maxSamplers, NULL);
     test_error(error, "Unable to get max sampler count from device");
 
+    if (checkForImageSupport(device))
+    {
+        test_failure_error_ret(
+            maxSamplers, 0,
+            "Missing image support but CL_DEVICE_MAX_SAMPLERS query "
+            "did not return 0",
+            TEST_FAIL);
```
rjodinchr: `test_failure_error_ret` will try to print `maxSamplers` as an error code in the failing path. I don't think that's what we want. I think we should not use that helper here.

test_conformance/api/test_api_min_max.cpp
shajder: cstring removed

test_conformance/api/test_api_min_max.cpp
shajder: helper replaced

rjodinchr (dismissed):

bashbaug:
Discussed in the September 8th teleconference. OK to merge when the conflict is resolved.

rjodinchr (approved):

PR #2645, opened 2026-03-25
Add validation for CL_IMAGE_ROW_PITCH and CL_IMAGE_SLICE_PITCH query results

### LINKED ISSUE #1396: Value returned for `CL_IMAGE_ROW_PITCH` and `CL_IMAGE_SLICE_PITCH` not validated
These queries are loosely specified (see https://github.com/KhronosGroup/OpenCL-Docs/issues/763 for the spec discussion) and the value returned by implementations is not validated by the CTS (except in the testing for `cl_khr_dx9_media_sharing`).

shajder:
Fixes #1396


test_conformance/images/clGetInfo/test_2D.cpp
```diff
@@ -148,9 +147,82 @@ int test_get_image_info_single( cl_context context, image_descriptor *imageInfo,
...
 
-  size_t outSlicePitch;
-  error = clGetImageInfo( image, CL_IMAGE_SLICE_PITCH, sizeof( outSlicePitch ), &outSlicePitch, NULL );
-  test_error( error, "Unable to get image info (slice pitch)" );
+    size_t outSlicePitch;
+    error = clGetImageInfo(image, CL_IMAGE_SLICE_PITCH, sizeof(outSlicePitch),
+                           &outSlicePitch, NULL);
+    test_error(error, "Unable to get image info (slice pitch)");
+
+    if (imageInfo->type == CL_MEM_OBJECT_IMAGE2D
```
rjodinchr: Shouldn't we have `CL_MEM_OBJECT_IMAGE1D_ARRAY` here as well?

test_conformance/images/clGetInfo/test_2D.cpp
```diff
@@ -148,9 +147,82 @@ int test_get_image_info_single( cl_context context, image_descriptor *imageInfo,
...
+                    "ERROR: CL_IMAGE_ROW_PITCH is 0 for COPY_HOST_PTR image\n");
+                return 1;
+            }
+
+            if (imageInfo->type != CL_MEM_OBJECT_IMAGE2D && outSlicePitch == 0)
+            {
+                log_error("ERROR: CL_IMAGE_SLICE_PITCH is 0 for COPY_HOST_PTR "
+                          "3D/array image\n");
+                return 1;
+            }
```
rjodinchr: For `IMAGE2D` slice must be zero. Can we add a check here to ensure that?

test_conformance/images/clGetInfo/test_2D.cpp
shajder: Indeed per spec `CL_IMAGE_SLICE_PITCH` refers to `CL_MEM_OBJECT_IMAGE1D_ARRAY` as well (implementation computes it as `CL_IMAGE_ROW_PITCH` not `height * row_pitch`, since `height` is 0 for a 1D array).

That said, I think there is a more fundamental issue here. `test_get_image_info_single` looks generic, with branches for all six image types, but in this file the only `CL_MEM_OBJECT_IMAGE2D` actually reaches this code; the 1D/3D/array branches are effectively dead in this translation unit, and the other image types are covered by their own drivers/files. Adding `1D_ARRAY` to the condition here would just add another unreachable branch.

I will have to apply more general refactoring here, unless I am missing something obvious. I would appreciate your confirmation, thanks.

rjodinchr: I disagree, I see `test_get_image_info_single` being called in all the image types.
```
$ git grep test_get_image_info_single
test_conformance/images/clGetInfo/test_1D.cpp:18:extern int test_get_image_info_single( cl_context context, image_descriptor *imageInfo, MTdata d, cl_mem_flags flags, size_t row_pitch, size_t slice_pitch );
test_conformance/images/clGetInfo/test_1D.cpp:66:                if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D.cpp:69:                    if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D.cpp:92:                if( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D.cpp:95:                    if( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D.cpp:128:                if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D.cpp:131:                    if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:18:extern int test_get_image_info_single( cl_context context, image_descriptor *imageInfo, MTdata d, cl_mem_flags flags, size_t row_pitch, size_t slice_pitch );
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:67:                    if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:70:                        if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:98:              if( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:101:                if( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:133:                if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:136:                    if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:200:                        if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:203:                            if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:232:              if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:235:                if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:275:                if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_1D_2D_array.cpp:278:                    if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_1D_buffer.cpp:19:extern int test_get_image_info_single(cl_context context,
test_conformance/images/clGetInfo/test_1D_buffer.cpp:62:            if (test_get_image_info_single(context, &imageInfo, seed, flags, 0,
test_conformance/images/clGetInfo/test_1D_buffer.cpp:86:            if (test_get_image_info_single(context, &imageInfo, seed, flags, 0,
test_conformance/images/clGetInfo/test_1D_buffer.cpp:121:            if (test_get_image_info_single(context, &imageInfo, seed, flags, 0,
test_conformance/images/clGetInfo/test_2D.cpp:18:int test_get_image_info_single( cl_context context, image_descriptor *imageInfo, MTdata d, cl_mem_flags flags, size_t row_pitch, size_t slice_pitch )
test_conformance/images/clGetInfo/test_2D.cpp:342:                    if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_2D.cpp:345:                        if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_2D.cpp:370:                if( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_2D.cpp:373:                    if( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_2D.cpp:407:                if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_2D.cpp:410:                    if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, 0 ) )
test_conformance/images/clGetInfo/test_3D.cpp:18:extern int test_get_image_info_single( cl_context context, image_descriptor *imageInfo, MTdata d, cl_mem_flags flags, size_t row_pitch, size_t slice_pitch );
test_conformance/images/clGetInfo/test_3D.cpp:73:                        if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_3D.cpp:76:                            if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, imageInfo.slicePitch ) )
test_conformance/images/clGetInfo/test_3D.cpp:105:                if( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_3D.cpp:108:                    if( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, imageInfo.slicePitch ) )
test_conformance/images/clGetInfo/test_3D.cpp:148:                if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], 0, 0 ) )
test_conformance/images/clGetInfo/test_3D.cpp:151:                    if ( test_get_image_info_single( context, &imageInfo, seed, all_host_ptr_flags[j], imageInfo.rowPitch, imageInfo.slicePitch ) )
```

shajder: Thanks, correction applied.

test_conformance/images/clGetInfo/test_2D.cpp
shajder: Done

test_conformance/images/clGetInfo/test_2D.cpp
rjodinchr: What about `CL_MEM_OBJECT_IMAGE1D` & `CL_MEM_OBJECT_IMAGE1D_BUFFER`?
`1D_BUFFER` seems to be handled below, is there anything in this if statement that would not apply to `1D_BUFFER`?

Basically, I'm asking whether we really need that check. I don't see the specification stating that what we are checking here are just for those image types, but I might be missing it from the specification.

test_conformance/images/clGetInfo/test_2D.cpp
```diff
@@ -148,9 +147,91 @@ int test_get_image_info_single( cl_context context, image_descriptor *imageInfo,
...
+                    "ERROR: CL_IMAGE_SLICE_PITCH mismatch "
+                    "(flags=0x%lx slice_pitch=%zu): expected %zu, got %zu\n",
+                    (unsigned long)flags, slice_pitch, expected_slice,
+                    outSlicePitch);
+                return 1;
+            }
+        }
+        else
+        {
+            // COPY_HOST_PTR: may relayout -> only require non-zero
```
rjodinchr: `COPY_HOST_PTR` does not require non-zero only I think when reading the spec:

> If image was created with a non-zero value for image_row_pitch, then the value provided for image_row_pitch by the application is returned, otherwise the returned value is calculated as CL_IMAGE_WIDTH × CL_IMAGE_ELEMENT_SIZE.

test_conformance/images/clGetInfo/test_2D.cpp
shajder: We can drop that guard, but then a few cases need to be handled explicitly instead of being implicitly skipped - the main one being `CL_MEM_OBJECT_IMAGE1D_BUFFER`.

A 1D image buffer is created from a buffer, so host_ptr is NULL at creation and `image_row_pitch` must be 0. That puts `CL_IMAGE_ROW_PITCH` right in the image-from-buffer pitch semantics discussed in KhronosGroup/OpenCL-Docs#674, which is still unresolved. Implementations legitimately return 0 there, so we cannot assert `width * element_size` for it - that is in fact exactly what fails on real drivers once the guard is removed.

rjodinchr: If the issue is with 1D_buffer, then I think we should make that clear and check for 1D_buffer, not everything except 1D_buffer.

shajder: Keeping the guard and adding an else for the remaining flat types (1D, 1D_BUFFER) turned out cleaner than dropping it: in the else we assert `slice == 0`, but intentionally do not assert row pitch there, where drivers legitimately return 0. The else also folds in the old standalone IMAGE1D slice check. Let me know if that works for you.

rjodinchr: No it does not.
We are now checking `outSlicePitch` twice for 1D_buffer. One time in the "else" that has been added, and right below it....

It is still unclear to me why we would single out IMAGE1D. It feels that it should be part of the normal path. And that path should be `imageInfo->type != CL_MEM_OBJET_IMAGE1D_BUFFER`, not an OR of everything else.

shajder: Just to be clear, by "checking `outSlicePitch` twice for 1D_buffer" you ment for IMAGE1D, right?

As I mentioned the `else` sections seems cleaner readability-wise (less intrusive changes to the code) but obviously we can go unified path with additional condition for special cases.

rjodinchr: Yes, I meant IMAGE1D.

Thank you for the change, it looks much better 🙏

test_conformance/images/clGetInfo/test_2D.cpp
rjodinchr: This is still relevant I believe. I don't think we want to single out COPY_HOST_PTR. At least that's not what I understand from the specification.

shajder: sorry for delay - it took some time to recall the origin of `CL_MEM_COPY_HOST_PTR` branch. Previously I assumed it lets the implementation relay out the data, so its row pitch was treated as implementation defined (non-zero only) while USE/default were checked exactly. Reviewing that, indeed it looks like `CL_IMAGE_ROW_PITCH` is **deterministic** for USE/COPY/default and the branch was an unnecessary relaxation of specification requirements.

Dropping it gives one deterministic check for all flags with 1D_BUFFER the only exception, left unasserted per #674.

rjodinchr (approved):

lakshmih (approved):

PR #2646, opened 2026-03-25
Fix offline compiler handling: skip missing subtests and guard cl_khr…

bananAshkar:
This PR skips three compiler subtests (get_program_info_kernel_names, get_linked_program_info_kernel_names, get_program_info_mult_devices) when using offline/binary compilation, because they rely on source creation or multi-device behavior that does not match that path. It also changes the SPIR test main so that if cl_khr_spir is missing, the suite logs a clear skip and exits successfully instead of failing.


bashbaug (approved):

bashbaug:
Merging as discussed in the April 28th teleconference.

PR #2647, opened 2026-03-27
Added floating point negation coverage

### LINKED ISSUE #2119: Missing floating point negation coverage
As a followup to the issue that was opened in: https://github.com/KhronosGroup/OpenCL-Docs/issues/1266

It was clarified regarding **floating point negation** that it should behave like every other floating point operation (i.e. add, sub, mul, div, fabs) in the sense that:
- Denorms are preserved on devices that support denorms
- Denorms may be flushed to zero on devices that don't support them.

A further conclusion of the discussion was that CTS coverage for the operation is encouraged to emphasize this behavior.

shajder:
Fixes #2119


svenvh (approved):

lakshmih (approved):

neiltrevett:

PR #2650, opened 2026-03-28
samplerlessReads: fix use_pitches test for devices with read_write images supported

karolherbst:
The test specified a image_row_pitch on the read_write image without a host_ptr or parent buffer.

The read_write image doesn't get inialized like the read_only does, so it's fine to just reset image_row_pitch back to 0 for its creation.


rjodinchr (commented):
LGTM, can you rebase on main and update the comments to refer to `ctx.enablePitch` instead of `gEnablePitch` which has been removed.

PR #2655, opened 2026-04-01
Add cl_khr_cooperative_matrix tests

svenvh:
Add tests for the `cl_khr_cooperative_matrix` extension.

The extension defines various operations on cooperative matrix types. The list of tested operations is in `cooperative_matrix.def`.  Each entry in that `.def` file results in a subtest.

Each subtest will iterate over all supported variants reported by the runtime and test each variant.  An individual variant can be run using the `--variant` command line option.

This CTS suite generates textual SPIR-V and uses spirv-tools to assemble the textual SPIR-V into a SPIR-V binary.


test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
```diff
@@ -0,0 +1,895 @@
...
+    OpCooperativeMatrixStoreKHR %out %result %layoutRes %strideRes
+            )";
+            break;
+
+        case CoopMatOp::copy:
+        case CoopMatOp::copy_stride0:
+        case CoopMatOp::multicomponent_load:
+        case CoopMatOp::multicomponent_store:
+            spirv_text << R"(
+    %matSrc = OpCooperativeMatrixLoadKHR )" << getResTy() << R"( %in)" << variant.order << R"( %layout)" << variant.order << R"( %stride)" << variant.order << R"(
```
joselopeqti, line 409: `matAty` should be the type always used for `OpCooperativeMatrixLoadKHR` in the copy tests. 
When a unary variant is created, `variant.inputA`, `variant.inputB` and `variant.inputC` matrices use the same size (rows x cols). 
For example, for a supported variant that uses MatrixA (64 x 16) and MatrixB (16 x 64), if the order of the copy operation is B, `variant.inputA` and `variant.inputB` will be 16x64. This results on the next SPIR-V being generated:
```  
%sizeM = OpConstant %i32 16
%sizeK = OpConstant %i32 64
%sizeN = OpConstant %i32 64
```

And then `matBty` will be used resulting in storing a 64x64 matrix in a buffer of size 16x64

svenvh: Nice catch, thanks!  Always using `matAty` would mean the `OpCooperativeMatrixLoadKHR` operations are only tested with `Use = MatrixAKHR`, whereas we should cover all `Use` values.  I'll prepare a different fix.

svenvh: I've pushed a fix now; let me know if this works for you.

test_conformance/extensions/cl_khr_cooperative_matrix/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+      cooperative_matrix.cpp
+      main.cpp
+      program.cpp
+  )
+
+  include(../../CMakeCommon.txt)
+  target_include_directories(${${MODULE_NAME}_OUT} PRIVATE ${SPIRV_TOOLS_INCLUDE_DIRS})
+  target_link_libraries(${${MODULE_NAME}_OUT} ${SPIRV_TOOLS_LDFLAGS})
+else()
+  message(STATUS "SPIRV-Tools not found; cl_khr_cooperative_matrix tests will not be built")
```
bashbaug, line 40: Asking as a review comment for better threading.

I am interested in this kind of SPIR-V tools integration for general-purpose SPIR-V testing via generated SPIR-V files vs. the current SPIR-V testing using pre-written .spvasm files.  Especially for cases where testing iterates over a combination of source and destination types, it's much more maintainable to generate each test case on-the-fly rather than commit tens (or more) different files.

My questions are specifically:

1. Are we happy with this general approach? Are we OK adding a built dependency on SPIR-V tools for general-purpose SPIR-V testing? Do we want to support building SPIR-V testing (specifically, `test_spirv_new`) without support for SPIR-V tools, or are we OK requiring SPIR-V tools for SPIR-V testing?
2. If so, are we OK separating out the SPIR-V tools integration into a separate PR? Is there anything we would want to change in the process?

Note that (2) is not needed if we can get this PR merged soon, but there hasn't been a lot of activity on it lately, so it will take some concerted effort and/or risk taking to get it merged in the near future.

bashbaug: Somewhat related: We might want to revisit https://github.com/KhronosGroup/OpenCL-CTS/pull/2527 if we proceed with SPIR-V tools integration, especially if we add a build dependency on SPIR-V tools for `test_spirv_new` but not for the entire CTS.  With access to SPIR-V tools, we could generate SPIR-V modules for the SPIR-V extensions, extended instruction sets, and capabilities returned by SPIR-V queries, but we'd want to do that testing as part of `test_spirv_new` and not as part of `test_api` (where the SPIR-V query testing is currently).

svenvh: I'm happy to factor the SPIR-V tools integration out into a separate PR.  I always felt that's a substantial change warranting a dedicated PR anyway.  As part of that, I can migrate an existing spirv_new test as a proof-of-concept, so that folks can test the new dependency in their environments without needing a cooperative matrix implementation.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+        int64_t min = CL_SHRT_MIN;
+        return { min, max };
+    }
+    if (type == CL_DEVICE_COOPERATIVE_MATRIX_COMPONENT_TYPE_SINT32_KHR)
+    {
+        int64_t max = CL_INT_MAX;
+        int64_t min = CL_INT_MIN;
+        return { min, max };
+    }
+    if (type == CL_DEVICE_COOPERATIVE_MATRIX_COMPONENT_TYPE_SINT32_KHR)
```
rjodinchr, line 815:
```suggestion
    if (type == CL_DEVICE_COOPERATIVE_MATRIX_COMPONENT_TYPE_SINT64_KHR)
```

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+            break;
+
+        case CoopMatOpKind::Ternary:
+            // For ternary operations (notably multiply-accumulate), derive the
+            // variants to test directly from the runtime-reported variants.
+            for (auto &rv : gTestContext->variants)
+            {
+                if (op == CoopMatOp::matrixmuladd_wrapping
+                    && isFloatType(rv.result_type))
+                    break;
```
rjodinchr, line 2409: That break statement would skip all the remaining variants
```suggestion
                    continue;
```

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+        {
+            return bounds;
+        }
+    }
+}
+
+template <typename T, std::enable_if_t<is_bounds_type_v<T>, int> = 0>
+T boundsIntersection(T left, T right)
+{
+    auto min = std::min(left.min, right.min);
```
rjodinchr, line 1019:
```suggestion
    auto min = std::max(left.min, right.min);
```

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+                // Signed inputs must be sign extended.
+                uint64_t a = isSignedType(inputs[0]->elementType)
+                    ? inputs[0]->get<int64_t>(AIndex)
+                    : inputs[0]->get<uint64_t>(AIndex);
+                uint64_t b = isSignedType(inputs[1]->elementType)
+                    ? inputs[1]->get<int64_t>(BIndex)
+                    : inputs[1]->get<uint64_t>(BIndex);
+
+                const uint64_t multiplication = (a * b) & RESULT_MASK;
+                acc = saturating_add(multiplication, acc, resultWidthInBytes,
```
rjodinchr, line 1916: Masking `(a * b) & RESULT_MASK` converts negative signed products (e.g. `-5` as 8-bit `0xFB` / `251`) into positive `uint64_t` values. In `saturating_add`, `const int64_t signedA = a` sees `251 > 0`, causing negative multiplication results to erroneously saturate to the positive maximum.

test_conformance/extensions/cl_khr_cooperative_matrix/halffp.hpp
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    HalfFP operator/(const HalfFP &o) const
+    {
+        return HalfFP(static_cast<float>(*this) / static_cast<float>(o));
+    }
+
+    bool operator!=(const HalfFP &o) const
+    {
+        return static_cast<float>(data) != static_cast<float>(o.data);
```
rjodinchr, line 66: I'm not sure this is correct.
In half `+0.0f == -0.0f`, which does not feel to be the case with that expression.
Why not use the same (almost) as for `==`?

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+        kernel, device, CL_KERNEL_LOCAL_SIZE_FOR_SUB_GROUP_COUNT,
+        sizeof(numSubgroups), (void *)&numSubgroups, sizeof(globalSize),
+        (void *)&globalSize, nullptr);
+    test_error_fail(err, "Failed to get required work group size");
+    variant.globalSize = globalSize;
+
+    // Enqueue the work. Cooperative matrices need full subgroups so we set the
+    // global size to the (local) size of a single subgroup.
+    err = clEnqueueNDRangeKernel(queue, kernel, 1, nullptr, &globalSize,
+                                 nullptr, 0, nullptr, nullptr);
```
rjodinchr, line 2206: I'm not sure to understand the comment, should this set the local size using `globalSize` as well?

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+            return "i32";
+        case CL_DEVICE_COOPERATIVE_MATRIX_COMPONENT_TYPE_UINT64_KHR:
+            return "i64";
+        default: return "???";
+    }
+}
+
+std::string
+describeBufferKind(cl_device_cooperative_matrix_component_type_khr kind)
+{
```
rjodinchr, line 566: Second definition of that function

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+        out << "]\n";
+    }
+
+    log_info("%s\n", out.str().c_str());
+}
+
+template <class Ty>
+Ty getHelper(size_t i,
+             cl_device_cooperative_matrix_component_type_khr elementType,
+             std::vector<uint8_t> data)
```
rjodinchr, line 665:
```suggestion
             const std::vector<uint8_t>& data)
```

test_conformance/extensions/cl_khr_cooperative_matrix/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+pkg_search_module(SPIRV_TOOLS SPIRV-Tools)
+if (NOT SPIRV_TOOLS_FOUND)
+  # then try locating SPIRV-Tools via cmake (the new way)
+  find_package(SPIRV-Tools)
+  find_package(SPIRV-Tools-tools)
+  if (SPIRV-Tools_FOUND AND SPIRV-Tools-tools_FOUND)
+    set(SPIRV_TOOLS_FOUND TRUE)
+    # check for the existence of library targets in the found packages
+    if(TARGET SPIRV-Tools-shared)
+      # use the shared libary target if present
```
rjodinchr, line 13:
```suggestion
      # use the shared library target if present
```

test_conformance/opencl_conformance_tests_full.csv
```diff
@@ -109,4 +109,5 @@ cl_ext_cxx_for_opencl,extensions/cl_ext_cxx_for_opencl/test_cl_ext_cxx_for_openc
 cl_khr_semaphore,extensions/cl_khr_semaphore/test_cl_khr_semaphore
 cl_khr_command_buffer,extensions/cl_khr_command_buffer/test_cl_khr_command_buffer
 cl_khr_command_buffer_mutable_dispatch,extensions/cl_khr_command_buffer/cl_khr_command_buffer_mutable_dispatch/test_cl_khr_mutable_dispatch
+cl_khr_cooperative Matrix,test_conformance/extensions/cl_khr_cooperative_matrix/test_cl_khr_cooperative_matrix
```
rjodinchr, line 112:
```suggestion
cl_khr_cooperative_matrix,extensions/cl_khr_cooperative_matrix/test_cl_khr_cooperative_matrix
```

.github/workflows/presubmit.yml
```diff
@@ -161,6 +177,8 @@ jobs:
                   CMAKE_OPENCL_LIBRARIES_OPTION="${CMAKE_OPENCL_LIBRARIES_OPTION} -lpthread"
               fi
           fi
+          PKG_CONFIG_LIBDIR='${{ github.workspace }}'/SPIRV-Tools/install/lib/pkgconfig/
+          PKG_CONFIG_PATH=
```
rjodinchr, line 181: We fetch and build SPIRV-Tools only for `aarch64` but set this for all target, are we sure about this?

test_conformance/extensions/cl_khr_cooperative_matrix/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+// Export a const pointer to the test context so it cannot be modified
+// during testing.
+const TestContext *gTestContext = &writableTestContext;
+
+int main(int argc, const char *argv[])
+{
+    if (parseTestArgs(&argc, argv) != TEST_PASS)
+    {
+        return TEST_FAIL;
+    }
```
rjodinchr, line 189: To be removed.
Use `runTestHarnessWithCheckAndParse` instead

test_conformance/extensions/cl_khr_cooperative_matrix/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+    err = clGetDeviceInfo(
+        device, CL_DEVICE_COOPERATIVE_MATRIX_POINTER_ALIGNMENT_KHR,
+        sizeof(cl_uint), &writableTestContext.devicePointerAlignment, nullptr);
+    test_error_fail(
+        err,
+        "clGetDeviceInfo for "
+        "CL_DEVICE_COOPERATIVE_MATRIX_POINTER_ALIGNMENT_KHR failed\n");
+    log_info("Pointer alignment is %d bytes.\n",
+             writableTestContext.devicePointerAlignment);
```
rjodinchr, line 60: We need `REQUIRE_EXTENSION("cl_khr_cooperative_matrix")` before that call I believe

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
```diff
@@ -0,0 +1,906 @@
...
+            }
+            else
+            {
+                convOp = "OpConvertUToF";
+            }
+        }
+        else
+        {
+            // it must be a bitwidth conversion
+            convOp = "OpUConvert";
```
rjodinchr, line 717: If the source integer is signed, `OpSConvert` must be emitted instead of `OpUConvert`. Using `OpUConvert` zero-extends signed negative integers, causing conversion tests for negative integers to fail.

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
```diff
@@ -0,0 +1,906 @@
...
+    %matB = OpCooperativeMatrixLoadKHR %matBty %inB %layoutB %strideB
+    %matC = OpCooperativeMatrixLoadKHR %matCty %inC %layoutC %strideC
+    %matAArr = OpCompositeConstruct %matAArrty %filledA %matA
+    %matBArr = OpCompositeConstruct %matBArrty %filledB %matB
+    %matCArr = OpCompositeConstruct %matCArrty %filledC %matC
+    %matAArr1 = OpCompositeExtract %matAty %matAArr 1
+    %matBArr1 = OpCompositeExtract %matBty %matBArr 1
+    %matCArr1 = OpCompositeExtract %matCty %matCArr 1
+
+    %result = OpCooperativeMatrixMulAddKHR %matCty %matA %matB %matC)" << muladdOperand << R"(
```
rjodinchr, line 591: Shouldn't it be using the `%matXArr1` instead of `%matX`?

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
```diff
@@ -0,0 +1,906 @@
...
+            break;
+        case CoopMatOp::constant:
+        case CoopMatOp::composite:
+            spirv_text << R"(
+    %fillValue = OpConstant %)"
+                       << showScalarType(variant.inputA.elementType)
+                       << " 123\n";
+            if (op == CoopMatOp::constant)
+                spirv_text << R"(
+    %result = OpConstantComposite %matAty %fillValue
```
rjodinchr, line 258: I'm not sure about the type here. What if `variant.order` is `OpB` or `OpC`?

test_conformance/extensions/cl_khr_cooperative_matrix/CMakeLists.txt
bashbaug: Awesome!  If you don't mind factoring out the SPIR-V tools integration into a separate PR, I'll definitely take you up on the offer.  I would be interested in using this functionality soon, though, so I don't mind doing this myself if it's going to be some time.  Just let me know - thanks!

svenvh: Here's an initial attempt: https://github.com/KhronosGroup/OpenCL-CTS/pull/2782
It might take a few iterations to get it working on CI, so I've created it as a draft for now.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+        case CoopMatOp::multicomponent_store:
+        case CoopMatOp::func:
+            variant.inputA.fill(0);
+            variant.inputB.fill(1);
+            variant.inputC.fill(2);
+            break;
+        case CoopMatOp::mul: {
+            MatrixType outType(variant.output.elementType, variant.output.nRows,
+                               variant.output.nCols, MatrixType::Use::Acc);
+            Bounds bounds = getBounds(outType.type);
```
joselopeqti, line 2037: This uses the bounds based only on the outputType, but the type of the Inputs can be different.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+            Bounds newBounds = std::visit(
+                [&](auto x) -> Bounds {
+                    auto min = x.min;
+                    auto max = x.max;
+                    auto maxMag = min >= 0 ? max : std::min(-min, max);
+                    // Allows some but not all to overflow.
+                    maxMag = std::sqrt(maxMag) / 4 * 5;
+                    x.max = maxMag;
+                    x.min = -maxMag;
```
joselopeqti, line 2046: In the case of unsigned bounds, `x.min = -maxMag;` could produce a large positive value instead.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+                    return x;
+                },
+                bounds);
+            variant.inputA.fill(0, newBounds);
+            variant.inputB.fill(1, newBounds);
+            variant.inputC.fill(2, newBounds);
+            break;
+        }
+        case CoopMatOp::matrixmuladd:
+        case CoopMatOp::matrixmuladd_array: {
```
joselopeqti, line 2056: Same comments as the ones for `CoopMatOp::mul`

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Oops, thanks.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Thanks, fixed.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Thanks, fixed.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Nice catch, thanks.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Thanks, fixed by doing sign extension in `saturating_add`.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: This is the ordinary `OpIMul` case, for which we only construct variants where input and output types match (i.e., the first `Variant` constructor which takes a single type).

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Thanks, fixed.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Done.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Good point, fixed.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
svenvh: Nice catch, fixed.

test_conformance/extensions/cl_khr_cooperative_matrix/halffp.hpp
svenvh: Indeed this was wrong, thanks.

> Why not use the same (almost) as for ==?

`operator==` is defined in terms of `operator!=`, so we need to cast to float to avoid infinite recursion.

test_conformance/extensions/cl_khr_cooperative_matrix/main.cpp
svenvh: Done, thanks.

test_conformance/opencl_conformance_tests_full.csv
svenvh: Done.

test_conformance/extensions/cl_khr_cooperative_matrix/main.cpp
svenvh: Done.

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
svenvh: Thanks, fixed.

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
svenvh: Indeed, fixed.

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
svenvh: Thanks, fixed.

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
```diff
@@ -0,0 +1,897 @@
...
+
+    %numElems = OpConstant %i32 )"
+                       << variant.output.elementCount() << R"(
+    %arrayTy = OpTypeArray %)"
+                       << showScalarType(variant.output.elementType)
+                       << R"( %numElems
+    %arrayPtr = OpTypePointer Function %arrayTy
+    %workgroupPtr = OpTypePointer Workgroup %arrayTy
+    %sharedBufferLoad = OpVariable %workgroupPtr Workgroup
+    %sharedBufferStore = OpVariable %workgroupPtr Workgroup
```
joselopeqti: Do we need two local buffers for this test?

E.g. consider a device that reports the minimum value for [CL_DEVICE_LOCAL_MEM_SIZE](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_DEVICE_LOCAL_MEM_SIZE) (32KB). In this case, if a matrix of size 64 x 64 and type `float` is supported, this test would already require those 32KB only for these two buffers.

svenvh: Actually not, I've simplified the SPIR-V for the copy_workgroup test.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+
+    // We need to map the buffer to learn the actual alignment.
+    void *bufPointer =
+        clEnqueueMapBuffer(queue, bufHandle, CL_TRUE, CL_MAP_READ, 0, bufSize,
+                           0, nullptr, nullptr, &err);
+    test_error_ret(err, "Unable to map buffer.\n", std::nullopt);
+    const uintptr_t bufAddress = reinterpret_cast<uintptr_t>(bufPointer);
+    const uint32_t alignmentRemainder = bufAddress % bufAlignment;
+    const uint32_t alignmentPadding =
+        alignmentRemainder == 0 ? 0 : bufAlignment - alignmentRemainder;
```
rjodinchr, line 356: Are we sure `clCreateSubBuffer` allows an arbitrary `alignmentPadding`? According to the OpenCL spec, for buffers created without `CL_MEM_USE_HOST_PTR`, origin must be aligned to `CL_DEVICE_MEM_BASE_ADDR_ALIGN`, wouldn't passing an arbitrary small offset (e.g. 4 or 8 bytes) fail with `CL_MISALIGNED_SUB_BUFFER_OFFSET`?

Also, since `bufPointer` is a host address returned by `clEnqueueMapBuffer`, does its address modulo actually tell us anything about the device buffer's alignment on discrete GPUs? I don't think so, but I'm not sure.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+                case CL_DEVICE_COOPERATIVE_MATRIX_COMPONENT_TYPE_UINT8_KHR:
+                case CL_DEVICE_COOPERATIVE_MATRIX_COMPONENT_TYPE_UINT16_KHR:
+                case CL_DEVICE_COOPERATIVE_MATRIX_COMPONENT_TYPE_UINT32_KHR:
+                case CL_DEVICE_COOPERATIVE_MATRIX_COMPONENT_TYPE_UINT64_KHR:
+                    UnsignedBounds uBounds = bounds
+                        ? (assert(
+                               std::holds_alternative<UnsignedBounds>(*bounds)),
+                           std::get<UnsignedBounds>(*bounds))
+                        : getUBounds(elementType);
+                    uint64_t value = genrand_int64(rng) % uBounds.max;
```
rjodinchr, line 1059: Could `genrand_int64(rng) % uBounds.max` trigger a division by zero if `uBounds.max == 0`?

Also, shouldn't `uBounds.`min be taken into account here like it is for SignedBounds? Similarly, in SignedBounds, could `sBounds.max - sBounds.min` divide by zero if `max == min`?

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.hpp
```diff
@@ -0,0 +1,597 @@
...
+              type, elementsPerStride(outputLayout, nRows, nCols),
+              numberOfStrides(outputLayout, nRows, nCols),
+              IndexedBufferElementType<1>(type))),
+          layoutA(inputLayout), layoutB(inputLayout), layoutC(inputLayout),
+          layoutRes(outputLayout), strideA(inputADesc.stride),
+          strideB(inputBDesc.stride), strideC(inputCDesc.stride),
+          strideRes(outputDesc.stride), isConversion(false),
+          isSaturating(false), isMulticomponent(false)
+    {
+        assert(o != OperandOrder::OpABC
```
rjodinchr, line 403: Shouldn't `strideA` (and the other strides) include `stridePadding`? `inputADesc.stride` only contains the unpadded element count, but `Matrix::cooperativeMatrixLoad` advances rows by `bufStrideSize` (which includes padding). Since the SPIR-V spec defines the `Stride` operand to `OpCooperativeMatrixLoadKHR` as the distance between consecutive rows/columns in memory elements, won't passing the unpadded stride cause the kernel to access rows at the wrong memory offsets whenever `stridePadding > 0`?

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
```diff
@@ -0,0 +1,874 @@
...
+        variant.inputC.elementType == variant.output.elementType ? "%matCty"
+                                                                 : "%matResty";
+
+    // clang-format off
+    switch (op)
+    {
+        case CoopMatOp::length:
+            spirv_text << R"(
+    %1 = OpCooperativeMatrixLengthKHR %i32 )" << getResTy() << R"(
+    OpStore %out_slid_offset %1 Aligned 32)";
```
rjodinchr, line 396: What happens in `spirv_length` if the device's subgroup size (globalSize) is larger than `nRows * nCols` (e.g. subgroup size 64 for a 4x4 matrix)? In the kernel, each work-item stores at `%out[slid]`, but the buffer is only allocated for `nRows * nCols * sizeof(uint32_t)`. Wouldn't work-items with `slid >= nRows * nCols` write past the end of the buffer? Shouldn't the output buffer be allocated with `std::max(elementCount, globalSize) * sizeof(uint32_t)`?

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
```diff
@@ -0,0 +1,874 @@
...
+
+    spirv_text << "    %matAty = OpTypeCooperativeMatrixKHR %"
+               << showScalarType(variant.inputA.elementType)
+               << " %i32_subgroup " << getDims(0) << " %i32_useA\n"
+               << "    %matBty = OpTypeCooperativeMatrixKHR %"
+               << showScalarType(variant.inputB.elementType)
+               << " %i32_subgroup " << getDims(1) << " %i32_useB\n"
+               << "    %matCty = OpTypeCooperativeMatrixKHR %"
+               << showScalarType(variant.inputC.elementType)
+               << " %i32_subgroup " << getDims(2) << " %i32_useC\n";
```
rjodinchr, line 618: Could unconditionally declaring `%matAty`, `%matBty`, and `%matCty` for non-ternary operations emit matrix types that the device doesn't support? For example, when testing matrix `B` with shape `K*N`, `%matAty` is declared with `Use::A` and shape `K*N`. If the device only supports `Use::A` with `M*K` (`M!=K`), wouldn't the driver reject the module with `CL_BUILD_PROGRAM_FAILURE`? Shouldn't we only declare the matrix type actually used by the operation?

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
```diff
@@ -0,0 +1,2512 @@
...
+                             &buf.subBufferHandle);
+        test_error_fail(err, "Unable to set kernel argument");
+        i++;
+    }
+
+    // Query the local size needed for a single subgroup, so we can
+    // use this as the global size for our work.
+    const size_t numSubgroups = 1;
+    size_t globalSize;
+    err |= clGetKernelSubGroupInfo(
```
rjodinchr, line 2196: I don't think we want to OR the error code here.
```suggestion
    err = clGetKernelSubGroupInfo(
```

test_conformance/extensions/cl_khr_cooperative_matrix/program.cpp
```diff
@@ -0,0 +1,874 @@
...
+        case CoopMatOp::add:
+        case CoopMatOp::sub:
+        case CoopMatOp::mul:
+        case CoopMatOp::div: {
+            std::ostringstream binOp;
+            binOp << "Op"
+                  << (isFloatType(variant.output.elementType) ? "F" :
+                      op != CoopMatOp::div ? "I" :
+                      isSignedIntType(variant.output.elementType) ? "S" : "U")
+                  << (op == CoopMatOp::negate ? "Negate" :
```
rjodinchr, line 511: Can `op == CoopMatOp::negate` ever be reached here? Since `CoopMatOp::negate` is already handled in its own case earlier, can we remove this check?

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.hpp
```diff
@@ -0,0 +1,597 @@
...
+    {
+        assert((o == OperandOrder::OpA || o == OperandOrder::OpB
+                || o == OperandOrder::OpC)
+               && "constructor must not be used for non-unary operations");
+    }
+
+    // Construct a Variant for a conversion.
+    Variant(cl_device_cooperative_matrix_component_type_khr srcType,
+            cl_device_cooperative_matrix_component_type_khr dstType,
+            cl_uint nRows, cl_uint nCols, MatrixType::Use u, OperandOrder o,
```
rjodinchr, line 446: Is the parameter `MatrixType::Use u` needed in this `Variant` constructor? It doesn't look like it's used in the constructor body.

test_conformance/extensions/cl_khr_cooperative_matrix/cooperative_matrix.cpp
rjodinchr: what is the difference between this function and `showScalarType`? They look identical to me, apart from the return type

Issue #1667, opened 2023-03-09
Add testing for use of device-side enqueue in a command-buffer


EwanC:
The test plan for command-buffers in https://github.com/KhronosGroup/OpenCL-CTS/issues/1369 missed out testing of the `CL_COMMAND_BUFFER_CAPABILITY_DEVICE_SIDE_ENQUEUE_KHR` capability from [cl_khr_command_buffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Ext.html#cl_khr_command_buffer).  This reports that an implementation supports the ability to use kernel commands in a command-buffer which use [device-side enqueue](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#device-side-enqueue).

It is likely that this feature will be less well-supported than an equivalent feature like the printf capability, so finding an implementation to write tests against may be an issue.

shajder:
@EwanC do we need any negative or mutable test for this issue ?

EwanC:
I think we already have a negative test for this in [CommandNDRangeKernelWithKernelEnqueueCall](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/extensions/cl_khr_command_buffer/negative_command_nd_range_kernel.cpp#L351) and this characteristic isn't something that's mutable so the scope of what you have in your PR should be fine for completing this task. I'll go review that PR 👀

shajder:
@EwanC can we close this issue or is there anything more to add? Thanks

PR #2657, opened 2026-04-02
Add device-side execution test for cl_khr_command_buffer

shajder:
Related to #1667

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
@@ -33,7 +33,7 @@ struct CommandBufferDeviceExecutionTest : public BasicCommandBufferTest
                                      cl_command_queue queue)
         : BasicCommandBufferTest(device, context, queue)
     {
-        buffer_size_multiplier = device_execution_size_mult;
+        buffer_size_multiplier = 2;
     }
 
     bool Skip() override
@@ -79,20 +79,13 @@ struct CommandBufferDeviceExecutionTest : public BasicCommandBufferTest
         cl_int error = BasicCommandBufferTest::SetUp(elements);
         test_error(error, "BasicCommandBufferTest::SetUp failed");
 
-        cl_uint preferred_size = 0;
-        error =
-            clGetDeviceInfo(device, CL_DEVICE_QUEUE_ON_DEVICE_PREFERRED_SIZE,
-                            sizeof(preferred_size), &preferred_size, NULL);
-        test_error(error,
-                   "Unable to query CL_DEVICE_QUEUE_ON_DEVICE_PREFERRED_SIZE");
-
         // create default device-side queue
         cl_queue_properties dev_queue_props[] = {
             CL_QUEUE_PROPERTIES,
             (cl_queue_properties)(CL_QUEUE_ON_DEVICE
                                   | CL_QUEUE_ON_DEVICE_DEFAULT
                                   | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE),
-            CL_QUEUE_SIZE, preferred_size, 0
+            0
         };
         device_queue = clCreateCommandQueueWithProperties(
             context, device, dev_queue_props, &error);
@@ -160,7 +153,7 @@ struct CommandBufferDeviceExecutionTest : public BasicCommandBufferTest
                            sizeof(cl_int), &offset, &error);
         test_error(error, "clCreateBuffer failed");
 
-        cl_int in_arg = device_execution_size_mult;
+        cl_int in_arg = buffer_size_multiplier;
         error = clSetKernelArg(kernel, 0, sizeof(cl_int), &in_arg);
         test_error(error, "clSetKernelArg failed");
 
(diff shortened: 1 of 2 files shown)


test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
```diff
@@ -0,0 +1,231 @@
...
+        test_error(error,
+                   "Unable to query CL_DEVICE_QUEUE_ON_DEVICE_PREFERRED_SIZE");
+
+        // create default device-side queue
+        cl_queue_properties dev_queue_props[] = {
+            CL_QUEUE_PROPERTIES,
+            (cl_queue_properties)(CL_QUEUE_ON_DEVICE
+                                  | CL_QUEUE_ON_DEVICE_DEFAULT
+                                  | CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE),
+            CL_QUEUE_SIZE, preferred_size, 0
```
EwanC: Can we skip manually setting `CL_QUEUE_SIZE` to make the test less verbose?

The spec says

> If [CL_QUEUE_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_QUEUE_SIZE) is not specified, the device queue is created with [CL_DEVICE_QUEUE_ON_DEVICE_PREFERRED_SIZE](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_DEVICE_QUEUE_ON_DEVICE_PREFERRED_SIZE) as the size of the queue.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
```diff
@@ -0,0 +1,231 @@
...
+                                 nullptr, &error);
+        test_error(error, "clCreateBuffer failed");
+
+        cl_int offset = 0;
+        off_mem =
+            clCreateBuffer(context, CL_MEM_READ_ONLY | CL_MEM_COPY_HOST_PTR,
+                           sizeof(cl_int), &offset, &error);
+        test_error(error, "clCreateBuffer failed");
+
+        cl_int in_arg = device_execution_size_mult;
```
EwanC: it's a bit confusing using `device_execution_size_mult` here but `buffer_size_multiplier` multiplier above to represent the same thing, would update this to be consistent.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
shajder: Done

EwanC (dismissed):
I haven't got a device handy that can run this test but LGTM

test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+        error = create_single_kernel_helper_create_program(context, &program, 1,
+                                                           &kernel_str);
+        test_error(error, "Failed to create program with source");
+
+        error = clBuildProgram(program, 1, &device, nullptr, nullptr, nullptr);
+        test_error(error, "Failed to build program");
+
+        kernel = clCreateKernel(program, "device_enqueue", &error);
+        test_error(error, "Failed to create device execution kernel");
```
bashbaug: This is overly verbose and it doesn't work - device-side enqueue requires OpenCL C 2.0 or newer, and passing `nullptr` as build options means we're getting the default latest version of OpenCL C 1.x.

How about instead:

```suggestion
        error = create_single_kernel_helper(context, &program, &kernel, 1,
                                            &kernel_str, "device_enqueue");
        test_error(error, "Failed to create test kernel");
```

test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+        return CL_SUCCESS;
+    }
+
+    clCommandQueueWrapper device_queue;
+};
+
+} // anonymous namespace
+
+REGISTER_TEST(device_execution)
```
bashbaug: Nitpick: I think the name of this test is misleading, since almost all of the command buffer tests (except maybe the negative tests) are executing on the device.

Should the name of this test be "device_enqueue" instead?

bashbaug (changes requested):
I didn't do a very thorough review but I did ensure that this test works on the Intel CPU OpenCL device when the OpenCL C version issue is fixed.

test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
shajder: Done

test_conformance/extensions/cl_khr_command_buffer/command_buffer_device_execution.cpp
shajder: Done

bashbaug (approved):

bashbaug:
Merging (offline) as discussed in the May 5th teleconference.

PR #2659, opened 2026-04-08
Add cl_khr_cooperative_matrix to known extensions list

svenvh:
Avoid spurious failures for the new extension while the extension tests themselves are still under review in #2655 .

### DIFF
--- a/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
+++ b/test_conformance/compiler/test_compiler_defines_for_extensions.cpp
@@ -99,7 +99,8 @@ const char *known_extensions[] = {
     "cl_khr_command_buffer_multi_device",
     "cl_khr_external_memory_android_hardware_buffer",
     "cl_khr_unified_svm",
-    "cl_khr_spirv_queries"
+    "cl_khr_spirv_queries",
+    "cl_khr_cooperative_matrix",
 };
 // clang-format on
 


test_conformance/compiler/test_compiler_defines_for_extensions.cpp
```diff
@@ -104,7 +105,7 @@ const char *known_extensions[] = {
 // clang-format on
 
 size_t num_known_extensions = ARRAY_SIZE(known_extensions);
-size_t first_API_extension = 32;
+size_t first_API_extension = 34;
```
bashbaug: This is consistent with the comment in the table above, but I think the prior change to add the `cl_khr_icd_unloadable` extension was incorrect, so it's in the wrong position in the table (`cl_khr_icd_unloadable` does not modify OpenCL C, so it should be considered an "API-only" extension).  

Because `first_API_extension` was left unmodified, it wasn't causing a problem previously, but I think it could be after incrementing this value by two, since we would now be checking for a compiler define for this extension.

I think we should move the entry for `cl_khr_icd_unloadable` lower in the table, perhaps immediately below `cl_khr_icd`, then we would only need to increment this value by one.

However, even the addition for `cl_khr_cooperative_matrix` is in a bit of a weird state.  Right now, it just adds SPIR-V support and API queries.  Should it be considered an "API-only" extension, at least currently?  Or, do we expect there to be an OpenCL C compiler define for it, even though the current extension does not add any new OpenCL C types or functions?

svenvh: > `cl_khr_icd_unloadable` does not modify OpenCL C, so it should be considered an "API-only" extension

That's a very good point; I was blindly assuming the extension was added in the right place without looking at the extension itself.  I've opened [a separate PR ](https://github.com/KhronosGroup/OpenCL-CTS/pull/2661) to fix this, to keep it separate from any cooperative matrix work in the git history.

> However, even the addition for cl_khr_cooperative_matrix is in a bit of a weird state. Right now, it just adds SPIR-V support and API queries. Should it be considered an "API-only" extension, at least currently? Or, do we expect there to be an OpenCL C compiler define for it, even though the current extension does not add any new OpenCL C types or functions?

I agree it would be weird to require an OpenCL C compiler define without any new OpenCL C types or functions.  It sounds reasonable to make it an API-only extension.

bashbaug (approved):

joselopeqti (approved):

bashbaug:

PR #2660, opened 2026-04-08
Fix memory leak in compiler_defines_for_extensions

svenvh:
`extensions` was not freed on all possible error paths.  Fix by converting to a `std::vector`, so that manual freeing is no longer needed.


bashbaug (approved):

rjodinchr (approved):

bashbaug:

PR #2663, opened 2026-04-15
[NFC] math_brute_force: use common TestInfoBase base

svenvh:
Use `TestInfoBase` as the base for all `TestInfo` classes in math_brute_force, so that individual files do not have to declare common fields.  This reduces code duplication.


EwanC (approved):

rjodinchr (approved):

bashbaug:

PR #2664, opened 2026-04-16
Add dynamic PR testing with golden reference comparison

rjodinchr:
This commit introduces a new targeted testing capability to the GitHub Actions pipeline, allowing developers to run specific CTS commands on Pull Requests and automatically compare the results against a known baseline.

Key additions and changes:
* Dynamic Test Triggers: Added a `tests` job to `presubmit.yml` that parses PR descriptions and commit messages for the `[run-test: <command>]` syntax. The CI environment is only provisioned and executed if a trigger is detected.
* pocl Test Environment: The workflow sets up an Ubuntu 24.04 runner, installs necessary dependencies (LLVM 20, Vulkan SDK), and builds `pocl` (v7.1), OpenCL-ICD-Loader, and the OpenCL CTS with experimental features enabled.
* Automated Result Comparison: Added `ci/compare_results.py`, a Python utility that compares the JSON output of the triggered tests against a predefined golden reference. It flags missing references and categorizes differences as "FIX", "REGRESSION", or "DIFFERENCE". Any differences marks the check as failing.
* Golden Baseline: Introduced `ci/pocl/golden.json` to store the expected outcomes (pass, fail, skip) for tests on the `pocl` implementation. It is initially populated with baseline expectations for the `test_computeinfo` command. It will need to be populated by PR using the feature for all the other tests in the future.

[run-test: test_computeinfo]

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -182,7 +182,6 @@ jobs:
   tests:
     name: Run Tests (pocl)
     runs-on: ubuntu-24.04
-    if: github.event_name == 'pull_request'
     steps:
       - uses: actions/checkout@v6
       - name: Parse triggers
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -178,6 +178,134 @@ jobs:
                 -DVULKAN_LIB_DIR='${{ github.workspace }}'/Vulkan-Loader/build/loader/ \
                 ${CMAKE_ADDITIONAL_CONFIG_ARGS}
           cmake --build . --parallel
+
+  tests:
+    name: Run Tests (pocl)
+    runs-on: ubuntu-24.04
+    if: github.event_name == 'pull_request'
+    steps:
+      - uses: actions/checkout@v6
+      - name: Parse triggers
+        id: check_triggers
+        env:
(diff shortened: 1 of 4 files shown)


bashbaug:
Discussed in the May 19th teleconference.  We will wait one more week for review comments and aim to merge in the May 26th teleconference.

franz:
I've taken a quick look, and noticed one problem. The `Build pocl` step uses:

```
cmake -B pocl/build -S pocl -G Ninja \
            -DCMAKE_INSTALL_PREFIX=${{ env.pocl }} \
            -DCMAKE_BUILD_TYPE=Release \
            -DENABLE_ICD=ON
```

... this will result in autodetection of CPU at CMake time, and the builtin library will be built using flags for a single particular CPU. Since the build is cached only on `key: ubuntu-24.04-pocl-v7.1`, the cache could be then restored on Github machine with different CPU, which would likely result in SIGILL at runtime. There are multiple way to resolve this, in our CI jobs I have used this cmake option: `-DDKERNELLIB_HOST_CPU_VARIANTS=distro`

FWIW, Github currently uses machines with at least two CPU types (Zen3 and Zen4; only the latter has AVX512).

franz (approved):

bashbaug (approved):

PR #2665, opened 2026-04-16
Added edge case behavior testing for math bruteforce

shajder:
Fixes #2641

### DIFF
--- a/test_conformance/math_brute_force/edge_cases.cpp
+++ b/test_conformance/math_brute_force/edge_cases.cpp
@@ -531,7 +531,7 @@ inline cl_int run_accumulated_cases(cl_context context, cl_command_queue queue)
                     }
                     log_error(") - expected NaN, got 0x");
 
-                    for (std::size_t i = ec.expected.byte_size; i !=0; --i)
+                    for (std::size_t i = ec.expected.byte_size; i != 0; --i)
                         log_error("%02x", result[byte_offset + i - 1]);
                     log_error("\n");
                     err = -1;
@@ -554,7 +554,7 @@ inline cl_int run_accumulated_cases(cl_context context, cl_command_queue queue)
                     log_anyvalue(ec.expected);
 
                     log_error(", got 0x");
-                    for (std::size_t i = ec.expected.byte_size; i !=0; --i)
+                    for (std::size_t i = ec.expected.byte_size; i != 0; --i)
                         log_error("%02x", result[byte_offset + i - 1]);
                     log_error("\n");
                     err = -1;
@@ -635,16 +635,15 @@ template <typename T> EdgeCaseSpec make_edge_case(const AbstractEdgeCase &aec)
 }
 
 template <typename T>
-cl_int flush_group (const AbstractEdgeCase *cases, std::size_t count,
-                      cl_context context, cl_command_queue queue, std::size_t i)
+cl_int flush_group(const AbstractEdgeCase *cases, std::size_t count,
+                   cl_context context, cl_command_queue queue, std::size_t i)
 {
-    cl_int ret=0;
-    if (!batch_cases.empty() &&
-        ((i == count - 1) ||
-         std::strcmp(cases[i].func_name, cases[i+1].func_name) != 0))
+    cl_int ret = 0;
+    if (!batch_cases.empty()
+        && ((i == count - 1)
+            || std::strcmp(cases[i].func_name, cases[i + 1].func_name) != 0))
     {
-        if (run_accumulated_cases<T>(context, queue) != CL_SUCCESS)
-            ret = -1;
+        if (run_accumulated_cases<T>(context, queue) != CL_SUCCESS) ret = -1;
         batch_cases.clear();
         kernel_src.clear();
     }
@@ -662,20 +661,20 @@ inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
         for (std::size_t i = 0; i < count; ++i)
         {
             const auto &aec = cases[i];
-            bool skip=false;
+            bool skip = false;
             if (gIsEmbedded)
             {
                 if (aec.requires_denorm && !(gFloatCapabilities & CL_FP_DENORM))
                 {
                     log_info("SKIP (no CL_FP_DENORM): %s\n", aec.func_name);
-                    skip=true;
+                    skip = true;
                 }
 
                 if (aec.requires_inf_nan
                     && !(gFloatCapabilities & CL_FP_INF_NAN))
                 {
                     log_info("SKIP (no CL_FP_INF_NAN): %s\n", aec.func_name);
-                    skip=true;
+                    skip = true;
                 }
 
                 if (aec.requires_rte
@@ -683,17 +682,18 @@ inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
                 {
                     log_info("SKIP (no CL_FP_ROUND_TO_NEAREST): %s\n",
                              aec.func_name);
-                    skip=true;
+                    skip = true;
                 }
             }
 
-            if(!skip)
+            if (!skip)
             {
                 const EdgeCaseSpec ec = make_edge_case<cl_float>(aec);
                 accumulate_edge_case(ec);
             }
 
-            if(flush_group<cl_float>(cases, count, context, queue, i)!=CL_SUCCESS)
+            if (flush_group<cl_float>(cases, count, context, queue, i)
+                != CL_SUCCESS)
                 overall = -1;
         }
     }
@@ -707,17 +707,17 @@ inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
         for (std::size_t i = 0; i < count; ++i)
         {
             const auto &aec = cases[i];
-            bool skip=false;
+            bool skip = false;
             if (aec.requires_denorm && !(gHalfCapabilities & CL_FP_DENORM))
             {
                 log_info("SKIP fp16 (no CL_FP_DENORM): %s\n", aec.func_name);
-                skip=true;
+                skip = true;
             }
 
             if (aec.requires_inf_nan && !(gHalfCapabilities & CL_FP_INF_NAN))
             {
                 log_info("SKIP fp16 (no CL_FP_INF_NAN): %s\n", aec.func_name);
-                skip=true;
+                skip = true;
             }
 
             if (aec.requires_rte
@@ -725,7 +725,7 @@ inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
             {
                 log_info("SKIP fp16 (no CL_FP_ROUND_TO_NEAREST): %s\n",
                          aec.func_name);
-                skip=true;
+                skip = true;
             }
 
             if (!skip)
@@ -748,17 +748,17 @@ inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
         for (std::size_t i = 0; i < count; ++i)
         {
             const auto &aec = cases[i];
-            bool skip=false;
+            bool skip = false;
             if (aec.requires_denorm && !(gDoubleCapabilities & CL_FP_DENORM))
             {
                 log_info("SKIP fp64 (no CL_FP_DENORM): %s\n", aec.func_name);
-                skip=true;
+                skip = true;
             }
 
             if (aec.requires_inf_nan && !(gDoubleCapabilities & CL_FP_INF_NAN))
             {
                 log_info("SKIP fp64 (no CL_FP_INF_NAN): %s\n", aec.func_name);
-                skip=true;
+                skip = true;
             }
 
             if (aec.requires_rte
@@ -766,16 +766,17 @@ inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
             {
                 log_info("SKIP fp64 (no CL_FP_ROUND_TO_NEAREST): %s\n",
                          aec.func_name);
-                skip=true;
+                skip = true;
             }
 
-            if(!skip)
+            if (!skip)
             {
                 const EdgeCaseSpec ec = make_edge_case<cl_double>(aec);
                 accumulate_edge_case(ec);
             }
 
-            if(flush_group<cl_double>(cases, count, context, queue, i)!=CL_SUCCESS)
+            if (flush_group<cl_double>(cases, count, context, queue, i)
+                != CL_SUCCESS)
                 overall = -1;
         }
     }
--- a/test_conformance/math_brute_force/edge_cases.cpp
+++ b/test_conformance/math_brute_force/edge_cases.cpp
@@ -378,6 +378,10 @@ inline void accumulate_edge_case(const EdgeCaseSpec &ec)
         }
         src += ")\n{";
     }
+    else
+    {
+        assert(batch_cases.front().inputs.size() == ec.inputs.size());
+    }
 
     src += "\n    out[";
     src += std::to_string(ind);
@@ -457,7 +461,7 @@ inline cl_int run_accumulated_cases(cl_context context, cl_command_queue queue)
                       * batch_cases.size());
 
         size_t byte_offset = 0;
-        for (auto &elem : batch_cases)
+        for (const auto &elem : batch_cases)
         {
             std::memcpy(&inData[byte_offset], elem.inputs[i].data.data(),
                         elem.inputs[i].byte_size);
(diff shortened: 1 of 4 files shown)


test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+            }
+        }
+    }
+
+    return err == -1 ? TEST_FAIL : TEST_PASS;
+}
+
+template <typename T> AnyValue abstract_to_anyvalue(const AbstractValue &av)
+{
+    // Int jest niezależny od precyzji — zawsze cl_int
```
svenvh: Suggesting to translate this comment to English. :)
(and avoid the non-ASCII em dash)

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+    std::size_t byte_size = 0;
+    std::array<uint8_t, CL_VALUE_MAX_BYTES> data{};
+
+    template <typename T> static AnyValue make(const T &val)
+    {
+        static_assert(sizeof(T) <= CL_VALUE_MAX_BYTES,
+                      "CLValue: type exceeds CL_VALUE_MAX_BYTES");
+        AnyValue v;
+        v.cl_type = "half";
+        if constexpr (std::is_same_v<T, cl_float>)
```
svenvh, line 50: Needs a `#include <type_traits>`.

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+const AbstractValue TWO = AV_F(2.0);
+const AbstractValue NEG_TWO = AV_F(-2.0);
+
+struct AbstractEdgeCase
+{
+    const char *func_name;
+    std::vector<AbstractValue> inputs;
+    AbstractValue expected;
+
+    bool expect_nan = false; // check isnan() instead of memcmp
```
svenvh: where is `isnan` being called?

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+        ec.inputs.push_back(abstract_to_anyvalue<T>(av));
+    ec.func_name = aec.func_name;
+    return ec;
+}
+
+inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
+                             cl_context context, cl_command_queue queue)
+{
+    cl_int overall = CL_SUCCESS;
+    log_info("float\n");
```
svenvh: We should probably gate this on `gTestFloat` so that it can be toggled with `-f` like all other tests.

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+    AbstractValue expected;
+
+    bool expect_nan = false; // check isnan() instead of memcmp
+
+    bool requires_inf_nan = false; // CL_FP_INF_NAN
+    bool requires_denorm = false; // CL_FP_DENORM
+    bool requires_rte = false; // CL_FP_ROUND_TO_NEAREST
+};
+
+static const AbstractEdgeCase edge_case_table[] = {
```
svenvh, line 174: Maybe worth adding something like:

```suggestion
// Taken from OpenCL C Section 7.5.1. Additional Requirements Beyond ...
static const AbstractEdgeCase edge_case_table[] = {
```

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+            AnyValue got;
+            got.byte_size = ec.expected.byte_size;
+            std::memcpy(got.data.data(), result.data(), got.byte_size);
+            if (!got.all_elements_nan<T>())
+            {
+                log_error("FAIL: %s(", ec.func_name);
+                for (std::size_t i = 0; i < ec.inputs.size(); ++i)
+                {
+                    if (i) log_error(", ");
+                    log_error("0x");
```
svenvh: Doesn't `log_anyvalue` already print the `0x`?
Please check other `log_anyvalue` calls too.

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+                    log_anyvalue(ec.inputs[i]);
+                }
+                log_error(") — expected NaN, got 0x");
```
svenvh:
```suggestion
                log_error(") - expected NaN, got 0x");
```

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+}
+
+inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
+                             cl_context context, cl_command_queue queue)
+{
+    cl_int overall = CL_SUCCESS;
+    log_info("float\n");
+    for (std::size_t i = 0; i < count; ++i)
+    {
+        auto aec = cases[i];
```
svenvh: Does this really need to be a copy?
Same for the half and double cases further down.

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+    bool expect_nan = false; // check isnan() instead of memcmp
+
+    bool requires_inf_nan = false; // CL_FP_INF_NAN
+    bool requires_denorm = false; // CL_FP_DENORM
+    bool requires_rte = false; // CL_FP_ROUND_TO_NEAREST
+};
+
+static const AbstractEdgeCase edge_case_table[] = {
+
+    { "acospi", { ONE }, POS_ZERO },
```
svenvh, line 176: This table only covers the cases listed in the OpenCL C specification as fas as I can tell?  Should we also check for the cases from Section F.9 of the C99 Specification?  I'd suggest keeping the OpenCL and C99 lists somehow distinct (i.e., not interleaved).

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+    }
+
+    return overall;
+}
+
+} // anonymous namespace
+
+REGISTER_TEST(math_edge_cases)
+{
+    return run_edge_cases(
```
svenvh, line 721: The `-l` flag has no effect for this test currently.

Maybe we should simply skip the edge case tests entirely if `gSkipCorrectnessTesting` is set.  Any sort of "link check" is already covered by the other individual tests.

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -0,0 +1,708 @@
...
+                std::memcpy(&bits, v.data.data() + off, 2);
+                log_error("0x%04" PRIx16, bits);
+                break;
+            }
+        }
+    }
+}
+
+template <typename T>
+inline cl_int run_edge_case(const EdgeCaseSpec &ec, cl_context context,
```
svenvh, line 386: This function is pretty inefficient because it builds + enqueues a new kernel for every edge case.  I understand that we want to keep the test code simple, so I'm not asking for a change (yet), but would like to hear some other opinions.

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: I wonder, shouldn't this be detected with CI/CD ? Line added

test_conformance/math_brute_force/edge_cases.cpp
shajder: Removed in the development process when I added support for half and double, comment removed.

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: Certainly we should support `half_` versions. As for the rest I would appreciate WG expertise.

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: As mentioned in the issue, we can batch edge-case kernels into groups. This will make the code slightly less readable, so I’d appreciate your input on whether it’s the acceptable tradeoff. Thanks for your review @svenvh

test_conformance/math_brute_force/edge_cases.cpp
svenvh: Can they be `const` references?

test_conformance/math_brute_force/edge_cases.cpp
svenvh: I'd expect that per-builtin-function batching shouldn't make the test code too complicated, and it would give a good improvement (e.g. the 22 `atan2pi` tests will require just 1 kernel execution).  If we'll be adding more edge cases (as per my other comment about the C99 edge cases), I'd say it's worth considering.

test_conformance/math_brute_force/edge_cases.cpp
shajder: Corrected

test_conformance/math_brute_force/edge_cases.cpp
shajder: > 22 `atan2pi` tests will require just 1 kernel execution

I just pushed a correction, the least intrusive solution is to group kernels based on function name (equal number of kernel arguments). On my machine ~120ms decreased to ~20ms.

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -382,12 +358,52 @@ void log_anyvalue(const AnyValue &v)
...
+        src += "[";
+        src += std::to_string(ind);
+        src += "]";
+    }
+    src += ");";
+
+    batch_cases.push_back(ec);
+}
+
+template <typename T>
```
svenvh: It would be nice to have a comment about the high-level strategy, now that we're batching executions together.

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -402,44 +418,59 @@ inline cl_int run_edge_case(const EdgeCaseSpec &ec, cl_context context,
...
 
-        err =
-            clEnqueueWriteBuffer(queue, buf, CL_TRUE, 0, ec.inputs[i].byte_size,
-                                 ec.inputs[i].data.data(), 0, nullptr, nullptr);
+        static std::vector<uint8_t> inData;
+        inData.resize(batch_cases.front().inputs[i].byte_size
+                      * batch_cases.size());
+
+        size_t byte_offset = 0;
+        for (auto &elem : batch_cases)
```
svenvh: `const`?

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -591,43 +634,66 @@ template <typename T> EdgeCaseSpec make_edge_case(const AbstractEdgeCase &aec)
...
+    return ret;
+}
+
 inline cl_int run_edge_cases(const AbstractEdgeCase *cases, std::size_t count,
                              cl_context context, cl_command_queue queue)
 {
     cl_int overall = CL_SUCCESS;
     if (gTestFloat)
     {
         log_info("float test\n");
```
svenvh: Remove ` test` to be consistent with the `double` and `half` cases.

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -461,64 +492,76 @@ inline cl_int run_edge_case(const EdgeCaseSpec &ec, cl_context context,
...
             log_error("ERROR: clEnqueueNDRangeKernel failed for '%s': %d\n",
-                      ec.func_name, err);
+                      batch_cases.front().func_name, err);
             return TEST_FAIL;
         }
     }
 
     {
-        std::array<uint8_t, CL_VALUE_MAX_BYTES> result{};
+        static std::vector<uint8_t> result;
```
svenvh: Why static?

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -382,12 +358,52 @@ void log_anyvalue(const AnyValue &v)
...
+        src += " *out";
+        for (std::size_t i = 0; i < ec.inputs.size(); ++i)
+        {
+            src += ",\n    __global const ";
+            src += ec.inputs[i].cl_type;
+            src += " *in";
+            src += std::to_string(i);
+        }
+        src += ")\n{";
+    }
```
svenvh: Perhaps worth having an else with `assert(batch_cases.front().inputs.size() == ec.inputs.size());` here?

test_conformance/math_brute_force/edge_cases.cpp
```diff
@@ -402,44 +418,59 @@ inline cl_int run_edge_case(const EdgeCaseSpec &ec, cl_context context,
...
-                      ec.func_name, err);
+                      batch_cases.front().func_name, err);
             return TEST_FAIL;
         }
         in_bufs.push_back(buf);
 
-        err =
-            clEnqueueWriteBuffer(queue, buf, CL_TRUE, 0, ec.inputs[i].byte_size,
-                                 ec.inputs[i].data.data(), 0, nullptr, nullptr);
+        static std::vector<uint8_t> inData;
```
svenvh: Why static?

test_conformance/math_brute_force/edge_cases.cpp
svenvh: Thanks, that brings it down from 20 to 6 seconds in my (emulated) environment, so that's a good enough improvement for me.

test_conformance/math_brute_force/edge_cases.cpp
shajder: To avoid continuous allocations/deallocations, the vector retains its capacity between calls so resize() does not reallocate as long as the size stays within the previously allocated capacity.

test_conformance/math_brute_force/edge_cases.cpp
shajder: as above

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

test_conformance/math_brute_force/edge_cases.cpp
shajder: moreover, [this](https://github.com/KhronosGroup/OpenCL-Docs/issues/1538) discussion may be relevant

test_conformance/math_brute_force/edge_cases.cpp
svenvh: The potential benefit here looks marginal in the bigger scheme of things, and I don’t think it outweighs the readability/maintainability cost of introducing hidden persistent state.  So I would encourage dropping `static`.

shajder: Done, working buffers turned into class members.

test_conformance/math_brute_force/edge_cases.cpp
shajder: Done

svenvh (approved):
LGTM, thanks!

There's still the open question about the C99 edge cases, but I would suggest merging the current state and adding more edge cases in a followup PR (potentially after ironing out the specification wrt listing edge cases).

neiltrevett:
Merged as discussed in call on 5May26

PR #2666, opened 2026-04-17
fix(non_uniform_work_group): respect per-dimension work item size lim…

Lurie97:
…its when selecting localSize

Problem:
Test cases selected localSize based only on CL_KERNEL_WORK_GROUP_SIZE, without considering CL_DEVICE_MAX_WORK_ITEM_SIZES per-dimension limits.

This caused issues because:
1. Constructor rounds globalSize based on the original localSize
2. prepareDevice() later trims _enqueuedLocalSize to device limits
3. Result: globalSize was rounded with wrong localSize, causing mismatch

For example, with maxWorkItemSizes=[256,256,64] and localSize={512,1}:
- globalSize rounded assuming localSize[0]=512
- prepareDevice() limits localSize[0] to 256
- globalSize and localSize no longer match

Fix:
Query CL_DEVICE_MAX_WORK_ITEM_SIZES in test cases and compute:
  effectiveMax = min(maxWgSize, maxWorkItemSizes[dim])
before selecting localSize for each dimension.

This ensures localSize is valid before rounding, so globalSize and localSize remain consistent throughout the test.

Affected files:
- test_advanced_2d.cpp
- test_advanced_3d.cpp
- test_advanced_other.cpp
- test_basic.cpp

### DIFF
--- a/test_conformance/non_uniform_work_group/test_advanced_2d.cpp
+++ b/test_conformance/non_uniform_work_group/test_advanced_2d.cpp
@@ -139,8 +139,14 @@ REGISTER_TEST(non_uniform_2d_basic)
 
   // non_uniform_2d_three_prime_numbers_basic
   {
+      size_t maxWorkItemSizes[3];
+      clGetDeviceInfo(device, CL_DEVICE_MAX_WORK_ITEM_SIZES,
+                      sizeof(maxWorkItemSizes), maxWorkItemSizes, NULL);
+
+      size_t effectiveMax = std::min(maxWgSize, maxWorkItemSizes[0]);
+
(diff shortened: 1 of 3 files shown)


rjodinchr:
I wonder if https://github.com/KhronosGroup/OpenCL-CTS/pull/2598 is fixing the same issue or not.

Lurie97:
Yes, PR #2598 fixes the same issue. That said, I believe both changes can coexist. My patch addresses the root cause at test generation time, while the prepareDevice() change provides a runtime safeguard. Together, they improve robustness by both preventing and handling the issue.

PR #2667, opened 2026-04-17
clCopyImage: clean up image resources on failure

xinjin01:
The clCopyImage failure paths did not always release image resources promptly. If verification failed after mapping the destination image, the test returned before unmapping it. The image creation helpers also held raw cl_mem handles after clCreateImage(), so later setup failures could bypass cleanup.

This could retain image-backed resources across repeated failing formats and drive the test into OOM.

Fix this by unmapping the verification mapping before returning on ompare failure, and by taking ownership of created images with clMemWrapper immediately so later failures release them through RAII.

### DIFF
--- a/test_conformance/images/clCopyImage/test_copy_generic.cpp
+++ b/test_conformance/images/clCopyImage/test_copy_generic.cpp
@@ -17,6 +17,28 @@
 #include <CL/cl.h>
 #include "../common.h"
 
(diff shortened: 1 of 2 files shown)


rjodinchr (approved):

lakshmih (approved):

bashbaug:

PR #2668, opened 2026-04-17
USVM test for multithreaded alloc & free

EwanC:
Allocation and free entry-points USVM entry-points should be threadsafe as they may be used in a threaded manner.

Stresses the bug found in the USVM emulation layer https://github.com/bashbaug/SimpleOpenCLSamples/pull/165


bashbaug (approved):
Thanks!  Latest version LGTM.

bashbaug:

PR #2671, opened 2026-04-21
Fix invalid alignment value for _aligned_malloc in Windows

joselopeqti:
The Device Maximum Alignment is set up to zero when using the USM wrapper to allocate a System type allocation. 
This value is used for the `alignment` param in the `_aligned_malloc` function if the test is run on Windows. In this case, `alignment` must be a power of 2 [_aligned_malloc](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=msvc-170) :

> If alignment isn't a power of 2 or size is zero, this function invokes the invalid parameter handler, as described in [Parameter validation](https://learn.microsoft.com/en-us/cpp/c-runtime-library/parameter-validation?view=msvc-170).

### DIFF
--- a/test_conformance/SVM/test_unified_svm_mem_cpy.cpp
+++ b/test_conformance/SVM/test_unified_svm_mem_cpy.cpp
@@ -211,7 +211,7 @@ struct UnifiedSVMOPs : UnifiedSVMBase
                                PSEUDO_CAPABILITY_USE_SYSTEM_ALLOCATOR
                                    | CL_SVM_CAPABILITY_HOST_READ_KHR
                                    | CL_SVM_CAPABILITY_HOST_WRITE_KHR,
-                               0, nullptr, nullptr, nullptr, nullptr));
+                               1, nullptr, nullptr, nullptr, nullptr));
     }
 
     bool caps_compatibility_check(cl_uint srcTypeIndex, cl_uint dstTypeIndex)
--- a/test_conformance/SVM/test_unified_svm_mem_cpy.cpp
+++ b/test_conformance/SVM/test_unified_svm_mem_cpy.cpp
@@ -211,7 +211,7 @@ struct UnifiedSVMOPs : UnifiedSVMBase
                                PSEUDO_CAPABILITY_USE_SYSTEM_ALLOCATOR
                                    | CL_SVM_CAPABILITY_HOST_READ_KHR
                                    | CL_SVM_CAPABILITY_HOST_WRITE_KHR,
-                               1, nullptr, nullptr, nullptr, nullptr));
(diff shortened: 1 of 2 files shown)


test_conformance/SVM/test_unified_svm_mem_cpy.cpp
```diff
@@ -211,7 +211,7 @@ struct UnifiedSVMOPs : UnifiedSVMBase
                                PSEUDO_CAPABILITY_USE_SYSTEM_ALLOCATOR
                                    | CL_SVM_CAPABILITY_HOST_READ_KHR
                                    | CL_SVM_CAPABILITY_HOST_WRITE_KHR,
-                               0, nullptr, nullptr, nullptr, nullptr));
+                               1, nullptr, nullptr, nullptr, nullptr));
```
bashbaug: I think I'd prefer to handle this in the wrapper itself, otherwise every case that wants to do a "hostptr" like this will need to remember to pass a nonzero alignment.  Something like:

```c++
                const auto alignment = deviceMaxAlignment ? deviceMaxAlignment : 1;
                data = static_cast<T*>(
                    align_malloc(count * sizeof(T), alignment));
```

What do you think?

bashbaug: Alternatively, would it make more sense to move this "hostptr" wrapper to UnifiedSVMBase, then it'd be present for any test that wanted to use it?

This would also let us use the same "deviceMaxAlignment" that we use for other USM allocations, which we might want to do, anyhow.

joselopeqti: > What do you think?

Yes, I think it might be better to change it on the wrapper itself.

> Alternatively, would it make more sense to move this "hostptr" wrapper to UnifiedSVMBase, then it'd be present for any test that wanted to use it?

I don't have a strong opinion about it, but I can see the benefit of having it in UnifiedSVMBase.

Another option would be modifying the `align_malloc` function in alloc.h. I can see a similar fix was done for `posix_memalign` but this change would imply that many tests would be affected by it. What do you think?

bashbaug: Let's move the check for zero to the wrapper itself, as shown above.  This seems like the path of least resistance.

We could always make further changes later (in a subsequent PR), either by moving the hostptr wrapper to the unified SVM base class, or by modifying the align_malloc function in the wrapper.

nikhiljnv (approved):
LGTM.

nikhiljnv:
Merging as discussed in OpenCL teleconference on June16th 2026.

PR #2673, opened 2026-04-24
Optimize test_vector_creation by pruning redundant test permutations

rjodinchr:
This commit reduces the overall execution time of
`test_vector_creation.cpp` by removing test cases that are not meaningful, as they are already sufficiently tested by previous smaller vector combinations.

Specifically, the `create_kernel` function now filters the generated `vload` permutations and only adds them to the test program if they meet at least one of the following criteria:
* The combination of loads spans across the middle index of the output vector (`covers_middle`).
* The permutation consists of fewer than 4 total `vload` operations (to cover all permutations for small vectors).
* The number of `vload` operations is exactly equal to the output size (e.g., loading element-by-element).

Additionally, the state tracking for splitting large kernels has been refactored. The `create_kernel` signature was updated to use a boolean `reset_skip_to_permutation` flag to manage internal static state, replacing the previous `skip_to_result` integer logic, and now returns a boolean indicating if the vector size is fully completed.

### DIFF
--- a/test_conformance/basic/test_vector_creation.cpp
+++ b/test_conformance/basic/test_vector_creation.cpp
@@ -33,13 +33,18 @@ static const char *size_names[] = { "",   "2",   "3",   "4",   "8",
                                     "16", "!!a", "!!b", "!!c", "!!d" };
 static char extension[128] = { 0 };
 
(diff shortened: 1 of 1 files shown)


lakshmih (approved):

bashbaug:

PR #2677, opened 2026-04-29
Add pattern lifetime tests for clEnqueueFillBuffer and clEnqueueSVMMemFill

joselopeqti:
Adds tests that modify the pattern used before the fill command is executed.

Adds coverage for [clEnqueueFillBuffer](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueFillBuffer) and [clEnqueueSVMMemFill](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clEnqueueSVMMemFill):

> The memory associated with pattern can be reused or freed after the function returns.

### DIFF
--- a/test_conformance/SVM/test_enqueue_api.cpp
+++ b/test_conformance/SVM/test_enqueue_api.cpp
@@ -349,3 +349,127 @@ REGISTER_TEST(svm_enqueue_api)
 
     return 0;
 }
+
+// Tests corrupting the pattern before clEnqueueSVMMemFill is executed
+REGISTER_TEST(svm_memfill_pattern_lifetime)
+{
+    constexpr cl_int FILL_PATTERN = 0x12345678;
+
+    clContextWrapper contextWrapper = nullptr;
+    clCommandQueueWrapper queues[MAXQ];
+    cl_uint num_devices = 0;
+    cl_int err = CL_SUCCESS;
+
+    const size_t svm_size = num_elements * sizeof(cl_int);
+
+    cl_int *svm_ptr = nullptr;
+    cl_int *pattern = nullptr;
+    cl_event user_event = nullptr;
+    cl_event fill_event = nullptr;
+
+    int result = TEST_FAIL;
+
+    err = create_cl_objects(device, nullptr, &contextWrapper, nullptr,
+                            &queues[0], &num_devices,
+                            CL_DEVICE_SVM_COARSE_GRAIN_BUFFER);
+    if (err) goto cleanup;
+
+    context = contextWrapper;
+    queue = queues[0];
+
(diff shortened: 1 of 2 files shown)


test_conformance/buffers/test_buffer_fill.cpp
```diff
@@ -709,6 +720,102 @@ static int test_buffer_fill(cl_device_id deviceID, cl_context context,
...
+    }
+
+    result = TEST_PASS;
+
+cleanup:
+    if (host_buffer) free(host_buffer);
+    if (pattern) free(pattern);
+    if (fill_event) clReleaseEvent(fill_event);
+    if (user_event) clReleaseEvent(user_event);
+    if (buffer) clReleaseMemObject(buffer);
```
bashbaug: Can we use the RAII wrappers and either std::vectors or smart pointers in these tests so we don't need to clean this up manually?

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -349,3 +349,122 @@ REGISTER_TEST(svm_enqueue_api)
...
+        goto cleanup;
+    }
+
+    result = TEST_PASS;
+
+cleanup:
+    if (fill_event) clReleaseEvent(fill_event);
+    if (user_event) clReleaseEvent(user_event);
+    if (pattern) free(pattern);
+    if (svm_ptr) clSVMFree(context, svm_ptr);
```
bashbaug: Same comment here - can we use wrappers so we don't need to do this cleanup manually?

test_conformance/buffers/test_buffer_fill.cpp
```diff
@@ -1510,3 +1596,17 @@ REGISTER_TEST(buffer_fill_float)
...
+    int err = 0;
+    if (test_fill_reused_pattern(device, context, queue, num_elements)
+        != TEST_PASS)
+    {
+        err++;
+    }
+
+    return err;
+
+} // end test_buffer_fill_pattern_lifetime()
```
rjodinchr, line 1612: Why not put the content of `test_fill_reused_pattern`  directly here?

test_conformance/SVM/test_enqueue_api.cpp
```diff
@@ -349,3 +350,106 @@ REGISTER_TEST(svm_enqueue_api)
...
+
+    err = clFinish(queue);
+    if (err != CL_SUCCESS)
+    {
+        print_error(err, "clFinish failed");
+        return TEST_FAIL;
+    }
+
+    return TEST_PASS;
+}
```
rjodinchr, line 455: missing EOL

PR #2678, opened 2026-04-30
Fix D3D11 log buffering on redirection

joselopeqti:
When the test log is redirected to a file, the `stdout` and `stderr` is buffered and therefore the complete log is not written to a file.
This tells the program to not use buffers.


bashbaug (approved):

PR #2680, opened 2026-05-04
subgroups: print cl_{,u}char as numbers in error messages

kpet:
So the output does not contain invalid UTF-8 values.


rjodinchr (approved):

bashbaug (approved):

PR #2681, opened 2026-05-04
c11_atomics: fix fp16 capability guard for atomic_exchange

svenvh:
The `CBasicTestExchange` class is never instantiated with `HostDataType` set to `HOST_ATOMIC_HALF`, so the condition was always false.

As a result, the atomic_exchange fp16 tests could run even when the required fp16 atomic load/store capabilities were not reported by the implementation.


shajder (approved):

neiltrevett:

PR #2682, opened 2026-05-06
fix(images/kernel_read_write): align mem flags between format query a…

Lurie97:
…nd image creation

The format query in test_loops.cpp uses CL_MEM_WRITE_ONLY, but test_write_*.cpp creates images with both CL_MEM_WRITE_ONLY and CL_MEM_READ_WRITE flags. This causes test failures for formats that are only supported with WRITE_ONLY flag (e.g., CL_RGB + CL_SIGNED_INT8).

Solution:
- Change write_only_mem_flag_types array from 2 elements to 1 element
- Only test CL_MEM_WRITE_ONLY flag for kWriteTests
- CL_MEM_READ_WRITE is already covered by kReadWriteTests which uses read_write kernel and queries formats with CL_MEM_KERNEL_READ_AND_WRITE

### DIFF
--- a/test_conformance/images/kernel_read_write/test_write_1D.cpp
+++ b/test_conformance/images/kernel_read_write/test_write_1D.cpp
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 #include "../testBase.h"
+#include "../common.h"
 #include "test_common.h"
 
 #if !defined(_WIN32)
@@ -71,6 +72,15 @@ int test_write_image_1D(cl_device_id device, cl_context context,
     }
     for( size_t mem_flag_index = 0; mem_flag_index < num_flags; mem_flag_index++ )
     {
+        if (!is_image_format_supported(context, imageInfo->type,
+                                       mem_flag_types[mem_flag_index],
+                                       imageInfo->format))
+        {
+            log_info("Format not supported for %s, skipping...\n",
+                     mem_flag_names[mem_flag_index]);
+            continue;
+        }
+
         int error;
         size_t threads[2];
         bool verifyRounding = false;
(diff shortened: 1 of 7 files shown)


Lurie97:
Hi @franz @karolherbst, could you please help review this patch?

rjodinchr (dismissed):

Lurie97:
Hi @rjodinchr, Thanks for the review! Could you please merge this when you get a chance?

rjodinchr:
Most of the time we want 2 reviewers to approve. I'll mention it at the next teleconference.

Lurie97:
Hi @karolherbst This Piglit MR is related to the opencl work - could you review?

karolherbst:
I think this needs to be fixed differently as this PR removes the coverage of testing images with `CL_MEM_READ_WRITE` for `write_only` image kernel parameters.

I don't have a good idea off-hand, but maybe it makes sense to pass in the format list down into the tests and check there if the the image can be created with `CL_MEM_READ_WRITE`.

Lurie97:
Updated the PR per your suggestion. Now instead of removing CL_MEM_READ_WRITE coverage, I added a runtime check using is_image_format_supported() before each image creation. Unsupported format + mem flag combinations are skipped gracefully, while supported ones are still fully tested.

rjodinchr (dismissed):

Lurie97:
Hi @karolherbst and @rjodinchr ,
Could you help review this MR? Thanks.

test_conformance/images/kernel_read_write/test_write_1D.cpp
```diff
@@ -1,5 +1,6 @@
 //
 // Copyright (c) 2017 The Khronos Group Inc.
+// Copyright 2026 NXP
```
rjodinchr: Is this really needed?

Lurie97: Removed. The changes are minor, so the copyright notice isn't needed. Done.

rjodinchr (approved):

bashbaug:

PR #2684, opened 2026-05-06
add cl_offline_compiler.example

karolherbst:
Now that we require SPIR-V everywhere, we should make it easier for implementations to actually test tests in the SPIR-V mode.

This file can be tweaked according to runtime details, but it should provide a good base already and it's what I'm using in my testing.

There are a few gaps in handling extensions properly, but my hope is that we can collectively improve on this one to make it viable to be used across many different implementations.

### DIFF
--- a/cl_offline_compiler.example
+++ b/cl_offline_compiler.example
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+
+# from test_common/harness/cl_offline_compiler-interface.txt:
+#
+# The cl_offline_compiler program used for offline compilation must
+# implement the following interface.
+#
+#    usage: cl_offline_compiler --source FILE --output FILE
+#                               --cl-device-info FILE --mode MODE
+#                               -- [BUILD_OPTIONS [BUILD_OPTIONS ...]]
+#
+#    positional arguments:
+#      BUILD_OPTIONS          additional options to pass to the compiler
+#
+#    optional arguments:
+#      --source FILE          OpenCL C source file to compile
+#      --output FILE          SPIR-V or binary file to create
+#      --cl-device-info FILE  OpenCL device info file
+#      --mode                 compilation mode (spir-v or binary)
+#
+# The --cl-device-info file is a list of KEY=VALUE pairs containing device
+# information relevant to the mode of offline compilation in question.
+# It is of the following form:
+#
+#    # OpenCL device info affecting <SPIR-V|binary> offline compilation:
+#    CL_DEVICE_ADDRESS_BITS=<32|64>
+#    CL_DEVICE_EXTENSIONS="<space separated list of CL extensions>"
+#    CL_DEVICE_IL_VERSION="<space separated list of IL versions>"
+#    CL_DEVICE_VERSION="OpenCL <version> <vendor info>"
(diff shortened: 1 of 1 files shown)


cl_offline_compiler.example
```diff
@@ -0,0 +1,91 @@
...
+#    usage: cl_offline_compiler --source FILE --output FILE
+#                               --cl-device-info FILE --mode MODE
+#                               -- [BUILD_OPTIONS [BUILD_OPTIONS ...]]
+#
+#    positional arguments:
+#      BUILD_OPTIONS          additional options to pass to the compiler
+#
+#    optional arguments:
+#      --source FILE          OpenCL C source file to compile
+#      --output FILE          SPIR-V or binary file to create
```
silverclaw, line 17: Usually, compilers use the `-o FILE` for output, so we should try to match that

cl_offline_compiler.example
```diff
@@ -0,0 +1,91 @@
...
+#
+#    usage: cl_offline_compiler --source FILE --output FILE
+#                               --cl-device-info FILE --mode MODE
+#                               -- [BUILD_OPTIONS [BUILD_OPTIONS ...]]
+#
+#    positional arguments:
+#      BUILD_OPTIONS          additional options to pass to the compiler
+#
+#    optional arguments:
+#      --source FILE          OpenCL C source file to compile
```
silverclaw, line 16: I'm not sure we need a `--source` command line opt at all.
Most compiler take this as a positional argument, so we could match that.

cl_offline_compiler.example
```diff
@@ -0,0 +1,91 @@
...
+#      --output FILE          SPIR-V or binary file to create
+#      --cl-device-info FILE  OpenCL device info file
+#      --mode                 compilation mode (spir-v or binary)
```
silverclaw, line 19: This doesn't really need to be explicit, right? We could infer this by the output file extension (.spv or .bin).

cl_offline_compiler.example
karolherbst: This is how the CTS is using it, I just copied this comment from from [test_common/harness/cl_offline_compiler-interface.txt](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_common/harness/cl_offline_compiler-interface.txt)

cl_offline_compiler.example
karolherbst: Yeah.. sure but I'd rather not change the interface in case anybody else is also using this interface already, so the plan was here to just implement it as expected.

cl_offline_compiler.example
```diff
@@ -0,0 +1,91 @@
...
+
+target=''
+spirv_extensions=[]
+spirv_max='1.0'
+
+for info in filter(len, args.cl_device_info.readlines()):
+    if info[0] == '#':
+        continue
+    info = info.strip('\r\n')
+    split = info.split('=')
```
karolherbst, line 52: I was wondering.. does anybody know of a good library to parse those kind of property files so we don't have to do it in a terrible way like here?

silverclaw: I think this feels like the INI config parsing that we have in Python's configparser: https://docs.python.org/3/library/configparser.html
Would that work?

PR #2686, opened 2026-05-07
math_brute_force: fix tgamma fp32 reference host-dependence

svenvh:
Avoid calling the `long double` tgamma helper for larger fp32 inputs that knowingly result in overflow or underflow.  long-double precision is ABI dependent and can produce different results for large values.

An example case I observed was for input `0x1.62e6p+9` (`709.796875f`) which must yield infinity as it is far beyond the fp32 overflow threshold for tgamma.  The host reference computation wrongly resulted in `NaN` on some platforms.

Fix by avoiding host long-double math when deciding a result that is already known from fp32 semantics.  Use `|x| >= 128` as a conservative cut-off point.  It is well past any finite fp32 tgamma result for both positive and negative inputs.

### DIFF
--- a/test_conformance/math_brute_force/reference_math.cpp
+++ b/test_conformance/math_brute_force/reference_math.cpp
@@ -2354,9 +2354,33 @@ smallarg:
 
 double reference_tgamma(double x)
 {
+    // This function is only used for fp32 reference calculations.
+    const float fx = x;
+
+    // Once |fx| >= 128, tgammaf has already overflowed or underflowed, so avoid
+    // host-dependent olm_tgammal behavior.
+    if (isfinite(fx) && fabsf(fx) >= 128.0f)
+    {
+        // Negative integer inputs are invalid for tgamma, so return NaN.
+        if (fx < 0.0f && fx == std::floor(fx))
+        {
+            return cl_make_nan();
+        }
+
+        if (fx < 0.0f)
+        {
+            // Negative non-integer large values underflow to zero; preserve the
+            // sign, which alternates between consecutive negative integers.
+            const long n = (long)std::floor(-fx);
+            const bool negative = (n % 2 == 0);
+            return negative ? -0.0 : 0.0;
+        }
+
+        return INFINITY;
+    }
+
     if (x < 0.0f)
     {
-        float fx = x;
         if (fx == std::floor(fx))
         {
             return std::numeric_limits<double>::signaling_NaN();


shajder:
FYI @bashbaug 

hmm, I have a feeling this has come up before, [here](https://github.com/KhronosGroup/OpenCL-CTS/pull/2251#discussion_r2318886968) is the original discussion (hidden as resolved). 

> ... The host reference computation wrongly resulted in NaN on some platforms.

I've tested x86_64/ARM with 0x1.62e6p+9 (709.796875f) and got correct `Inf` result, what platforms are you referring to?

svenvh:
> what platforms are you referring to?

I was seeing this on an ARMv7 platform where `long double` is only 64 bits.

shajder:
> > what platforms are you referring to?
> 
> I was seeing this on an ARMv7 platform where `long double` is only 64 bits.

Thanks, the libm `MAXGAML = 1755.455L` is sized for 80/128-bit `long double`. On ARMv7 where `long double` is 64-bit, `stirf()` overflows internally (expl/powl saturate to +Inf, then +Inf/+Inf = NaN) before the `goverf` path can trigger.

Would this work on your ARMv7 setup?

```c
#if LDBL_MANT_DIG <= DBL_MANT_DIG
#define MAXGAML 171.624L
#else
#define MAXGAML 1755.455L
#endif
```

svenvh:
> Would this work on your ARMv7 setup?
> 
> ```c
> #if LDBL_MANT_DIG <= DBL_MANT_DIG
> #define MAXGAML 171.624L
> #else
> #define MAXGAML 1755.455L
> #endif
> ```

That may fix the `0x1.62e6p+9` input, but breaks other cases as it will return `+/- INFINITY` for negative cases that now exceed the threshold, but are supposed to return 0.

*update*: I suppose this would require aligning the 1755 constants scattered in the other reference_functions.

svenvh:
> _update_: I suppose this would require aligning the 1755 constants scattered in the other reference_functions.

@shajder I think this might work too; I've opened https://github.com/KhronosGroup/OpenCL-CTS/pull/2689 as an alternative fix based on your suggestion.

shajder:
> > _update_: I suppose this would require aligning the 1755 constants scattered in the other reference_functions.
> 
> @shajder I think this might work too; I've opened #2689 as an alternative fix based on your suggestion.

Thanks for investigation and fix, #2689 looks a bit less invasive

svenvh:
Thanks, then I'll abandon this in favor of https://github.com/KhronosGroup/OpenCL-CTS/pull/2689.

PR #2688, opened 2026-05-08
math_brute_force: do not create signaling NaNs for tgamma

svenvh:
Since this is a reference implementation, we only need the NaN result and are not interested in signaling-NaN behavior, so a quiet NaN is preferred.  Use the same NaN helper that the other reference functions use.


rjodinchr (approved):

bashbaug (commented):

PR #2689, opened 2026-05-12
math_brute_force: fix tgamma reference MAXGAML

svenvh:
long-double precision is ABI dependent and some ABIs use 64-bit precision for the long double format, whereas `MAXGAML` is currently sized for 128-bit long double formats.  On such ABIs, `stirf()` overflows internally for some inputs causing an incorrect `NaN` reference result.

An example case I observed was for input `0x1.62e6p+9` (`709.796875f`) which must yield infinity as it is far beyond the fp32 overflow threshold for tgamma.  The host reference computation wrongly resulted in `NaN` on some platforms.

Fix by selecting a lower `MAXGAML` cutoff when `long double` has the same effective range as `double`.

### DIFF
--- a/test_conformance/math_brute_force/reference_math.cpp
+++ b/test_conformance/math_brute_force/reference_math.cpp
@@ -18,6 +18,7 @@
 #include "harness/compat.h"
 
 #include <climits>
+#include <float.h>
 
 #if !defined(_WIN32)
 #include <cstring>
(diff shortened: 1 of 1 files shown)


svenvh:
This is an alternative to https://github.com/KhronosGroup/OpenCL-CTS/pull/2686 .

shajder (dismissed):
Thanks for your investigation, LGTM

bashbaug (approved):

PR #2691, opened 2026-05-13
[NFC] math_brute_force: move ThreadInfo to common.h

svenvh:
Unify the various `ThreadInfo` structs into a single hierarchy and move them into `common.h`.  This reduces code duplication.

Some tests will not use every member variable of the new hierarchy (e.g. the macro_* tests do not compute `maxError`), but avoid specializing for every use for the sake of maintainability.


bashbaug (approved):

rjodinchr (approved):

bashbaug:

PR #2692, opened 2026-05-14
math_brute_force: test unary_u_half inputs only once

svenvh:
Avoid testing the same inputs for unary_u_half functions many times. The input domain of such functions only contains 2^16 possible elements.

This only affects the fp16 `nan` built-in functions.


rjodinchr:
Sorry, I missed that PR. It has been fixed by https://github.com/KhronosGroup/OpenCL-CTS/pull/2698

PR #2694, opened 2026-05-15
spirv_new: don't use Physical64 in spvasm32 files

svenvh:
spirv14_image_operand_signextend and spirv14_image_operand_zeroextend accidentally used the `Physical64` memory model for the 32-bit variants.


svenvh:
I've looked for other Phyical32/Physical64 discrepancies in the spvasm files, but these were the only two I could find.

bashbaug (approved):

bashbaug:
Obvious fix, no reason to wait to merge this one.

PR #2695, opened 2026-05-15
Fix sub_group_inverse_ballot testing

StuartDBrady:
The previous test would still pass even if the implementation always
returned true.

Now that we use BALLOT_BIT_OPS for inverse_ballot, we no longer need
sub_group_inverse_ballot_source. We use sub_group_ballot_bit_ops_source
instead.

This change also renames BALLOT_COUNT_SCAN_FIND to BALLOT_BIT_OPS,
along with renaming rft_arith to rft_bit_ops, test_params_arith to
test_params_bit_ops, and sub_group_ballot_bit_scan_find_source to
sub_group_ballot_bit_ops_source.

It also uses cl_uint4_to_bs128 in BALLOT_BIT_OPS::chk and adds
static_asserts for BallotOp values.

### DIFF
--- a/test_conformance/subgroups/test_subgroup_ballot.cpp
+++ b/test_conformance/subgroups/test_subgroup_ballot.cpp
@@ -339,92 +339,6 @@ template <typename Ty, BallotOp operation> struct BALLOT_BIT_EXTRACT
     }
 };
 
-template <typename Ty, BallotOp operation> struct BALLOT_INVERSE
-{
-    static void log_test(const WorkGroupParams &test_params,
-                         const char *extra_text)
-    {
-        log_info("  sub_group_inverse_ballot...%s\n", extra_text);
-    }
-
-    static void gen(Ty *x, Ty *t, cl_int *m, const WorkGroupParams &test_params)
-    {
-        // no work here
-    }
-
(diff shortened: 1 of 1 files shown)


test_conformance/subgroups/test_subgroup_ballot.cpp
```diff
@@ -339,95 +339,11 @@ template <typename Ty, BallotOp operation> struct BALLOT_BIT_EXTRACT
...
-            y += lws;
-            m += 4 * lws;
-        }
-
-        return TEST_PASS;
-    }
-};
-
+// Used for static_asserts below
+template <auto...> inline constexpr bool always_false_v = false;
```
joselopeqti:
```suggestion
template <auto...> [[maybe_unused]] inline constexpr bool always_false_v = false;
```

joselopeqti (changes requested):

test_conformance/subgroups/test_subgroup_ballot.cpp
StuartDBrady: Good catch, thankyou for this!

joselopeqti (approved):

kpet:

PR #2696, opened 2026-05-18
images: Remove global variables to allow for parallel execution

rjodinchr:
This patch removes the use of global state across the image conformance tests and replaces it with a passed `context_t` structure.

The goal of this change is to be able to run `image_streams` in parallel by removing global variables as they were not thread-safe. This refactoring had to be applied everywhere because `image_streams` needs to update `test_format_set_fn`, which is used by all the image tests.

Key changes:
* Removed global flags such as `gTestSmallImages`, `gTestMipmaps`, `gDebugTrace`, `gEnablePitch`, etc.
* Initialized a `context_t` struct in the `main.cpp` files to store test configuration.
* Updated test function signatures across `clCopyImage`, `clFillImage`, `clGetInfo`, and `kernel_read_write` to accept `const context_t &ctx`.

[run-test: test_image_streams --num-worker-threads 2 CL_A CL_FILTER_NEAREST 1D]

### DIFF
--- a/test_conformance/images/testBase.h
+++ b/test_conformance/images/testBase.h
@@ -22,14 +22,52 @@
 #include "harness/clImageHelper.h"
 #include "harness/imageHelpers.h"
 
-extern bool gDebugTrace;
-extern bool gTestSmallImages;
-extern bool gEnablePitch;
-extern bool gTestMaxImages;
-extern bool gTestMipmaps;
-
 // Amount to offset pixels for checking normalized reads
 #define NORM_OFFSET 0.1f
+
+typedef struct
+{
+    bool debugTrace;
+    bool testSmallImages;
+    bool testMaxImages;
+    bool enablePitch;
+    bool testMipmaps;
+    bool extraValidateInfo;
+    bool disableOffsets;
+    bool testImage2DFromBuffer;
+    bool useKernelSamplers;
+    bool testReadWrite;
+    int typesToTest;
+    int testTypesToRun;
+    int normalizedModeToUse;
+    cl_addressing_mode addressModeToUse;
+    cl_mem_flags memFlagsToUse;
+    cl_filter_mode filterModeToUse;
+    cl_channel_type channelTypeToUse;
+    cl_channel_order channelOrderToUse;
+} context_t;
+
+static inline void init_context(context_t &ctx)
+{
+    ctx.debugTrace = false;
+    ctx.testSmallImages = false;
+    ctx.testMaxImages = false;
+    ctx.enablePitch = false;
+    ctx.testMipmaps = false;
+    ctx.extraValidateInfo = false;
+    ctx.disableOffsets = false;
+    ctx.testImage2DFromBuffer = false;
+    ctx.useKernelSamplers = false;
+    ctx.testReadWrite = false;
+    ctx.typesToTest = 0;
+    ctx.testTypesToRun = 0;
+    ctx.normalizedModeToUse = 7;
+    ctx.addressModeToUse = (cl_addressing_mode)-1;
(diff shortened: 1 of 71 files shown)


rjodinchr:
Many diffs are needed to comply with `clang-format`.

rjodinchr:

rjodinchr:
ref #2723

rjodinchr:
This PR is also fixing 2 minor bugs:
- in `test_conformance/images/kernel_image_methods/main.cpp`: missing parsing of channel order.
- in `test_conformance/images/kernel_read_write/test_cl_ext_image_raw10_raw12.cpp`: stop mutating `gtestTypesToRun` without reverting it back to its original value.

ahesham-arm:
Sorry but this PR needs to be broken down into commits (even if the intermediate state is not clang-formatted, for example) if we are to get a reasonable code review in time.

rjodinchr:
> Sorry but this PR needs to be broken down into commits (even if the intermediate state is not clang-formatted, for example) if we are to get a reasonable code review in time.

I completely understand. This one will take some time either way. I went through it to split as much as I could. I hope this will make it easier🙏

test_conformance/images/testBase.h
```diff
@@ -22,15 +22,53 @@
...
-extern bool gDebugTrace;
-extern bool gTestSmallImages;
-extern bool gEnablePitch;
-extern bool gTestMaxImages;
-extern bool gTestMipmaps;
-
 // Amount to offset pixels for checking normalized reads
 #define NORM_OFFSET 0.1f
 
+typedef struct
```
ahesham-arm: This is C++ so no need for the typedef.

test_conformance/images/testBase.h
```diff
@@ -22,15 +22,53 @@
...
+    ctx.useKernelSamplers = false;
+    ctx.testReadWrite = false;
+    ctx.typesToTest = 0;
+    ctx.testTypesToRun = 0;
+    ctx.normalizedModeToUse = 7;
+    ctx.addressModeToUse = (cl_addressing_mode)-1;
+    ctx.memFlagsToUse = CL_MEM_USE_HOST_PTR;
+    ctx.filterModeToUse = (cl_filter_mode)-1;
+    ctx.channelTypeToUse = (cl_channel_type)-1;
+    ctx.channelOrderToUse = (cl_channel_order)-1;
```
ahesham-arm: The struct above can just have these values set as its default, in the definition itself. No need for a dedicated init_context function.

test_conformance/images/testBase.h
```diff
@@ -22,15 +22,53 @@
...
+    bool testReadWrite;
+    int typesToTest;
+    int testTypesToRun;
+    int normalizedModeToUse;
+    cl_addressing_mode addressModeToUse;
+    cl_mem_flags memFlagsToUse;
+    cl_filter_mode filterModeToUse;
+    cl_channel_type channelTypeToUse;
+    cl_channel_order channelOrderToUse;
+} context_t;
```
ahesham-arm: I'd prefer if we use a narrower name for the type to avoid potential collisions, something that indicates that this is the context for the image tests, not the OpenCL context, general test context, etc. For example `image_test_context_t`.

test_conformance/images/testBase.h
rjodinchr: Done

test_conformance/images/testBase.h
rjodinchr: Done

test_conformance/images/testBase.h
rjodinchr: Done

ahesham-arm (approved):

PR #2697, opened 2026-05-19
math_brute_force: Refactor input generation & reduce default test size

rjodinchr:
This patch consolidates the hardcoded arrays of edge-case test values
from multiple individual test files into a single, shared location.


rjodinchr:

joselopeqti:
Hi @rjodinchr, would it be possible to rebase to main?

rjodinchr:
ref #2723

ahesham-arm:
LGTM. Generally do we need these to be `std::vector` or is `std::array` a better fit here?

rjodinchr:
> LGTM. Generally do we need these to be `std::vector` or is `std::array` a better fit here?

Even though C++17 allows for variable initialization without specifying the size, function prototypes still require it. 
This means `get*SpecialValues` must explicitly expose the element count, creating a maintenance burden whenever elements are added or removed, despite providing little to no performance benefit.

ahesham-arm:
Makes sense, thanks for checking.

ahesham-arm (approved):

rjodinchr:
@bashbaug @kpet, is this fine to merge right now, as this is a non-functional change with an approval?
It would allow me to push the next one for `math_brute_force` which will really be reducing the workload.

Issue #2669, opened 2026-04-20
math_brute_force: Optimize test execution time and improve coverage methodology


rjodinchr:
**TL;DR:** Most `math_brute_force` tests compute 4 billion (`1<<32`) values, which results in excessive execution times, particularly on mobile GPUs. This issue proposes a strategy to reduce these workloads while maintaining a high level of confidence in the correctness of the implementations.

### Current Implementation

**1. Tests evaluating all 64K (`1<<16`) values:**
*(This behavior is correct and working as intended)*
- `unary_half`
- `unary_two_results_half`
- `unary_two_results_i_half`
- `i_unary_half`
- `macro_unary_half`

**2. Tests evaluating 4 billion (`1<<32`) values redundantly:**
*(These test the exact same inputs multiple times)*
- `unary_u_half`

**3. Tests evaluating 4 billion (`1<<32`) values using combinations:**
*(These test all possible combinations of special values, followed by combinations of randomly selected values)*
- `binary_double`
- `binary_float`
- `binary_half`
- `binary_i_double`
- `binary_i_float`
- `binary_i_half`
- `binary_operator_double`
- `binary_operator_float`
- `binary_operator_half`
- `macro_binary_double`
- `macro_binary_float`
- `macro_binary_half`
- `ternary_double`
- `ternary_float`
- `ternary_half`

**4. Tests evaluating 4 billion (`1<<32`) values using purely random combinations:**
*(These completely ignore special values)*
- `binary_two_results_i_double`
- `binary_two_results_i_float`
- `binary_two_results_i_half`
- `mad_double`
- `mad_float`
- `mad_half`
- `unary_u_double`

**5. Tests evaluating 4 billion (`1<<32`) values uniformly:**
*(These spread values uniformly across the range, but completely ignore special values)*
- `i_unary_double`
- `macro_unary_double`
- `unary_double`
- `unary_two_results_double`
- `unary_two_results_i_double`

**6. Tests evaluating all 4 billion (`1<<32`) values exhaustively:**
- `i_unary_float`
- `macro_unary_float`
- `unary_float`
- `unary_two_results_float`
- `unary_two_results_i_float`
- `unary_u_float`

### Identified Issues
- **Redundancy:** Group 2 tests the exact same values multiple times.
- **Missing Edge Cases:** Groups 4 and 5 completely ignore special values (e.g., NaN, Infinity, zero).
[...]

PR #2698, opened 2026-05-19
math_brute_force: fix loop range in unary_u_half.cpp.

rjodinchr:
The test loop for half precision values should iterate through all possible 16-bit patterns. This change corrects the loop upper bound from 1ULL << 32 to 1ULL << 16.

Ref #2669


bashbaug (approved):

PR #2702, opened 2026-05-25
Fix non-uniform workgroups atomics tests

ahesham-arm:
On devices that don't support non-uniform workgroups, recompute `_numOfGlobalWorkItems` after any global-size adjustments.

This fixes the issue introduced by #2598.


rjodinchr (approved):

bashbaug (approved):

PR #2704, opened 2026-05-26
disable nextafter edge case testing in math bruteforce

### LINKED ISSUE #2703: mathbruteforce: nextafter edge case testing is too strict
I was testing this and saw it failing for `nextafter`:

```
FAIL: nextafter(0x80000000, 0x3f800000) - expected 0x00000001, got 0x00800000
FAIL: nextafter(0x00000000, 0xbf800000) - expected 0x80000001, got 0x80800000
```

But according to the Spec it's legal to either return the smallest normal or denormal value:

[7.5.1. Additional Requirements Beyond C99 TC2](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#additional-requirements-beyond-c99-tc2) states:

> nextafter(-0, y > 0) returns smallest positive denormal value.
> nextafter(+0, y < 0) returns smallest negative denormal value.

[7.5.3. Edge Case Behavior in Flush to Zero Mode](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#edge-case-behavior-in-flush-to-zero-mode) states:

> If subnormals are flushed to zero, a device may choose to conform to the following edge cases for nextafter instead of those listed in the [additional requirements](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#additional-requirements-beyond-c99-tc2) section.
>    nextafter(+smallest normal, y < +smallest normal) = +0.
>    nextafter(-smallest normal, y > -smallest normal) = -0.
>    nextafter(-0, y > 0) returns smallest positive normal value.
>    nextafter(+0, y < 0) returns smallest negative normal value.

karolherbst:
The test is too strict on decides that to not support denorms as on those a runtime is allowed to return the smallest normal values according to 7.5.3. Edge Case Behavior in Flush to Zero Mode.

Closes: https://github.com/KhronosGroup/OpenCL-CTS/issues/2703


ahesham-arm:
Change is fine but it causes unused variable/function warnings/errors.

karolherbst:
> Change is fine but it causes unused variable/function warnings/errors.

I think I should just go for this patch though: https://github.com/KhronosGroup/OpenCL-CTS/issues/2703#issuecomment-4546049545

karolherbst:
Please look at https://github.com/KhronosGroup/OpenCL-CTS/pull/2710 instead

PR #2705, opened 2026-05-26
ci: restrict tests job to run only on pull requests

rjodinchr:
The 'tests' job relies on the GitHub CLI (`gh`) to parse the commits and description of an associated pull request to find triggers. When running on a standard branch push (such as to 'main'), there is no pull request context, making it impossible to determine which commits to parse.

Also add missing if on steps to make sure nothing runs when
appropriate.


bashbaug (approved):

PR #2706, opened 2026-05-27
ci: Update result tracking with new golden format

rjodinchr:
The primary goal of this commit is to improve CI tracking by
introducing a new golden format that can differentiate test results
based on command-line arguments. To cleanly extract and pass these
arguments into the JSON result outputs, the command-line parsing
infrastructure across the CTS required a significant refactoring.

Key changes include:
* Enhanced CI Tracking: Updates `ci/compare_results.py`,
`ci/pocl/golden.json`, and `saveResultsToJson` to include and evaluate
an `args` key. The golden JSON now uses a nested format mapping
specific argument strings (e.g., `--wimpy -1`) to their expected
results, allowing the CI to validate the same binary run under
different parameters.
* Centralized Parsing Infrastructure: Introduces the `ParseArgsFn`
callback and `runTestHarnessWithCheckAndParse`. This offloads custom
argument parsing from individual test `main()` functions and safely
extracts the arguments used so they can be logged by the test harness.
* Help Text Consolidation: Replaces fragmented `printUsage()`
functions with unified `help` string references populated directly by
the standard parsing callbacks.


rjodinchr:
ref #2723

ahesham-arm:
Hi @rjodinchr, is it possible to separate the work into multiple commits, e.g. one for each of the bullet points you listed in the description. There is a lot of value in your recent PRs but they can be exhuasting to review because of the volume of unrelated/non-core changes required.

rjodinchr:
> Hi @rjodinchr, is it possible to separate the work into multiple commits, e.g. one for each of the bullet points you listed in the description. There is a lot of value in your recent PRs but they can be exhuasting to review because of the volume of unrelated/non-core changes required.

I understand the difficulty of reviewing such a large PR. 
It feels very complicated to break this one down into smaller pieces. Those commits would either not be able to compile or would add unused code.

I can suggest the following strategy for the review:
- First, let's focus on the `test_common/harness` part.
- Once we agree on that, each individual test makes use of the new harness changes. They are all independent but heavily depend on the harness changes.
- We can have a look at the `ci` part at the end.

The harness which is the core of this PR should not be too big to start with:
```
4 files changed, 144 insertions(+), 72 deletions(-)
```

test_common/harness/parseParameters.cpp
```diff
@@ -91,43 +91,38 @@ For spir-v mode only:
...
         else if (!strcmp(argv[i], "--list") || !strcmp(argv[i], "-list"))
         {
             delArg++;
+            removed_args.push_back(argv[i]);
             gListTests = true;
         }
         else if (!strcmp(argv[i], "--wimpy") || !strcmp(argv[i], "-w"))
         {
             delArg++;
+            removed_args.push_back("--wimpy");
```
ahesham-arm, line 125: Why is this the only argument that gets pushed back as a hardcoded string and not `argv[i]`?

test_common/harness/parseParameters.cpp
```diff
@@ -91,43 +91,38 @@ For spir-v mode only:
...
         {
-            // Note: we don't increment delArg to delete this argument,
-            // to allow the caller's argument parsing routine to see the
-            // option and print its own help.
-            helpInfo();
+            delArg++;
+            if (!help)
+            {
+                help = true;
+                removed_args.push_back("--help");
```
ahesham-arm, line 112: And this one too I guess.

test_common/harness/testHarness.h
```diff
@@ -191,11 +191,25 @@ extern int runTestHarnessWithCheck(int argc, const char *argv[], int testNum,
                                    cl_command_queue_properties queueProps,
                                    DeviceCheckFn deviceCheckFn);
 
+typedef test_status (*ParseArgsFn)(int &argc, const char *argv[],
```
ahesham-arm: Personal perference but I think `using` is easier to read for functions, i.e.
```
using ParseArgsFn = test_status (*)(
    int &argc,
    const char *argv[],
    std::vector<std::string> &removed_args,
    std::string &help_description
);
```

test_common/harness/testHarness.cpp
```diff
@@ -337,6 +386,17 @@ int runTestHarnessWithCheck(int argc, const char *argv[], int testNum,
         return EXIT_SUCCESS;
     }
 
+    gWimpyMode |= (getenv("CL_WIMPY_MODE") != nullptr);
+    if (gWimpyMode)
+    {
+        log_info("\n");
+        log_info("**************************\n");
+        log_info("***    !! WARNING !!   ***\n");
```
ahesham-arm, line 394: Nit: this is different than the existing message. I don't mind it, just making sure this was intentional.

test_common/harness/parseParameters.cpp
rjodinchr: Same as for wimpy, even so I believe it is less of a problem with the `help` argument.

test_common/harness/parseParameters.cpp
rjodinchr: This ensures consistency, as we don't need both `-w` and `--wimpy` to be reported. It forces all reports to use a single format so that two identical runs aren't mistaken for being different just because they used different naming conventions.

test_common/harness/testHarness.cpp
rjodinchr: This is intentional. Some tests trigger an additional message when wimpy mode is enabled. I wanted to ensure that any tests not printing an additional message can rely on that message to explicitly signal the warning part that most other messages display.

test_common/harness/testHarness.h
rjodinchr: I'm indifferent. `clang-format` should be trusted to handle repository requirements. If multiple formats are permitted, we should either configure `clang-format` to enforce a specific one or simply accept whatever `clang-format` allows.

test_common/harness/parseParameters.cpp
ahesham-arm: You didn't do this for `--list` and `-list` which is what probably confused me.

test_common/harness/testHarness.h
ahesham-arm: I don't think clang-format can change the syntax from `typedef` to `using`.

rjodinchr: Oh I misread your comment, sorry about that. I agree, I'll change that.

test_common/harness/parseParameters.cpp
rjodinchr: If one passes `-list`, the JSON file will not be created. The same could be said for `-h`.

I think I will change `-list` so that, regardless of what happens in the future, this ensures consistency in what is reported for options that have multiple formats.

ahesham-arm (dismissed):
LGTM, approved. I reiterate the suggestion of planning the work for these type of changes over multiple commits to ease reviewing and speed up approvals.

rjodinchr:
My rebase on top-of-tree with the latest changes we discussed removed your approval.

ahesham-arm (approved):

karolherbst:
Not sure if this was intentional, but it broke it broke the `-list` parameter for `test_conversions`

rjodinchr:
Yes, that was an error on my side. I noticed it a while ago, but I was curious to know whether anyone would notice it.

I intend to fix it with https://github.com/KhronosGroup/OpenCL-CTS/commit/5a1ec424e4c9d431121b0977d563c87eff7b5e31 (I'll make a PR for it once https://github.com/KhronosGroup/OpenCL-CTS/pull/2754 is merged, as it depends on it).

Let me know if you are fine waiting for that fix to get it, or if you would prefer to have it fixed earlier.

karolherbst:
> Yes, that was an error on my side. I noticed it a while ago, but I was curious to know whether anyone would notice it.
> 
> I intend to fix it with [5a1ec42](https://github.com/KhronosGroup/OpenCL-CTS/commit/5a1ec424e4c9d431121b0977d563c87eff7b5e31) (I'll make a PR for it once #2754 is merged, as it depends on it).
> 
> Let me know if you are fine waiting for that fix to get it, or if you would prefer to have it fixed earlier.

Seems like a rather big commit to only fix the `-list` problem.

rjodinchr:
> Seems like a rather big commit to only fix the -list problem.

While that commit indirectly tackles the `-list` issue, addressing it wasn't the main objective. Still, it seems like the right move for a long-term fix.

If a quicker solution is required, we could implement a smaller, temporary workaround, though it would be a hack slated for eventual removal.

PR #2707, opened 2026-05-28
Change the allowed error for tan(half) to match the spec.

b-sumner:
The OpenCL working group changed the max allowable ULP error for tan(half) to 2.5, as specified in table 69 of the C spec.  However the current CTS complains about any error above 2.0 ULP.  This patch sets the specified limit.


bashbaug:
Linking:

https://github.com/KhronosGroup/OpenCL-Docs/issues/1303
https://github.com/KhronosGroup/OpenCL-Docs/pull/1387

rjodinchr (approved):

PR #2711, opened 2026-06-02
conversions: Refactor tests to reduce dataset size

rjodinchr:
This commit reduces the overall dataset size for conversion tests
 while maintaining high confidence in edge-case coverage. By replacing
 hardcoded arrays with a dynamically generated set of problematic
 corner cases, the test execution time and memory footprint are
 optimized without sacrificing strict conformance validation.

 It also introduces an exhaustive testing mode (`-a` flag) to test 2^32
 inputs when full coverage is explicitly desired, and corrects the
 workload reduction math for Wimpy and Embedded modes. Thread safety is
 guaranteed via a global `std::recursive_mutex`, and exact bit-level
 deduplication using `std::unordered_set` prevents redundant
 floating-point testing (e.g., collapsing NaNs or signed zeros).

 Special Value Selection Strategies:

 * Integer Types (Base and Conversions)
   Focuses on a dense cluster of known boundary triggers rather than
   sparse randoms. The generator populates a baseline pool (0 to 255),
   calculates powers of 2, 3, 5, 7, and 10 up to the sign bit, and
   creates bitwise combinations of repeating patterns and shifting
   masks. Crucially, to catch off-by-one and alignment errors, every
   base value generates its immediate neighbors (offsets from -3 to
   +3), along with their bitwise NOT and sign-bit XORs.

 * Floating-Point Types
   Focuses strictly on precision loss and rounding boundaries. It seeds
   a core set of known difficult values (-/+ INF, NaN, subnormals, and
   type limits). To test exact rounding thresholds, the generator
   calculates the +/- 1 ULP neighbors for every seeded value via
   integer bitcasting. Finally, to guarantee coverage of saturation and
   overflow thresholds, cross-type boundary values corresponding to the
   specific destination type's domain limits (e.g., injecting exact
   char/short limits into float inputs) are dynamically added to the
   test set.

### DIFF
--- a/test_conformance/conversions/basic_test_conversions.cpp
+++ b/test_conformance/conversions/basic_test_conversions.cpp
@@ -51,6 +51,7 @@
 #include <vector>
 #include <type_traits>
 #include <cmath>
+#include <mutex>
 
 #include "basic_test_conversions.h"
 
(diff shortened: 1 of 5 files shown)


rjodinchr:

ahesham-arm:
I would be interested to know if you have any metrics that you can share, e.g. execution times and peak memory usage before and after your change.

rjodinchr:
I don't have memory usage numbers, but for the execution time it is about 40 times faster.

rjodinchr:
ref #2723

test_conformance/conversions/conversions_data_info.h
```diff
@@ -87,6 +89,249 @@ struct DataInitBase : public DataInitInfo
...
+            }
+        }
+        else if constexpr (std::is_same<OutType, cl_long>::value)
+        {
+            for (const auto val : GetIntSpecialValues<uint64_t>())
+            {
+                push(static_cast<InType>(bitcast<uint64_t, OutType>(val)));
+            }
+        }
+        else if constexpr (std::is_same<OutType, cl_half>::value)
```
ahesham-arm: This is unreachable because `cl_half` and `cl_ushort` are both `uint16_t`.

test_conformance/conversions/conversions_data_info.h
```diff
@@ -87,6 +89,249 @@ struct DataInitBase : public DataInitInfo
...
+        else if constexpr (std::is_same<OutType, cl_ushort>::value)
+        {
+            for (uint32_t i = 0; i < (64 * 1024); i++)
+            {
+                push(static_cast<InType>(i));
+            }
+        }
+        else if constexpr (std::is_same<OutType, cl_short>::value)
+        {
+            for (int32_t i = -(64 * 1024 / 2 - 1); i <= (64 * 1024 / 2); i++)
```
ahesham-arm: This is `-32767:32768` it should be `-32768:32767` as this excludes `SHRT_MIN`.

test_conformance/conversions/conversions_data_info.h
```diff
@@ -87,6 +89,249 @@ struct DataInitBase : public DataInitInfo
...
+            for (const auto &mask : masks)
+            {
+                push(pattern & mask);
+            }
+        }
+        return vec;
+    }
+
+    template <typename InType, typename InIntegerType, typename OutType>
+    std::vector<InType> GetFpSpecialValues()
```
ahesham-arm: This returns a copy of a static vector, it should be `std::vector<InType>&` instead.

test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -690,29 +634,42 @@ int ConversionsTest::DoTest(Type outType, Type inType, SaturationMode sat,
...
-        if (0 == (i & ((lastCase >> 3) - 1)))
+        if (0 == (i & ((nbInputs >> 3) - 1)))
         {
             vlog(".");
             fflush(stdout);
         }
 
-        cl_uint count = (uint32_t)std::min((uint64_t)blockCount, lastCase - i);
-        writeInputBufferInfo.count = count;
+        writeInputBufferInfo.count = blockCount;
```
ahesham-arm: I think this stay as `min(blockCount, nbInputs - i)`.

test_conformance/conversions/conversions_data_info.h
```diff
@@ -761,187 +1006,123 @@ template <typename InType, typename OutType, bool InFP, bool OutFP>
...
+            {
+                for (uint32_t i = 0; i < size; i++)
+                {
+                    o[i] = (uint32_t)(i + ulStart);
+                }
+                return;
+            }
+            const auto &specialValues = GetIntSpecialValues<uint32_t>();
+            uint32_t i;
+            for (i = 0; i < size && (i + ulStart) < specialValues.size(); i++)
```
ahesham-arm, line 1052: Worker jobs offset their destination buffer by `job_id`, but not their source-value index. Every worker repeats the same slice; later iterations jump by `blockCount`, skipping special values.

This also means `-a` is not exhaustive. I think we need to add `job_id * size` to the logical index throughout init().

test_conformance/conversions/conversions_data_info.h
```diff
@@ -87,6 +89,241 @@ struct DataInitBase : public DataInitInfo
...
+            for (const auto &mask : masks)
+            {
+                push(pattern & mask);
+            }
+        }
+        return vec;
+    }
+
+    template <typename InType, typename InIntegerType, typename OutType>
+    std::vector<InType> &GetFpSpecialValues()
```
ahesham-arm, line 213: I believe you need to pass `OutFP/outType` to your generator below (constexpr if) so you can tell apart types that are aliases of `uint16_t`, i.e. `cl_half` and `cl_ushort`, because otherwise the case for `cl_half` now matches and uses integers instead of floating point values.

test_conformance/conversions/conversions_data_info.h
```diff
@@ -87,6 +89,241 @@ struct DataInitBase : public DataInitInfo
...
+        };
+
+        // Add all the values close to 0, MIN, and MAX.
+        for (unsigned i = 0; i < 256; i++)
+        {
+            push(i);
+        }
+
+        // Add powers of 2, 3, 5, 7, 10.
+        for (const T base : { 2, 3, 5, 7, 10 })
```
ahesham-arm, line 159: This loop exits before pushing the last power below the sign bit. It misses 2^30 for `uint32_t` and 2^62 for `uint64_t`. Just making sure this is an intentional behaviour.

test_conformance/conversions/basic_test_conversions.cpp
```diff
@@ -690,29 +634,43 @@ int ConversionsTest::DoTest(Type outType, Type inType, SaturationMode sat,
...
+    {
+        if (gTestAll)
+        {
+            nbInputs = (1ULL << 32);
+        }
+        else
+        {
+            if (gWimpyMode)
+            {
+                nbInputs /= gWimpyReductionFactor;
```
ahesham-arm, line 652: Accepted large wimpy factors can reduce `nbInputs` to zero. This will pass parsing but executes no conversion inputs and reports success. I think we should clamp/reject the factor or enforce at least one block.

rjodinchr: Done

test_conformance/conversions/conversions_data_info.h
rjodinchr: Indeed, thank you 🙏

test_conformance/conversions/conversions_data_info.h
rjodinchr: fixed

ahesham-arm (dismissed):

lakshmih (dismissed):

rjodinchr:
I've just rebased it to fix the conflicts

ahesham-arm (approved):

rjodinchr:
Merging as discussed during July 28 teleconference

PR #2712, opened 2026-06-03
api/clone_kernel: Fix kernel compilation failure

rjodinchr:
Isolate 'set_kernel_exec_info_kernel' into its own program source string to prevent compilation failures on OpenCL 3.0 devices that only support OpenCL C 1.2.

The 'set_kernel_exec_info_kernel' uses a struct containing a pointer member (`__global BufPtr* buffer`), which is passed directly as a kernel argument. This construct triggers a multi-level pointer evaluation (`__global int **`) that violates Section 6.11.a of the OpenCL C 1.2 Specification.

Previously, this kernel was bundled inside
`clone_kernel_test_kernel`. On devices running OpenCL 3.0 but defaulting to OpenCL C 1.2 build options, the compiler would reject the entire source string, causing all unrelated `clone_kernel` API tests to fail program compilation.

This change extracts the problematic kernel into a dedicated `clone_kernel_exec_info_kernel` source, ensuring it is only compiled when SVM features are verified and active, preserving test suite conformance on 1.2-only compilers.

### DIFF
--- a/test_conformance/api/test_clone_kernel.cpp
+++ b/test_conformance/api/test_clone_kernel.cpp
@@ -57,10 +57,6 @@ const char* clone_kernel_test_kernel[] = {
         int i;
         float f;
     } structArg;
-    
-    typedef struct {
-        __global int *store;
-    } BufPtr;
 
     // value type test
     __kernel void clone_kernel_test0(int iarg, float farg, structArg sarg,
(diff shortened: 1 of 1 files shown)


lakshmih:
Could this be rebased please?

rjodinchr:
> Could this be rebased please?

That's all set. Out of curiosity, could you help me understand the context for that change? Since I rebased the commit without any conflicts, I'm trying to get a better sense of why that step is needed.

bashbaug (approved):
LGTM.

For any other reviewers, the same restriction is in the latest unified OpenCL C spec also, see:

https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_C.html#restrictions

> Arguments to kernel functions in a program cannot be declared as a pointer to a pointer(s). Variables inside a function or arguments to non-kernel functions in a program can be declared as a pointer to a pointer(s). This restriction only applies to OpenCL C 1.2 or below.

PR #2714, opened 2026-06-03
Fix error handling logic in test_kernel_call_kernel_function

joselopeqti:
Fixes the following issues:

1. Incorrect suppression condition in sections 2 & 3
Both sections use if (errors > 10) twice instead of having a trigger condition (errors == 10). As a result, the "Suppressing further results..." message never prints, and errors are skipped silently.

2. Incorrect pass reporting in section 3
Since errors are global, earlier sub-tests may already push it beyond 10. Section 3 then immediately skips error handling (due to errors > 10), including setting pass = 0, and incorrectly reports "Passed" even when errors occur.

3. Repeated suppression message in section 1
When errors == 10, the code prints the suppression message and continues without incrementing errors. This causes the condition to remain true, printing the message repeatedly for every subsequent error instead of just once, while doing nothing else.

### DIFF
--- a/test_conformance/basic/test_kernel_call_kernel_function.cpp
+++ b/test_conformance/basic/test_kernel_call_kernel_function.cpp
@@ -127,14 +127,18 @@ REGISTER_TEST(kernel_call_kernel_function)
     pass = 1;
     int subtest_errors = 0;
     for (int i=0; i<num_elements; i++) {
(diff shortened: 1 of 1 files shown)


rjodinchr (approved):

kpet:

PR #2715, opened 2026-06-03
vectors: Parallelize alignment tests

rjodinchr:
Modernizes the vector alignment conformance tests by replacing slow nested procedural loops with thread pool parallelization, significantly improving test execution times on multi-core systems.

### DIFF
--- a/test_conformance/vectors/test_vec_align.cpp
+++ b/test_conformance/vectors/test_vec_align.cpp
@@ -19,6 +19,7 @@
 #include "harness/conversions.h"
 #include "harness/typeWrappers.h"
 #include "harness/testHarness.h"
+#include "harness/ThreadPool.h"
 
 #include "structs.h"
 
@@ -26,6 +27,7 @@
 
 #include "type_replacer.h"
 
+#include <array>
 
 size_t get_align(size_t vecSize)
 {
@@ -243,7 +245,7 @@ int test_vec_internal(cl_device_id deviceID, cl_context context,
 }
 
 
-static const char* patterns[] = {
+static const std::vector<const char*> patterns = {
     ".PRAGMA..STATE.\n"
     "__kernel void test_vec_align_array(.SRC_SCOPE. .TYPE..NUM. *source, "
     ".DST_SCOPE. uint *dest)\n"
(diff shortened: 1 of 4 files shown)


rjodinchr:
ref #2723

test_conformance/vectors/test_vec_align.cpp
```diff
@@ -306,50 +308,102 @@ static const char* patterns[] = {
...
+};
 
 
 // alignments of everything in post_substitution_arr as raw alignments
 // 0 if such a thing is meaningless
-size_t post_align_arr[] = { 0, sizeof(cl_char),
-                            0, // taken care of in type_multiple_post_align_arr
-                            0, 0,
-                            0 };
+static const std::array<size_t, ARR_SIZE> post_align_arr = {
```
ahesham-arm, line 353: I don't understand what we gained by changing these to `std::array`.

rjodinchr: All those arrays are used inside `test_vec_thread`. By switching to `std::array<T, ARR_SIZE>`, we enforce compile-time safety to ensure that they all stay tightly bound to the exact same size, protecting against accidental out-of-bounds access.
Written this way, the dimensions are completely explicit and transparent, making the codebase much easier to understand and maintain I believe.

ahesham-arm: I see, thanks. If the information shared across these arrays is tightly coupled in this way, would it make more sense to express this as a `struct` instead? For example:
```
struct foo
{
    const char* pre_sub = nullptr;
    size_t pre_align = 0;
    size_t type_multiple = 0;
    const char* post_sub = nullptr;
};
```
And then define an array of foo like so:
```
static constexpr foo array[] =
{
    {"", 0, 0, ""},
    {"char c;\n", sizeof(char), 0, "char cPost;\n"},
    // etc
};
```

rjodinchr: For sure we could refactor the test even more. That's not what I was after, so I stuck to small improvements.
But I don't think that test really needs more than that right now.

test_conformance/vectors/test_vec_align.cpp
```diff
@@ -306,50 +308,102 @@ static const char* patterns[] = {
...
+    cl_command_queue queue;
+    const char* testName;
+    size_t bufSize;
+    bool packed;
+    const char* source;
+};
+
+cl_int test_vec_thread(cl_uint job_id, cl_uint thread_id, void* userInfo)
+{
+    struct test_vec_thread_info* info = (struct test_vec_thread_info*)userInfo;
```
ahesham-arm: This file is C++, there is no need to prepend `struct` when declaring a variable of type `test_vec_thread_info` (here and elsewhere).

test_conformance/vectors/test_vec_align.cpp
```diff
@@ -306,50 +308,102 @@ static const char* patterns[] = {
     // __attribute__ ((packed))
 };
 
+#define ARR_SIZE (6)
```
ahesham-arm: I prefer constexpr over define because it is typed, but that's a personal preference.

ahesham-arm:
Generally LGTM, but I'll need to test this before approving.

test_conformance/vectors/test_vec_align.cpp
rjodinchr: Done

test_conformance/vectors/test_vec_align.cpp
rjodinchr: Done

ahesham-arm (dismissed):
Test passes. Approved.

If anyone is interested in numbers, I see the CPU utilisation up from ~13% to ~47%. Runtime down from ~29 seconds to ~15 seconds.

test_conformance/vectors/test_vec_align.cpp
shajder: I think it's worth replacing the `is_extension_available` call with a pre-collected flag instead of calling it from each thread.

shajder, line 125: minor: I think `gHasLong` already accounts for embedded platforms via `cles_khr_int64` support, so it'd be more accurate to use here than gIsEmbedded.

shajder (commented):
In general LGTM, test works on my Intel/Nvidia implementations, I am slightly concerned about missing `-m` flag (and possibly `-t`). Take a look at `math_brute_force`, `conversions` or `c11_atomics`, similar logic with `ThreadPool_Do`, all have this flag (SetThreadCount(1) under -m), and it makes debugging driver issues a lot easier.

test_conformance/vectors/test_vec_align.cpp
rjodinchr: That would change the behavior of that test. Right now, if an embedded device runs it, it does not test `long` types. With that change, if the device supports `cles_khr_int64`, it would run for `long` types.

I think that would be fine, but I would prefer it to be a separate PR so that the group can discuss it.

rjodinchr: Agreed, done

rjodinchr:
> In general LGTM, test works on my Intel/Nvidia implementations, I am slightly concerned about missing `-m` flag (and possibly `-t`). Take a look at `math_brute_force`, `conversions` or `c11_atomics`, similar logic with `ThreadPool_Do`, all have this flag (SetThreadCount(1) under -m), and it makes debugging driver issues a lot easier.

I share your concerns, I have created https://github.com/KhronosGroup/OpenCL-CTS/pull/2744 to address it.

shajder (approved):

ahesham-arm (approved):

PR #2718, opened 2026-06-08
half: Improve tests using special values migrated to harness

rjodinchr:
Moves floating-point and integer special value arrays from the conversions
conformance suite into the centralized test harness, making them globally
accessible for the first time.

Leverages these newly available datasets to rewrite the vStoreHalf and
vStoreaHalf reference loops. Instead of using a slow, sequential bit-pattern
loop, the tests now seed blocks with targeted edge cases (NaNs, infinities,
subnormals) before padding with random data—drastically improving precision
coverage and execution efficiency for half-float storage.


rjodinchr:
Depends on #2706 & #2711

rjodinchr:
I'll reopen once/if #2711 gets merged.

PR #2720, opened 2026-06-09
select: Stop testing every integers as cmp value

rjodinchr:
Previously, the select conformance test attempted to sequentially iterate through
large ranges of 32-bit and 64-bit integers for comparison values. This approach
was inefficient and required a "wimpy mode" to arbitrarily skip values in order
to finish within a reasonable timeframe.

Since the OpenCL `select` built-in evaluates the Most Significant Bit (MSB) for
vector types, exhaustive iteration of the entire integer space is unnecessary.
This patch refactors `initCmpBuffer` to systematically generate bitmasks that
toggle the MSB across all vector lanes, filling the remaining bits with random
data. This ensures complete coverage of the vector selection mechanics within a
bounded execution time.

Additionally, this patch cleans up the test harness by:
- Removing the now-obsolete wimpy mode and its custom argument parsing.
- Hoisting `src1` and `src2` host-to-device buffer writes outside of the main
  vector loop, as these buffers only depend on the source type and do not need
  to be re-written for every vector size iteration.

### DIFF
--- a/test_conformance/select/test_select.cpp
+++ b/test_conformance/select/test_select.cpp
@@ -47,8 +47,9 @@ static void initSrcBuffer(void* src1, Type stype, MTdata);
 
 // initialize the valued used to compare with in the select with
 // vlaues [start, count)
-static void initCmpBuffer(void *cmp, Type cmptype, uint64_t start,
-                          const size_t count);
+static void initCmpBuffer(void *cmp, Type cmptype, uint32_t start,
+                          const size_t count, const uint32_t vecsize,
+                          MTdataHolder &d);
 
 // make a program that uses select for the given stype (src/dest type),
 // ctype (comparison type), veclen (vector length)
@@ -82,12 +83,6 @@ static int doTest(cl_command_queue queue, cl_context context,
         }                                                                      \
     }
 
-// When we indicate non wimpy mode, the types that are 32 bits value will
-// test their entire range and 64 bits test will test the 32 bit
-// range.  Otherwise, we test a subset of the range
-// [-min_short, min_short]
-static int s_wimpy_reduction_factor = 256;
-
 //-----------------------------------------
 // Static helper functions
 //-----------------------------------------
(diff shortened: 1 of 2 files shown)


rjodinchr:
Depends on #2706

rjodinchr:
ref #2723

test_conformance/select/test_select.cpp
```diff
@@ -402,79 +418,73 @@ static int doTest(cl_command_queue queue, cl_context context, Type stype, Type c
...
-        err = clEnqueueWriteBuffer(queue, cmp, CL_FALSE, 0, BUFFER_SIZE,
-                                   cmp_host.data(), 0, NULL, NULL);
-        test_error_count(err, "Error: Could not write cmp");
+    for (int vector_idx = 0; vector_idx < VECTOR_SIZE_COUNT; ++vector_idx)
+    {
+        uint32_t vecsize = element_count[vector_idx];
+        size_t vector_size = vecsize * type_size[stype];
+        size_t vector_count = (BUFFER_SIZE + vector_size - 1) / vector_size;
+        uint32_t nb_elements =
+            std::max((uint32_t)64 * 1024, vecsize << vecsize);
```
ahesham-arm: Need `#include <algorithm>` for `std::max`. Is the cast to `uint32_t` needed?

ahesham-arm: This not wrong but I feel like this is the core of the change and it is slightly hidden. For example this is more verbose but it gets the point across of what we're calculating here:
```
const uint32_t full_msb_mask_elements = vecsize * (1u << vecsize);
const uint32_t min_cmp_elements = 64 * 1024;
const uint32_t nb_elements = std::max(min_cmp_elements, full_msb_mask_elements);
```

rjodinchr: I think so, otherwise `64 * 1024` is considered as `int` by some compiler, which creates a mismatch with `vecsize << vecsize` which is `uint32_t` leading to the compiler not finding a `std::max` function that takes `int, uint32_t` as arguments.

rjodinchr: Looks good, I will change that, it will also remove the need for the cast.

ahesham-arm (dismissed):

ahesham-arm (approved):

rjodinchr:
`Check code format` was failing because some commits had been pushed after the previous rebase. I've rebased one more time on tip-of-tree (no other changes). CI should be green.

rjodinchr:
Merging as approved during the July 14th teleconference

PR #2721, opened 2026-06-11
clCopyImage: improve performance by removing redundant copies

rjodinchr:
This patch restructures the image copy conformance tests to eliminate unnecessary
host-side memory allocations and data duplication, significantly improving the
execution speed of the test loops.

Key changes include:
* Removed the redundant `srcHost` buffer allocation and its associated `memcpy`
  operation. The test loop now directly uses `srcData` for validation.
* Split the monolithic `test_copy_image_generic` into `test_copy_init_images`
  (for setup) and `test_copy_image_generic` (for execution). This hoists
  buffer sizing, random data generation, and image creation out of the inner
  testing loops.
* Extracted memory mapping and initialization logic from `create_image` into a
  standalone `init_image` function in `common.cpp`. This allows destination
  images to be efficiently reset during test iterations without fully
  recreating the OpenCL memory objects.
* Consolidated scattered `extern` function declarations across all dimension
  tests into a new `test_copy_generic.h` header.
* Fixed a minor typo in `test_copy_1D_buffer.cpp` where `srcImageInfo` was
  incorrectly passed as the destination descriptor.


rjodinchr:
Depends on #2706 & #2696

rjodinchr:
I'll reopen once/if the dependencies get merged

PR #2724, opened 2026-06-19
Migrate D3D11 suite to use the new test registration framework

ahesham-arm:
Contributes to #2181

### DIFF
--- a/test_conformance/common/directx_wrapper/directx_wrapper.cpp
+++ b/test_conformance/common/directx_wrapper/directx_wrapper.cpp
@@ -92,12 +92,12 @@ DirectX11Wrapper::DirectX11Wrapper()
         ++i;
 
         ComPtr<ID3D11Device> device;
-        hr = D3D11CreateDevice(adapter.Get(), D3D_DRIVER_TYPE_HARDWARE, nullptr,
+        hr = D3D11CreateDevice(adapter.Get(), D3D_DRIVER_TYPE_UNKNOWN, nullptr,
                                0, nullptr, 0, D3D11_SDK_VERSION,
                                device.GetAddressOf(), nullptr, nullptr);
         if (FAILED(hr))
         {
-            throw std::runtime_error("Failed to create DirectX 10 device");
+            throw std::runtime_error("Failed to create DirectX 11 device");
         }
 
         devices.push_back({ adapter, device });
--- a/test_conformance/d3d11/main.cpp
+++ b/test_conformance/d3d11/main.cpp
@@ -274,43 +274,32 @@ void TestAdapterEnumeration(
 
      cl_int result;
 
-    HarnessD3D11_TestBegin("cl_device_id Enumeration");
+     log_info("cl_device_id Enumeration\n");
 
-    // get the cl_device_ids for the adapter
-    {
-        result = clGetDeviceIDsFromD3D11KHR(
-            platform,
-            CL_D3D11_DXGI_ADAPTER_KHR,
-            pAdapter,
-            CL_ALL_DEVICES_FOR_D3D11_KHR,
-            0,
-            NULL,
-            &num_adapter_devices);
-        TestRequire(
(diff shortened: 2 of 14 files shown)


marekkozl (approved):
LGTM

joselopeqti:
@ahesham-arm I'm still reviewing it. I'm observing that if a subtest fails, the result is not captured and the overall test counts as passed. I'll post more comments once I finish the review

.github/workflows/presubmit.yml
```diff
@@ -317,7 +317,7 @@ jobs:
                 -DUSE_CL_EXPERIMENTAL=ON \
                 -DGL_IS_SUPPORTED=OFF \
                 -DVULKAN_IS_SUPPORTED=OFF
-          cmake --build build
+          cmake --build build --parallel
```
rjodinchr: Not sure if you really meant to have that change here?

test_conformance/d3d11/main.cpp
```diff
@@ -13,93 +13,263 @@
...
+}
 
-        // run the test on the adapter
-        HarnessD3D11_CreateDevice(pAdapter, &pDevice, &pDC);
+test_status InitD3D11Device(cl_device_id device)
+{
+    if (!is_extension_available(device, "cl_khr_d3d11_sharing"))
+    {
+        return TEST_SKIPPED_ITSELF;
+    }
```
rjodinchr: REQUIRE
```suggestion
    REQUIRE_EXTENSION("cl_khr_d3d11_sharing");
```

rjodinchr (commented):
LGTM overall, but I have not tested it (I don't have a platform to do it)

test_conformance/d3d11/main.cpp
ahesham-arm: Thanks, updated.

.github/workflows/presubmit.yml
ahesham-arm: You're right, removed. Thanks!

lakshmih:
Reviewing, might need changes/

test_conformance/common/directx_wrapper/directx_wrapper.cpp
joselopeqti, line 81: `CreateDXGIFactory1` instead, as DXGI1.1 features are required

test_conformance/common/directx_wrapper/directx_wrapper.cpp
```diff
@@ -92,12 +92,12 @@ DirectX11Wrapper::DirectX11Wrapper()
         ++i;
```
joselopeqti, line 93: Adapters with the flag `DXGI_ADAPTER_FLAG_SOFTWARE` should be skipped

test_conformance/common/directx_wrapper/directx_wrapper.cpp
```diff
@@ -92,12 +92,12 @@ DirectX11Wrapper::DirectX11Wrapper()
...
 
         ComPtr<ID3D11Device> device;
-        hr = D3D11CreateDevice(adapter.Get(), D3D_DRIVER_TYPE_HARDWARE, nullptr,
+        hr = D3D11CreateDevice(adapter.Get(), D3D_DRIVER_TYPE_UNKNOWN, nullptr,
                                0, nullptr, 0, D3D11_SDK_VERSION,
                                device.GetAddressOf(), nullptr, nullptr);
         if (FAILED(hr))
         {
-            throw std::runtime_error("Failed to create DirectX 10 device");
+            throw std::runtime_error("Failed to create DirectX 11 device");
```
joselopeqti, line 100: Instead of throwing an error here, we should keep checking all adapters and only throw if at the end there are no devices.

test_conformance/common/directx_wrapper/directx_wrapper.cpp
joselopeqti, line 89: Use `ReleaseAndGetAddressOf` instead to avoid leaking adapter references.

joselopeqti (changes requested):

PR #2725, opened 2026-06-24
compiler: Add test for clLinkProgram returning program on linker failure

woodychow:
According to the OpenCL specification (Section 5.6.2), clLinkProgram must return a valid non-zero cl_program object when a linker failure occurs (setting errcode_ret to CL_LINK_PROGRAM_FAILURE). This allows applications to query the build log via clGetProgramBuildInfo to diagnose the linking errors.

See: https://registry.khronos.org/OpenCL/specs/unified/refpages/man/html/clLinkProgram.html

This commit adds a test case that compiles a kernel with an unresolved symbol, links it, and asserts that a valid program object is returned along with CL_LINK_PROGRAM_FAILURE, and that the build log is queryable.

### DIFF
--- a/test_conformance/compiler/test_compile.cpp
+++ b/test_conformance/compiler/test_compile.cpp
@@ -16,6 +16,7 @@
 #include "testBase.h"
 
 #include <filesystem>
+#include <vector>
 
 #if defined(_WIN32)
 #include <time.h>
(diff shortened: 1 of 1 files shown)


test_conformance/compiler/test_compile.cpp
```diff
@@ -1626,6 +1626,77 @@ REGISTER_TEST(simple_link_only)
...
+                                  CL_PROGRAM_BUILD_LOG, 0, NULL, &size_ret);
+    test_error(error, "clGetProgramBuildInfo CL_PROGRAM_BUILD_LOG failed");
+    if (size_ret == 0) {
+        log_error("ERROR: Expected build log to be non-empty\n");
+        clReleaseProgram(my_newly_linked_program);
+        clReleaseProgram(program);
+        return -1;
+    }
+
+    char *log = (char *)malloc(size_ret);
```
rjodinchr: use `std::vector<char>` to avoid having to deal with `malloc`/`free`.

test_conformance/compiler/test_compile.cpp
```diff
@@ -1626,6 +1626,77 @@ REGISTER_TEST(simple_link_only)
...
+            "ERROR: Unable to create a simple test program! (%s in %s:%d)\n",
+            IGetErrorString(error), __FILE__, __LINE__);
+        return -1;
+    }
+
+    error =
+        clCompileProgram(program, 1, &device, NULL, 0, NULL, NULL, NULL, NULL);
+    test_error(error, "Unable to compile a simple program");
+
+    cl_program my_newly_linked_program = clLinkProgram(
```
rjodinchr: `clProgramWrapper`

test_conformance/compiler/test_compile.cpp
```diff
@@ -1626,6 +1626,77 @@ REGISTER_TEST(simple_link_only)
     return 0;
 }
 
+REGISTER_TEST(link_error_returns_program)
+{
+    int error;
+    cl_program program;
```
rjodinchr: Use `clProgramWrapper` (from `harness/typeWrappers.h`) instead of `cl_program`. It will take care of releasing the object when needed.
That would clean that patch a lot.

rjodinchr (changes requested):
Looking good, thank you for the contribution.
Just some nits to make it easier to read/maintain. I understand that this current format is following what we have in the file, but let's make it better for that new test.

test_conformance/compiler/test_compile.cpp
woodychow: Resolved.

test_conformance/compiler/test_compile.cpp
woodychow: Resolved.

test_conformance/compiler/test_compile.cpp
woodychow: Resolved.

woodychow:
Thanks very much for the review. Please take another look!

rjodinchr (approved):

PR #2728, opened 2026-06-25
Format fixes commonfns

Antonios-C:
For replication, setup info:
GCC: 16.1
glibc: 2.43
target architecture build: aarch64
Config used to setup Yocto layers: [Config](https://git.ti.com/cgit/arago-project/oe-layersetup/tree/configs/arago-master-config.txt)

While building Yocto wrynose, recipe in meta-oe layer, following error appears multiple times:

```
test_conformance/commonfns/test_binary_fn.cpp:274:40:   required from here
  274 |         error = test_binary_fn<cl_half>(device, context, queue, num_elems,
      |                 ~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  275 |                                         fnName.c_str(), vecParam,
      |                                         ~~~~~~~~~~~~~~~~~~~~~~~~~
  276 |                                         max_verify<cl_half>);
      |                                         ~~~~~~~~~~~~~~~~~~~~

test_conformance/commonfns/test_binary_fn.cpp:219:72: error: format '%g' expects argument of type 'double', but argument 8 has type 'int' [-Werror=format=]
  219 |                     log_error("x[%d]=%g y[%d]=%g out[%d]=%g, expected %g. "
      |                                                                       ~^
      |                                                                        |
      |                                                                        double
      |                                                                       %d
......
  223 |                               conv_to_flt(out[k]), v, k, i, j, vecSize);
      |                                                    ~
      |                                                    |
      |                                                    int


```
The [commonfns: fix type mismatches across function templates](https://github.com/KhronosGroup/OpenCL-CTS/commit/d85a92794aca418359d60cedc56ad1288e4dbd12) patch fixes the above. This is also the only format error when building, so 
[commonfns: Remove -Wno-format flag](https://github.com/KhronosGroup/OpenCL-CTS/commit/5b099b9c63ebf7b0ded4bc2470326de22c7296a4) re-enables -W-format

### DIFF
--- a/test_conformance/commonfns/test_binary_fn.cpp
+++ b/test_conformance/commonfns/test_binary_fn.cpp
@@ -220,12 +220,15 @@ int max_verify(const T* const x, const T* const y, const T* const out,
                               "(index %d is "
                               "vector %d, element %d, for vector size %d)\n",
                               k, conv_to_flt(x[k]), l, conv_to_flt(y[l]), k,
-                              conv_to_flt(out[k]), v, k, i, j, vecSize);
+                              conv_to_flt(out[k]), conv_to_flt(v), k, i, j,
+                              vecSize);
                 else
                     log_error("x[%d]=%g y[%d]=%g out[%d]=%g, expected %g. "
                               "(index %d is "
                               "vector %d, element %d, for vector size %d)\n",
-                              k, x[k], l, y[l], k, out[k], v, k, i, j, vecSize);
+                              k, conv_to_flt(x[k]), l, conv_to_flt(y[l]), k,
+                              conv_to_flt(out[k]), conv_to_flt(v), k, i, j,
+                              vecSize);
                 return -1;
             }
         }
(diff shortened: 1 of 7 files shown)


rjodinchr (approved):

ahesham-arm (dismissed):

test_conformance/commonfns/test_binary_fn.cpp
```diff
@@ -251,12 +254,15 @@ int min_verify(const T* const x, const T* const y, const T* const out,
...
                               k, conv_to_flt(x[k]), l, conv_to_flt(y[l]), k,
-                              conv_to_flt(out[k]), v, k, i, j, vecSize);
+                              conv_to_flt(out[k]), conv_to_flt(v), k, i, j,
+                              vecSize);
                 else
                     log_error("x[%d]=%g y[%d]=%g out[%d]=%g, expected %g. "
                               "(index %d is "
                               "vector %d, element %d, for vector size %d)\n",
-                              k, x[k], l, y[l], k, out[k], v, k, i, j, vecSize);
+                              k, conv_to_flt(x[k]), l, conv_to_flt(y[l]), k,
```
ahesham-arm, line 263: If the warning/error is complaining that the argument to `%g`, we should use `conv_to_dbl` instead of `conv_to_flt`.

In addition (unrelated to this patch), `conv_to_flt` is wrong. The return value is cast to `float`, but the return type of the function is `double`, so it is immediately promoted to a double.

cc: @rjodinchr

rjodinchr: The warning asked for `double`, but I think it is fine to use `float` with `%g`. It is just the warning message that needs to specify something, and `double` is the default for variadic functions.

I agree there is a bug in `conv_to_flt`, but I believe it should not block that PR.

This is removing warnings from the logs; I think it is fine to merge as it is. Even if it is not perfect, it would definitely be better than the current state.

Antonios-C: I also noticed the behavior of  `conv_to_flt` working on  these patches. https://github.com/KhronosGroup/OpenCL-CTS/pull/2736

ahesham-arm: Thanks for fixing this as well. Agree with @rjodinchr that this shouldn't really block this change.

ahesham-arm (approved):

PR #2731, opened 2026-06-29
Fix the "set but not used" error

cdai2:
When "-Werror=unused-but-set-variable" flag is set.


rjodinchr (approved):

svenvh (approved):

svenvh:
Merging immediately as this should be non-controversial.

PR #2732, opened 2026-06-30
printf: cast `t` and `z` test args to their types

svenvh:
The OpenCL C 3.1 printf length modifiers `t` and `z` require arguments of type `ptrdiff_t` and `size_t` respectively.  For implementations where `CL_DEVICE_ADDRESS_BITS == 64`, these types are 64-bit, but the `ptrdiff_t` and `size_t` printf tests pass integer literals such as 10 as a 32-bit `int` argument.

For example, the following calls use the same format string, but pass arguments with different widths:

    printf("%5td\n", 10);            // int
    printf("%5td\n", 10000000000L);  // long

A variadic printf implementation cannot work around this type mismatch, as it cannot know whether to read 32 or 64 bits from the vararg slot.  Fix by explicitly casting the argument to the tested type.

### DIFF
--- a/test_conformance/printf/test_printf.h
+++ b/test_conformance/printf/test_printf.h
@@ -93,8 +93,6 @@ static std::vector<std::string> correctBufferDouble;
 static std::vector<std::string> correctBufferOctal;
 static std::vector<std::string> correctBufferUnsigned;
 static std::vector<std::string> correctBufferHexadecimal;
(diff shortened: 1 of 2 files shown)


svenvh:
Clang warns about this for 64-bit targets: https://godbolt.org/z/7z3E7cbeG

rjodinchr (approved):

ahesham-arm (approved):

PR #2734, opened 2026-07-01
Added checks to query support for MEMORY_SCOPE_ALL_SVM_DEVICES

joselopeqti:
A few kernels in the SVM and C11_atomics tests use MEMORY_SCOPE_ALL_SVM_DEVICES without querying for its support.

Added a fix to either use a narrower scope or skip the subtest altogether if the support is not present.


rjodinchr (approved):

bashbaug:

PR #2738, opened 2026-07-02
Fix binaries path in test_spirv_new

ahesham-arm:
Update the behaviour of `test_spirv_new` so that the default relative path is resolved in relation to the path of the executable, not the current working directory.

`--spirv-binaries-path` can now handle absolute and relative paths.

Remove the redundant path override from the CSV files, as it is setting it to the default.


rjodinchr (approved):
LGTM

ahesham-arm:
This fixes, for example, failures observed in Mesa CI: https://gitlab.freedesktop.org/ahesham/mesa/-/jobs/103360690/artifacts/external_file/results/SPIRV%20new.c23.r1.log, where the workflow runs the tests from a different working directory, causing all tests to fail with an error of failing to find the SPIR-V binaries.

PR #2739, opened 2026-07-06
Split test_allocations image kernels into multiple dispatches

ahesham-arm:
The purpose of test_allocations is to check that the device can allocate the requested memory sizes successfully. To ensure that the allocated memory is correctly allocated and writable by the device, a checksum kernel is then enqueued on the entire buffer/image.

Without subdividing the work into multiple dispatches, image tests risk triggering timeouts on slower devices. This can trigger, for example, ANRs on
Android (https://developer.android.com/topic/performance/vitals/anr).

Maintain the core logic of the test, but split the verification kernel into multiple dispatches, where each can finish safely within a timeout.

[run-test: test_allocations single 5 image2d_read]
[run-test: test_allocations multiple 5 image2d_read]

### DIFF
--- a/test_conformance/allocations/allocation_execute.cpp
+++ b/test_conformance/allocations/allocation_execute.cpp
@@ -16,7 +16,10 @@
 #include "allocation_execute.h"
 #include "allocation_functions.h"
 
+#include <algorithm>
 #include <vector>
+
+#define MAX_WORK_ITEMS_PER_DISPATCH (64 * 1024)
 
 
 const char *buffer_kernel_pattern = {
@@ -39,7 +42,7 @@ const char *accumulate_pattern = {
 };
 
 const char *image_kernel_pattern = {
-    "__kernel void sample_test(%s __global uint *result)\n"
+    "__kernel void sample_test(%s __global uint *result, uint stride)\n"
     "{\n"
     "\tuint4 color;\n"
     "\tcolor = (uint4)(0);\n"
@@ -52,7 +55,7 @@ const char *image_kernel_pattern = {
 
 const char *read_pattern = {
     "\tfor(y=0; y<get_image_height(image%d); y++)\n"
-    "\t\tif (y %s get_global_size(0) == get_global_id(0))\n"
+    "\t\tif (y %s stride == get_global_id(0))\n"
     "\t\t\tfor (x=0; x<get_image_width(image%d); x++) {\n"
     "\t\t\t\tcolor += read_imageui(image%d, sampler, (int2)(x,y));\n"
     "\t\t\t}\n"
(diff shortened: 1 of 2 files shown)


ahesham-arm:
@rjodinchr you might be interested in this. I think it fits in with the spirit of your recent PRs about reducing needless complexity and time wasted in tests while maintining the core of what is being tested.

test_conformance/allocations/allocation_execute.cpp
```diff
@@ -380,24 +383,48 @@ int execute_kernel(cl_context context, cl_command_queue *queue,
...
+                    "\t\tExecute kernel failed: %s (global dim: %zu, local "
+                    "dim: %zu)\n",
+                    IGetErrorString(error), chunk_dims[0], local_dims[0]);
+            else
+                print_error(error, "clEnqueueNDRangeKernel failed");
+            return result;
+        }
+
+        events.push_back(clEventWrapper(event));
+        clFlush(*queue);
```
rjodinchr: I would have put only one flush after all the kernels have been enqueued

rjodinchr (commented):
Can you set up a run with GitHub CI, please?https://github.com/KhronosGroup/OpenCL-CTS#running-targeted-ci-tests-on-pull-requests

test_conformance/allocations/allocation_execute.cpp
ahesham-arm: Good point, I removed it. There's a `clFinish` call immediately after the loop.

rjodinchr:
You need to set `ci/pocl/golden.json` with what to expect from `test_allocations` as this has never been done before.

rjodinchr:
Because we want the arguments to always match, independently of the order in which they were given to the executable by the user, the harness sorts them. That is why we have a mismatch in the CI at the moment.

ahesham-arm:
@rjodinchr unrelated to this change, building the CTS for pocl could really use `--parallel` as an argument to the CTS build command in presubmit.yml :)

rjodinchr:
> @rjodinchr unrelated to this change, building the CTS for pocl could really use `--parallel` as an argument to the CTS build command in presubmit.yml :)

I would expect `ninja` to build with some level of parallelism by default, but maybe that's not the case on GitHub's runners.

The main issue is that the [pocl cache](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/.github/workflows/presubmit.yml#L238) is cold. I need to do something so that it gets filled on `push` as well, allowing `pull_request` to use it instead of always rebuilding it.

rjodinchr (approved):

ahesham-arm:
That would definitely help. Thanks for the approval.

PR #2740, opened 2026-07-06
[ci] Build pocl and its dependencies in parallel

ahesham-arm:
Pass `--parallel` to the CMake build commands.

[run-test: test_bruteforce exp -w -1]


ahesham-arm:
No measurable difference in build times. Ninja already uses available cores. Closing.

PR #2741, opened 2026-07-06
ci: allow POCL build on push to populate cache

rjodinchr:
Split the `tests` job into three separate jobs:
- `check-triggers`: Parses triggers on pull requests.
- `build-pocl`: Builds and caches POCL. Runs on pushes to any branch or when tests are triggered on PRs.
- `tests`: Runs the tests using the cached POCL.

This allows the POCL build to run on pushes to populate the cache, ensuring subsequent PRs have a cache hit and don't need to rebuild POCL.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -231,16 +231,7 @@ jobs:
       - name: Setup Ninja
         uses: seanmiddleditch/gha-setup-ninja@master
       - name: Install pocl dependencies
-        run: |
-          sudo apt-get update
-          export LLVM_VERSION=20
-          sudo apt-get install -y \
-            python3-dev libpython3-dev build-essential ocl-icd-libopencl1 \
-            cmake git pkg-config libclang-${LLVM_VERSION}-dev clang-${LLVM_VERSION} \
-            llvm-${LLVM_VERSION} make ninja-build ocl-icd-libopencl1 ocl-icd-dev \
-            ocl-icd-opencl-dev libhwloc-dev zlib1g zlib1g-dev clinfo dialog apt-utils \
-            libxml2-dev libclang-cpp${LLVM_VERSION}-dev libclang-cpp${LLVM_VERSION} \
-            llvm-${LLVM_VERSION}-dev
+        uses: ./.github/actions/install-pocl-deps
       - name: Cache pocl
         id: cache-pocl
         uses: actions/cache@v6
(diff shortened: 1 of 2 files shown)


ahesham-arm (dismissed):

.github/workflows/presubmit.yml
```diff
@@ -253,8 +258,42 @@ jobs:
...
+        uses: seanmiddleditch/gha-setup-ninja@master
+      - name: Install Vulkan SDK
+        uses: humbletim/install-vulkan-sdk@main
+        with:
+          version: 1.4.309.0
+          cache: true
+      - name: Restore pocl from cache
+        uses: actions/cache@v5
+        with:
+          path: ${{ env.pocl }}
```
ahesham-arm, line 287: Because this is a running on a different instance of the runner, I think that you're only restoring/installing POCL itself, but it's missing any of its dynamically linked dependencies, i.e. do we need to replicate `Install pocl dependencies` here?

rjodinchr: Indeed, I have push a commit to fix that, let me know what you think

rjodinchr: I've also added a CI run (not sure why I did not add one before in that PR), to make sure CI can run a test with pocl properly

.github/workflows/presubmit.yml
```diff
@@ -265,10 +297,11 @@ jobs:
           cmake --build OpenCL-ICD-Loader/build
           cmake --install OpenCL-ICD-Loader/build
       - name: Build OpenCL CTS
-        if: steps.check_triggers.outputs.run_tests == 'true'
         shell: bash
         run: |
           cmake -B build -S . -G Ninja \
+                -DCMAKE_C_COMPILER_LAUNCHER=sccache \
```
ahesham-arm, line 308: Where is `sccache` configured?

rjodinchr: good point, thank you.

.github/workflows/presubmit.yml
```diff
@@ -264,11 +296,17 @@ jobs:
             -DOPENCL_ICD_LOADER_HEADERS_DIR='${{ env.cl-headers }}'
           cmake --build OpenCL-ICD-Loader/build
           cmake --install OpenCL-ICD-Loader/build
+      - name: Setup ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          variant: sccache
+          key: ubuntu-24.04-aarch64
```
ahesham-arm: I don't think this is correct, we're running pocl natively (x86_64) on this x86_64 runner, right?

ahesham-arm (approved):

PR #2744, opened 2026-07-11
harness: Unify thread pool disabling option

rjodinchr:
Replace test-specific thread pool control with a global option. Previously, tests like conversions and math_brute_force had their own `-m` option to disable multi-threading and called `SetThreadCount(1)`. This API is removed and replaced by a global `-m` / `--disable-threadpool` option parsed by the common test harness.

The `SetThreadCount` API is removed from `ThreadPool.h` and the threaded implementation in `ThreadPool.cpp` as it is no longer needed.

### DIFF
--- a/test_common/harness/ThreadPool.cpp
+++ b/test_common/harness/ThreadPool.cpp
@@ -16,6 +16,7 @@
 #include "ThreadPool.h"
 #include "errorHelpers.h"
 #include "fpcontrol.h"
+#include "parseParameters.h"
 #include <stdio.h>
 #include <stdlib.h>
 
@@ -273,7 +274,7 @@ pthread_cond_t caller_cond_var;
 std::atomic<cl_int> gRunning{ 0 };
 
 // The total number of threads launched.
-std::atomic<cl_int> gThreadCount{ 0 };
+static std::atomic<cl_int> gThreadCount{ 0 };
 
 #ifdef _WIN32
 void ThreadPool_WorkerFunc(void *p)
@@ -470,31 +471,6 @@ exit:
 #endif
 }
 
-// SetThreadCount() may be used to artifically set the number of worker threads
-// If the value is 0 (the default) the number of threads will be determined
-// based on the number of CPU cores.  If it is a unicore machine, then 2 will be
-// used, so that we still get some testing for thread safety.
-//
-// If count < 2 or the CL_TEST_SINGLE_THREADED environment variable is set then
-// the code will run single threaded, but will report an error to indicate that
-// the test is invalid.  This option is intended for debugging purposes only. It
-// is suggested as a convention that test apps set the thread count to 1 in
-// response to the -m flag.
-//
-// SetThreadCount() must be called before the first call to GetThreadCount() or
-// ThreadPool_Do(), otherwise the behavior is indefined.
-void SetThreadCount(int count)
-{
-    if (threadPoolInitErr == CL_SUCCESS)
-    {
-        log_error("Error: It is illegal to set the thread count after the "
-                  "first call to ThreadPool_Do or GetThreadCount\n");
-        abort();
-    }
-
-    gThreadCount = count;
-}
-
 void ThreadPool_Init(void)
 {
     cl_int i;
@@ -503,10 +479,14 @@ void ThreadPool_Init(void)
 
     // Check for manual override of multithreading code. We add this for better
     // debuggability.
-    if (getenv("CL_TEST_SINGLE_THREADED"))
+    if (getenv("CL_TEST_SINGLE_THREADED") || !gThreadPoolEnabled)
     {
-        log_error("ERROR: CL_TEST_SINGLE_THREADED is set in the environment. "
-                  "Running single threaded.\n*** TEST IS INVALID! ***\n");
+        log_info("\n");
+        log_info("*******************************************************\n");
+        log_info("***                  !! WARNING !!                  ***\n");
+        log_info("*** ThreadPool is disabled, running single threaded ***\n");
+        log_info("*******************************************************\n");
+        log_info("\n");
         gThreadCount = 1;
         return;
     }
@@ -591,16 +571,6 @@ void ThreadPool_Init(void)
         gThreadCount = 12;
     }
 #endif
-
-    // Allow the app to set thread count to <0 for debugging purposes.
-    // This will cause the test to run single threaded.
-    if (gThreadCount < 2)
-    {
-        log_error("ERROR: Running single threaded because thread count < 2. "
-                  "\n*** TEST IS INVALID! ***\n");
-        gThreadCount = 1;
-        return;
-    }
 
 #if defined(_WIN32)
     InitializeCriticalSection(gThreadPoolLock);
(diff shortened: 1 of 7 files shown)


test_common/harness/parseParameters.cpp
```diff
@@ -35,6 +35,7 @@ std::string gCompilationProgram = DEFAULT_COMPILATION_PROGRAM;
 bool gDisableSPIRVValidation = false;
 std::string gSPIRVValidator = DEFAULT_SPIRV_VALIDATOR;
 unsigned gNumWorkerThreads;
+unsigned gThreadPoolEnabled = true;
```
shajder: bool ?

test_common/harness/ThreadPool.cpp
```diff
@@ -503,10 +479,14 @@ void ThreadPool_Init(void)
...
+    if (getenv("CL_TEST_SINGLE_THREADED") || !gThreadPoolEnabled)
     {
-        log_error("ERROR: CL_TEST_SINGLE_THREADED is set in the environment. "
-                  "Running single threaded.\n*** TEST IS INVALID! ***\n");
+        log_info("\n");
+        log_info("*******************************************************\n");
+        log_info("***                  !! WARNING !!                  ***\n");
+        log_info("*** ThreadPool is disabled, running single threaded ***\n");
+        log_info("*******************************************************\n");
+        log_info("\n");
```
shajder: I think it's worth discussing whether removing "ERROR" from the message is completely safe. Enabling `ThreadPool_Do` changes the behaviour of the test (e.g. thread-safety of using the same `cl_command_queue` across multiple threads). In that sense, a `run_conformance.py` run in single-threaded mode may slip through unflagged, since this line:

`match = re.search(".*(FAILED|ERROR).*", line)`

will no longer match this message, unlike before.

test_conformance/math_brute_force/main.cpp
```diff
@@ -451,16 +447,6 @@ static test_status ParseArgs(int &argc, const char *argv[],
...
-                    case 'm': singleThreaded ^= 1; break;
-
-                    case 't':
-                        removed_args.pop_back();
-                        removed_args.push_back(std::string(argv[i]) + " "
-                                               + argv[i + 1]);
-                        forcedWorkerThreads = atoi(argv[++i]);
-                        vlog(" %d", forcedWorkerThreads);
-                        break;
-
```
shajder, line 463: I've never needed to use a specific number of threads, so I'm not sure whether it's important or not, but I can imagine this could be useful for tracking down some corner-case races - just wanted to flag that before we remove it.

shajder (commented):
Works as expected, tested with `conversions` and `math_brute_force`. One concern below (removing "ERROR" from the log message) affects how `run_conformance.py` flags invalid runs and probably needs a decision before merge - the other two are minor.

test_common/harness/ThreadPool.cpp
rjodinchr: That's interesting. I've never used `run_conformance.py`.

Note that this is more or less the same layout message as for the wimpy mode, which `run_conformance.py` would slip through unflagged (and I think it would always have done so).

I do believe this is not an error. Maybe we should look for `!! WARNING !!` in `run_conformance.py` to deal with both wimpy and single-threaded? What do you think. I could do that in that PR, it makes sense I believe.

test_conformance/math_brute_force/main.cpp
rjodinchr: I would not expect that removal to be a blocker (anyone could easily change the code if they really need it).
And I don't see a better way to know whether it is used or not than removing it (for sure we want to ask the group as the next teleconference as well).

test_common/harness/parseParameters.cpp
rjodinchr: For sure, good catch. I'll change that.

test_common/harness/ThreadPool.cpp
shajder: It would be best to ask the author, but the "ERROR" approach dates back to the initial commit 9 years ago, so it's hard to say what the original motivation was:

image

As for matching on "!! WARNING !!" - I checked, and that exact string currently only appears in one other place (the wimpy-mode banner in `testHarness.cpp`), so it looks like this message intentionally reused that convention. A plain WARNING match would be too noisy (75+ unrelated hits across the codebase), but the literal "!! WARNING !!" string is precise and would cover both cases. Sounds good to me.

rjodinchr: I'm not sure what people would expect from `run_conformance.py` when running with wimpy and/or single-threaded.
I could see usecases where one would run that script with one of those modes enabled. It is not great if the result is that everything is failing because of the mode enabled.
Maybe having something dedicated to those 2 modes in `run_conformance.py` make more sense. To keep individual test  status, but keeping the information that this is not a valid run to pass OpenCL compliance.

shajder: I see your point, but `failures_this_run` is only driven by a separate .*FAILED.* check. Still, a dedicated flag is cleaner than relying on a highlighted line. Follow-up PR sounds fine to me.

test_conformance/math_brute_force/main.cpp
gwawiork: It is used by Intel :). In this way, we need to limit the work for pre-silicon runs. Otherwise, tests last too long.

rjodinchr: Understood, I'll go ahead and build an equivalent implementation for the harness using the new pattern.

However, it's unclear why the tests drag on without it. Limiting the threads used for host computation could prevent memory issues or reveal bugs, but it's unexpected for this to increase total test duration.

test_common/harness/parseParameters.cpp
```diff
@@ -125,6 +131,32 @@ int parseCommonParamAndGetRemovedArgs(int argc, const char *argv[],
             removed_args.push_back("--wimpy");
             gWimpyMode = true;
         }
+        else if (!strcmp(argv[i], "-m")
```
ahesham-arm, line 134: What happens if we pass both options?

rjodinchr: With both `-m` and `--disable-threadpool` present, the CSV file ends up containing `--disable-threadpool` twice. 

Ensuring that duplicates are removed from the `removed_args` list before CSV output is an option, though it falls outside the current PR's scope. If you'd like, we can open a separate issue to address this independently.

ahesham-arm: Sorry I meant what if we specify -m and -t and the same time.

rjodinchr: I believe the final option provided on the command line takes precedence. Would you like me to ensure they are mutually exclusive?

ahesham-arm: I think that would be better, yes.

rjodinchr: I'm not sure that making them mutually exclusive is a good idea. I push a commit to add a warning message when we override `gNumThreadPoolThreads`. Which means that we have at least 2 options modifying that variable (either twice the same, or both of them).
Let me know what you think

ahesham-arm (approved):

PR #2745, opened 2026-07-14
compiler: skip link_error_returns_program in offline mode

svenvh:
The goal of the `link_error_returns_program` test is to verify that the API returns a usable program object even when linking fails (so that a build log can still be queried).

In offline compilation mode, there isn't necessarily a returned program object to inspect, as an offline compiler may produce just a diagnostic without any binary output.  In that case there won't be a binary to load, so skip this test for offline compilation.


rjodinchr (approved):

bashbaug:

Issue #2723, opened 2026-06-18
Reducing CTS runtime


rjodinchr:
This issue explains all the changes I've made to optimize the OpenCL-CTS to reduce the time it takes to run by more than 10x (up to 20x depending on the actual driver/device).

The overall changes can be found in [my branch](https://github.com/rjodinchr/OpenCL-CTS/tree/main)

----------------------------------------------------------
### Special Values Management
* Group float and double special values into a single file (`conversions.cpp`).
* Expose `GetFpSpecialValues` (for floats) and `GetIntSpecialValues` (for integers) in `conversions.h`.
* Implement a static buffer with a mutex to cache computations, ensuring thread-safe concurrent access.

### Test Execution & Parsing
* Rename `parseCustomParam` to `parseCommonParamAndGetRemovedArgs` to better reflect its actual behavior.
* Add `removed_args` and `help` as output parameters to improve usage clarity and track function actions later in the flow.
* Introduce a `ParseArgsFn` parameter to `runTestHarnessWithCheckAndParse` to reduce boilerplate code while retaining necessary data for JSON generation.
* Clean up memory management by removing excessive C-style `malloc`, `calloc`, and `free` operations.
* Fix parsing issues for previously unreachable arguments.

### Registry & Output
* Remove `testNum` and `testList` from `runTestHarness*` functions to retrieve them directly from the test registry instead.
* Update `test_registry` templates to remove unnecessary private fields and classes, streamlining dynamic registration.
* Refine the test listing behavior (`--list`) to only print subtests when explicitly requested, reducing visual clutter.
* Implement `cl_channel_type_to_string` for finer-grained image sub-tests to improve parallelization performance.

### CI
* Update `saveResultsToJson` to include meaningful arguments, making it easier for dynamic CI testing to track expected outcomes.
* Update CI comparison scripts (`compare_results.py`) to validate against specific execution arguments recorded in `golden.json`.

----------------------------------------------------------
## Tests Only Modified to Match New Harness

Updated the following tests solely to match the new harness:
[...]

PR #2748, opened 2026-07-15
clCopyImage: improve performance by removing redundant copies

rjodinchr:
This patch restructures the image copy conformance tests to eliminate unnecessary host-side memory allocations and data duplication, significantly improving the execution speed of the test loops.

Key changes include:
* Removed the redundant `srcHost` buffer allocation and its associated `memcpy` operation. The test loop now directly uses `srcData` for validation.
* Split the monolithic `test_copy_image_generic` into `test_copy_init_images` (for setup) and `test_copy_image_generic` (for execution). This hoists buffer sizing, random data generation, and image creation out of the inner testing loops.
* Extracted memory mapping and initialization logic from `create_image` into a standalone `init_image` function in `common.cpp`. This allows destination images to be efficiently reset during test iterations without fully recreating the OpenCL memory objects.
* Consolidated scattered `extern` function declarations across all dimension tests into a new `test_copy_generic.h` header.
* Fixed a minor typo in `test_copy_1D_buffer.cpp` where `srcImageInfo` was incorrectly passed as the destination descriptor.

Ref #2723

[run-test: test_cl_copy_images --num-worker-threads 2 small_images]

### DIFF
--- a/test_conformance/images/clCopyImage/test_copy_1D.cpp
+++ b/test_conformance/images/clCopyImage/test_copy_1D.cpp
@@ -13,15 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 //
-#include "../testBase.h"
-
-extern int test_copy_image_generic(cl_context context, cl_command_queue queue,
-                                   image_descriptor *srcImageInfo,
-                                   image_descriptor *dstImageInfo,
-                                   const size_t sourcePos[],
-                                   const size_t destPos[],
-                                   const size_t regionSize[], MTdata d,
-                                   const image_test_context_t &ctx);
+#include "test_copy_generic.h"
 
 int test_copy_image_size_1D(cl_context context, cl_command_queue queue,
                             image_descriptor *srcImageInfo,
@@ -69,9 +61,19 @@ int test_copy_image_size_1D(cl_context context, cl_command_queue queue,
         regionSize[ 0 ] = width_lod;
     }
 
+    clMemWrapper srcImage, dstImage;
+    BufferOwningPtr<char> srcData, dstData;
     retCode =
-        test_copy_image_generic(context, queue, srcImageInfo, dstImageInfo,
-                                sourcePos, destPos, regionSize, d, ctx);
+        test_copy_init_images(context, queue, srcImageInfo, dstImageInfo,
+                              srcImage, dstImage, srcData, dstData, d, ctx);
+    if (retCode != CL_SUCCESS)
+    {
+        return retCode;
+    }
+    retCode = test_copy_image_generic(
+        context, queue, srcImageInfo, dstImageInfo, srcImage, dstImage, srcData,
+        dstData, sourcePos, destPos, regionSize, d, ctx);
+
     if( retCode < 0 )
       return retCode;
     else
@@ -107,9 +109,9 @@ int test_copy_image_size_1D(cl_context context, cl_command_queue queue,
 
 
       // Go for it!
-      retCode =
-          test_copy_image_generic(context, queue, srcImageInfo, dstImageInfo,
-                                  sourcePos, destPos, regionSize, d, ctx);
+      retCode = test_copy_image_generic(
+          context, queue, srcImageInfo, dstImageInfo, srcImage, dstImage,
+          srcData, dstData, sourcePos, destPos, regionSize, d, ctx);
       if( retCode < 0 )
         return retCode;
       else
--- a/test_conformance/images/clCopyImage/test_copy_1D.cpp
+++ b/test_conformance/images/clCopyImage/test_copy_1D.cpp
@@ -61,18 +61,15 @@ int test_copy_image_size_1D(cl_context context, cl_command_queue queue,
         regionSize[ 0 ] = width_lod;
     }
 
-    clMemWrapper srcImage, dstImage;
-    BufferOwningPtr<char> srcData, dstData;
-    retCode =
-        test_copy_init_images(context, queue, srcImageInfo, dstImageInfo,
-                              srcImage, dstImage, srcData, dstData, d, ctx);
+    copy_image_env_t env{ context, queue, d, ctx };
+    copy_image_buffers_t buffers;
+    retCode = test_copy_init_images(env, srcImageInfo, dstImageInfo, buffers);
     if (retCode != CL_SUCCESS)
     {
         return retCode;
     }
-    retCode = test_copy_image_generic(
-        context, queue, srcImageInfo, dstImageInfo, srcImage, dstImage, srcData,
-        dstData, sourcePos, destPos, regionSize, d, ctx);
+    retCode = test_copy_image_generic(env, srcImageInfo, dstImageInfo, buffers,
+                                      sourcePos, destPos, regionSize);
 
     if( retCode < 0 )
       return retCode;
@@ -109,11 +106,11 @@ int test_copy_image_size_1D(cl_context context, cl_command_queue queue,
 
 
       // Go for it!
-      retCode = test_copy_image_generic(
-          context, queue, srcImageInfo, dstImageInfo, srcImage, dstImage,
-          srcData, dstData, sourcePos, destPos, regionSize, d, ctx);
-      if( retCode < 0 )
-        return retCode;
+      retCode =
+          test_copy_image_generic(env, srcImageInfo, dstImageInfo, buffers,
+                                  sourcePos, destPos, regionSize);
+      if (retCode < 0)
+          return retCode;
       else
         ret += retCode;
     }
--- a/test_conformance/images/clCopyImage/test_copy_generic.cpp
+++ b/test_conformance/images/clCopyImage/test_copy_generic.cpp
@@ -47,16 +47,6 @@ int test_copy_init_images(copy_image_env_t &env, image_descriptor *srcImageInfo,
     int error;
 
     // Generate some data to test against
-    size_t srcBytes = 0;
-    if (env.ctx.testMipmaps)
-    {
-        srcBytes = (size_t)compute_mipmapped_image_size( *srcImageInfo );
-    }
-    else
-    {
-        srcBytes = get_image_size(srcImageInfo);
-    }
-
     if (env.ctx.debugTrace) log_info(" - Resizing random image data...\n");
 
     generate_random_image_data(srcImageInfo, buffers.srcData, env.d);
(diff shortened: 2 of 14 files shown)


test_conformance/images/clCopyImage/test_copy_generic.cpp
```diff
@@ -105,38 +87,42 @@ int test_copy_image_generic(cl_context context, cl_command_queue queue,
...
+                            const size_t sourcePos[], const size_t destPos[],
+                            const size_t regionSize[], MTdata d,
+                            const image_test_context_t &ctx)
+{
+    int error;
+
+    if (ctx.debugTrace) log_info(" ++ Entering inner test loop...\n");
+
+    memset(dstHost, 0xff, dstHost.getSize());
+    init_image(queue, dstImage, dstImageInfo, dstHost, ctx.testMipmaps);
```
shajder: Shouldn't the return value of `init_image()` be checked here, the same way it is at the `create_image()` call site above, otherwise a failed map/unmap would silently leave `dstImage` in a stale state before the comparison?

test_conformance/images/clCopyImage/test_copy_1D.cpp
```diff
@@ -69,9 +61,19 @@ int test_copy_image_size_1D(cl_context context, cl_command_queue queue,
...
+        test_copy_init_images(context, queue, srcImageInfo, dstImageInfo,
+                              srcImage, dstImage, srcData, dstData, d, ctx);
+    if (retCode != CL_SUCCESS)
+    {
+        return retCode;
+    }
+    retCode = test_copy_image_generic(
+        context, queue, srcImageInfo, dstImageInfo, srcImage, dstImage, srcData,
+        dstData, sourcePos, destPos, regionSize, d, ctx);
+
```
shajder: It is nothing crucial, just something I would do. Both calls here take a long list of positional parameters (`context`, `queue`, `d`, `ctx`, invariant for the whole size test plus `srcImage`/`dstImage`/`srcData`/`dstData`, which travel together as one unit from `test_copy_init_images` to `test_copy_image_generic`), and this exact pattern repeats in every `test_copy_*.cpp` file touched by this PR. Would it be worth grouping these into two small structs (e.g. an invariant "env" struct for `context/queue/d/ctx`, and a "buffers" struct for the four `srcImage/dstImage/srcData/dstData` members) instead of passing them all individually? It would shrink both signatures considerably and remove a class of copy-paste mistakes from having several same-typed parameters in a row.
Moreover, making `test_copy_init_images`/`test_copy_image_generic` methods on the structs themselves. context, queue, d, ctx and the `srcImage/dstImage/srcData/dstData` would become members, and the calls would shrink to something like `buffers.init(srcImageInfo, dstImageInfo)` and `buffers.copy(sourcePos, destPos, regionSize)`, with nothing left in the signature except what actually changes per call.

test_conformance/images/clCopyImage/test_copy_1D.cpp
```diff
@@ -69,9 +61,19 @@ int test_copy_image_size_1D(cl_context context, cl_command_queue queue,
         regionSize[ 0 ] = width_lod;
     }
 
+    clMemWrapper srcImage, dstImage;
+    BufferOwningPtr<char> srcData, dstData;
```
shajder: Wouldn't it be a real benefit to initialize `srcData/dstData` once for each test, sized to the maximum dimensions this test entry will ever exercise, instead of allocating and randomizing them fresh inside every call to `test_copy_image_*`? `test_copy_image_set_*` calls this function once per size (potentially hundreds of times across the small/max/random size loops), and each call currently pays for a fresh `malloc` plus `generate_random_image_data` even though the content only needs to be large enough to back the current image, not freshly random per size. A buffer allocated once at the top of `test_copy_image_set_*` and passed down as a sized view could remove most of that per-size host-side cost, on top of what this PR already removes at the per-region level. It's a bigger change than the rest of the review, so happy to raise it as a separate follow-up if you'd rather keep this PR scoped to what it already covers.

shajder (commented):
LGTM, test passes on my local implementations. A few comments below and one general consideration for discussion:

`test_copy_image_set_2D_2D_array`, for example, iterates roughly 585 combinations of width/height/arraySize (4..17 x 4..13 x 4..9) purely sequentially, and the equivalent loops exist in every `test_copy_image_set_*` in this directory.

Would it be worth dispatching those iterations through `ThreadPool_Do` instead of a plain nested loops? It looks like a clean fit for the existing pattern (build the list of size/parameter combinations, then hand it to `ThreadPool_Do` as jobs), though it would need per-thread RNG state and ideally per-thread command queues to be safe. This is well beyond the scope of what this PR already does, so not a request to expand it here, just pointing it out since it sounds like a good candidate for larger refactoring. Might be worth its own issue.

test_conformance/images/clCopyImage/test_copy_generic.cpp
rjodinchr: Good catch, I'll change that

test_conformance/images/clCopyImage/test_copy_1D.cpp
rjodinchr: Let's keep that idea for another PR.
To be honest, I would not expect much perf improvement out of the change you propose, but it would still be worth trying.
I'm still looking for ways to improve image tests, but I have other PRs after this one that I verified which will bring good perf improvement.

test_conformance/images/clCopyImage/test_copy_1D.cpp
rjodinchr: Sounds good to me, I'll make the change

rjodinchr:
> LGTM, test passes on my local implementations. A few comments below and one general consideration for discussion:
> 
> `test_copy_image_set_2D_2D_array`, for example, iterates roughly 585 combinations of width/height/arraySize (4..17 x 4..13 x 4..9) purely sequentially, and the equivalent loops exist in every `test_copy_image_set_*` in this directory.
> 
> Would it be worth dispatching those iterations through `ThreadPool_Do` instead of a plain nested loops? It looks like a clean fit for the existing pattern (build the list of size/parameter combinations, then hand it to `ThreadPool_Do` as jobs), though it would need per-thread RNG state and ideally per-thread command queues to be safe. This is well beyond the scope of what this PR already does, so not a request to expand it here, just pointing it out since it sounds like a good candidate for larger refactoring. Might be worth its own issue.

I like the idea, I'll explore that for another PR. Thank you for the review

shajder (dismissed):

test_conformance/images/clCopyImage/test_copy_1D.cpp
shajder: I put together a quick and dirty prototype of this on `test_copy_image_set_2D_2D_array` only (not the full suite) to get real numbers instead of guessing. Wanted to be upfront that it's a rough hack, not PR-quality: hardcoded max format assumption, a static buffer that only gets initialized on the first call.

With those fixes, combined across both directions of this one test (2Darrayto2D + 2Dto2Darray), it's up to 26% faster overall (23069.04ms vs 17075.67ms). Not sure if it's worth the effort given the numbers and refactoring scope but wanted to report for completeness.

test_conformance/images/clCopyImage/test_copy_generic.cpp
```diff
@@ -68,35 +57,23 @@ int test_copy_image_generic(cl_context context, cl_command_queue queue,
         srcBytes = get_image_size(srcImageInfo);
     }
```
shajder: While testing performance I've noticed it is probably dead code, `srcBytes` is never used

test_conformance/images/clCopyImage/test_copy_generic.cpp
```diff
@@ -105,44 +82,50 @@ int test_copy_image_generic(cl_context context, cl_command_queue queue,
         destImageSize = get_image_size(dstImageInfo);
     }
```
shajder: As above, `destImageSize` no longer needed, it seems.

shajder (commented):
Minor suggestions

ahesham-arm (dismissed):

test_conformance/images/clCopyImage/test_copy_1D.cpp
rjodinchr: Good to know! I'll have a closer look at it then. Thank you for trying it out 🙏

ahesham-arm (approved):

shajder (approved):

bashbaug:

test_conformance/images/clCopyImage/test_copy_1D.cpp
rjodinchr: Can you try your prototype with https://github.com/KhronosGroup/OpenCL-CTS/pull/2754?
Make sure to run with `--num-worker-threads 2`, and maybe even more (4, 8).

shajder: @rjodinchr 

--num-worker-threads 2: common buffer 16 027.52 ms vs. separate buffer 17 318.03 ms, 7.45% faster
--num-worker-threads 4: common buffer 15 048.01 ms vs. separate buffer 15 865.45 ms, 5.15% faster
--num-worker-threads 8: common buffer 15 567.49 ms ms vs. separate buffer 15 813.85 ms, 1.5% faster

In this case implementation cost is rising: we can no longer use a shared `dstData` buffer, and the `srcData` buffer must be initialized with a mutex guard and secured agains premature release. I’d say these two concepts contradict each other.

rjodinchr: `common buffer` corresponds to your prototype + #2754, and `separate buffer` #2754 without your prototype. Is that correct?

Side note: the number you are reporting should not be that precise, the time it takes to run the test is not stable enough to report that many digits. 3 or 4 should be more than enough: `16 027.52 ms => 16.0 s`, the rest is just meaningless.

shajder: > `common buffer` corresponds to your prototype + #2754, and `separate buffer` #2754 without your prototype. Is that correct?

that is correct

> Side note: the number you are reporting should not be that precise, the time it takes to run the test is not stable enough to report that many digits. 3 or 4 should be more than enough: `16 027.52 ms => 16.0 s`, the rest is just meaningless.

Indeed, I put additional chrono measurements to get results with a better precision.

PR #2749, opened 2026-07-15
math_brute_force: Refactor input generation & reduce default test size

rjodinchr:
This patch addresses execution time and input coverage in the math_brute_force tests by centralizing input generation and scaling down the default number of tested values.

Key changes:
* Abstracted input generation: Replaced duplicated array-filling logic across test files with shared `fill*Input` helper functions (for unary, binary, and ternary operations). This ensures that special edge-case values (NaNs, Infs, denormals) and random payloads are consistently injected across all tests.
* Reduced default test size: The default loop bounds have been reduced from an exhaustive `1ULL << 32` values to a randomized subset of `1ULL << 27` inputs. This drastically speeds up standard test runs.
* Added exhaustive `-a` flag: Introduced a new `-a` command-line option that restores the exhaustive testing behavior, allowing validation against all 2^32 sequential values when needed.
* Improved wimpy mode scaling: Removed the `scale` and `step` skip logic, which could inadvertently skip combinations of special values. Instead, the patch now directly reduces the total `input_count_power_of_two` by using a De Bruijn sequence bit-hack to calculate the `log2` of the `wimpyReductionFactor`.

Ref #2723
[run-test: test_bruteforce -w -1]


test_conformance/math_brute_force/common.cpp
```diff
@@ -848,20 +842,199 @@ static const std::vector<cl_half> halfSpecialValues = {
...
+    if (gWimpyMode)
+    {
+        input_count_power_of_two -= getExponentDeBruijn(wimpyReductionFactor);
+    }
+    if (gIsEmbedded)
+    {
+        input_count_power_of_two -=
+            getExponentDeBruijn(EMBEDDED_REDUCTION_FACTOR);
+    }
+}
```
shajder, line 881: This looks like a real change to test semantics, not just a wimpy-mode tweak: `input_count_power_of_two` defaults to 27 regardless of mode, and only reaches 32 with `-a`. `opencl_conformance_tests_full.csv`, the list used for official conformance submissions, invokes `math_brute_force/test_bruteforce` with no flags, so submissions would move from exhaustive **2^32** coverage to a random **2^27** subset without any change to that file or a WG decision. I think this needs WG sign-off before merging, either by adding `-a` to that entry or by discussing the coverage trade-off explicitly.

test_conformance/math_brute_force/common.cpp
```diff
@@ -848,20 +842,199 @@ static const std::vector<cl_half> halfSpecialValues = {
...
+        data_short[idx] = (cl_ushort)(gen & 0xffff);
+        data_short[idx + 1] = (cl_ushort)(gen >> 16);
+    }
+    if (idx < num_elems)
+    {
+        data_short[idx] = (cl_ushort)(genrand_int32(d) & 0xffff);
+    }
+}
+void fillFloatUnaryInput(float *data, size_t num_elems, size_t base_elem,
+                         MTdata d, bool testAll)
```
shajder: Some of those fill* functions seem to be a great candidate for unification, there is plenty of redundant code and it could be turned into templates. I am aware this is not exactly the point of this PR, just something worth mentioning.

test_conformance/math_brute_force/unary_float.cpp
```diff
@@ -460,18 +459,7 @@ int TestFunc_Float_Float(const Func *f, MTdata d, bool relaxedMode)
...
-    if (test_info.step / test_info.subBufferSize != test_info.scale)
-    {
-        // there was overflow
-        test_info.jobCount = 1;
-    }
-    else
-    {
-        test_info.jobCount = (cl_uint)((1ULL << 32) / test_info.step);
-    }
+    test_info.jobCount = getInputCount() / test_info.subBufferSize;
```
shajder: `std::max` for consistency?

test_conformance/math_brute_force/i_unary_double.cpp
```diff
@@ -45,8 +45,6 @@ int TestFunc_Int_Double(const Func *f, MTdata d, bool relaxedMode)
     KernelMatrix kernels;
     int ftz = f->ftz || gForceFTZ;
     uint64_t step = getTestStep(sizeof(cl_double), BUFFER_SIZE);
-    int scale =
-        (int)((1ULL << 32) / (16 * BUFFER_SIZE / sizeof(cl_double)) + 1);
```
shajder, line 48: I wonder about the scope of the change in test behavior caused by this removal. As far as I understand, wimpy mode previously reduced coverage two ways: `getTestStep` returned a much larger step to skip most buffers, and within each processed buffer, scale spread the samples across nearly the whole 32-bit range. getTestStep is now unconditional (bufferSize / typeSize regardless of gWimpyMode), and buffer count reduction is driven entirely by `getInputCount()` instead. I don't see a replacement for the scale-based structured spread: the buffer is now filled with a few special values plus `genrand_int32`, so coverage of the reduced set is random rather than deterministic and evenly spaced. Can you confirm this was an intentional simplification and not just dropped in the refactor?

test_conformance/math_brute_force/binary_double.cpp
```diff
@@ -446,18 +411,8 @@ int TestFunc_Double_Double_Double(const Func *f, MTdata d, bool relaxedMode)
...
-    {
-        // there was overflow
-        test_info.jobCount = 1;
-    }
-    else
-    {
-        test_info.jobCount = (cl_uint)((1ULL << 32) / test_info.step);
-    }
+    test_info.jobCount = std::max(
+        (cl_uint)1, (cl_uint)(getInputCount() / test_info.subBufferSize));
```
shajder: nitpick: I bet test_info.subBufferSize is never meant to be zero, but since we're touching this line, maybe it's worth putting in some assertion? Plenty of such calls in this PR

rjodinchr: Done

test_conformance/math_brute_force/unary_float.cpp
rjodinchr: Done as part of `test_info.jobCount` refactoring

test_conformance/math_brute_force/common.cpp
rjodinchr: Good idea, done

test_conformance/math_brute_force/common.cpp
rjodinchr: Thanks for raising this concern. This change in default test scope was discussed at the last F2F meeting. There were no objections at the time, and the group generally agreed that the reduction in execution time brings more benefit than the reduction in coverage harms.

One of the main arguments is that for FP64 or pairs of FP32 (binary operations), the testing strategy is already non-exhaustive; we rely on a subset designed with special edge-case values and random payloads. Since we are confident with this approach for those configurations, we should be able to apply the same strategy to unary FP32 tests as well to align the behaviors.

While formal approval from the Working Group is still pending, please note that the conversion tests have already adopted this same logic in PR #2711.

test_conformance/math_brute_force/i_unary_double.cpp
rjodinchr: Yes, this was an intentional simplification. My primary focus for the wimpy runs is to ensure all special values (edge cases like NaNs, Infs, denormals) are verified.

For the remainder of the reduced test space, we transition to random generation. Since pseudo-random values are by design evenly distributed across the value space, they provide good coverage of normal ranges without the need for the complex scale-based spacing logic. Note that the total number of tested values (and thus the density of the random coverage) can still be adjusted using the wimpy reduction factor option.

ci/pocl/golden.json
```diff
@@ -45,7 +45,7 @@
             "fabs": "pass",
             "fdim": "fail",
             "floor": "pass",
-            "fma": "pass",
+            "fma": "fail",
```
ahesham-arm, line 48: Why are we introducing a refactor that makes passing tests fail?

rjodinchr: Actually, wimpy mode is now utilizing all special values instead of testing against sparse values from the exhaustive list. This new strategy enables the testing of meaningful values even within wimpy mode, which helps capture failures that the current implementation misses (in wimpy mode).

Issue #2750, opened 2026-07-15
Extensions Testing and Conformance Verification


rjodinchr:
## The Problem

The current OpenCL conformance process relies on running tests listed in CSV files (such as `opencl_conformance_tests_full.csv`), as described in the [generating a conformance report](https://github.com/KhronosGroup/OpenCL-CTS#generating-a-conformance-report) section of the README.

This approach has several limitations:
1.  **Outdated Conformance Lists:** The CSV files often lag behind, omitting tests for newly added extensions.
2.  **Maintenance Overhead:** Updating and maintaining these CSV files manually is awkward and error-prone (refer to discussions in PR #2096).
3.  **Build-Time vs. Runtime Gating:** Many extension tests are gated behind CMake flags and are not built by default. This makes it difficult for implementers to know what needs to be run to claim full conformance, and complicates the build process.
4.  **Poor Discoverability:** It is not obvious to external implementers which extension tests exist and should be run.

## Proposed Solutions / Discussion Points

We propose addressing these issues by moving away from build-time gating and CSV-based test selection, towards consolidated binaries and runtime feature detection.

### 1. Consolidate Test Binaries
*   **One CTS Binary:** Ideally, collapse the entire Conformance Test Suite into a single executable.
*   **One Extensions Binary:** As an intermediate step, consolidate all extension tests into a single executable, separate from the core CTS binary.

### 2. Always Build, Runtime Skip (with Stub Implementations)
Instead of hiding extension tests behind CMake flags:
*   **Always build all extension tests** by default.
*   **Runtime Skipping:** Tests should dynamically query the device (e.g., via `CL_DEVICE_EXTENSIONS` or `CL_DEVICE_EXTENSIONS_WITH_VERSION`) to check if the extension is supported. If not supported, the test should be gracefully skipped.
*   **Stub Implementations:** To facilitate building tests for extensions that might not be supported by the host OpenCL loader/runtime at link time, we should employ stub implementations.
    *   If a test is run for an unsupported extension, it must skip.
    *   If the test fails to skip and attempts to call the extension API through the stub, the stub implementation should ensure a clear error is returned (and the test fails), rather than crashing or silently passing.

### 3. Long-Term Direction: dEQP-like Architecture
*   Consider a long-term migration towards a dEQP-like framework. This would involve a single runner with dynamic test discovery and filtering capabilities, eliminating the need for maintaining static CSV files for conformance runs.

PR #2751, opened 2026-07-15
Always build all test binaries and validate full csv file

rjodinchr:
This PR enables build-time validation of all conformance test binaries against the CSV list, and introduces a preprocessor guard strategy for compiling tests when their corresponding features are not supported by the build configuration.
### Key Changes:
* **Always Build All Binaries**: CMake is modified to always configure and build all test modules (Vulkan, D3D, GL, GLES, and various interop extensions) instead of conditionally excluding them based on build-time support.
* **CSV Validation**: Added a build-time check to ensure that all compiled test binaries are present in `test_conformance/opencl_conformance_tests_full.csv`. The build will fail if any binary is missing.
* **Preprocessor Guard Strategy**: 
  * Test sources that directly depend on unavailable external headers or APIs are excluded from compilation in CMake.
  * The remaining compiled sources (including the test entry points) are guarded with preprocessor macros to disable test logic when build-time support is missing, avoiding the need for mock headers or stubs.
* **Runtime Mismatch Detection**: 
  * Guarded binaries check at runtime if the device supports the corresponding extension.
  * The test fails if the device supports the extension (indicating a build/device mismatch where a supported feature is not being tested).
  * The test is gracefully skipped if the device does not support the extension.

Ref #2750


test_common/harness/deviceInfo.cpp
```diff
@@ -185,3 +186,51 @@ cl_ulong get_device_info_max_constant_buffer_size(cl_device_id device,
...
+                                      "stub_check_extensions" } };
+
+    return runTestHarness(argc, argv, 1, test_list, false, 0);
+}
+
+int report_skipped_stub(const char* testName)
+{
+    log_info("%s test skipped (built as stub).\n", testName);
+    return 0;
+}
```
bashbaug: Is this function needed? It seems like the tests that are stubbed should always be tied to at least one extension that we can check.

For example, the D3D10 test is currently skipped when it is built as a stub, but it seems like could be checking for cl_khr_d3d10_sharing ~~and cl_khr_d3d11 sharing~~ instead, and then call run_extension_stub.

test_conformance/d3d10/CMakeLists.txt
```diff
@@ -14,5 +14,8 @@ list(APPEND CLConform_LIBRARIES directx_wrapper)
 
 include(../CMakeCommon.txt)
 else()
-message(STATUS "D3D10 tests are only supported on Windows.")
+set(${MODULE_NAME}_SOURCES
+    stub.cpp
+)
```
bashbaug: I'd find it useful here to log a message to indicate that the test is built as a stub.  Same goes for the other tests that are built as a stub.

bashbaug (dismissed):
This is great!  Just a few suggestions for additional improvements.

test_common/harness/deviceInfo.cpp
rjodinchr: I agree in principle.
I find it weird then not to find them in the `extensions` folder. But I guess it's not a big deal, it can stay like that.
It is not obvious for each of them which extensions are the proper ones to look for. 

- `d3d10`:
  - `cl_khr_d3d10_sharing`
- `d3d11`:
  - `cl_khr_d3d11_sharing`
- `gl`:
  - `cl_khr_gl_event`
  - `cl_khr_gl_msaa_sharing`
  - `cl_khr_gl_depth_images`
  - `cl_khr_gl_msaa_sharing`
- `gles`:
  - Same as `gl`?
- `vulkan`:
  - `cl_khr_external_memory_win32`
  - `cl_khr_external_memory_opaque_fd`
  - `cl_khr_external_memory`
  - `cl_khr_external_semaphore`

I guess for `d3d10`, `d3d11` & `gl` it is clear, but for `gles` & `vulkan`, I'm not sure what should be done.

test_conformance/d3d10/CMakeLists.txt
rjodinchr: Good idea, I'll do that

test_common/harness/deviceInfo.cpp
rjodinchr: The lists above are the extensions I found in each test folder (using `git grep`).

bashbaug: Ah right, there are separate test executables for d3d10 and d3d11.

This list looks right to me, or at least good enough to start.  I'm not completely sure about `gles` either, but checking for the same extensions as `gl` seems reasonable.

For `vulkan`, we're probably OK only checking for `cl_khr_external_memory` and `cl_khr_external_semaphore` since the extensions for the specific handle types depend on the base extensions.  If it's simpler to just list all of them though, that's fine too.

bashbaug (dismissed):

rjodinchr:
I was not happy about the stub I previously made. I fixed it with the second commit. Have a look at the commit message itself for more explanation on what it achieves.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
```diff
@@ -54,6 +54,18 @@
 
 static const char* source = "__kernel void empty() {}";
 
+static bool ooo_queues_supported(cl_device_id device)
```
bashbaug: Might be good to move this to the harness so other tests could use it also.

test_conformance/gl/CMakeLists.txt
```diff
@@ -1,16 +1,20 @@
...
-else(WIN32)
-    list(APPEND CLConform_LIBRARIES  GL glut GLEW GLU)
-endif(WIN32)
+#     list(APPEND CLConform_LIBRARIES opengl32 freeglut glew32)
+#else(WIN32)
+#     list(APPEND CLConform_LIBRARIES  GL glut GLEW GLU)
+#endif(WIN32)
+#else()
+build_with_stub(${MODULE_NAME})
+#endif()
```
bashbaug: Are these lines intended to remain commented out?  I'm guessing this was a temporary change.

test_conformance/gles/CMakeLists.txt
```diff
@@ -15,14 +15,20 @@ set (${MODULE_NAME}_SOURCES
         ../../test_common/gles/helpers.cpp
     )
 
-list(APPEND CLConform_LIBRARIES EGL GLESv2)
+#if(GLES_IS_SUPPORTED)
+#  list(APPEND CLConform_LIBRARIES EGL GLESv2)
+#else()
+  build_with_stub(${MODULE_NAME})
+#endif()
```
bashbaug: Commented out lines here, too.

bashbaug (commented):
Are the stub files auto-generated somehow?  I'm a bit concerned that they'll be tricky to maintain.

On the plus side, it looks like compiling with other compilers on other platforms caught some legitimate issues, so maybe the maintenance cost is worth paying.

> they will gracefully skip execution at runtime
when the required interop extensions are not supported by the driver

What happens if the tests are stubbed and the required interop extensions **are** supported by the driver?

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
rjodinchr: Done

test_conformance/gl/CMakeLists.txt
rjodinchr: yes

rjodinchr:
> Are the stub files auto-generated somehow? I'm a bit concerned that they'll be tricky to maintain.

The stubs are only used to compile, I would not expect any trickyness maintaining them. When we update the tests, if something new is needed, the CI GitHub CI will catch it, and we just need to add a new stub function basically.

> On the plus side, it looks like compiling with other compilers on other platforms caught some legitimate issues, so maybe the maintenance cost is worth paying.
> 
> > they will gracefully skip execution at runtime
> > when the required interop extensions are not supported by the driver
> 
> What happens if the tests are stubbed and the required interop extensions **are** supported by the driver?

If the driver supports all the extensions needed by a test, I would expect the test to be compiled without the stubs and run to get conformance.
You can expose extensions without testing them and expect to be OpenCL compliant, right?

rjodinchr:
I don't really have the time to figure out `gl`, `gles`, `vulkan`, `d3d10` & `d3d11` right now. I've implemented it for the `extensions`. I've kept the full stubbing for the other.

rjodinchr:
I've updated the PR description.
This is now ready for review.
Thank you

test_conformance/d3d10/CMakeLists.txt
```diff
@@ -1,18 +1,17 @@
-if(WIN32)
 set(MODULE_NAME D3D10)
 
-set(${MODULE_NAME}_SOURCES
+set(${MODULE_NAME}_SOURCES main.cpp)
+
+if(WIN32 AND D3D10_IS_SUPPORTED)
+  list(APPEND ${MODULE_NAME}_SOURCES
```
shajder, line 6: The PR's stated mechanism for d3d10 is correct: `main.cpp` compiles unconditionally, `buffer.cpp/texture2d.cpp/texture3d.cpp/misc.cpp/harness.cpp` compile only when `WIN32 AND D3D10_IS_SUPPORTED`, and every REGISTER_TEST is guarded with `#ifdef D3D10_IS_SUPPORTED` / `#else return TEST_SKIPPED_ITSELF`, with the device init callback (`InitD3D10Device`) using `REQUIRE_EXTENSION("cl_khr_d3d10_sharing")` followed by `#else return TEST_FAIL` to catch a real build/device mismatch.

That mismatch detection **works as intended**. On a machine where `D3D10_IS_SUPPORTED` is left at its default OFF, but the OpenCL driver does advertise `cl_khr_d3d10_sharing`:

> Test failed while initialization
FAILED 6 of 6 tests.

test_conformance/d3d11/CMakeLists.txt
```diff
@@ -1,18 +1,17 @@
-if(WIN32)
 set(MODULE_NAME D3D11)
 
-set(${MODULE_NAME}_SOURCES
+set(${MODULE_NAME}_SOURCES main.cpp)
+
+if(WIN32 AND D3D11_IS_SUPPORTED)
+  list(APPEND ${MODULE_NAME}_SOURCES
```
shajder, line 6: d3d11 has the identical `CMakeLists.txt` structure (`main.cpp` always, `harness.cpp/buffer.cpp`/... only under `WIN32 AND D3D11_IS_SUPPORTED`), but `test_conformance/d3d11/main.cpp` was not updated to match. It is still entirely the pre-PR file, unconditionally including "harness.h" and calling `HarnessD3D11_Initialize/HarnessD3D11_TestBegin/TestDeviceBuffer/TestDeviceTexture2D/...` under `#if defined(_WIN32)`, with no `#ifdef D3D11_IS_SUPPORTED` anywhere. Since `D3D11_IS_SUPPORTED` defaults to OFF, `harness.cpp` is never compiled, and the module fails to link on any default Windows build:

> main.obj : error LNK2019: unresolved external symbol "void __cdecl HarnessD3D11_Initialize(...)"
main.obj : error LNK2019: unresolved external symbol "void __cdecl HarnessD3D11_TestBegin(...)"
main.obj : error LNK2019: unresolved external symbol "void __cdecl TestDeviceBuffer(...)"
main.obj : error LNK2019: unresolved external symbol "void __cdecl TestDeviceTexture2D(...)"
...


I presume, `d3d11/main.cpp` needs the same restructuring `d3d10/main.cpp` already received: replace the top-level `#if defined(_WIN32)` split with per-symbol `#ifdef D3D11_IS_SUPPORTED` guards mirroring d3d10's pattern, so a Windows build with `D3D11_IS_SUPPORTED` off degrades to a `REQUIRE_EXTENSION`-driven `SKIP/FAIL` instead of a link error. The non-Windows fallback does not need any extension query of its own.

test_conformance/gl/CMakeLists.txt
```diff
@@ -22,18 +10,33 @@ set (${MODULE_NAME}_SOURCES
...
+    list(APPEND CLConform_LIBRARIES  GL glut GLEW GLU)
+endif(WIN32)
+
+list (APPEND ${MODULE_NAME}_SOURCES
+    test_images_getinfo_common.cpp
+    test_images_read_common.cpp
+    test_images_write_common.cpp
+    ../../test_common/gl/helpers.cpp
+)
 if (WIN32)
```
shajder, line 40: with `GL_IS_SUPPORTED` set to off (default), `test_gl.exe` returns 0 (pass) without querying any device for `cl_khr_gl_sharing`. This may look like the same "opted out, API not applicable" case as d3d10 on non-Windows, but it is not: GL exists on every platform this suite targets, and d3d10 itself implements exactly this runtime check via `REQUIRE_EXTENSION` + `TEST_FAIL` when `D3D10_IS_SUPPORTED` is off. `test_buffers.cpp, test_fence_sync.cpp, test_image_methods.cpp` already carry the same per-function guard pattern as d3d10, but `main.cpp` never reaches them. It looks a bit like an unfinished migration rather than an intended difference, is that correct?

test_conformance/gl/main.cpp
```diff
@@ -170,6 +174,24 @@ int main(int argc, const char *argv[])
         return -1;
     }
 
+    if (gListTests)
+    {
+        log_info("Available 2.x tests:\n");
```
shajder, line 179: This branch runs before the `#ifdef GL_IS_SUPPORTED` check further down in main(), so `--list` reports all 24 tests even when GL_IS_SUPPORTED is off (confirmed on Windows). The rest of the function does not run in that case: main() returns 0 without querying the device for `cl_khr_gl_sharing`, so the per-function guards already in `test_buffers.cpp, test_fence_sync.cpp` etc. never get tested.

Could we move device iteration and the `REQUIRE_EXTENSION` checks out of that `#ifdef` block, same as `d3d10/main.cpp`, so those guards actually run?

test_conformance/vulkan/CMakeLists.txt
```diff
@@ -28,10 +16,29 @@ set (${MODULE_NAME}_SOURCES
...
+
+list(APPEND CLConform_INCLUDE_DIR ${VULKAN_INCLUDE_DIR})
+
+include_directories (${CLConform_INCLUDE_DIR})
+
 include_directories("../common/vulkan_wrapper")
 
 add_subdirectory(shaders)
+add_compile_definitions(VULKAN_IS_SUPPORTED)
+
```
shajder, line 37: `external_semaphores_import_export` reports PASS while skipping all its actual assertions, on real hardware with the extension genuinely supported. (NVIDIA RTX 3070 laptop, `cl_khr_external_semaphore` supported, confirmed by `external_semaphores_simple_1` passing for real in the same run):

> external_semaphores_import_export...
Could not find a handle type that supports both import and export.
Could not find a handle type that supports both import and export.
external_semaphores_import_export passed


Shouldn't we have `external_semaphores_import_export` propagate `SKIP` like cross_context does, instead of defaulting total_status to `PASS`?

shajder (commented):
Disclaimer before the comments below: **I don't have a clear picture of the overall intent behind this PR** (why every module needs to always build, what downstream consumes opencl_conformance_tests_full.csv or the CSV check, what "conformance" is supposed to mean for a stub/guarded binary). 

What follows are inconsistencies and deviations from what I'd naively expect, reproduced where I could. **I'm not confident any of them are actual problems** given the intended use case, just flagging what looked surprising from the outside!

test_conformance/d3d10/CMakeLists.txt
rjodinchr: It feels like you are saying that everything works as intended. If that's the case, please resolve the conversation. Otherwise could you please clarify what the issue is?

test_conformance/vulkan/CMakeLists.txt
rjodinchr: I don't understand.
First of all, that comment in on `vulkan/CMakeLists.txt` but it seems to be about `extensions/cl_khr_external_semaphpores`, can you confirm?

Then it is not clear to me how that PR is impacting the behavior of that test. I think this is the current test behavior, please correct me if that is not the case.

test_conformance/d3d11/CMakeLists.txt
rjodinchr: I agree, I will keep it like that for the time being. I would like https://github.com/KhronosGroup/OpenCL-CTS/pull/2724 to get merged before wasting time on that test.

test_conformance/gl/CMakeLists.txt
rjodinchr: Since `gl` isn't utilizing the harness, I opted for a straightforward approach by ensuring the test remains executable even when `GL_IS_SUPPORTED` is disabled. In that scenario, the test performs no checks, which is acceptable as it avoids introducing any regressions.
But for sure, that test is not bringing the improvement other do. That would be for a future PR once `gl` uses the harness.

test_conformance/gl/main.cpp
rjodinchr: As long as the harness remains unused, I'd prefer to minimize changes to `gl`. It's perfectly fine for the test to report every subtest without actually executing any when `GL_IS_SUPPORTED` is disabled.

rjodinchr:
> Disclaimer before the comments below: **I don't have a clear picture of the overall intent behind this PR** (why every module needs to always build, what downstream consumes opencl_conformance_tests_full.csv or the CSV check, what "conformance" is supposed to mean for a stub/guarded binary).

Please read https://github.com/KhronosGroup/OpenCL-CTS/issues/2750 to get a clearer picture of the intent behind that PR.

test_conformance/api/test_zero_sized_enqueue.cpp
```diff
@@ -210,36 +211,27 @@ int test_zero_sized_enqueue_helper(cl_device_id device, cl_context context,
 
 REGISTER_TEST_VERSION(zero_sized_enqueue, Version(2, 1))
 {
+    REQUIRE_QUEUE_PROPERTIES(CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE);
+
     int res =
         test_zero_sized_enqueue_helper(device, context, queue, num_elements);
     if (res != 0)
```
bashbaug, line 216: I don't think this is a functionally equivalent change.  Previously, we would run this test with the default (in-order) queue unconditionally, and then conditionally run the test again on an out-of-order queue when it is supported.  Now, we will skip the entire test when out-of-order queues are unsupported.

I don't have a strong opinion how we fix this, but shouldn't be reducing test coverage when out-of-order queues are unsupported.

test_conformance/extensions/cl_khr_dx9_media_sharing/test_functions_api.cpp
```diff
@@ -673,9 +675,11 @@ int api_functions(cl_device_id deviceID, cl_context context,
 
     return result.Result();
 }
+#endif
 
 REGISTER_TEST(api)
 {
+#ifdef _WIN32
```
bashbaug, line 677: Minor: There's already an `#if defined(_WIN32)` in this function.  Can we just use it?  If not, should we remove it?  Same goes for a bunch of the other tests in this suite.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
```diff
@@ -195,15 +202,17 @@ REGISTER_TEST_VERSION(external_semaphores_queries, Version(1, 2))
     }
 
     return TEST_PASS;
+#else
+    return TEST_FAIL;
```
bashbaug, line 206: This might be a bit too strong, in theory at least, since it requires a implementation that supports `cl_khr_external_semaphore` to build with Vulkan or the test will fail.  This will be the case even if, say, there is no Vulkan implementation to interop with.

Should this be considered a test skip instead?  Or, is there some other criteria we could check for that would reduce the risk of "false failures"?  Or, do we think this won't be an issue in practice, and all implementations that support `cl_khr_external_semaphore` really should be required to build with Vulkan?

bashbaug (commented):
Thanks!  I suspect there are still some gremlins lurking here, but this looks like a nice net improvement.

Please check the first comment.  The others are nice-to-haves, not must-haves.

test_conformance/api/test_zero_sized_enqueue.cpp
rjodinchr: > I don't think this is a functionally equivalent change. Previously, we would run this test with the default (in-order) queue unconditionally, and then conditionally run the test again on an out-of-order queue when it is supported. Now, we will skip the entire test when out-of-order queues are unsupported.

While I agree with the global idea. I think this statement is partially wrong. Right now we do not run conditionally when an out-of-order queue is supported. We run it even if out-of-order queue is not supported.

I will adjust the test to run with the OOO queue only if it is supported.

test_conformance/extensions/cl_khr_dx9_media_sharing/test_functions_api.cpp
rjodinchr: Yes, I must have been in automatic mode when doing that part. I have cleaned it up.

test_conformance/extensions/cl_khr_external_semaphore/test_external_semaphore.cpp
rjodinchr: I have not thought about that.
So we could have implementations supporting `cl_khr_external_semaphore` that aren't used with Vulkan. I guess that's possible.
To confirm this, I think a good experiment would be to keep it as `TEST_FAIL` for now and see if anyone is bothered by it. I can add a comment next to it so that anyone hitting it understands they need to speak up. What do you think?

rjodinchr: My biggest concern would be for an implementation to expose `cl_khr_external_semaphore` and have nothing testing it in the CTS. If it is not to be used with Vulkan, then does it mean we are missing a test for that something else interacting with `cl_khr_external_semaphore`?

bashbaug: Yes, exactly.  Some examples would be: an implementation that supports external semaphores to interop with DirectX, or even an implementation that supports external semaphores to work across OpenCL contexts with no other API involved at all.

We have coverage for at least some of these scenarios, see for example `external_semaphores_import_export` and `external_semaphores_cross_context`.

bashbaug: I'm fine with this code the way it is for now.  We can always re-address later if this turns out to be a real problem.  There are other issues with these semaphore tests (and external memory tests) that should also be addressed...

test_conformance/vulkan/CMakeLists.txt
shajder: > First of all, that comment in on vulkan/CMakeLists.txt but it seems to be about extensions/cl_khr_external_semaphpores, can you confirm?

Sorry, wrong anchor point.

> Then it is not clear to me how that PR is impacting the behavior of that test. I think this is the current test behavior, please correct me if that is not the case.

What the PR changes is exposure, right? The module used to build only under `if(VULKAN_IS_SUPPORTED)`, so it never ran without Vulkan. Now it always builds and runs, so this false PASS shows up in every environment lacking `cl_khr_external_semaphore` instead of the test simply not existing there.

test_conformance/d3d10/CMakeLists.txt
shajder: That comment was only context, the actual issue is on d3d11/main.cpp.

test_conformance/d3d11/CMakeLists.txt
shajder: Sounds good to me

test_conformance/gl/CMakeLists.txt
shajder: Clear, thanks

test_conformance/gl/main.cpp
shajder: Sounds consistent, thanks

test_conformance/vulkan/CMakeLists.txt
rjodinchr: Correct, you can find more information about the intent of that PR in #2750.

shajder:
> > Please read #2750 to get a clearer picture of the intent behind that PR.

> Build-Time vs. Runtime Gating: Many extension tests are gated behind CMake flags and are not built by default. This makes it difficult for implementers to know what needs to be run to claim full conformance, and complicates the build process.

Isn't a CMake option itself already a clear signal that a given piece is optional? An implementer who supports `cl_khr_gl_sharing` knows that already, and turns `GL_IS_SUPPORTED` on as part of their submission, same as they already select which binaries to run via the CSV today. #2750's point 3 ("implementer doesn't know what to run") seems to assume someone testing functionality they don't support in the first place, which they would not run either way regardless of how the option is exposed.

One more concern, isn't there also a case where an implementer's device reports an extension they don't want included in this submission (in-progress driver work, a shared CI machine, etc)? Previously the CMake flag gave them full control over that regardless of what the device reports. Now that control seems to shift entirely to `is_extension_supported`, is that right?

rjodinchr:
> Isn't a CMake option itself already a clear signal that a given piece is optional? An implementer who supports `cl_khr_gl_sharing` knows that already, and turns `GL_IS_SUPPORTED` on as part of their submission, same as they already select which binaries to run via the CSV today. 

That's what I want to avoid. I don't want to have to know that if the driver supports `cl_khr_gl_sharing`, I need to build with a specific CMake flag. The same goes for knowing which binaries to run.

I think the OpenCL-CTS would be in a much better place if one didn't have to know those kinds of things, and everyone could just run everything. That way, we (the OpenCL-CTS contributors) could ensure that the right things are being tested.

> #2750's point 3 ("implementer doesn't know what to run") seems to assume someone testing functionality they don't support in the first place, which they would not run either way regardless of how the option is exposed.

It could be a case of not knowing, but it could also be a case of forgetting, having a CI bug, etc. Adding automatic protection against this is an improvement, in my opinion.

> One more concern, isn't there also a case where an implementer's device reports an extension they don't want included in this submission (in-progress driver work, a shared CI machine, etc)? Previously the CMake flag gave them full control over that regardless of what the device reports. Now that control seems to shift entirely to `is_extension_supported`, is that right?

I think this is wrong, a compliant driver should not expose an extension that has not been tested by the CTS.

If one wants to enable an extension locally to test things, that's fine. They will just get more feedback from the CTS about their support of that extension.

test_conformance/check_binaries_in_csv.py
```diff
@@ -0,0 +1,48 @@
...
+
+        for field in fields:
+            parts = field.split()
+            if not parts:
+                continue
+            path = parts[0] # The first part should be the path
+            binary_name = os.path.basename(path)
+            csv_binaries.add(binary_name)
+
+    missing_binaries = [b for b in binaries if b not in csv_binaries]
```
joselopeqti, line 36: This script is not considering that the `CONFORMANCE_PREFIX` and `CONFORMANCE_SUFIX` could have been overridden by the CMakeVendor.txt file, and it expects the binary to start with `test_`.

ci/check_binaries_in_csv.py
```diff
@@ -0,0 +1,48 @@
...
+
+        for field in fields:
+            parts = field.split()
+            if not parts:
+                continue
+            path = parts[0] # The first part should be the path
+            binary_name = os.path.basename(path)
+            csv_binaries.add(binary_name)
+
+    missing_binaries = [b for b in binaries if b not in csv_binaries]
```
rjodinchr: That script is not expected to be run on individual machines, only in the CI, to make sure nothing is missed. Thus, it does not need to handle `CONFORMANCE_PREFIX` or `CONFORMANCE_SUFFIX`.

In the CI, all binaries start with `test_`, so this is expected.

rjodinchr: I will move that script in the `ci` directory to make it clearer.

rjodinchr: I've also removed it from the `ALL` cmake target, to have it run explicitly in the github CI.

PR #2752, opened 2026-07-15
subgroups: Fix maximum subgroup size for tail calculation

riccardo-zanetti-arm:
Use get_max_sub_group_size() when calculating the size of the final subgroup in a non-uniform workgroup.

get_sub_group_size() may already return the size of the current partial subgroup. Using it as the modulus divisor can therefore calculate an incorrect size, split the subgroup at the wrong lane, and pass non-uniform indices to sub_group_non_uniform_broadcast.


ahesham-arm (approved):

rjodinchr (approved):

bashbaug:
Discussed in the July 21st teleconference.  OK to merge offline with a Qualcomm review approval.

lakshmih (approved):

PR #2754, opened 2026-07-22
images: refactor into finer-grained pieces for parallel execution

rjodinchr:
Monolithic test suites can create significant work imbalances when executing
tests in parallel, as a single large test block can bottleneck the test runner.
This commit refactors the image conformance tests to allow efficient
parallel execution and better workload distribution.

To achieve this, the test harness is upgraded to support generic user data
payloads, and the test registration is flattened:
* Updated the test_function_pointer signature and test_registry::add_test
  to accept a trailing `void *args` parameter.
* Removed the object-oriented `test` class wrapper abstraction in favor of
  direct function pointer test definitions.
* Cleaned up redundant `test_registry::getInstance()` boilerplate from main
  entry points across all conformance test suites.

Ref #2723
  
[run-test: test_cl_copy_images small_images --num-worker-threads 2 ]


rjodinchr:
@ahesham-arm I believe those changes should be submitted together, but I've split it into many smaller commits to help the review.

test_common/harness/testHarness.h
```diff
@@ -80,13 +80,15 @@ Version get_device_cl_version(cl_device_id device);
 #define ARRAY_SIZE(arr) (sizeof(arr) / sizeof((arr)[0]))
 
 typedef int (*test_function_pointer)(cl_device_id deviceID, cl_context context,
-                                     cl_command_queue queue, int num_elements);
+                                     cl_command_queue queue, int num_elements,
+                                     void *arg);
```
shajder, line 84: Quick note: gles no longer builds after this change.

shajder (changes requested):

test_common/harness/testHarness.h
rjodinchr: Make sense, not good news that we do not even build it in the CI.
I'll fix it, thank you

shajder (dismissed):
LGTM

test_common/harness/testHarness.h
ahesham-arm: This is why I'd rather we focus on completing https://github.com/KhronosGroup/OpenCL-CTS/issues/2181 before making changes to the harness.

test_common/harness/imageHelpers.cpp
```diff
@@ -163,6 +163,40 @@ uint32_t get_channel_order_channel_count(cl_channel_order order)
             return 0;
     }
 }
+const char *cl_channel_type_to_string(cl_channel_type type)
```
ahesham-arm: nit: I'd rather make use of C++ overloads and just have these helpers named `to_string` instead, but that's a personal choice.

test_conformance/images/common.cpp
```diff
@@ -626,3 +635,26 @@ clMemWrapper create_image(cl_context context, cl_command_queue queue,
 
     return img;
 }
+
+void register_test_configs(const std::vector<struct image_type> &tests,
```
ahesham-arm: In C++ you don't need to prepend structs with the keyword `struct`.

test_common/harness/testHarness.cpp
```diff
@@ -233,16 +233,28 @@ static std::string removed_args_to_string(std::vector<std::string> &args)
...
+        limit = 12;
+    }
     for (const auto &name : names)
     {
         log_info("\t%s\n", name.c_str());
+        if (!limit--)
+        {
+            log_info("\t...\n");
+            log_info("\n");
+            log_info("List truncated; to view the full list use `--list`.\n");
```
ahesham-arm: I don't understand the point of this commit, why are we truncating?

rjodinchr: Following that change, these tests now expose a vast number of subtests, exceeding 200 for cl_copy_images and nearly 100 for the others.
This makes it difficult to locate the help message when running `--help`, as it requires significant scrolling.
Truncating the output provides a much cleaner experience, while the dedicated `--list` argument remains available for those who need to view every item.

ahesham-arm: Not sure how I feel about this. I don't like the fact that a test has a different help message behaviour compared to other binaries, just because it has a lot of test cases. Long output can always be piped through `more`, `less` or even `grep` if needed.

ahesham-arm: I think it's best to keep the changes focused on what they actually are about, i.e. changes to the image tests, and keep these type of wider changes separate because they are bound to attract comments (like this one :D)

rjodinchr: This is not specific to that test, it applies to all binaries.
If that's the only blocker, I'll remove it from that PR.
I do believe that in the case of the OpenCL-CTS binaries where the main part of the help message can fit on screen, it is not a great user experience to have to scroll to get to it.

test_common/harness/testHarness.h
rjodinchr: Prioritizing harness usage for all tests is essential, but it won't safeguard against build-related issues.
Issue #2751 aims to close that gap; clearly, for the gles case, we require both full harness integration and the updates from #2751 to reach a secure state.

test_conformance/images/clCopyImage/main.cpp
```diff
@@ -127,13 +80,164 @@ static test_status parseArgs(int &argc, const char *argv[],
...
+                    break;
+                case k1DBufferTo1D:
+                    test_configs.emplace_back(
+                        "1DBufferTo1D", CL_MEM_OBJECT_IMAGE1D_BUFFER,
+                        CL_MEM_READ_ONLY, CL_MEM_OBJECT_IMAGE1D,
+                        ctx.enablePitch ? CL_MEM_USE_HOST_PTR | CL_MEM_READ_ONLY
+                                        : CL_MEM_READ_ONLY,
+                        test_copy_image_set_1D_1D_buffer, channel_type);
+                    break;
+            }
```
shajder, line 227: Might be worth adding a `default` case with a warning, in case this enum gets extended without updating this switch.

test_conformance/images/clFillImage/main.cpp
```diff
@@ -16,31 +16,25 @@
...
-REGISTER_TEST(1Darray)
-{
-    return test_image_set(device, context, queue, k1DArray, ctx);
-}
-REGISTER_TEST(2Darray)
-{
-    return test_image_set(device, context, queue, k2DArray, ctx);
-}
-REGISTER_TEST(1Dbuffer)
+static std::vector<test_configs> test_configs;
```
shajder: The variable name shadows the type name here, unlike `clCopyImage/main.cpp`, which avoids this by naming the type `TestConfigs`. Might be worth applying the same naming to `test_configs/common.h` for consistency.

test_conformance/images/clFillImage/main.cpp
```diff
@@ -16,31 +16,25 @@
...
+static int doTest(cl_device_id device, cl_context context,
+                  cl_command_queue queue, int, void *args)
 {
-    return test_image_set(device, context, queue, k1DBuffer, ctx);
+    auto &test = test_configs[(uintptr_t)args];
```
shajder, line 34: Preference to guard against an out-of-range index, here and couple of similar locations below.

test_conformance/images/kernel_read_write/main.cpp
```diff
@@ -215,33 +214,21 @@ static int doTest(cl_device_id device, cl_context context,
...
 
     return ret;
 }
 
-REGISTER_TEST(1D)
-{
-    return doTest(device, context, queue, CL_MEM_OBJECT_IMAGE1D, ctx);
-}
-REGISTER_TEST(2D)
+std::vector<test_configs> test_configs;
```
shajder: Missed `static`?

shajder (commented):
A few minor suggestions

test_conformance/images/clCopyImage/main.cpp
rjodinchr: If a case is added, but not in the switch, I would expect a compiler to complain about it.
I remove one case locally to test:
```
test_conformance/images/clCopyImage/main.cpp:91:21: error: enumeration value 'k1DBufferTo1D' not handled in switch [-Werror,-Wswitch]
   91 |             switch (testMethod)
      |                     ^~~~~~~~~~
1 error generated.
```
Maybe we should make sure that at least 1 compiler in our CI would complain about that?

ahesham-arm: I'd prefer having `-Wswitch`, and not having a default case, because then it's a compile-time error instead of a runtime one.

shajder (dismissed):
LGTM from a technical standpoint: the PR does what it says in the description, and everything raised during review has been addressed. I can't say much about the necessity of this refactoring, but technically it looks solid.

rjodinchr:
> I can't say much about the necessity of this refactoring

This has a non-negligeable impact when I run the CTS on various devices (laptops, phones) using the harness multi-threading (through `--num-worker-threads`).
The performance gains can reach several hours on slower devices.

It is also better to monitor failing tests using the JSON output as it makes it more detailed.

test_common/harness/testHarness.cpp
ahesham-arm: I don't really agree with this approach so let's remove this part from this PR since it's not really the core of it and we would benefit from merging the actual improvements.

ahesham-arm (changes requested):

ahesham-arm (dismissed):

rjodinchr:
The `license/cla` check seems stuck, so I'm going to do a simple rebase to trigger it again.

rjodinchr:
Agree to merge offline once the conflicts are resolved as discussed during the August 11th teleconference.

test_common/harness/testHarness.cpp
```diff
@@ -354,6 +344,16 @@ int runTestHarnessWithCheckAndParse(int argc, const char *argv[], int testNum,
                  != TEST_PASS);
     }
 
+    int testNum = test_registry::getInstance().num_tests();
+    test_definition *testList = test_registry::getInstance().definitions();
```
bashbaug, line 348: Previously, we were explicitly passing the number of tests and the test definitions to the harness for execution.  Now, we are unconditionally getting this implicitly from the test registration framework.  I think that's probably OK, and this fits with our general trend to convert more tests to the new test registration framework, but IMHO it's still worth checking to confirm this is what we want to do.

Note that this change appears to be driving a lot of the changes in this PR.

test_conformance/images/clCopyImage/main.cpp
```diff
@@ -127,13 +81,164 @@ static test_status parseArgs(int &argc, const char *argv[],
...
+        }
+    }
+    for (uint32_t i = 0; i < test_configs.size(); i++)
+    {
+        test_registry::getInstance().add_test(
+            test_image_set, test_configs[i].name.c_str(), Version(1, 2),
+            (void *)(uintptr_t)i);
```
bashbaug, line 234: Ah, I kind of get it now. Since we are now generating test variations as part of this "parseArgs" function, we cannot pass the number of tests and the test definitions to the harness like we did previously, since it hasn't been generated yet.

I'm not sure if this is clever or a bit of a hack (or maybe both :smile:).  Should we separate the argument parsing from the test generation? Is this even possible?

test_common/harness/testHarness.cpp
rjodinchr: The previous (current) implementation feels like a hack to me. I think this proposal is much cleaner. It gives a proper interface to register.

test_conformance/images/clCopyImage/main.cpp
rjodinchr: > Ah, I kind of get it now. Since we are now generating test variations as part of this "parseArgs" function, we cannot pass the number of tests and the test definitions to the harness like we did previously, since it hasn't been generated yet.

Not exactly. I wanted to avoid making too much change, and since the `ctx.enablePitch` is used to set the flag for the config, I opted to register those test in `parseArgs`, but we could refactor that test more to register them in main. Leaving the possibility to pass the test definitions to the harness. But I personally don't feel that's good/better than this proposal.

> I'm not sure if this is clever or a bit of a hack (or maybe both 😄). Should we separate the argument parsing from the test generation? Is this even possible?

It might be possible, but I don't think that's necessary. At least not right now. We can revisit once we feel the need for it.

bashbaug:
Discussed in the September 8th teleconference. Aiming to merge next week (September 15th teleconference).

PR #2755, opened 2026-07-24
run_conformance: add --conformance-results-dir option

kwk:
Add an optional `--conformance-results-dir=<DIR>` argument that, when given, sets `CL_CONFORMANCE_RESULTS_FILENAME` to `<DIR>/<test>.json` before each test so test binaries write machine-readable JSON results there. The directory `<DIR>` is created if it does not exist. Without the option the existing behaviour is unchanged.

I created this option because I noticed that the script https://github.com/KhronosGroup/OpenCL-CTS/blob/main/ci/compare_results.py apparently needs a special `CL_CONFORMANCE_RESULTS_FILENAME` set when running tests which required quite a [manual work](https://github.com/KhronosGroup/OpenCL-CTS/blob/db8e9279b4a9203be9b635f797be237fa84980b4/.github/workflows/presubmit.yml#L321-L340). want to use the `run_conformance.py` script and still be able to get those result files so that we can verify our own conformance to the OpenCL specification downstream in Fedora and RHEL distributions.

### DIFF
--- a/test_conformance/run_conformance.py
+++ b/test_conformance/run_conformance.py
@@ -18,6 +18,7 @@ import time
 import tempfile
 
 DEBUG = 0
+conformance_results_dir = None
 
 log_file_name = "opencl_conformance_results_" + time.strftime("%Y-%m-%d_%H-%M", time.localtime()) + ".log"
 process_pid = 0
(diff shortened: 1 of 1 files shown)


kwk:
@franz @karolherbst @lakshmih can you please take a look at this?

kwk:
@franz @karolherbst @lakshmih can you please have a look at this PR, please?

PR #2756, opened 2026-07-27
geometrics: test that cross(float4, float4).w equals 0

svenvh:
The OpenCL C Specification explicitly calls out that "The w component of float4 result returned will be 0.0." but the CTS was not testing for this.

[run-test: test_geometrics geom_cross]

### DIFF
--- a/test_conformance/geometrics/test_geometrics.cpp
+++ b/test_conformance/geometrics/test_geometrics.cpp
@@ -238,20 +238,22 @@ REGISTER_TEST(geom_cross)
             // On an embedded device w/ round-to-zero, 3 ulps is the worst-case tolerance for cross product
             cross_product( inDataA + i * vecsize, inDataB + i * vecsize, testVector, errorTolerances, 3.f );
 
-        // RTZ devices accrue approximately double the amount of error per operation.  Allow for that.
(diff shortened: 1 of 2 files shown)


rjodinchr:
Could you add `[run-test: test_geometrics geom_cross]` to the PR description or commit message and update https://github.com/KhronosGroup/OpenCL-CTS/blob/main/ci/pocl/golden.json accordingly?

svenvh:
> Could you add `[run-test: test_geometrics geom_cross]` to the PR description or commit message and update https://github.com/KhronosGroup/OpenCL-CTS/blob/main/ci/pocl/golden.json accordingly?

Sure, done (well, assuming the test will pass).

test_conformance/geometrics/test_geometrics.cpp
rjodinchr, line 260: We test `outData[i * vecsize + 3]` against `0.0f`, but we print `testVector[3]`. I think we should use the same thing everywhere (maybe a const value equals to`0.0f`?)

svenvh: Thanks, I think we should compare against `testVector[3]` then, as `cross_product()` already sets the reference to `0.0f`.

rjodinchr (approved):

PR #2757, opened 2026-07-29
Command-buffer testing for recoding with invalid queue props

EwanC:
Adds CTS coverage for spec change https://github.com/KhronosGroup/OpenCL-Docs/pull/1588

## Negative Test
The error condition for
`CL_DEVICE_COMMAND_BUFFER_REQUIRED_QUEUE_PROPERTIES_KHR` is [already tested](https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/extensions/cl_khr_command_buffer/negative_command_buffer_enqueue.cpp#L351) for both a valid and nullptr queue parameter in the `negative_enqueue_queue_without_reqd_properties` test, so that doesn't need extra coverage.

```cpp
        error = clEnqueueCommandBufferKHR(0, nullptr, command_buffer, 0,
                                          nullptr, nullptr);
        test_failure_error_ret(error, CL_INCOMPATIBLE_COMMAND_QUEUE_KHR,
                               "clEnqueueCommandBufferKHR should return "
                               "CL_INCOMPATIBLE_COMMAND_QUEUE_KHR",
                               TEST_FAIL);

        error = clEnqueueCommandBufferKHR(1, &queue, command_buffer, 0, nullptr,
                                          nullptr);
        test_failure_error_ret(error, CL_INCOMPATIBLE_COMMAND_QUEUE_KHR,
                               "clEnqueueCommandBufferKHR should return "
                               "CL_INCOMPATIBLE_COMMAND_QUEUE_KHR",
                               TEST_FAIL);
```

However, the existing `negative_enqueue_with_unsupported_queue_property` test only verifies with a non-null queue parameter for the `CL_DEVICE_COMMAND_BUFFER_SUPPORTED_QUEUE_PROPERTIES_KHR` check. So this PR extends that test case to also verify the correct error is reported when the queue parameter is nullptr.

## Positive Tests

Introduces explicit testing for 2 cases about using invalid command-buffer enqueue properties during command-buffer recording:
1) Recording into a command-buffer created from a command-queue with missing required properties,
   and then executing on a substitute queue with required properties.
2) Recording into a command-buffer created from a command-queue with unsupported properties,
   and then executing on a substitute queue with only required properties.

**Note:** I haven't got a device with non-zero required properties, or any unsupported queue properties, so review from a vendor that doesn't have these restrictions would be helpful.

### DIFF
--- a/test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
+++ b/test_conformance/extensions/cl_khr_command_buffer/command_buffer_queue_substitution.cpp
@@ -293,6 +293,237 @@ struct QueueOrderTest : public BasicCommandBufferTest
     const cl_int overwritten_pattern = 0xACDC;
     const cl_int pattern_pri = 42;
 };
+
+// Command-queue substitution test which handles the case where a command-buffer
+// is recorded on a command-queue with an unsupported property, and then a
+// substituted queue with required properties used for execution.
+struct RecordUnsupportedPropTest : public BasicCommandBufferTest
+{
+    using BasicCommandBufferTest::BasicCommandBufferTest;
+
+    RecordUnsupportedPropTest(cl_device_id device, cl_context context,
+                              cl_command_queue queue)
+        : BasicCommandBufferTest(device, context, queue), recording_prop(),
+          recording_queue(nullptr)
+    {}
+
+    // Record using sync points in case the command-buffer is using an
+    // out-of-order queue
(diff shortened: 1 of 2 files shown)


ahesham-arm (approved):

ahesham-arm:
LGTM, thank you for adding the tests.

bashbaug:
Merging as discussed in the August 11th teleconference.

PR #2758, opened 2026-07-29
half: Improve tests using special values migrated to harness

rjodinchr:
This PR improves the half-float storage conformance tests by leveraging targeted special values (edge cases) and centralizing the dataset generation in the test harness:

*   **Centralize Special Values**: Migrated the floating-point and integer special value datasets (`specialValuesFloat`, `specialValuesDouble`) and their generator helper functions from the conversions conformance suite into the centralized test harness (`test_common/harness`). This makes these edge-case datasets globally accessible for other conformance tests.
*   **Thread-safe Initialization**: Refactored the initialization of the special value datasets to use C++11 thread-safe static local initialization. This removes the need for explicit locking via `std::recursive_mutex` and associated includes.
*   **Enhance Half-Float Storage Tests**: Updated the reference loops in `Test_vStoreHalf.cpp` (`vStoreHalf` and `vStoreaHalf` tests) to use the migrated special values. Instead of using a slow sequential bit-pattern loop, the tests now seed blocks with targeted edge cases (NaNs, infinities, subnormals) before padding with random data.

Ref #2723


test_common/harness/conversions.h
```diff
@@ -134,4 +138,251 @@ static inline int IsHalfSubnormal(cl_half x)
     return ((x & 0x7fffU) - 1U) < 0x03ffU;
 }
 
+extern const std::vector<float> specialValuesFloat;
+extern const std::vector<double> specialValuesDouble;
+
+template <typename InType, typename OutType> OutType bitcast(InType in)
+{
+    OutType out;
+    assert(sizeof(InType) == sizeof(OutType));
```
shajder: Preference for `static_assert` here, since it's evaluated at compile time regardless of whether `NDEBUG` is defined.

test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -844,14 +869,9 @@ int Test_vStoreHalf_private(cl_device_id device, f2h referenceFunc,
...
-    uint64_t lastCase = 1ULL << (8 * sizeof(float)); // number of floats.
-    size_t stride = blockCount;
-
-    if (gWimpyMode)
-        stride = (uint64_t)blockCount * (uint64_t)gWimpyReductionFactor;
+    uint64_t lastCase = 1ULL << NB_INPUT_POWER_OF_TWO;
 
-    // we handle 64-bit types a bit differently.
-    if (lastCase == 0) lastCase = 0x100000000ULL;
+    if (gWimpyMode) lastCase /= gWimpyReductionFactor;
```
shajder: The baseline coverage itself dropped from full 32-bit exhaustive (2^32) to a fixed budget of 2^26, regardless of wimpy mode, this affects not just wimpy, right? On top of that, wimpy mode further divides this already-reduced budget by gWimpyReductionFactor. Worth confirming this two-tier reduction is intentional and acceptable, may be worth flagging for WG discussion.

test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -81,16 +86,26 @@ static cl_int ReferenceF(cl_uint jid, cl_uint tid, void *userInfo)
     cl_ushort *r = cri->r + off;
     f2h f = cri->f;
     cl_ulong i = cri->i + off;
-    cl_uint j;
+    const auto &specialValues =
+        GetFpSpecialValues<float, uint32_t, cl_half, true>();
+    MTdata m = init_genrand(jid);
```
shajder, line 105: Correct me if I am mistaken, but it looks like every block sets `jid` to the same sequence of values (`0..threadCount-1`), and since `init_genrand` is deterministic, this reseeds the RNG to the exact same state every block, so the "random" padding produces the identical sequence of values in every block instead of new ones. If that's correct, maybe something like `init_genrand(jid ^ (cl_uint)i)` would fix it?

shajder (commented):
In general LGTM, few minor comments to discuss below.

test_conformance/half/Test_vStoreHalf.cpp
rjodinchr: It is intentional. Having a reduced coverage for the default mode is what I'm looking for.

test_conformance/half/Test_vStoreHalf.cpp
rjodinchr: Good catch, thank you!
I changed it for `init_genrand((cl_uint)i)` as `i` already takes `jid` into account through `off`.

test_conformance/half/Test_vStoreHalf.cpp
rjodinchr: I've added the exhaustive mode as for conversion test.

shajder (dismissed):

bashbaug (dismissed):
I've verified that the test time is reduced by a factor of ~64 due to the fixed budget of 2^26 values vs. exhaustive testing of all 2^32 values.  I've also verified that all values are tested, same as before, when the exhaustive `-a` flag is passed.

I'm not sure how valuable "wimpy" mode is anymore, given that the entire test suite runs in less than one minute by default on my system, and of that one minute, most of the time is dominated by kernel compilation.  That's a pretty good "problem" to have, though.

Nice work!

test_conformance/half/Test_vStoreHalf.cpp
```diff
@@ -1691,11 +1740,9 @@ int Test_vStoreaHalf_private(cl_device_id device, f2h referenceFunc,
     // Figure out how many elements are in a work block
     size_t elementSize = std::max(sizeof(cl_ushort), sizeof(float));
     size_t blockCount = BUFFER_SIZE / elementSize;
-    uint64_t lastCase = 1ULL << (8 * sizeof(float));
-    size_t stride = blockCount;
+    uint64_t lastCase = 1ULL << NB_INPUT_POWER_OF_TWO;
```
svenvh, line 1743: Should we be checking `gTestAll` here?

test_conformance/half/Test_vStoreHalf.cpp
svenvh: This needs fixing for the `double` case below too.

test_conformance/half/Test_vStoreHalf.cpp
rjodinchr: Indeed, I'll fix that, thank you 🙏

bashbaug: I'm a bit annoyed that I didn't discover this when I tried to verify that all values are tested with the exhaustive `-a` flag.

Regardless, I'm glad it's fixed, and I did (re-)verify that all values are tested with the `-a` flag now (tested with `vstorea_half` specifically).

bashbaug (approved):

svenvh (approved):

bashbaug:

PR #2759, opened 2026-07-29
Fix compilation error in test_conversions

ahesham-arm:
Change `sign` to `constexpr`, allowing its use without capturing across all compilers.


svenvh (approved):

svenvh (changes requested):
Actually no as figured out by the CI. :)

rjodinchr (approved):

rjodinchr (changes requested):
In fact, windows is still complaining

svenvh (approved):

rjodinchr (approved):

rjodinchr:
This is a simple compilation error fix, I think it is fine merging offline

PR #2760, opened 2026-07-30
printf: test successful return value

svenvh:
The printf suite never tested the printf() return value.  Add coverage for a successful return value of 0.

The failing return value of -1 cannot be tested, because the specification defines no portable way to make printf fail.

[run-test: test_printf return_value]


rjodinchr (dismissed):

bashbaug (dismissed):

bashbaug:
Discussed in the August 11th teleconference.  OK to merge.

rjodinchr (approved):

bashbaug (approved):

PR #2763, opened 2026-08-04
Remove MAX_DEVICE_MEMORY_SIZE_DIVISOR from min_max_mem_alloc test while quering maxAllocSize

joselopeqti:
Removed `MAX_DEVICE_MEMORY_SIZE_DIVISOR` from `get_device_info_max_mem_alloc_size()` call so the test uses the actual reported max allocation size instead of a divided value.
This fixes an issue with devices with low memory as the test expects to allocate at least 128mb.

### DIFF
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -573,8 +573,7 @@ REGISTER_TEST(min_max_mem_alloc_size)
 
     /* Get the max mem alloc size, limit the alloc to half of the available
      * memory */
-    maxAllocSize = get_device_info_max_mem_alloc_size(
-        device, MAX_DEVICE_MEMORY_SIZE_DIVISOR);
+    maxAllocSize = get_device_info_max_mem_alloc_size(device);
     memSize =
         get_device_info_global_mem_size(device, MAX_DEVICE_MEMORY_SIZE_DIVISOR);
 


rjodinchr:
I think we talked about `MAX_DEVICE_MEMORY_SIZE_DIVISOR` during the teleconference a while ago. What I remember from it is that `CL_DEVICE_MAX_MEM_ALLOC_SIZE` is not the maximum size that one can allocate. It is the size above which you should not expect to be able to allocate. Failing to allocate `CL_DEVICE_MAX_MEM_ALLOC_SIZE -1` is compliant with the OpenCL specification if I remember correctly.

Thus removing `MAX_DEVICE_MEMORY_SIZE_DIVISOR` does not seem a proper fix. Instead we could make sure that we report at least what OpenCL expects as its minimum?

joselopeqti:
Is there any reason for the `requiredAllocSize` of 128mb used by the test?

rjodinchr:
> Is there any reason for the `requiredAllocSize` of 128mb used by the test?

I think `128mb` comes from `CL_DEVICE_GLOBAL_MEM_SIZE` for the full profile.

I agree that the current implementation of this test is weird (wrong?).

We should at least apply the same formula as we find in the specification:

> Max size of memory object allocation in bytes. The
minimum value is max(min(1024 × 1024 × 1024, 1/4th of
CL_DEVICE_GLOBAL_MEM_SIZE), 32 × 1024 × 1024).

But for the real allocation (using `clCreateBuffer`) in the test, we should keep `MAX_DEVICE_MEMORY_SIZE_DIVISOR`.

PR #2764, opened 2026-08-04
relationals: print any/all function being tested

svenvh:
The function name is already available in `fnName`, so print it to make the output log clearer.


rjodinchr (approved):
The output from pocl run is looking good!

PR #2761, opened 2026-07-31
add SPIR-V tests for various access chain instructions

bashbaug:
This PR adds targeted SPIR-V tests for various access chain instructions, including access chain instructions without any indices.  See: https://github.com/KhronosGroup/SPIRV-Registry/issues/425

Test variants include:

* Access chains into arrays (OpTypeArray) and vectors (OpTypeVector).
* Access chains via OpAccessChain, OpInBoundsAccessChain, OpPtrAccessChain, and OpInBoundsPtrAccessChain.

[run-test: test_spirv_new access_chain* ptr_access_chain*]

PR #2767, opened 2026-08-06
add SPIR-V support to the automated PoCL builds

bashbaug:
[run-test: test_spirv_new basic_versions]


bashbaug:
This seems to _mostly_ be doing what I want it to do, but there's a problem right now - because the PoCL cache key is just `ubuntu-24.04-pocl-v7.1`, we're hitting the cache, and therefore we aren't rebuilding a PoCL version with SPIR-V support.

I think I'd like to include the SPIR-V LLVM translator SHA into the cache key, which would cause it to get rebuilt.  Does that make sense?  We could also include just the SPIR-V LLVM translator branch in the cache key, though then we won't pick up the latest SPIR-V LLVM translator changes - this is both good and bad.

Regardless, having the PoCL dependencies in a separate file is a bit awkward right now.  @rjodinchr , how would you feel about removing `actions.yml` and moving this stuff into the main `presubmit.yml`?  This will make it easier to keep things like the right LLVM version and the SPIR-V LLVM translator release branch in sync, and easier to get the SHA for the cache key.  I'm open to other suggestions as well - maybe we could move all of the PoCL stuff to its own file?

rjodinchr:
The dependencies for pocl are kept in separate files because they are referenced in two locations within `presubmit.yml`, primarily to prevent code duplication. I am open to removing the file, provided we can still avoid duplicating that code.

It seems beneficial for the pocl cache to include those SPIRV-LLVM-Translator libraries to prevent the need for re-cloning or rebuilding them repeatedly.

Looking ahead, I would eventually like to incorporate other implementations, such as clvk over lavapipe. At that stage, the yaml will need increased flexibility to easily accommodate multiple backends, each with its own dedicated cache.

rjodinchr:
Also, pocl released 7.2, I would like to update to that version which should fix some tests, maybe that would be a good time to include support for SPIR-V?

bashbaug:
Cool, I think I got this working now.  I kept the previous install-pocl-deps step, so there is no new redundancy, and instead I just pass the LLVM version as an input so it should stay in sync with the SPIR-V LLVM translator version.

I did end up separating the PoCL testing into its own workflow file (vs. including it in the presubmit workflow file), but I'm flexible about this if we'd prefer to keep it in the presubmit file for whatever reason.

A few other notes worth mentioning:

1. I'm "installing" the CTS test binaries now vs. relying on CMAKE_RUNTIME_OUTPUT_DIRECTORY.  Without the install target, it is much harder to find the assembled SPIR-V files.
2. I'm including the Git SHA for the SPIR-V LLVM translator in the PoCL cache key, as I described previously, so SPIR-V LLVM translator changes should be picked up automatically.  If we decide that this is too much thrash, we could switch to SPIR-V LLVM translator tags instead.
3. I couldn't quite figure out how and when the testing should be done.  Previously, it definitely ran on pull requests.  It looks like there were some provisions to run on pushes also, but I couldn't quite follow this.  It should parse for testing triggers on both pushes and pull requests now, but there's a definite possibility I've messed this up, or I've misunderstood how this should work.
4. I did a bit of modernization to the pocl workflow file that should be functionally equivalent.  If we're happy with this, I can make similar changes to presubmit also, or I suppose I can revert these changes (but I would prefer not to).
5. The "basic_versions" SPIR-V test is passing.  If we're feeling daring, I can run all of the "spirv_new" tests, also.
6. It should be a tiny change to move to PoCL v7.2 (vs. v7.1), but I'd prefer to do that in a separate PR.

rjodinchr:
Looking good overall, but we need to revert the run conditions for `check-triggers` and `build-pocl`:

The cache needs to be created on the main branch to be accessible from all pull requests. Otherwise when a new pull request is created, it has to build its pocl cache the first time which is not ideal.
Thus `build-pocl` needs to keep the `always()` mark with the condition.

On the main branch (push) we don't want to run anything. There is no commit list we can consider, and just looking at the last commit does not bring much value I believe. `check-triggers` just needs to run on pull requests.

bashbaug:
> we need to revert the run conditions for `check-triggers` and `build-pocl`

OK, no problem, like I said I wasn't entirely sure how this was supposed to work.

I'd like to add some clarifications to the README, then, to describe the expected workflow.  We currently say:

https://github.com/KhronosGroup/OpenCL-CTS#how-to-trigger-tests

> Testing is triggered by adding a special tag to either your **Pull Request description** or in any of your **commit messages**.
> 
> The CI parses the text for the following syntax: `[run-test: <command>]`
> 
> Multiples tags for a single Pull Request is supported.

Some questions:

* Is the part about "in any of your commit messages" incorrect?
* Is there a way to trigger tests prior to creating a PR?
* Is there a way to trigger additional tests after a PR is created?

rjodinchr:
> Is the part about "in any of your commit messages" incorrect?

The actual implementation parses the commit message from all the commits that the PR is trying to merge in the targeted branch.

> Is there a way to trigger tests prior to creating a PR?
> Is there a way to trigger additional tests after a PR is created?

Tests are triggered when the PR is created or updated.
It is also possible to trigger the tests by manually forcing the pocl workflow to run.

bashbaug:
OK, I see.  I think all comments have been addressed now.

At some point it might be nice to support automated testing without creating a PR, but that can be saved for a separate PR.

.github/workflows/pocl.yml
```diff
@@ -0,0 +1,232 @@
...
+          SHA=$(git ls-remote https://github.com/KhronosGroup/SPIRV-LLVM-Translator.git refs/heads/${TRANSLATOR_BRANCH} | awk '{print $1}')
+          echo "sha=$SHA" >> $GITHUB_OUTPUT
+      - name: Setup Ninja
+        uses: seanmiddleditch/gha-setup-ninja@master
+      - name: Cache PoCL
+        id: cache-pocl
+        uses: actions/cache@v6
+        with:
+          path: ${{ env.pocl }}
+          key: ubuntu-24.04-pocl-${{ env.POCL_VERSION }}-${{ steps.translator_sha.outputs.sha }}
```
rjodinchr:
```suggestion
      - name: Cache PoCL
        id: cache-pocl
        uses: actions/cache@v6
        with:
          path: ${{ env.pocl }}
          key: ubuntu-24.04-pocl-${{ env.POCL_VERSION }}-${{ steps.translator_sha.outputs.sha }}
      - name: Setup Ninja
        uses: seanmiddleditch/gha-setup-ninja@master
```
I understand that this part has just been moved, but I think it is a good time to move it to avoid setting up Ninja if PoCL is already cached.

.github/workflows/pocl.yml
```diff
@@ -0,0 +1,232 @@
...
+      pocl: ${{ steps.pocl_env.outputs.pocl }}
+      translator_sha: ${{ steps.translator_sha.outputs.sha }}
+    steps:
+      - name: Checkout OpenCL CTS (for composite action)
+        uses: actions/checkout@v7
+      - name: Environment variables
+        id: pocl_env
+        run: |
+          echo "pocl=${{ github.workspace }}/pocl-install" >> $GITHUB_ENV
+          echo "pocl=${{ github.workspace }}/pocl-install" >> $GITHUB_OUTPUT
```
rjodinchr: I would prefer something like that, but that's fine if you prefer to keep it that way.
```suggestion
          POCL_INSTALL_PATH="${{ github.workspace }}/pocl-install"
          echo "pocl=$POCL_INSTALL_PATH" >> $GITHUB_ENV
          echo "pocl=$POCL_INSTALL_PATH" >> $GITHUB_OUTPUT
```

rjodinchr (dismissed):
I'm looking for a way to avoid the duplication of the `ubuntu-24.04` string. But I'll keep that for later, let's get this one in first.

LGTM

rjodinchr (approved):
I would be fine merging that before the next teleconference if you want. This is just CI stuff, that is not critical as well.

.github/workflows/pocl.yml
bashbaug: Done.  I moved the code checkout step also, fyi.  It didn't take too long to execute, but it's not needed either, so may as well skip it too.

.github/workflows/pocl.yml
bashbaug: Sure, done.

bashbaug:
> I would be fine merging that before the next teleconference if you want. This is just CI stuff, that is not critical as well.

I do think this would be helpful since it would enable automated testing for things like #2761, plus I have some other similar changes where I haven't submitted PRs yet.

IMHO it'd also make it easier to experiment with PoCL 7.2.

So, +1 from me!

PR #2770, opened 2026-08-12
compiler: skip options_uniform_work_group_size in SPIR-V mode

joselopeqti:
The -cl-uniform-work-group-size flag has no standard representation in SPIR-V: the OpenCL SPIR-V environment spec defines no ExecutionMode that carries this constraint, so when building from SPIR-V the runtime cannot enforce uniform work-group sizes and the negative path of this test (expecting CL_INVALID_WORK_GROUP_SIZE) will always fail.

Skip the test when gCompilationMode == kSpir_v so that it is reported as skipped rather than as a spurious failure.

### DIFF
--- a/test_conformance/compiler/test_build_options.cpp
+++ b/test_conformance/compiler/test_build_options.cpp
@@ -17,6 +17,7 @@
 #include "harness/kernelHelpers.h"
 #include "harness/os_helpers.h"
 #include "harness/testHarness.h"
+#include "harness/parseParameters.h"
 
 extern std::string spvIncludeTestDirectory;
 extern std::string spvSecondIncludeTestDirectory;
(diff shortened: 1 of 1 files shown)


rjodinchr (approved):

bashbaug (commented):
This doesn't seem right - can you check the expected behavior here?

Because `-cl-uniform-work-group-size` is a build option it doesn't need a SPIR-V representation, just like it doesn't need an OpenCL C representation.  The way I would expect this to work is:

1. In SPIR-V mode, the OpenCL C kernel is compiled to SPIR-V, all good.
2. The SPIR-V module is used to create the program object (using `clCreateProgramWithIL`) in place of the OpenCL C string (using `clCreateProgramWithSource`), still good.
3. The program is built with the `-cl-uniform-work-group-size` build option, which enforces the uniform work-group size requirement.

bashbaug (changes requested):
Please address the question above before merging - thanks!

alegal-qc:
In this negative test, the error is expected to come from clEnqueueNDRangeKernel (line 477). There is just nothing supplied to clEnqueueNDRangeKernel  that can help to throw the expected error.

bashbaug:
> There is just nothing supplied to clEnqueueNDRangeKernel that can help to throw the expected error.

The kernel object itself should be sufficient to throw the error, correct?  This is the way the uniform work-group size requirement is conveyed when the program object is created from OpenCL C source (AKA when not in "SPIR-V mode").

joselopeqti:
After some discussions and re-reviewing, the `-cl-uniform-work-group-size` build should be enough to enforce the uniform work-group size requirement. Closing this PR.

PR #2771, opened 2026-08-12
change kernel_preprocessor_macros backslash handling for __FILE__

### LINKED ISSUE #2769: Backslash handling for __FILE__ and #line
The current CTS test for the pre-processor `__FILE__` macro uses the `#line` directive to set the name of the current source file.  There is special handling for the backslash character `\` on Windows only, presumably because backslashes are typically part of a full Windows file name:

https://github.com/KhronosGroup/OpenCL-CTS/blob/bc44b25d7b6266a848d1acc8a8f2998fbd1dafc5/test_conformance/basic/test_preprocessors.cpp#L102-L119

In the past, this produced the desired behavior on Clang-based implementations, because the filename was treated as a string literal and hence subject to escape handling.  For example, in this example https://godbolt.org/z/o8oWq4n57, the filename in the error message is `C:\test:1:1: error: unknown type name 'foo'` for the line directive `#line 1 "C:\\test"`.  In the latest versions of Clang, though (as of now, August 10th 2026), the filenames in the line directive are no longer subject to escape handling, see for example https://godbolt.org/z/e4h5sn7q9, which instead prints the error message as `C:\\test:1:1: error: unknown type name 'foo'`.  This change in behavior means that the test fails for implementations using the latest Clang.

How do we want to handle this?  Some options:

1. This is a Clang bug, the C99 spec says that "the string literal of a #line directive, if present, shall be a character string literal" (section 6.10.4) and is hence subject to escape handling (section 6.4.5).  Note, the Clang change was introduced by https://github.com/llvm/llvm-project/pull/211512.
2. This behavior was always ambiguous (or perhaps, the escape behavior was never intended), so we should accept both the old behavior (with escape handling) and the new behavior (no escape handling).  For example, we could update the test to check for both strings, or replace the backslash character with forward slashes instead, or some other change.  If we go this route, we might also want to add some informative text (footnote?) to the spec cautioning about this difference in behavior.
3. The previous behavior with the special backslash handling was working around a Clang bug and should no longer be necessary.  We should remove this handling from the test and require that the strings do not process escape sequences.
4. Something else.

Regardless, we may want to temporarily stop failing the test while we figure out what to do.  Interestingly, a mismatch is currently logged as a "warning", not an "error", though the test still does fail when there is a mismatch:

https://github.com/KhronosGroup/OpenCL-CTS/blob/bc44b25d7b6266a848d1acc8a8f2998fbd1dafc5/test_conformance/basic/test_preprocessors.cpp#L330-L334

bashbaug:
fixes #2769 

Unconditionally replaces backslash characters `\` in the test filename with frontslash characters `/` instead.  By eliminating backslash characters, the test will work both with compilers that process filenames with escape handling and compilers that do not.

[run-test: test_basic kernel_preprocessor_macros]

### DIFF
--- a/test_conformance/basic/test_preprocessors.cpp
+++ b/test_conformance/basic/test_preprocessors.cpp
@@ -105,7 +105,7 @@ REGISTER_TEST(kernel_preprocessor_macros)
     // escape handling and some compilers do not.
     std::replace(testFileName.begin(), testFileName.end(), '\\', '/');
 
-    sprintf(programSource,preprocessor_test,testFileName.c_str());
+    sprintf(programSource, preprocessor_test, testFileName.c_str());
 
     // Create the kernel
     if( create_single_kernel_helper( context, &program, &kernel, 1,  (const char **)&programPtr, "test" ) != 0 )
--- a/test_conformance/basic/test_preprocessors.cpp
+++ b/test_conformance/basic/test_preprocessors.cpp
@@ -14,6 +14,7 @@
 // limitations under the License.
 //
 #include <ctype.h>
+#include <algorithm>
 
 #include "testBase.h"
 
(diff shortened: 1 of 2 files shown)


test_conformance/basic/test_preprocessors.cpp
```diff
@@ -96,28 +97,15 @@ REGISTER_TEST(kernel_preprocessor_macros)
...
-#endif
-    sprintf(programSource,preprocessor_test,curFileName);
+    std::string testFileName(__FILE__);
+
+    // Replace all backslash characters with forward slashes.
+    // This avoid possible issues where some compilers process filenames with
+    // escape handling and some compilers do not.
+    std::replace(testFileName.begin(), testFileName.end(), '\\', '/');
+
+    sprintf(programSource, preprocessor_test, testFileName.c_str());
```
svenvh, line 120: You may want to change this into an `snprintf`, as we could now overflow `programSource` for long enough `__FILE__` contents (whereas before that wouldn't happen due to `curFileName` being limited to 511 by the `snprintf`).

bashbaug: Good point.  I actually went a slightly different route and limited the length of the test file name to 512 characters, same as before.  I think this is better than using `snprintf` for two reasons:

1. If we truncate the test file name the test is still valid and should pass (albeit with a truncated value for `__FILE__`), whereas if we truncate the program source it probably won't build.
2. The size of the output buffer is also 512 characters, so even if the program source happened to build (and it probably won't), we wouldn't read the appropriate number of characters for the `__FILE__`.

Let me know if this works for you.

Ultimately, this test should be refactored to be more dynamic to avoid these fixed-size arrays that "should be big enough", but that'd be a bigger change.

svenvh: Looks perfectly reasonable to me, thanks!  Indeed we could modernize/improve this test further, but that's beyond the scope of what you're fixing here.

svenvh (approved):

rjodinchr (approved):

PR #2772, opened 2026-08-13
add testing for SPV_KHR_non_semantic_info

bashbaug:
Adds targeted SPIR-V testing for the SPV_KHR_non_semantic_info extension, and specifically for SPIR-V modules that declare unknown (bogus) non-semantic extended instruction sets.

The testing on PoCL is skipped right now because PoCL does not currently advertise support for the SPV_KHR_non_semantic_info extension (via SPIR-V queries).

Even if PoCL did advertise support for SPV_KHR_non_semantic_info, the test will currently fail, because the fix for unknown non-semantic extended instruction sets has not been back-ported to all of the release branches.  This should be fixed shortly, though, see e.g.: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3958

[run-test: test_spirv_new extinstimport_nonsemantic]

### DIFF
--- a/ci/pocl/golden.json
+++ b/ci/pocl/golden.json
@@ -216,7 +216,7 @@
     "test_spirv_new": {
         "": {
             "basic_versions": "pass",
-            "extimport_nonsemantic": "skip"
+            "extinstimport_nonsemantic": "skip"
         }
     },
     "test_vectors": {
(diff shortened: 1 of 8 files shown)


bashbaug:
Marking as a draft - I still need to add 32-bit SPIR-V files.

PR #2773, opened 2026-08-14
relationals: do not test deprecated scalar any/all

svenvh:
Scalar inputs to the `all` and `any` builtin functions are deprecated by OpenCL 3.0.

### DIFF
--- a/test_conformance/relationals/test_relationals.cpp
+++ b/test_conformance/relationals/test_relationals.cpp
@@ -202,7 +202,7 @@ int anyVerifyFn( ExplicitType vecType, unsigned int vecSize, void *inData )
 REGISTER_TEST(relational_any)
 {
     ExplicitType vecType[] = { kChar, kShort, kInt, kLong };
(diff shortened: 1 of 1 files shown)


svenvh:
Leaving this as a draft for now, as I don't know when we generally remove testing of deprecated functionality.

bashbaug:
> I don't know when we generally remove testing of deprecated functionality.

I'm not saying this is the right decision necessarily, but I don't think we've generally removed testing much ever.  Our definition of deprecation is a signal, not a removal, so the testing should probably remain in place until the feature is removed.

Note though, we still have testing for the rare cases where a feature was removed too, e.g. `clSetCommandQueueProperty`, although the testing has been updated to handle cases where the feature is no longer present in newer OpenCL implementations.

svenvh:
Thanks for the clarification.  I noticed we had no testing for some other deprecated builtins (e.g. `read_mem_fence`), hence I opened this.

It seems reasonable to leave the testing in place for now, so I'll close this.

PR #2775, opened 2026-08-20
basic: Add missing application constant tests

svenvh:
These 5 host constants were not tested yet.  Compare the NaN and infinity constants semantically using `isnan`/`isinf` because equality is unsuitable for NaN and bitwise checks are unnecessarily representation-specific.

[run-test: test_basic host_numeric_constants]

### DIFF
--- a/test_conformance/basic/test_numeric_constants.cpp
+++ b/test_conformance/basic/test_numeric_constants.cpp
@@ -16,6 +16,7 @@
 #include "testBase.h"
 
 #include <cinttypes>
+#include <cmath>
 
 #define TEST_VALUE_POSITIVE( string_name, name, value ) \
 { \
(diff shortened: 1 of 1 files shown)


test_conformance/basic/test_numeric_constants.cpp
```diff
@@ -140,6 +141,15 @@ REGISTER_TEST(host_numeric_constants)
     TEST_VALUE_EQUAL( "CL_M_SQRT2_F",    CL_M_SQRT2_F,   MAKE_HEX_FLOAT(0x1.6a09e6p+0f, 0x16a09e6L, -24));
     TEST_VALUE_EQUAL( "CL_M_SQRT1_2_F",  CL_M_SQRT1_2_F, MAKE_HEX_FLOAT(0x1.6a09e6p-1f, 0x16a09e6L, -25));
 
+    TEST_VALUE_EQUAL_LITERAL( "CL_NAN", std::isnan(CL_NAN), 1 )
+    TEST_VALUE_EQUAL_LITERAL( "CL_HUGE_VALF", (std::isinf(CL_HUGE_VALF) && CL_HUGE_VALF > 0), 1 )
+#if !defined(_MSC_VER)
+    // Skip under MSVC to avoid error C2177: constant too big.
+    TEST_VALUE_EQUAL_LITERAL( "CL_HUGE_VAL", (std::isinf(CL_HUGE_VAL) && CL_HUGE_VAL > 0), 1 )
+#endif
```
bashbaug: Does this imply that there is a problem with the headers on MSVC that we should fix?

The test itself seems fine and I don't want this comment to hold up this PR, but if this is a real issue I don't want this it to slip through the cracks, either.

bashbaug (dismissed):

test_conformance/basic/test_numeric_constants.cpp
svenvh: I think there is indeed, [confirmed with latest MSVC](https://compiler-explorer.com/z/EoTn4K9T4).  I have raised https://github.com/KhronosGroup/OpenCL-Headers/issues/313 .

rjodinchr (dismissed):

test_conformance/basic/test_numeric_constants.cpp
bashbaug: I've fixed the MSVC issue in the latest headers.

https://github.com/KhronosGroup/OpenCL-Headers/pull/314

Could we try removing this special handling now?

```suggestion
    TEST_VALUE_EQUAL_LITERAL( "CL_HUGE_VAL", (std::isinf(CL_HUGE_VAL) && CL_HUGE_VAL > 0), 1 )
```

svenvh: Thanks for pushing that.  The MSVC CI builds seem to be working now so I've removed the MSVC guard again.

bashbaug (approved):

rjodinchr (approved):

joselopeqti (approved):

lakshmih (approved):

bashbaug:

PR #2776, opened 2026-08-21
conversions: Return 0 when converting NaN to integer with saturation

woodychow:
Per the OpenCL C specification (section 6.4.3.3 "Out-of-Range Behavior and Saturated Conversions" [1]):
"When in saturated mode, values that are outside the representable range are clamped to the nearest representable value in the destination format, and NaN is converted to zero."

In DataInfoSpec::conv_sat, computing the reference values for cl_ulong and cl_long previously evaluated comparisons on NaN, resulting in undefined behavior when casting NaN to an integer type.

Explicitly check for NaN input when converting to integer types in conv_sat and set the output reference to 0.

[1] https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#out-of-range-behavior

[run-test: test_conversions -w]

### DIFF
--- a/test_conformance/conversions/conversions_data_info.h
+++ b/test_conformance/conversions/conversions_data_info.h
@@ -856,6 +856,24 @@ void DataInfoSpec<InType, OutType, InFP, OutFP>::conv_sat(OutType *out,
         cl_float inVal = *in;
         if (is_in_half()) inVal = HTF(*in);
 
+        if (!(std::is_floating_point<OutType>::value || is_out_half()))
+        {
+            bool is_nan = false;
+            if constexpr (std::is_same<InType, cl_half>::value && InFP)
+            {
+                is_nan = std::isnan(inVal);
+            }
+            else
+            {
+                is_nan = std::isnan(*in);
+            }
+            if (is_nan)
+            {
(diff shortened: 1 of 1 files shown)


rjodinchr (dismissed):

svenvh (dismissed):

test_conformance/conversions/conversions_data_info.h
```diff
@@ -856,6 +856,24 @@ void DataInfoSpec<InType, OutType, InFP, OutFP>::conv_sat(OutType *out,
...
+            bool is_nan = false;
+            if constexpr (std::is_same<InType, cl_half>::value && InFP)
+            {
+                is_nan = std::isnan(inVal);
+            }
+            else
+            {
+                is_nan = std::isnan(*in);
+            }
+            if (is_nan)
```
bashbaug: Can we simplify this?  We've already handled the "half" case above, so I think this will work for all floating-point types.

```suggestion
            if (std::isnan(inVal))
```

rjodinchr (approved):

svenvh (approved):

bashbaug (approved):

PR #2777, opened 2026-08-21
commonfns: broaden x & y mix inputs beyond [0, 1]

svenvh:
The `mix` tests restricted the `x` and `y` inputs to the interval `[0, 1]`, while the specification only puts that restriction on the `a` input.

[run-test: test_commonfns mix]
[run-test: test_commonfns mixf]

### DIFF
--- a/test_conformance/commonfns/test_mix.cpp
+++ b/test_conformance/commonfns/test_mix.cpp
@@ -181,17 +181,28 @@ int test_mix_fn(cl_device_id device, cl_context context, cl_command_queue queue,
         pragma_str = "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n";
         for (i = 0; i < num_elements; i++)
         {
-            input_ptr[0][i] = conv_to_half((float)genrand_real1(d));
-            input_ptr[1][i] = conv_to_half((float)genrand_real1(d));
+            input_ptr[0][i] =
+                conv_to_half(get_random_float(-32752.0f, 32752.0f, d));
+            input_ptr[1][i] =
+                conv_to_half(get_random_float(-32752.0f, 32752.0f, d));
             input_ptr[2][i] = conv_to_half((float)genrand_real1(d));
+        }
+    }
+    else if (std::is_same<T, float>::value)
+    {
+        for (i = 0; i < num_elements; i++)
+        {
+            input_ptr[0][i] = get_random_float(-0x20000000, 0x20000000, d);
+            input_ptr[1][i] = get_random_float(-0x20000000, 0x20000000, d);
+            input_ptr[2][i] = (T)genrand_real1(d);
         }
     }
     else
     {
         for (i = 0; i < num_elements; i++)
         {
-            input_ptr[0][i] = (T)genrand_real1(d);
-            input_ptr[1][i] = (T)genrand_real1(d);
+            input_ptr[0][i] = get_random_double(-0x20000000, 0x20000000, d);
+            input_ptr[1][i] = get_random_double(-0x20000000, 0x20000000, d);
             input_ptr[2][i] = (T)genrand_real1(d);
         }
     }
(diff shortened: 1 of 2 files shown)


test_conformance/commonfns/test_mix.cpp
rjodinchr: I think this formula is not correct anymore with `x` & `y` outside of `[0;1]`.
`r` can now be negative, so can `delta`, leading to silently passing negative results.
Also `r` could now be null (division by zero).

```suggestion
double x = conv_to_dbl(inptrX[i]);
double y = conv_to_dbl(inptrY[i]);
double a = conv_to_dbl(inptrA[i]);
r = x + (y - x) * a;

o = conv_to_dbl(outptr[i]);
double max_scale = std::max({ fabs(r), fabs(x), fabs(y) });
delta = max_scale > 0.0 ? (fabs(r - o) / max_scale) : 0.0;
```

rjodinchr: Same here

test_conformance/commonfns/test_mix.cpp
```diff
@@ -181,17 +181,28 @@ int test_mix_fn(cl_device_id device, cl_context context, cl_command_queue queue,
         pragma_str = "#pragma OPENCL EXTENSION cl_khr_fp16 : enable\n";
         for (i = 0; i < num_elements; i++)
         {
-            input_ptr[0][i] = conv_to_half((float)genrand_real1(d));
-            input_ptr[1][i] = conv_to_half((float)genrand_real1(d));
+            input_ptr[0][i] =
+                conv_to_half(get_random_float(-32752.0f, 32752.0f, d));
+            input_ptr[1][i] =
+                conv_to_half(get_random_float(-32752.0f, 32752.0f, d));
```
rjodinchr: It might be either for future maintenance to use something like: `const float half_bound = CL_HALF_MAX / 2.0f;` instead of raw values.

test_conformance/commonfns/test_mix.cpp
```diff
@@ -181,17 +181,28 @@ int test_mix_fn(cl_device_id device, cl_context context, cl_command_queue queue,
...
             input_ptr[2][i] = conv_to_half((float)genrand_real1(d));
         }
     }
+    else if (std::is_same<T, float>::value)
+    {
+        for (i = 0; i < num_elements; i++)
+        {
+            input_ptr[0][i] = get_random_float(-0x20000000, 0x20000000, d);
+            input_ptr[1][i] = get_random_float(-0x20000000, 0x20000000, d);
```
rjodinchr: I'm not sure if the values here are the one you really expect.
```suggestion
            input_ptr[0][i] = get_random_float(-0x1.0p29f, 0x1.0p29f, d);
            input_ptr[1][i] = get_random_float(-0x1.0p29f, 0x1.0p29f, d);
```
Maybe with a const variable as well to avoid the duplication?

test_conformance/commonfns/test_mix.cpp
svenvh: Thanks, this changes the check from absolute error (as required by the spec) to relative error though.  The problem is that for large inputs the pure absolute error against an ideal result becomes impossible because float spacing itself exceeds `1e-3`.  I’ll try to find an alternative.

test_conformance/commonfns/test_mix.cpp
svenvh: Thanks for the suggestion, applied.

test_conformance/commonfns/test_mix.cpp
svenvh: Thanks for the suggestion, applied.

test_conformance/commonfns/test_mix.cpp
svenvh: Fixed.

rjodinchr (dismissed):
Looking good, just need to update the CI expectations

rjodinchr (approved):

bashbaug:

PR #2778, opened 2026-08-24
add a utility function to query if a SPIR-V extension is available

bashbaug:
This PR adds a utility function to check if a SPIR-V extension is available (using the SPIR-V query from OpenCL 3.1 or cl_khr_spirv_queries).  This utility function is not currently used, but I would like to use it for testing in the future.  I think it makes sense to review and merge it independently of any testing for a specific SPIR-V extension, but if it would be easier to review with testing for a SPIR-V extension, that's fine too.

I've kept this specific to the SPIR-V test executable for now (test_spirv_new), similar to the SPIR-V version check, but I could consider moving it to the harness for broader availability in the future.

As an aside, note that this PR also bumps the OpenCL headers from v3.0 to v3.1 in order to use the un-suffixed enums for the SPIR-V extension query.

[run-test: test_spirv_new basic_versions]


bashbaug:
As an example, the (currently draft) test for SPV_KHR_non_semantic_info would use this functionality: https://github.com/KhronosGroup/OpenCL-CTS/pull/2772

test_conformance/spirv_new/main.cpp
```diff
@@ -85,6 +87,60 @@ std::vector<unsigned char> readSPIRV(const char *file_name)
...
+    if (version < Version(3, 1)
+        && !is_extension_available(device, CL_KHR_SPIRV_QUERIES_EXTENSION_NAME))
+    {
+        return false;
+    }
+
+    cl_int err;
+    size_t sz = 0;
+    err = clGetDeviceInfo(device, CL_DEVICE_SPIRV_EXTENSIONS, 0, nullptr, &sz);
+    if (err != CL_SUCCESS) return false;
```
svenvh: We would get a silent "not supported" if this query fails.  Would it be worth logging something like "clGetDeviceInfo failed"?

test_conformance/spirv_new/main.cpp
```diff
@@ -85,6 +87,60 @@ std::vector<unsigned char> readSPIRV(const char *file_name)
...
+
+    std::vector<const char *> extensions(sz / sizeof(const char *));
+    err = clGetDeviceInfo(device, CL_DEVICE_SPIRV_EXTENSIONS, sz,
+                          extensions.data(), nullptr);
+    if (err != CL_SUCCESS) return false;
+
+    for (auto &ext : extensions)
```
svenvh:
```suggestion
    for (const auto &ext : extensions)
```

svenvh (commented):
LGTM and landing this ahead of actual uses seems fine to me.

test_conformance/spirv_new/main.cpp
bashbaug: Sure, I've added the log for now.

Note that this function was modeled heavily after `get_extension_version`, which throws an exception (`std::runtime_error`) if any of these calls fail.  I generally prefer not to throw exceptions from CTS tests, but I could match that behavior, if desired.

test_conformance/spirv_new/main.cpp
bashbaug: done - I fixed the same code in `get_extension_version` also.

test_conformance/spirv_new/main.cpp
svenvh: Thanks; I think non-fatal logs are better than throwing exceptions here.

svenvh (dismissed):

rjodinchr (dismissed):

bashbaug:
Discussed in the September 8th teleconference.  OK to merge when the conflict is resolved.

rjodinchr (approved):

PR #2782, opened 2026-08-27
spirv_new: assemble SPIR-V using libSPIRV-Tools

svenvh:
Allow SPIR-V tests to assemble SPIR-V modules using libSPIRV-Tools instead of relying on checked-in assembly.  Tests that iterate over combinations of e.g. types and vector widths can generate each case on the fly, avoiding many .spvasm files with a lot of duplication.

This makes the SPIRV-Tools library a required build dependency.

Migrate the fmath tests as an initial proof of concept.


test_common/harness/spirvTools.h
```diff
@@ -0,0 +1,49 @@
...
+                         std::vector<uint32_t> &spirv_binary);
+
+// Create an IL program from assembled SPIR-V and build it for the device.
+cl_int
+create_program_from_spirv_binary(clProgramWrapper &program, cl_device_id device,
+                                 cl_context context,
+                                 const std::vector<uint32_t> &spirv_binary);
+
+// Assemble textual SPIR-V, create an IL program, and build it for the device.
+cl_int create_program_from_spirv_text(clProgramWrapper &program,
```
rjodinchr: I don't see that function being used, am I missing something?

test_common/harness/spirvTools.cpp
```diff
@@ -0,0 +1,111 @@
...
+    auto message_consumer = [](spv_message_level_t, const char *,
+                               const spv_position_t &, const char *message) {
+        log_error("SPIR-V Tools: %s\n", message);
+    };
+
+    spvtools::SpirvTools spirv_tools(SPV_ENV_OPENCL_2_0);
+    spirv_tools.SetMessageConsumer(message_consumer);
+
+    if (!spirv_tools.Assemble(spirv_text, &spirv_binary)) return false;
+    return spirv_tools.Validate(spirv_binary.data(), spirv_binary.size());
```
rjodinchr, line 62: Should we guard it behind `gDisableSPIRVValidation` that is set by `--disable-spirv-val` in the harness?

test_common/harness/spirvTools.cpp
```diff
@@ -0,0 +1,111 @@
...
+    output << "OpMemoryModel Physical" << address_bits << " OpenCL\n";
+    return CL_SUCCESS;
+}
+
+bool assemble_spirv_text(const std::string &spirv_text,
+                         std::vector<uint32_t> &spirv_binary)
+{
+    auto message_consumer = [](spv_message_level_t, const char *,
+                               const spv_position_t &, const char *message) {
+        log_error("SPIR-V Tools: %s\n", message);
```
rjodinchr:
```suggestion
        log_error("SPIR-V Tools: %zu:%zu: %s\n", position.line, position.column, message);
```

test_common/harness/spirvTools.cpp
```diff
@@ -0,0 +1,111 @@
...
+                    platform, "clCreateProgramWithILKHR"));
+        if (clCreateProgramWithILKHR == nullptr)
+        {
+            log_error("Failed to get clCreateProgramWithILKHR\n");
+            return CL_INVALID_OPERATION;
+        }
+        program = clCreateProgramWithILKHR(context, spirv_binary.data(),
+                                           binary_size, &error);
+    }
+    if (error != CL_SUCCESS) return error;
```
rjodinchr: We may want to print something before returning.

test_conformance/spirv_new/test_op_fmath.cpp
```diff
@@ -16,20 +16,145 @@
...
+    else
+    {
+        log_error("Unsupported fmath function: %s\n", function_name.c_str());
+        return -1;
+    }
+
+    const std::string value_type = vector_width == 1
+        ? scalar_name
+        : "v" + std::to_string(vector_width) + scalar_name;
+    const unsigned alignment = scalar_bits * vector_width / 8;
```
rjodinchr: It does not feel right for 3-element vector, but I'm not sure if that's an issue or not here.

test_common/harness/spirvTools.cpp
svenvh: Great suggestion, done!

test_common/harness/spirvTools.cpp
svenvh: Applied (in a slightly more verbose form).

test_common/harness/spirvTools.cpp
svenvh: Done.

test_common/harness/spirvTools.h
svenvh: Good find, removed.

test_conformance/spirv_new/test_op_fmath.cpp
svenvh: Currently not, as we don't test 3-element vectors here.  But I've fixed it anyway in case we ever decide to add 3-element vectors.

rjodinchr (commented):
LGTM, please get a second approval as I'm not very confident about that test.

test_conformance/spirv_new/procs.h
```diff
@@ -55,6 +56,11 @@ int get_unbuilt_program_with_il(clProgramWrapper &prog,
                                 const cl_context context, const char *fileName);
 int get_program_with_il(clProgramWrapper &prog, const cl_device_id deviceID,
                         const cl_context context, const char *fileName);
+int get_program_with_generated_il(clProgramWrapper &prog,
+                                  const cl_device_id deviceID,
+                                  const cl_context context,
+                                  const char *prog_name,
+                                  const std::string &spirv_text);
```
bashbaug: I think we need to think a bit more about how to structure the various helper functions to get (create?) program objects from SPIR-V. We don't _need_ to do this as part of this PR, but if we don't do it now, we should do it soon.

For reference, for creating programs from source, we have (in the harness):

`create_single_kernel_helper`: Calls `create_single_kernel_helper_create_program`, then `build_program_create_kernel_helper`.

`create_single_kernel_helper_create_program`: Calls `clCreateProgramWithSource` for online compilation, or `create_single_kernel_helper_create_program_offline` otherwise.

`create_single_kernel_helper_create_program_offline`: Gets the offline compiler output and then calls either `clCreateProgramWithBinary` (for binary offline compilation) or `clCreateProgramWithIL` (or `clCreateProgramWithILKHR`) (for SPIR-V offline compilation).

`build_program_create_kernel_helper`: Calls `clBuildProgram`, checks the program build log, optionally calls `clCreateKernel`.

We don't have any helper functions that accept file names.

For creating programs from SPIR-V, we currently have, prior to this PR (in the spirv_new test):

`get_program_with_il`: Calls `get_unbuilt_program_with_il`, then `clBuildProgram`.

`get_unbuilt_program_with_il`: Calls `offline_get_program_with_il` for binary offline compilation, otherwise reads the SPIR-V binary from a file and then calls `clCreateProgramWithIL` (or `clCreateProgramWithILKHR`).

`offline_get_program_with_il`: Invokes a script to generate an output file, reads the output file, then calls `clCreateProgramWithBinary` followed by `clBuildProgram`.
  * Aside: Since this helper builds the program, it doesn't seem right to call it from `get_unbuilt_program_with_il`.

Now, with this PR, we additionally have (in the harness directory, but compiled into a separate `spirv_tools_harness` library):

`create_program_from_spirv_binary`: Calls `clCreateProgramWithIL` (or `clCreateProgramWithILKHR`), then `clBuildProgram`.

and (in the spirv_new test):

`get_program_with_generated_il`: Assembles the SPIR-V text, then calls `offline_get_program_with_il` for binary offline compilation, otherwise calls `create_program_from_spirv_binary`.

`offline_get_program_with_il` is also enhanced to optionally generate an input file when called with SPIR-V text, before invoking the script to generate the output file.

I'm guessing the answer is "no", but worth asking anyways: do we need all of these helper functions?

If we keep all of them, can we rearrange and rename them so there's more reuse and the purpose for each of the helpers is clearer? For example, can `get_unbuilt_program_with_il` call `create_program_from_spirv_binary`? Should we rename `get_program_with_il` so it is clearer that the IL is coming from a file?

I'd really like to fix the `get` vs. `create` difference also, and perhaps consolidate these helper functions into the harness, but I think this would be better done as a separate PR.

test_conformance/spirv_new/CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ set(TEST_HARNESS_SOURCES
 set(${MODULE_NAME}_SOURCES ${${MODULE_NAME}_SOURCES} ${TEST_HARNESS_SOURCES})
 
 include(../CMakeCommon.txt)
+target_link_spirv_tools(${${MODULE_NAME}_OUT})
```
bashbaug: Is there any way to add this dependency to the new SPIR-V harness `spirv_tools_harness` so we don't need to add it to every executable?

bashbaug (commented):
Thank you for doing this! I think it will make it much easier to permute different SPIR-V types and opcodes in the future.

I added one long comment below on a specific line for better threading, but it's really more of a general comment.

I would be fine making some of the improvements in a separate PR.

test_conformance/spirv_new/procs.h
svenvh: A fair point; indeed doing the wider consolidation/renaming is better done in a separate PR.  But at least I can rename the new helpers added in this PR to have more consistent names, so I've renamed:

- `create_program_from_spirv_binary` into `create_and_build_program_from_spirv_il`, and
- `get_program_with_generated_il` into `create_and_build_program_from_spirv_text`.

test_conformance/spirv_new/CMakeLists.txt
svenvh: I think we should still keep the dependency opt-in, to avoid linking all of the non-SPIR-V executables against libSPIRV-Tools unnecessarily.  But we can make this a little less awkward with `set(HARNESS_LIB spirv_tools_harness)` instead.  Updated accordingly.

PR #2783, opened 2026-08-28
Created new test to verify __OPENCL_C_VERSION__ preprocessor macro

### LINKED ISSUE #2670: missing testing for __OPENCL_C_VERSION__ preprocessor macro
We have tests that ensure a program can be compiled for all of the OpenCL C versions advertised by a device, see:

https://github.com/KhronosGroup/OpenCL-CTS/blob/main/test_conformance/compiler/test_opencl_c_versions.cpp

However, we never check that the `__OPENCL_C_VERSION__` preprocessor macro is set properly when compiling for each OpenCL C version.

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#preprocessor-directives-and-macros

In addition to checking that a program can be compiled for a specific OpenCL C version, we should also check that the preprocessor macro is set correctly, for example by executing the kernel, writing the value of the preprocessor macro to a buffer, and checking that the value written is correct.

shajder:
Fixes #2670

[run-test: test_compiler opencl_c_version_macro]

### DIFF
--- a/test_conformance/compiler/test_opencl_c_versions.cpp
+++ b/test_conformance/compiler/test_opencl_c_versions.cpp
@@ -17,9 +17,105 @@
 #include "testBase.h"
 #include "harness/featureHelpers.h"
 
+#include <string>
 #include <vector>
 
-static const char* test_kernel = R"CLC(
+namespace {
+
+const char* version_macro_kernel = R"CLC(
+__kernel void test(__global int* dst) {
+    dst[0] = __OPENCL_C_VERSION__;
+}
+)CLC";
+
+// __OPENCL_C_VERSION__ requires support for OpenCL C 1.2 or newer
+Version min_macro_clc_version() { return Version(1, 2); }
+
+// This sub-test verifies if __OPENCL_C_VERSION__ preprocessor macro matches
+// specific version the kernel was built for.
+int test_version_macro(cl_context context, cl_command_queue queue,
+                       const Version& clc_version)
+{
+    std::string buildOptions = "-cl-std=CL";
+    buildOptions += std::to_string(clc_version.get_major());
+    buildOptions += ".";
+    buildOptions += std::to_string(clc_version.get_minor());
+
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+    cl_int error = create_single_kernel_helper(context, &program, &kernel, 1,
+                                               &version_macro_kernel, "test",
+                                               buildOptions.c_str());
+    test_error(error, "Unable to build program!");
+
+    cl_int value = 0;
+    clMemWrapper dst =
+        clCreateBuffer(context, CL_MEM_READ_WRITE | CL_MEM_COPY_HOST_PTR,
+                       sizeof(value), &value, &error);
+    test_error(error, "clCreateBuffer failed");
+
+    error = clSetKernelArg(kernel, 0, sizeof(cl_mem), &dst);
+    test_error(error, "clSetKernelArg failed");
+
+    const size_t global_size = 1;
+    error = clEnqueueNDRangeKernel(queue, kernel, 1, NULL, &global_size, NULL,
+                                   0, NULL, NULL);
+    test_error(error, "clEnqueueNDRangeKernel failed");
+
+    error = clEnqueueReadBuffer(queue, dst, CL_TRUE, 0, sizeof(value), &value,
+                                0, NULL, NULL);
+    test_error(error, "clEnqueueReadBuffer failed");
+
+    const cl_int expected =
+        clc_version.get_major() * 100 + clc_version.get_minor() * 10;
+    if (value != expected)
+    {
+        log_error("got __OPENCL_C_VERSION__ %d when building with '%s', "
+                  "expected %d!\n",
+                  value, buildOptions.c_str(), expected);
+        return TEST_FAIL;
+    }
+
+    return TEST_PASS;
+}
+
+// Collects all supported OpenCL C versions to test
+int get_all_clc_versions(cl_device_id device, std::vector<Version>& versions)
+{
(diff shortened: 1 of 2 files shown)


test_conformance/compiler/test_opencl_c_versions.cpp
```diff
@@ -17,9 +17,105 @@
...
+namespace {
+
+const char* version_macro_kernel = R"CLC(
+__kernel void test(__global int* dst) {
+    dst[0] = __OPENCL_C_VERSION__;
+}
+)CLC";
+
+// __OPENCL_C_VERSION__ requires support for OpenCL C 1.2 or newer
+Version min_macro_clc_version() { return Version(1, 2); }
```
rjodinchr: Do we really need a function for that? We could just have a const variable, right?

test_conformance/compiler/test_opencl_c_versions.cpp
```diff
@@ -285,10 +380,54 @@ static int test_CL_DEVICE_OPENCL_C_VERSION_versions(cl_device_id device,
         }
     }
 
-
     return TEST_PASS;
 }
 
+}
```
rjodinchr: I understand that using namespaces is a better practice in C++. However, we could close the namespace earlier to avoid modifying existing functions that should not be affected by that change.

test_conformance/compiler/test_opencl_c_versions.cpp
```diff
@@ -17,9 +17,105 @@
...
+                  "expected %d!\n",
+                  value, buildOptions.c_str(), expected);
+        return TEST_FAIL;
+    }
+
+    return TEST_PASS;
+}
+
+// Collects all supported OpenCL C versions to test
+int get_all_clc_versions(cl_device_id device, std::vector<Version>& versions)
```
rjodinchr: The name does not really match what the function does.
Maybe `get_macro_supported_clc_versions`?

test_conformance/compiler/test_opencl_c_versions.cpp
shajder: A const variable at namespace scope triggers `clazy-non-pod-global-static`, so I replaced the function with `Version(1, 2)` at the call sites.

test_conformance/compiler/test_opencl_c_versions.cpp
shajder: Done. Noting for the record that this leaves the file in two styles, which I would rather have avoided.

test_conformance/compiler/test_opencl_c_versions.cpp
shajder: Renamed, comment updated.

test_conformance/compiler/test_opencl_c_versions.cpp
rjodinchr: I understand, but then we can have another PR to clean that file.

rjodinchr (approved):

svenvh (approved):

bashbaug:
Discussed in the September 8th teleconference. OK to merge.

test_conformance/compiler/test_opencl_c_versions.cpp
```diff
@@ -17,9 +17,105 @@
...
+            versions.push_back(version);
+        }
+    }
+
+    return TEST_PASS;
+}
+
+}
+
+const char* test_kernel = R"CLC(
```
bashbaug, line 118: Nitpick: I think this either should remain `static`, or it should move into the anonymous namespace.

```suggestion
static const char* test_kernel = R"CLC(
```

test_conformance/compiler/test_opencl_c_versions.cpp
```diff
@@ -285,10 +381,52 @@ static int test_CL_DEVICE_OPENCL_C_VERSION_versions(cl_device_id device,
...
+    if (get_device_cl_version(device) >= Version(3, 0))
+    {
+        if (get_macro_supported_clc_versions(device, versions) != TEST_PASS)
+        {
+            return TEST_FAIL;
+        }
+    }
+    else
+    {
+        versions.push_back(Version(1, 2));
```
bashbaug, line 411: I'm pretty sure we can still test OpenCL C 1.1 in this case, too (although most devices these days should take the OpenCL 3.0 or newer path).

```suggestion
        versions.push_back(Version(1, 1));
        versions.push_back(Version(1, 2));
```

shajder: Sorry for delayed answear, I missed that comment.

One question: per spec a 1.1 compiler is not required to provide the macro, so how should the test behave when the 1.1 iteration fails? Right now any failing version fails the whole test, which seems unjustified for 1.1. Would logging it and carrying on be an acceptable outcome?

bashbaug: Ah, I see now.

I was looking at the description of `CL_VERSION_1_1`, which "requires support for OpenCL C 1.1 or newer".

However, the description of `__OPENCL_C_VERSION__` says that it "requires support for OpenCL C 1.2 or newer".  So, even though the "OpenCL 1.1" version macro exists to test against in OpenCL C 1.1, the "OpenCL C version" macro itself doesn't exist in OpenCL C 1.1.

This seems a little weird, but I think this is because prior OpenCL C versions would (likely, incorrectly) test against `__OPENCL_VERSION__` (note: NOT `__OPENCL_C_VERSION__`), and this could feasibly be `CL_VERSION_1_1` for an OpenCL 1.1 device.

In any case, I think the test is fine, if anything it's testing too little vs. too much, and this is in a codepath that almost all devices won't take anyhow.  I'll just resolve this comment.

PR #2784, opened 2026-08-29
compiler: Add test for CL_PROGRAM_SOURCE with clCreateProgramWithBinary

funch0za:
The test verifies that `clGetProgramInfo` with `CL_PROGRAM_SOURCE` returns either an empty string or the source code for programs created from a binary file (per OpenCL spec).

[run-test: test_compiler get_program_source_from_binary]

### DIFF
--- a/test_conformance/compiler/test_build_helpers.cpp
+++ b/test_conformance/compiler/test_build_helpers.cpp
@@ -1146,25 +1146,25 @@ REGISTER_TEST(get_program_source_from_binary)
     }
 
     std::vector<unsigned char> binary(binary_size);
-    unsigned char* binary_ptr = binary.data();
+    unsigned char *binary_ptr = binary.data();
 
-    error = clGetProgramInfo(program, CL_PROGRAM_BINARIES,
-                             sizeof(binary_ptr), &binary_ptr, nullptr);
+    error = clGetProgramInfo(program, CL_PROGRAM_BINARIES, sizeof(binary_ptr),
+                             &binary_ptr, nullptr);
     test_error(error, "clGetProgramInfo for binary failed");
 
-    const unsigned char* binary_data = binary.data();
+    const unsigned char *binary_data = binary.data();
     cl_int binary_status = CL_SUCCESS;
 
-    program_from_binary = clCreateProgramWithBinary(
-        context, 1, &device, &binary_size, &binary_data, &binary_status,
-        &error);
+    program_from_binary =
+        clCreateProgramWithBinary(context, 1, &device, &binary_size,
+                                  &binary_data, &binary_status, &error);
     test_error(error, "clCreateProgramWithBinary failed");
     test_assert_error(binary_status == CL_SUCCESS,
                       "Binary status is not CL_SUCCESS");
 
     size_t src_size = 0;
-    error = clGetProgramInfo(program_from_binary, CL_PROGRAM_SOURCE,
-                             0, nullptr, &src_size);
+    error = clGetProgramInfo(program_from_binary, CL_PROGRAM_SOURCE, 0, nullptr,
+                             &src_size);
     test_error(error, "clGetProgramInfo for source size failed");
 
     if (src_size == 0)
@@ -1175,8 +1175,8 @@ REGISTER_TEST(get_program_source_from_binary)
 
     std::vector<char> src(src_size, static_cast<char>(0x7f));
 
-    error = clGetProgramInfo(program_from_binary, CL_PROGRAM_SOURCE,
-                             src_size, src.data(), nullptr);
+    error = clGetProgramInfo(program_from_binary, CL_PROGRAM_SOURCE, src_size,
+                             src.data(), nullptr);
     test_error(error, "clGetProgramInfo for source failed");
 
     if (src.back() != '\0')
@@ -1187,8 +1187,8 @@ REGISTER_TEST(get_program_source_from_binary)
 
     if (src_size != 1
         && (src_size != strlen(sample_kernel_code_single_line[0]) + 1
-            || memcmp(src.data(), sample_kernel_code_single_line[0],
-                      src_size) != 0))
+            || memcmp(src.data(), sample_kernel_code_single_line[0], src_size)
+                != 0))
     {
         log_error("CL_PROGRAM_SOURCE returned unexpected content\n");
         return TEST_FAIL;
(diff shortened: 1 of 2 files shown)


test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1120,3 +1120,79 @@ REGISTER_TEST(kernel_name_size)
...
+    cl_int error = CL_SUCCESS;
+    clProgramWrapper program;
+    clProgramWrapper program_from_binary;
+
+    program = clCreateProgramWithSource(
+        context, 1, sample_kernel_code_single_line, nullptr, &error);
+    test_error(error, "clCreateProgramWithSource failed");
+
+    error = clBuildProgram(program, 1, &device, nullptr, nullptr, nullptr);
+    test_error(error, "clBuildProgram failed");
```
rjodinchr: Please use `create_single_kernel_helper` instead

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1120,3 +1120,79 @@ REGISTER_TEST(kernel_name_size)
...
+
+    const unsigned char *binary_data = binary.data();
+    cl_int binary_status = CL_SUCCESS;
+
+    program_from_binary =
+        clCreateProgramWithBinary(context, 1, &device, &binary_size,
+                                  &binary_data, &binary_status, &error);
+    test_error(error, "clCreateProgramWithBinary failed");
+    test_assert_error(binary_status == CL_SUCCESS,
+                      "Binary status is not CL_SUCCESS");
```
rjodinchr:
```suggestion
    test_error(binary_status, "Binary status is not CL_SUCCESS");
```
So that the `binary_status` can be printed in a human readable format?

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1120,3 +1120,79 @@ REGISTER_TEST(kernel_name_size)
...
+    }
+
+    if (src_size != 1
+        && (src_size != strlen(sample_kernel_code_single_line[0]) + 1
+            || memcmp(src.data(), sample_kernel_code_single_line[0], src_size)
+                != 0))
+    {
+        log_error("CL_PROGRAM_SOURCE returned unexpected content\n");
+        return TEST_FAIL;
+    }
```
rjodinchr, line 1193:
```suggestion
    if (src_size != 1
        && (src_size != strlen(sample_kernel_code_single_line[0]) + 1
            || strcmp(src.data(), sample_kernel_code_single_line[0]) != 0))
    {
        log_error(
            "CL_PROGRAM_SOURCE returned unexpected content (size %zu):\n%s\n",
            src_size, src.data());
        return TEST_FAIL;
    }
```

rjodinchr (commented):
Add `[run-test: test_compiler get_program_source_from_binary]` to your PR description

You will need to update `ci/pocl/golden.json` as well with the expected status. If pocl fails that test, it is fine to set the expected status as `fail`.

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1120,3 +1120,79 @@ REGISTER_TEST(kernel_name_size)
...
+
+    size_t src_size = 0;
+    error = clGetProgramInfo(program_from_binary, CL_PROGRAM_SOURCE, 0, nullptr,
+                             &src_size);
+    test_error(error, "clGetProgramInfo for source size failed");
+
+    if (src_size == 0)
+    {
+        log_error("Source size is zero\n");
+        return TEST_FAIL;
```
bashbaug: Is this the proper check?  The spec for [CL_PROGRAM_SOURCE](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_PROGRAM_SOURCE) says:

> If program is created using [clCreateProgramWithBinary](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#clCreateProgramWithBinary), [clCreateProgramWithIL](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#clCreateProgramWithIL), [clCreateProgramWithILKHR](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#clCreateProgramWithILKHR), or [clCreateProgramWithBuiltInKernels](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#clCreateProgramWithBuiltInKernels), a null string or the appropriate program source code is returned depending on whether or not the program source code is stored in the binary.

Is "a null string" intended to be a valid string consisting of the null terminator only, or can it also return zero to indicate there are no characters at all?

Note, this is the only place in the spec where we use the term "null string".  Other places in the spec use the term "empty string".

FWIW, both the Intel CPU and Intel GPU implementations currently return a size of zero (no characters) for CL_PROGRAM_SOURCE when the program is created from a binary.

rjodinchr: I'm not sure how "null string" should be understood, but I would be fine supporting both a string with a null size or a string with only the null terminator.

rjodinchr:
PoCL tests were not run because of the leading space before `[run-test:...]`. I've removed it to trigger the PoCL tests.
I expect it to fail because the kernel inside `sample_kernel_code_single_line` is named `sample_test`, not `test`. It should cause `clCreateKernel` within the helper to fail with `CL_INVALID_KERNEL_NAME`.

funch0za:
@rjodinchr @bashbaug Could you please take another look at this PR? Thanks!

test_conformance/compiler/test_build_helpers.cpp
```diff
@@ -1120,3 +1120,77 @@ REGISTER_TEST(kernel_name_size)
...
+REGISTER_TEST(get_program_source_from_binary)
+{
+    cl_int error = CL_SUCCESS;
+    clProgramWrapper program;
+    clKernelWrapper kernel;
+    clProgramWrapper program_from_binary;
+
+    error = create_single_kernel_helper(context, &program, &kernel, 1,
+                                        sample_kernel_code_single_line,
+                                        "sample_test");
```
rjodinchr, line 1133:
```suggestion
    clProgramWrapper program_from_binary;

    error = create_single_kernel_helper(context, &program, nullptr, 1,
                                        sample_kernel_code_single_line,
                                        nullptr);
```
As we do not need the kernel, we can just not create it at all.

PR #2785, opened 2026-08-31
Add test for maximum number of read_write image arguments in a kernel

shajder:
Related to #594 and [this](https://github.com/KhronosGroup/OpenCL-CTS/pull/2632#discussion_r3882168173) discussion

Adds test_api min_max_read_write_image_args, covering CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS the same way the existing tests cover the read_only and write_only limits.

[run-test: test_api min_max_read_write_image_args]


test_conformance/api/test_api_min_max.cpp
```diff
@@ -558,6 +566,162 @@ REGISTER_TEST(min_max_write_image_args)
...
+    imageFormatDesc.image_channel_data_type = CL_FLOAT;
+
+    if (maxReadWriteImageArgs == 0)
+    {
+        log_info("WARNING: Device reports 0 for a max read/write image arg "
+                 "count (read/write image arguments unsupported). Skipping "
+                 "test (implicitly passes). This is only valid if the number "
+                 "of image formats is also 0.\n");
+        return TEST_SKIPPED_ITSELF;
+    }
```
```suggestion
    if (get_device_cl_version(device) >= Version(3, 0) && maxReadWriteImageArgs == 0)
    {
        log_info("Device does not support read-write images. Skipping test.\n");
        return TEST_SKIPPED_ITSELF;
    }
```
Read-write images are mandatory in OpenCL 2.x when images are supported, but optional in OpenCL 3.0. Skipping the test when `maxReadWriteImageArgs == 0` should therefore be gated by `get_device_cl_version(device) >= Version(3, 0)`, so that non-conformant 2.x devices fail rather than skip.

test_conformance/api/test_api_min_max.cpp
```diff
@@ -28,6 +28,14 @@ const char *sample_single_param_kernel[] = {
     "}\n"
 };
 
+const char *sample_read_write_image_kernel_pattern[] = {
```
```suggestion
static const char *sample_read_write_image_kernel_pattern[] = {
```

test_conformance/api/test_api_min_max.cpp
```diff
@@ -558,6 +566,162 @@ REGISTER_TEST(min_max_write_image_args)
...
+                 "count (read/write image arguments unsupported). Skipping "
+                 "test (implicitly passes). This is only valid if the number "
+                 "of image formats is also 0.\n");
+        return TEST_SKIPPED_ITSELF;
+    }
+
+    if (maxReadWriteImageArgs < minRequiredReadWriteImages)
+    {
+        log_error("ERROR: Reported max read/write image arg count is less than "
+                  "required! (%d)\n",
```
```suggestion
                  "required! (%u)\n",
```

rjodinchr (commented):
Could you trigger a pocl run for it please?

Also I understand that this is mainly a copy/paste from another test, but keeping the declaration of `i` at function scope instead of inside the loop, and other similar things could be modernize as this is a new test.

test_conformance/api/test_api_min_max.cpp
shajder: good find, corrected

test_conformance/api/test_api_min_max.cpp
shajder: Done, though it looks odd next to the other non-static literals. Should we put all of them in an anonymous namespace?

test_conformance/api/test_api_min_max.cpp
shajder: Done

shajder:
> Also I understand that this is mainly a copy/paste from another test, but keeping the declaration of `i` at function scope instead of inside the loop, and other similar things could be modernize as this is a new test.

Fixed the loop counters. I kept the rest matching the surrounding code, since a single test written differently from the other 2000 lines in the file is harder to follow than a consistent one. Happy to modernize the whole file if that is something we want, but that feels like a separate PR.

test_conformance/api/test_api_min_max.cpp
rjodinchr: I would say no.
If we want to clean up that code, let's have a PR dedicated for that.

rjodinchr (dismissed):

bashbaug:
Adding "focused review". Note, also has a merge conflict that needs to be resolved.

shajder:
close/open to retrigger CI

rjodinchr (approved):

PR #2786, opened 2026-08-31
vectors: add fp16 coverage

svenvh:

### DIFF
--- a/test_conformance/vectors/test_vec_align.cpp
+++ b/test_conformance/vectors/test_vec_align.cpp
@@ -421,16 +421,6 @@ cl_int test_vec_thread(cl_uint job_id, cl_uint thread_id, void* userInfo)
         typeMultiplePostSize, info->supports_fp64, info->supports_fp16);
 }
 
-bool supports_fp64(cl_device_id device)
-{
-    return is_extension_available(device, "cl_khr_fp64");
-}
-
-bool supports_fp16(cl_device_id device)
-{
-    return is_extension_available(device, "cl_khr_fp16");
-}
-
 // there hsould be a packed version of this?
 REGISTER_TEST(vec_align_array)
 {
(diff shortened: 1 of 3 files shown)


test_conformance/vectors/test_vec_align.cpp
```diff
@@ -400,17 +415,22 @@ cl_int test_vec_thread(cl_uint job_id, cl_uint thread_id, void* userInfo)
...
+        info->bufSize, preSize, typeMultiplePreSize, postSize,
+        typeMultiplePostSize, info->supports_fp64, info->supports_fp16);
 }
 
 bool supports_fp64(cl_device_id device)
 {
     return is_extension_available(device, "cl_khr_fp64");
 }
 
+bool supports_fp16(cl_device_id device)
```
rjodinchr: We could use `device_supports_half` from the harness instead of introducing that function

rjodinchr (dismissed):

test_conformance/vectors/test_vec_align.cpp
bashbaug: Noting for completeness: there's a `device_supports_double` in the harness, also.

bashbaug (dismissed):
Needs to update the PoCL "golden" results.

Testing fp16 is absolutely the right thing to do, but it does increase test time by ~10% to test the additional type.  Since this test compiles ~10K kernels it takes a while to run, even on a multi-core system. This suite would be a good candidate for optimization in the future.

test_conformance/vectors/test_vec_align.cpp
svenvh: Thanks, I've used the `device_supports_` variants for both now.

svenvh:
> it does increase test time by ~10% to test the additional type. Since this test compiles ~10K kernels it takes a while to run, even on a multi-core system. This suite would be a good candidate for optimization in the future.

I have some ideas to speed up this suite; I will followup with some PRs.  Update: I've opened #2788 and #2789 for the low-hanging fruit.

rjodinchr (approved):

bashbaug:

PR #2788, opened 2026-09-01
vectors: reduce test execution time by shrinking NDRanges

svenvh:
The result of `vec_step` depends only on its argument type.  Running 1024 work-items just repeats the same check, so use one work-item and shrink the buffers accordingly.

The alignment tests were launching up to 524288 work-items for each kernel, which resulted in a lot of repetition of the same compile-time layouts without adding meaningful coverage.  Limit NDRanges and shrink the buffers accordingly.

### DIFF
--- a/test_conformance/vectors/defines.h
+++ b/test_conformance/vectors/defines.h
@@ -30,10 +30,6 @@ extern const int g_arrVecSteps[NUM_VECTOR_SIZES];
 extern const char* const g_arrVecSizeNames[NUM_VECTOR_SIZES];
 extern const size_t g_arrVecAlignMasks[NUM_VECTOR_SIZES];
 
-// Define the buffer size that we want to block our test with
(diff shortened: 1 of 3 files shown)


rjodinchr (approved):
LGTM! 🥳🙏

bashbaug (approved):
FWIW, this set of changes didn't significantly improve test execution time on the device I tested (with "wimpy" mode), but they didn't hurt anything either, and they're good changes that I definitely agree could improve test execution on some devices.

svenvh:
> FWIW, this set of changes didn't significantly improve test execution time on the device I tested (with "wimpy" mode), but they didn't hurt anything either, and they're good changes that I definitely agree could improve test execution on some devices.

Thanks for trying!  The PoCL CI run also didn't see any significant reduction, but a local emulated run showed a significant reduction.

bashbaug:

PR #2789, opened 2026-09-01
vectors: reduce test execution time for vec_align_*struct

svenvh:
The vec_align_struct and vec_align_packed_struct tests currently exercise every combination of six preceding and six trailing member layouts.  This commit prunes the trailing cases to avoid up to 5400 kernel builds without losing meaningful coverage.

Trailing members cannot affect an earlier packed member's offset, so only keep the empty trailing-member case for vec_align_packed_struct. For vec_align_struct, keep the empty case and the trailing-vector case, since a trailing vector can increase the alignment of the structure as a whole.

### DIFF
--- a/test_conformance/vectors/test_vec_align.cpp
+++ b/test_conformance/vectors/test_vec_align.cpp
@@ -382,6 +382,20 @@ static const std::array<size_t, ARR_SIZE> type_multiple_post_align_arr = {
     0, 0, 3, 5, 4, 12
 };
 
+// Array-of-struct tests need every trailing-member layout because it can affect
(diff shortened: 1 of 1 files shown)


rjodinchr (dismissed):
LGTM! 🥳🙏

bashbaug (dismissed):
This reduced test execution time from ~5 minutes to a bit less than 3 minutes on the device I tested - nice!

bashbaug:
Discussed in the September 8th teleconference. OK to merge when the conflict is resolved.

rjodinchr (approved):

PR #2793, opened 2026-09-03
relationals: improve any/all input patterns

svenvh:
The any/all tests were using random input data.  This gave poor test coverage: `all()` almost always returned `false` unless the input had all MSBs set which becomes less likely as the vector width increases. Conversely, `any()` almost always returned `true`.

Fill the first few input buffer entries with directed patterns covering all-clear, all-set, alternating, and some isolated set or clear MSBs.

### DIFF
--- a/test_conformance/relationals/test_relationals.cpp
+++ b/test_conformance/relationals/test_relationals.cpp
@@ -19,8 +19,6 @@
 #include "harness/testHarness.h"
 
 #include <array>
-#include <cassert>
-#include <cstdlib>
 
 // clang-format off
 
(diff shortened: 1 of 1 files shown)


test_conformance/relationals/test_relationals.cpp
```diff
@@ -46,6 +50,78 @@ const char *anyAllTestKernelPatternVload =
...
+                                         vectorStride);
+            break;
+        case kLong:
+            write_any_all_patterns_typed((cl_long *)inData, vecSize,
+                                         vectorStride);
+            break;
+        default:
+            assert(false
+                   && "Unsupported element type for any/all input patterns");
+            abort();
```
rjodinchr: I would prefer to have a proper error propagation, something like:

```suggestion
            log_error("Unsupported element type for any/all input patterns: %s\n",
                      get_explicit_type_name(elementType));
            return false;
```
And we would need to update the places where that function is called.

svenvh: Sure, applied.

rjodinchr (approved):

bashbaug:

PR #2795, opened 2026-09-07
api: Fix min_max_mem_alloc_size minimum requirement check

rjodinchr:
Previously, get_device_info_max_mem_alloc_size was called with MAX_DEVICE_MEMORY_SIZE_DIVISOR (2), halving the reported size before checking whether it satisfied requiredAllocSize. For compliant devices reporting the exact specification minimum (e.g. 128 MB on full profile), this caused the test to falsely fail with:
"ERROR: Reported max allocation size is less than required"

Fix this by querying the reported device limits without a divisor, and then applying MAX_DEVICE_MEMORY_SIZE_DIVISOR to compute maxAllocToAttempt so that the buffer allocation retry loop preserves its original behavior.

Also add a missing newline to the log_error call.

[run-test: test_api min_max_mem_alloc_size]

### DIFF
--- a/ci/pocl/golden.json
+++ b/ci/pocl/golden.json
@@ -11,6 +11,7 @@
         "": {
           "consistency_read_write_images": "pass",
           "min_max_image_array_size": "pass",
+          "min_max_mem_alloc_size": "pass",
           "min_max_read_image_args": "pass",
           "min_max_write_image_args": "pass",
           "negative_set_kernel_arg_invalid_image_msaa": "skip",
--- a/test_conformance/api/test_api_min_max.cpp
+++ b/test_conformance/api/test_api_min_max.cpp
@@ -564,12 +564,8 @@ REGISTER_TEST(min_max_mem_alloc_size)
     else
         requiredAllocSize = 128 * 1024 * 1024;
 
-    /* Get the max mem alloc size, limit the alloc to half of the available
-     * memory */
-    maxAllocSize = get_device_info_max_mem_alloc_size(
-        device, MAX_DEVICE_MEMORY_SIZE_DIVISOR);
-    memSize =
-        get_device_info_global_mem_size(device, MAX_DEVICE_MEMORY_SIZE_DIVISOR);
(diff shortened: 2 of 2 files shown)


shajder:
> For compliant devices reporting the exact specification minimum (e.g. 128 MB on full profile)

I don't have objections to the fix itself, it correctly removes the double division. But since 128 MB is called out here as "spec minimum", let's be precise about that number. First of all as per spec for [CL_DEVICE_MAX_MEM_ALLOC_SIZE](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_DEVICE_MAX_MEM_ALLOC_SIZE):

> Max size of memory object allocation in bytes. The minimum value is max(min(1024 × 1024 × 1024, 1/4th of [CL_DEVICE_GLOBAL_MEM_SIZE](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#CL_DEVICE_GLOBAL_MEM_SIZE)), 32 × 1024 × 1024).

That puts the actual floor at 32 not 128 MB, right? If so we should decrease requirement even more (this PR already halves the effective requirement from 256 to 128). 

Separately, `CL_DEVICE_MAX_MEM_ALLOC_SIZE` has no entry in the embedded profile relaxation table, so the same 32 MB floor applies there too (instead of current 1MB requirement), or is it a doc issue? Given both of these I'd rather hold off on merging until `requiredAllocSize` gets revisited for both profiles. FYI @bashbaug.

rjodinchr:
This is not regressing anything if I'm not mistaken, it puts the floor closer to what it should be. I don't see a good reason to hold it in your reply.
I agree that there are other things to figure out and fix as well, but in my opinion, that is not a reason to hold off on that fix.

shajder:
> I agree that there are other things to figure out and fix as well, but in my opinion, that is not a reason to hold off on that fix.

I've created separate issue #2813 to track my findings, so this PR doesn't need to be put on hold.

shajder (approved):

PR #2796, opened 2026-09-07
Fix test_vulkan_interop_image validation error for oversized image2DList descriptor array

shajder:
Related to #2487

Fixes VUID-vkCmdDispatch-None-08114, reported by VK_LAYER_KHRONOS_validation 1.4.313.

The `image2DList` descriptor array in `run_test_with_one_queue` and `run_test_with_two_queue` was sized for hardcoded 11 mipmap levels while every image in this test is created with a single mip level. Added another specialization constant to address that problem.

### DIFF
--- a/test_conformance/vulkan/test_vulkan_interop_image.cpp
+++ b/test_conformance/vulkan/test_vulkan_interop_image.cpp
@@ -266,12 +266,13 @@ int run_test_with_two_queue(
     srcBufferPtr = (char *)malloc(maxImage2DSize);
     dstBufferPtr = (char *)malloc(maxImage2DSize);
 
+    const uint32_t numMipLevels = 1;
     VulkanDescriptorSetLayoutBindingList vkDescriptorSetLayoutBindingList;
     vkDescriptorSetLayoutBindingList.addBinding(
         0, VULKAN_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1);
     vkDescriptorSetLayoutBindingList.addBinding(
         1, VULKAN_DESCRIPTOR_TYPE_STORAGE_IMAGE,
-        MAX_2D_IMAGE_MIP_LEVELS * (useSingleImageKernel ? 1 : num2DImages));
+        (useSingleImageKernel ? 1 : num2DImages) * numMipLevels);
     VulkanDescriptorSetLayout vkDescriptorSetLayout(
         vkDevice, vkDescriptorSetLayoutBindingList);
     VulkanPipelineLayout vkPipelineLayout(vkDevice, vkDescriptorSetLayout);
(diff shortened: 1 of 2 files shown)


test_conformance/vulkan/test_vulkan_interop_image.cpp
rjodinchr: Now unused, I think we should remove them

rjodinchr (commented):
Adding a specialization constant for `MAX_2D_IMAGE_MIP_LEVELS` introduces flexibility that is not actually there in the shader. The shader loop assumes all descriptors share the same dimensions (`imageDim`) and dispatch grid, so it cannot handle multiple mip levels anyway, and the test only ever runs with 1 mip level.

I think it would be better to do either of the following:

1. **Use a single specialization constant**: Size `image2DList` with a single constant representing the descriptor count (matching what `buffer.comp` does with `MAX_BUFFERS`). This avoids the second constant, the anonymous struct, and the `offsetof(decltype(...))` boilerplate entirely.
2. **If keeping both constants**: Use a plain `uint32_t specData[2]` array instead of the anonymous struct:
   ```cpp
   uint32_t specData[2] = { numImages, numMipLevels };
   VkSpecializationMapEntry entries[2] = {
       { 0, 0 * sizeof(uint32_t), sizeof(uint32_t) },
       { 1, 1 * sizeof(uint32_t), sizeof(uint32_t) },
   };

shajder:
> Adding a specialization constant for `MAX_2D_IMAGE_MIP_LEVELS` introduces flexibility that is not actually there in the shader. The shader loop assumes all descriptors share the same dimensions (`imageDim`) and dispatch grid, so it cannot handle multiple mip levels anyway, and the test only ever runs with 1 mip level.

The shader loop reads imageSize() per slot and doesn't assume shared dimensions, so it already supports a per mip level image2DList entry if numMipLevels is ever raised - VulkanImageViewList already builds one view per mip level. 

> If keeping both constants: Use a plain uint32_t specData[2]...

Done

test_conformance/vulkan/test_vulkan_interop_image.cpp
shajder: Done

rjodinchr:
Fair point about `VulkanImageViewList` and `imageSize()`. Keeping `MAX_2D_IMAGE_MIP_LEVELS` makes sense in that context, and the array simplification looks much cleaner!

One issue remains with `--useSingleImageKernel`:
The layout binding is created with `(useSingleImageKernel ? 1 : num2DImages) * numMipLevels`:
```cpp
vkDescriptorSetLayoutBindingList.addBinding(
    1, VULKAN_DESCRIPTOR_TYPE_STORAGE_IMAGE,
    (useSingleImageKernel ? 1 : num2DImages) * numMipLevels);
```
When `useSingleImageKernel` is true, the layout binding has `descriptorCount = 1`. But `specData` sets constant 0 to `num2DImages` unconditionally:
```cpp
uint32_t specData[2] = { num2DImages, numMipLevels };
```
When `num2DImages > 1`, the shader array size (`num2DImages * numMipLevels`) will exceed the pipeline layout's `descriptorCount = 1`, which may triggers a validation error on pipeline creation (I think, I don't have a setup to test it myself right now).

Could you update `specData` to account for `useSingleImageKernel`?
```cpp
const uint32_t numImages = useSingleImageKernel ? 1 : num2DImages;
uint32_t specData[2] = { numImages, numMipLevels };
```

shajder:
> One issue remains with `--useSingleImageKernel`

That's corrected. Turns out, --useSingleImageKernel has a pre-existing, unrelated data-mismatch bug with num2DImages > 1 (reproducible at my machine, first fails at r8ui/2 images) - I will create separate issue.

rjodinchr (approved):

bashbaug:

PR #2798, opened 2026-09-08
ci: update pocl to v7.2

rjodinchr:


.github/workflows/pocl.yml
```diff
@@ -120,6 +120,7 @@ jobs:
             -DCMAKE_INSTALL_PREFIX=${{ env.pocl }} \
             -DCMAKE_BUILD_TYPE=Release \
             -DKERNELLIB_HOST_CPU_VARIANTS=distro \
+            -DLLVM_DIR=/usr/lib/llvm-${LLVM_VERSION}/lib/cmake/llvm \
```
rjodinchr, line 123: Needed because of https://github.com/pocl/pocl/issues/2312

bashbaug (approved):
Nice, more passing tests 👍

rjodinchr:
7.2 is OpenCL compliant, so I think it would be interesting to know why some tests are still failing.
Is it because we are not using the OpenCL-CTS version that has been used to pass conformance? Are we missing something when building PoCL?

bashbaug:
Discussed in the September 8th teleconference. OK to merge offline.

PR #2799, opened 2026-09-08
ci: Add Rusticl support and refactor driver testing

rjodinchr:
Generalize the GitHub Actions targeted PR testing workflow to support
multiple OpenCL drivers, refactor driver setup and testing steps into
modular composite actions, and introduce Rusticl (Mesa llvmpipe) alongside
PoCL.

Key highlights:
* Multi-Driver Workflow: Renamed pocl.yml to tests.yml and introduced a
  dynamic matrix based on the DRIVERS environment variable.
* Split Workflow Jobs & Optimized Caching:
  - check-triggers: Parses commit messages and PR descriptions for test
    triggers and determines the active drivers.
  - build-and-cache: Builds and caches driver artifacts on push or when PR
    triggers are matched. Uses sparse checkout (.github) to avoid cloning
    CTS source when only building drivers. Skips runtime packages and
    test utilities during cache population.
  - run: Restores driver cache, installs only lightweight runtime
    dependencies, builds OpenCL-CTS, and runs triggered test commands.
* Centralized Configuration:
  - Pinned driver versions and branches in workflow-level environment
    variables with clean prefixes (POCL_*, RUSTICL_*, VULKAN_SDK_VERSION)
    inherited directly by composite actions.
* Modular Composite Actions:
  - .github/actions/setup-pocl: Manages PoCL & SPIRV-LLVM-Translator build,
    caching, and runtime environment configuration.
  - .github/actions/setup-rusticl: Clones Mesa, builds Rusticl with the
    llvmpipe driver, SPIRV-Tools, and SPIRV-LLVM-Translator, and configures
    the ICD search path.
  - .github/actions/build-cts: Builds OpenCL Headers, ICD Loader, and CTS.
  - .github/actions/run-cts: Executes targeted test binaries and compares
    test output against golden reference baselines.
* Driver-Specific Triggers:
  - Extended syntax to allow targeting all drivers ([run-test: <cmd>]) or
    specific drivers ([run-test-pocl: <cmd>], [run-test-rusticl: <cmd>]).
* Rusticl Baseline: Added ci/rusticl/golden.json baseline results for
  test_computeinfo and test_api enqueue_task.
* Documentation: Updated README.md with the driver-specific trigger syntax
  and examples.

[run-test: test_computeinfo]
[run-test-rusticl: test_api enqueue_task]


rjodinchr:
@karolherbst , somehow I'm not able to add you in the reviewer list, but I would appreciate your feedback here. Thank you

.github/actions/setup-rusticl/action.yml
```diff
@@ -0,0 +1,146 @@
...
+      shell: bash
+      run: |
+        source $HOME/.cargo/env 2>/dev/null || true
+        export PATH="/usr/lib/llvm-${RUSTICL_LLVM_VERSION}/bin:$HOME/.cargo/bin:$PATH"
+        export PKG_CONFIG_PATH="${{ github.workspace }}/rusticl-install/lib/pkgconfig:${{ github.workspace }}/rusticl-install/lib/x86_64-linux-gnu/pkgconfig:${{ github.workspace }}/rusticl-install/share/pkgconfig:$PKG_CONFIG_PATH"
+        export LD_LIBRARY_PATH="${{ github.workspace }}/rusticl-install/lib:${{ github.workspace }}/rusticl-install/lib/x86_64-linux-gnu:$LD_LIBRARY_PATH"
+
+        meson setup mesa/build mesa \
+          --prefix='${{ github.workspace }}/rusticl-install' \
+          --buildtype=release \
```
karolherbst, line 108: I think it might be better to use `debugoptimized` here or at least enable assertions with `-Db_ndebug=false`. This will turn on more validation in the compiler and might spot some UB or other random bugs.

rjodinchr: I don't think I want to build and run with `debugoptimized`. The goal here is not to catch `rusticl` bugs.
I think we want the build/run to be as fast as it can get, thus keeping the `release` mode.
But adding `-Db_ndebug=false` is a great tips! I'll do that, thank you

PR #2801, opened 2026-09-08
Corrected bufferreadwriterect to take into account proper row and slice pitch

### LINKED ISSUE #2765: basic bufferreadwriterect does not detect overlaps correctly.
Caused by https://github.com/KhronosGroup/OpenCL-CTS/pull/2515

Given the example:
```
CL_DEVICE_MAX_MEM_ALLOC_SIZE = 1073741824 bytes.
Using maximum dimension      = 512.
Creating 8 pairs of random sized host and cl buffers.
Buffer[0] is (107,225,86) = 1 MB (truncated)
Buffer[1] is (438,353,408) = 60 MB (truncated)
Buffer[2] is (287,81,327) = 7 MB (truncated)
Buffer[3] is (229,511,75) = 8 MB (truncated)
Buffer[4] is (328,216,108) = 7 MB (truncated)
Buffer[5] is (435,290,68) = 8 MB (truncated)
Buffer[6] is (221,17,431) = 1 MB (truncated)
Buffer[7] is (335,36,315) = 3 MB (truncated)
Total size: 98 MB (truncated)
Initializing buffers
Initializing buffer 0
Initializing buffer 1
Initializing buffer 2
Initializing buffer 3
Initializing buffer 4
Initializing buffer 5
Initializing buffer 6
Initializing buffer 7
Executing 50 test operations selected at random.
0 Read 5 offset (182,170,5) -> 7 offset (26,14,29) region (3x17x3 = 153)
1 Copy 1 offset (14,69,100) -> 1 offset (176,181,171) region (224x5x226 = 253120)
CL ERROR: /home/kherbst/git/OpenCL-CTS/test_conformance/basic/test_bufferreadwriterect.cpp 220: clEnqueueCopyBufferRect failed between 1 and 1
```
`clEnqueueCopyBufferRect` gets called with `src_row_pitch` and `dst_row_pitch` of `438` and `src_slice_pitch` and `dst_slice_pitch` of `0`.

The implicit slice_pitch is then calculated as `row_pitch * region[1]` as `2190` according to the spec:
> If src_slice_pitch is 0, src_slice_pitch is computed as region[1] × src_row_pitch.

Now the CTS complains that an implementation rejects that copy. The runtime has the correct behavior here as the copy actually overlap:
- at offset `0x0x0` of the copy in the destination we are at byte offset `176 + 181*438 + 171*2190 = 453944`
- at offset `162x2x93` of the copy in the source we are at byte offset `14+162 + (69+2)*438 + (100+93)*2190 = 453944`

so the copy does overlap, but the CTS' own overlapping detection does not detect this.

shajder:
Fixes #2765

Corrected check_overlap_rect which did not compute proper row and slice pitch. As result some implementations properly reported `CL_MEM_COPY_OVERLAP`.

[run-test: test_basic bufferreadwriterect]
[run-test: test_basic immutable_bufferreadwriterect]

### DIFF
--- a/test_conformance/basic/test_bufferreadwriterect.cpp
+++ b/test_conformance/basic/test_bufferreadwriterect.cpp
@@ -113,12 +113,18 @@ void print_buffer(BufferType* buf, size_t w, size_t h, size_t d) {
 
 // Returns true if the two specified regions overlap.
 bool check_overlap_rect(size_t src_offset[3], size_t dst_offset[3],
-                        size_t region[3], size_t src)
+                        size_t region[3], size_t src_row_pitch,
+                        size_t src_slice_pitch, size_t dst_row_pitch,
+                        size_t dst_slice_pitch)
 {
-    // Copy between cl buffers.
-    size_t slice_pitch =
-        (width[src] * height[src] != 1) ? width[src] * height[src] : 0;
-    size_t row_pitch = width[src];
+    size_t src_effective_rpitch = effective_row_pitch(src_row_pitch, region[0]);
+    size_t src_effective_spitch = (src_slice_pitch == 0)
+        ? region[1] * src_effective_rpitch
+        : src_slice_pitch;
+    size_t dst_effective_rpitch = effective_row_pitch(dst_row_pitch, region[0]);
+    size_t dst_effective_spitch = (dst_slice_pitch == 0)
+        ? region[1] * dst_effective_rpitch
+        : dst_slice_pitch;
 
     const size_t src_min[] = { src_offset[0], src_offset[1], src_offset[2] };
     const size_t src_max[] = { src_offset[0] + region[0],
@@ -138,21 +144,23 @@ bool check_overlap_rect(size_t src_offset[3], size_t dst_offset[3],
             overlap && (src_min[i] < dst_max[i]) && (src_max[i] > dst_min[i]);
     }
 
-    size_t dst_start =
-        dst_offset[2] * slice_pitch + dst_offset[1] * row_pitch + dst_offset[0];
+    size_t dst_start = dst_offset[2] * dst_effective_spitch
+        + dst_offset[1] * dst_effective_rpitch + dst_offset[0];
     size_t dst_end = dst_start
-        + (region[2] * slice_pitch + region[1] * row_pitch + region[0]);
-    size_t src_start =
-        src_offset[2] * slice_pitch + src_offset[1] * row_pitch + src_offset[0];
+        + (region[2] * dst_effective_spitch + region[1] * dst_effective_rpitch
+           + region[0]);
+    size_t src_start = src_offset[2] * src_effective_spitch
+        + src_offset[1] * src_effective_rpitch + src_offset[0];
     size_t src_end = src_start
-        + (region[2] * slice_pitch + region[1] * row_pitch + region[0]);
+        + (region[2] * src_effective_spitch + region[1] * src_effective_rpitch
+           + region[0]);
     if (!overlap)
     {
-        size_t delta_src_x = (src_offset[0] + region[0] > row_pitch)
-            ? src_offset[0] + region[0] - row_pitch
+        size_t delta_src_x = (src_offset[0] + region[0] > src_effective_rpitch)
+            ? src_offset[0] + region[0] - src_effective_rpitch
             : 0;
-        size_t delta_dst_x = (dst_offset[0] + region[0] > row_pitch)
-            ? dst_offset[0] + region[0] - row_pitch
+        size_t delta_dst_x = (dst_offset[0] + region[0] > dst_effective_rpitch)
+            ? dst_offset[0] + region[0] - dst_effective_rpitch
             : 0;
         if ((delta_src_x > 0 && delta_src_x > dst_offset[0])
             || (delta_dst_x > 0 && delta_dst_x > src_offset[0]))
@@ -163,8 +171,8 @@ bool check_overlap_rect(size_t src_offset[3], size_t dst_offset[3],
         }
         if (region[2] > 1)
         {
-            size_t src_height = slice_pitch / row_pitch;
-            size_t dst_height = slice_pitch / row_pitch;
+            size_t src_height = src_effective_spitch / src_effective_rpitch;
+            size_t dst_height = dst_effective_spitch / dst_effective_rpitch;
             size_t delta_src_y = (src_offset[1] + region[1] > src_height)
                 ? src_offset[1] + region[1] - src_height
                 : 0;
@@ -203,7 +211,8 @@ int copy_region(size_t src, size_t soffset[3], size_t sregion[3], size_t dst,
     }
 
     cl_int err;
-    if (check_overlap_rect(soffset, doffset, sregion, src))
+    if (check_overlap_rect(soffset, doffset, sregion, src_row_pitch,
+                           src_slice_pitch, dst_row_pitch, dst_slice_pitch))
     {
         log_info("Copy overlap reported, skipping copy buffer rect\n");
         return CL_SUCCESS;
@@ -266,7 +275,8 @@ int immutable_copy_region(size_t src, size_t soffset[3], size_t sregion[3],
         (width[dst] * height[dst] != 1) ? width[dst] * height[dst] : 0;
 
     cl_int err;
-    if (check_overlap_rect(soffset, doffset, sregion, src))
+    if (check_overlap_rect(soffset, doffset, sregion, width[src],
+                           src_slice_pitch, width[dst], dst_slice_pitch))
     {
         log_info("Copy overlap reported, skipping copy buffer rect\n");
         return CL_SUCCESS;


test_conformance/basic/test_bufferreadwriterect.cpp
```diff
@@ -203,7 +211,8 @@ int copy_region(size_t src, size_t soffset[3], size_t sregion[3], size_t dst,
     }
 
     cl_int err;
-    if (check_overlap_rect(soffset, doffset, sregion, src))
+    if (check_overlap_rect(soffset, doffset, sregion, src_row_pitch,
+                           src_slice_pitch, dst_row_pitch, dst_slice_pitch))
```
rjodinchr, line 215: We only need to check the overlap is `src == dst`, otherwise there cannot be any overlap by definition, right?
Also, why add `{src|dst}_{row|slice}_pitch`? when `src == dst` they are equal due to the statement right above. We should be able to drop at least half of them I think.

rjodinchr (changes requested):
I've asked an AI about the `check_overlap_rect` function, I think its feedback is relevant and should be addressed as well:

### High rate of False Positives (~15% of operations skipped)

The heuristic in `if (!overlap)` attempts to detect overlap by checking if 1D bounding intervals `[src_start, src_end)` and `[dst_start, dst_end)` overlap:

```cpp
if (region[2] > 1)
{
    size_t src_height = src_effective_spitch / src_effective_rpitch;
    size_t dst_height = dst_effective_spitch / dst_effective_rpitch;
    size_t delta_src_y = (src_offset[1] + region[1] > src_height)
        ? src_offset[1] + region[1] - src_height
        : 0;
    size_t delta_dst_y = (dst_offset[1] + region[1] > dst_height)
        ? dst_offset[1] + region[1] - dst_height
        : 0;
    if ((delta_src_y > 0 && delta_src_y > dst_offset[1])
        || (delta_dst_y > 0 && delta_dst_y > src_offset[1]))
    {
        if ((src_start <= dst_start && dst_start < src_end)
            || (dst_start <= src_start && src_start < dst_end))
            overlap = true;
    }
}
```

* When `src_slice_pitch == 0`, `src_effective_spitch = region[1] * src_effective_rpitch`. Therefore, `src_height = region[1]`.
* This causes `delta_src_y = src_offset[1]` and `delta_dst_y = dst_offset[1]`.
* Consequently, `(delta_src_y > dst_offset[1] || delta_dst_y > src_offset[1])` evaluates to `true` whenever `src_offset[1] != dst_offset[1]` (almost always in random testing).
* When this condition is met, the code assumes that any overlap between the 1D spans `[src_start, src_end)` and `[dst_start, dst_end)` is a 3D rectangular overlap.
* **Counterexample**:
  - `region = (10, 5, 2)`
  - `src_offset = (0, 0, 0)`
  - `dst_offset = (500, 1, 0)`
  - `row_pitch = 1000`, `slice_pitch = 0` (`effective_slice_pitch = 5000`)
  - On every row, `src` occupies bytes `[0, 10)` and `dst` occupies bytes `[500, 510)`. They are separated by 490 bytes on every row and do not overlap at all.
  - `check_overlap_rect` computes `src_start = 0`, `src_end = 15010`, `dst_start = 1500`. Since `0 <= 1500 < 15010`, it sets `overlap = true` and skips the test.
* Across randomized testing simulating this test's distributions, ~14.8% of valid copies are falsely declared as overlapping and skipped.

---

### Off-by-one / overshoot in `src_end` and `dst_end`

In lines 147–156:

```cpp
size_t dst_end = dst_start
    + (region[2] * dst_effective_spitch + region[1] * dst_effective_rpitch
       + region[0]);
```

* The last byte in `dst` is at coordinate `(region[0] - 1, region[1] - 1, region[2] - 1)`:
  ```
  last_byte = dst_start + (region[2] - 1) * spitch + (region[1] - 1) * rpitch + (region[0] - 1)
  ```
* The half-open range upper bound `dst_end` should be:
  ```
  dst_start + (region[2] - 1) * spitch + (region[1] - 1) * rpitch + region[0]
  ```
* Adding `region[2] * spitch + region[1] * rpitch` overshoots the actual region by an extra slice and an extra row (`spitch + rpitch`). For a 1x1x1 region, it adds `spitch + rpitch + 1` instead of `1`.

shajder:
> I've asked an AI about the `check_overlap_rect` function, I think its feedback is relevant and should be addressed as well:

I went through the change history to understand this function's logic, which has stayed in roughly the same form since the start of CTS. It turns out the spec provides a [reference implementation](https://registry.khronos.org/OpenCL/specs/unified/html/OpenCL_API.html#check-copy-overlap) for exactly this case, so we should use it here, for clarity. I believe it fixes false positives your analisis refers to.

test_conformance/basic/test_bufferreadwriterect.cpp
shajder: Indeed. I wonder if there was a reason it's been implemented this way since the start (no clue). Secured it on `src == dst` and dropped `dst_row_pitch`/`dst_slice_pitch` as you suggested.

shajder:
re-trigger CI/CD

test_conformance/basic/test_bufferreadwriterect.cpp
```diff
@@ -111,76 +111,61 @@ void print_buffer(BufferType* buf, size_t w, size_t h, size_t d) {
 }
 #endif
 
-// Returns true if the two specified regions overlap.
+// Returns true if the two specified regions overlap. This is adapted version of
```
rjodinchr:
```suggestion
// Returns true if the two specified regions overlap. This is an adapted version of
```

shajder: Done.

As for failed CI/CD PoCL tests, I've reproduced the same numbers on both NVIDIA and Intel; op 36 passes on both, and `check_overlap_rect` returns the correct answer for every possible pitch draw for that operation. Looks like a race condition in PoCL's CPU driver, not a regression here.

rjodinchr: Then, let's update the expectation

rjodinchr (commented):
LGTM, I would like someone else to have a look, as this is a complex change.

anun333:
**Corrected:** the first version of this comment had point 2 backwards. `check_overlap_rect` *is* the current specification's appendix algorithm; PoCL's is the outdated one. The fault for the remaining failure is PoCL's, and the only change I'd suggest here is point 1.

Tested this on PoCL (upstream `main` at `aaaa2d67b`, CPU device, distro kernel-library variants with the `sse2` one selected; one build with `ENABLE_CONFORMANCE=ON`, one without). With this PR at `c116b81`, `test_basic bufferreadwriterect` still fails on both:

```
clEnqueueCopyBufferRect failed between 2 and 2    -> CL_MEM_COPY_OVERLAP
src_origin (11,277,388)  dst_origin (187,87,50)  region (6,97,845)
row pitch 0 (both), slice pitch 7620 (both)
```

1. **Zero pitches (a suggestion for this PR).** A row pitch of 0 means `region[0]` and a slice pitch of 0 means `region[1] * row_pitch`, which is what an implementation checks with. `check_overlap_rect` is given the 0 as is:

   ```cpp
   const size_t eff_row_pitch = src_row_pitch ? src_row_pitch : sregion[0];
   const size_t eff_slice_pitch =
       src_slice_pitch ? src_slice_pitch : sregion[1] * eff_row_pitch;
   ```

2. **The rest is PoCL's bug, not this PR's.** The current appendix code ("Checking for Memory Copy Overlap", the gap-based version `check_overlap_rect` follows) says this call does **not** overlap. PoCL's `check_copy_overlap` (`lib/CL/pocl_util.c`) is an older version of that appendix, copied in 2014 with a 2011 Khronos header. Its linear-range fallback reports an overlap here that the current algorithm, and the actual bytes, don't have. Fixed on the PoCL side in pocl/pocl#2336. It isn't a race: the same operation fails on every run (op 36 in both of this PR's PoCL CI runs). For op 36 under either pitch draw that this PR's check lets through, the regions share no bytes; PoCL's old check reports an overlap anyway, and the current algorithm doesn't. So the `bufferreadwriterect: fail` entry in `ci/pocl/golden.json` should become `pass` once the CI's PoCL includes pocl/pocl#2336.

Investigated with the assistance of Claude Code; the results above come from running the test and the appendix code.

shajder:
Thanks, it confirms the PR fixes overlap computations, both the missed cases (the original regression) and the false ones.

> 1. **Zero pitches (a suggestion for this PR).** A row pitch of 0 means `region[0]` and a slice pitch of 0 means `region[1] * row_pitch`, which is what an implementation checks with. `check_overlap_rect` is given the 0 as is:
>    ```c++
>    const size_t eff_row_pitch = src_row_pitch ? src_row_pitch : sregion[0];
>    const size_t eff_slice_pitch =
>        src_slice_pitch ? src_slice_pitch : sregion[1] * eff_row_pitch;
>    ```

Already applied by this PR - take a look at `effective_row_pitch` (plus the ternary right below it), it does exactly that.

FYI: @karolherbst

karolherbst:
> Thanks, it confirms the PR fixes overlap computations, both the missed cases (the original regression) and the false ones.
> 
> > 1. **Zero pitches (a suggestion for this PR).** A row pitch of 0 means `region[0]` and a slice pitch of 0 means `region[1] * row_pitch`, which is what an implementation checks with. `check_overlap_rect` is given the 0 as is:
> >    ```c++
> >    const size_t eff_row_pitch = src_row_pitch ? src_row_pitch : sregion[0];
> >    const size_t eff_slice_pitch =
> >        src_slice_pitch ? src_slice_pitch : sregion[1] * eff_row_pitch;
> >    ```
> 
> Already applied by this PR - take a look at `effective_row_pitch` (plus the ternary right below it), it does exactly that.
> 
> FYI: @karolherbst

Yeah I think using the reference implementation is a good idea. It doesn't take sub-buffers into account, but that's trivially to add. However, I also use code based on the reference implementation in rusticl.

I'll give it a try later.

anun333:
You're right, thanks. `check_overlap_rect` resolves zero pitches itself; I read the call site and not the function. I've struck that point from my comment. The remaining PoCL failure is PoCL's outdated check, and pocl/pocl#2336 (now approved) fixes it.

@karolherbst, on sub-buffers: when the two sub-buffers use the same row and slice pitch, adding each one's offset within the parent to `origin[0]` makes the reference algorithm exact (checked against the actual bytes on 250,000 random equal-pitch copies). When the pitches differ it no longer applies; I've posted an exact check for that case on KhronosGroup/OpenCL-Docs#887, verified the same way.

PR #2802, opened 2026-09-08
Add support for svm types without single address space to cl_khr_unified_svm test

joselopeqti:
- Adds the mechanisms to read and write from a Unified SVM allocation when it does not support `CL_SVM_CAPABILITY_SINGLE_ADDRESS_SPACE_KHR`.
- Fixes `CL_SVM_CAPABILITY_DEVICE_WRITE_KHR` test calling `test_CL_SVM_CAPABILITY_DEVICE_READ_KHR` instead.
- Adds test for `clEnqueueSVMMemcpyWithPropertiesKHR`.
- Adds missing `clFinish` in `test_SVMMemfill` test.

### DIFF
--- a/test_conformance/SVM/test_unified_svm_capabilities.cpp
+++ b/test_conformance/SVM/test_unified_svm_capabilities.cpp
@@ -291,8 +291,7 @@ struct UnifiedSVMCapabilities : UnifiedSVMBase
         test_error(err, "could not write to usvm memory");
 
         cl_int check;
-        err = clEnqueueSVMMemcpy(queue, CL_TRUE, &check, mem->get_ptr(),
-                                 sizeof(value), 0, nullptr, nullptr);
+        err = mem->read(check);
         test_error(err, "could not read from usvm memory with memcpy");
 
         test_assert_error(check == value,
@@ -351,6 +350,9 @@ struct UnifiedSVMCapabilities : UnifiedSVMBase
                                   sizeof(value), 0, nullptr, nullptr);
         test_error(err, "could not write to usvm memory with memfill");
 
+        err = clFinish(queue);
+        test_error(err, "clFinish failed");
+
         cl_int check;
         err = mem->read(check);
         test_error(err, "could not read from usvm memory");
@@ -359,17 +361,20 @@ struct UnifiedSVMCapabilities : UnifiedSVMBase
                           "read value with memfill does not match");
 
         // test writing via memcpy
-        value = genrand_int32(d);
-        err = clEnqueueSVMMemcpy(queue, CL_TRUE, mem->get_ptr(), &value,
-                                 sizeof(value), 0, nullptr, nullptr);
-        test_error(err, "could not write to usvm memory with memcpy");
+        if (mem->getCapabilities() & CL_SVM_CAPABILITY_SINGLE_ADDRESS_SPACE_KHR)
+        {
+            value = genrand_int32(d);
+            err = clEnqueueSVMMemcpyWithPropertiesKHR(
+                queue, NULL, CL_TRUE, mem->get_ptr(), &value, sizeof(value), 0,
+                nullptr, nullptr);
+            test_error(err, "could not write to usvm memory with memcpy");
 
-        err = mem->read(check);
-        test_error(err, "could not read from usvm memory");
+            err = mem->read(check);
+            test_error(err, "could not read from usvm memory");
 
-        test_assert_error(check == value,
-                          "read value with memcpy does not match");
-
+            test_assert_error(check == value,
+                              "read value with memcpy does not match");
+        }
         // test writing via kernel
         value = genrand_int32(d);
         clMemWrapper in =
@@ -586,7 +591,7 @@ struct UnifiedSVMCapabilities : UnifiedSVMBase
             if (caps & CL_SVM_CAPABILITY_DEVICE_WRITE_KHR)
             {
                 log_info("     testing CL_SVM_CAPABILITY_DEVICE_WRITE\n");
-                err = test_CL_SVM_CAPABILITY_DEVICE_READ_KHR(ti);
+                err = test_CL_SVM_CAPABILITY_DEVICE_WRITE_KHR(ti);
                 test_error(err, "CL_SVM_CAPABILITY_DEVICE_READ failed");
             }
             if (caps & CL_SVM_CAPABILITY_DEVICE_ATOMIC_ACCESS_KHR)
--- a/test_conformance/SVM/test_unified_svm_mem_cpy.cpp
+++ b/test_conformance/SVM/test_unified_svm_mem_cpy.cpp
@@ -27,7 +27,7 @@ struct UnifiedSVMOPs : UnifiedSVMBase
     // Test the clEnqueueSVMMemcpy function for random ranges
     // of a USM allocation and validate the results.
     cl_int test_SVMMemcpy(USVMWrapper<cl_uchar> *src,
-                          USVMWrapper<cl_uchar> *dst)
+                          USVMWrapper<cl_uchar> *dst, bool use_new_api)
     {
         cl_int err = CL_SUCCESS;
 
@@ -54,9 +54,19 @@ struct UnifiedSVMOPs : UnifiedSVMBase
             void *dst_ptr = &dst->get_ptr()[offset];
 
             clEventWrapper event;
-            err = clEnqueueSVMMemcpy(queue, CL_BLOCKING, dst_ptr, src_ptr,
-                                     length, 0, nullptr, &event);
-            test_error(err, "clEnqueueSVMMemcpy failed");
+            if (!use_new_api)
+            {
+                err = clEnqueueSVMMemcpy(queue, CL_BLOCKING, dst_ptr, src_ptr,
+                                         length, 0, nullptr, &event);
+                test_error(err, "clEnqueueSVMMemcpy failed");
+            }
+            else
+            {
+                err = clEnqueueSVMMemcpyWithPropertiesKHR(
+                    queue, NULL, CL_BLOCKING, dst_ptr, src_ptr, length, 0,
+                    nullptr, &event);
+                test_error(err, "clEnqueueSVMMemcpyWithPropertiesKHR failed");
+            }
 
             err = check_event_type(event, CL_COMMAND_SVM_MEMCPY);
             test_error(err,
@@ -95,7 +105,8 @@ struct UnifiedSVMOPs : UnifiedSVMBase
         return CL_SUCCESS;
     }
 
-    cl_int test_svm_memcpy(cl_uint srcTypeIndex, cl_uint dstTypeIndex)
+    cl_int test_svm_memcpy(cl_uint srcTypeIndex, cl_uint dstTypeIndex,
+                           bool use_new_api = false)
     {
         cl_int err;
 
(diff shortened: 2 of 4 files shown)


test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -22,6 +22,21 @@
 constexpr cl_bitfield PSEUDO_CAPABILITY_USE_SYSTEM_ALLOCATOR =
     ((cl_bitfield)1 << 63);
 
+typedef cl_properties cl_svm_memcpy_properties_khr;
+
+// clang-format off
+typedef cl_int CL_API_CALL clEnqueueSVMMemcpyWithPropertiesKHR_t(
```
bashbaug, line 28: I added support for clEnqueueSVMMemcpyWithPropertiesKHR to the unified SVM headers already, so we shouldn't need to add it here.

See: https://github.com/KhronosGroup/OpenCL-Headers/commit/485178b6095bbfc4bb853201b82a7ea428859838

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -184,11 +207,34 @@ template <typename T> class USVMWrapper {
...
+                0, nullptr, nullptr);
+            test_error(err, "clEnqueueSVMMemcpyWithPropertiesKHR failed");
+        }
         else if (caps & CL_SVM_CAPABILITY_DEVICE_WRITE_KHR)
         {
-            err = clEnqueueSVMMemcpy(queue, CL_TRUE, data + offset, source,
-                                     count * sizeof(T), 0, nullptr, nullptr);
-            test_error(err, "clEnqueueSVMMemcpy failed");
+            size_t char_count = sizeof(T);
+            cl_mem source_obj =
```
bashbaug, line 221: This should use a clMemWrapper, or some other mechanism to safely release this memory object.

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -184,11 +207,34 @@ template <typename T> class USVMWrapper {
...
+                clCreateBuffer(context, CL_MEM_READ_ONLY | CL_MEM_USE_HOST_PTR,
+                               count * char_count * sizeof(unsigned char),
```
bashbaug, line 223: Can this just use:

```suggestion
                               count * sizeof(T),
```

This is what all of the other paths are doing.

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -237,11 +283,37 @@ template <typename T> class USVMWrapper {
...
+                nullptr, nullptr);
+            test_error(err, "clEnqueueSVMMemcpyWithPropertiesKHR failed");
+        }
         else if (caps & CL_SVM_CAPABILITY_DEVICE_READ_KHR)
         {
-            err = clEnqueueSVMMemcpy(queue, CL_TRUE, dst, data + offset,
-                                     count * sizeof(T), 0, nullptr, nullptr);
-            test_error(err, "clEnqueueSVMMemcpy failed");
+            size_t char_count = sizeof(T);
+            cl_mem dst_obj = clCreateBuffer(
```
bashbaug, line 297: This should use a clMemWrapper or some other mechanism to release the memory object, also.

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -406,16 +511,19 @@ struct UnifiedSVMBase
...
+            clGetSVMPointerInfoKHR, clGetSVMSuggestedTypeIndexKHR,
+            clEnqueueSVMMemcpyWithPropertiesKHR));
     }
 
     MTdataHolder d;
     cl_context context = nullptr;
     cl_device_id device = nullptr;
     cl_command_queue queue = nullptr;
+    cl_kernel kernel_WriteOp = nullptr;
+    cl_kernel kernel_ReadOp = nullptr;
```
bashbaug, line 526: This should use a wrapper class or some other mechanism to delete these kernel objects.

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -290,6 +366,27 @@ struct UnifiedSVMBase
...
+        )";
+
+        clProgramWrapper program;
+        err = create_single_kernel_helper(context, &program, &kernel_WriteOp, 1,
+                                          &programString, "test_WriteOp");
+        test_error(err, "could not create WriteOp kernel");
+
+        err = create_single_kernel_helper(context, &program, &kernel_ReadOp, 1,
+                                          &programString, "test_ReadOp");
+        test_error(err, "could not create ReadOp kernel");
```
bashbaug, line 388: `create_single_kernel_helper` calls `clCreateProgramWithSource` and then `clBuildProgram`, so this code is creating the program object and building it twice, which probably isn't intended.

test_conformance/SVM/test_unified_svm_capabilities.cpp
```diff
@@ -291,8 +291,7 @@ struct UnifiedSVMCapabilities : UnifiedSVMBase
         test_error(err, "could not write to usvm memory");
 
         cl_int check;
-        err = clEnqueueSVMMemcpy(queue, CL_TRUE, &check, mem->get_ptr(),
-                                 sizeof(value), 0, nullptr, nullptr);
+        err = mem->read(check);
```
bashbaug, line 294: I think I'd prefer to keep this an explicit call to `clEnqueueSVMMemcpy` (or `clEnqueueSVMMemcpyWithPropertiesKHR`) and skip the "reading via memcpy" path if `CL_SVM_CAPABILITY_SINGLE_ADDRESS_SPACE_KHR` is not supported.  This aligns with `test_CL_SVM_CAPABILITY_DEVICE_WRITE_KHR` - see below.

test_conformance/SVM/test_unified_svm_capabilities.cpp
```diff
@@ -359,17 +361,20 @@ struct UnifiedSVMCapabilities : UnifiedSVMBase
...
 
         // test writing via memcpy
-        value = genrand_int32(d);
-        err = clEnqueueSVMMemcpy(queue, CL_TRUE, mem->get_ptr(), &value,
-                                 sizeof(value), 0, nullptr, nullptr);
-        test_error(err, "could not write to usvm memory with memcpy");
-
-        err = mem->read(check);
-        test_error(err, "could not read from usvm memory");
+        if (mem->getCapabilities() & CL_SVM_CAPABILITY_SINGLE_ADDRESS_SPACE_KHR)
```
bashbaug, line 364: I have a light preference toward using `deviceUSVMCaps[typeIndex]` here vs. adding `getCapabilities()` to the fixture. This is the pattern used by other USVM tests.

test_conformance/SVM/test_unified_svm_setarg.cpp
```diff
@@ -72,7 +72,8 @@ struct UnifiedSVMSetArg : UnifiedSVMBase
             test_error(err, "clEnqueueReadBuffer failed");
 
             // Validate result
-            const cl_uchar *expected_data = src_data.data() + offset;
+            std::vector<cl_uchar> expected_data(length, 0);
+            src->read(expected_data, offset);
```
bashbaug, line 76: I don't understand this change. Why do we need to read the source data again? It shouldn't have changed.

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -60,23 +75,31 @@ static inline void parseSVMAllocProperties(
 template <typename T> class USVMWrapper {
 public:
     USVMWrapper(cl_context context_, cl_device_id device_,
-                cl_command_queue queue_, cl_uint typeIndex_,
+                cl_command_queue queue_, cl_kernel kernel_WriteOp_,
+                cl_kernel kernel_ReadOp_, cl_uint typeIndex_,
```
bashbaug, line 79: Did you consider building kernel_WriteOp and kernel_ReadOp as part of this wrapper (if required for the passed-in caps) vs. passing the kernel handles into the wrapper?

I know there are tradeoffs:

1. If we build the kernels as part of this wrapper we could end up compiling them multiple times, especially for tests that instantiate the wrapper multiple times.
2. If we build the kernels as part of the wrapper and based on the passed-in caps, we might not need to compile any kernels ever, for implementations that always support a single address space though.
3. If we build the kernels as part of the wrapper then then the wrapper interface stays more-or-less the same, and we only need to change one place if we want to use a different kernel instead (this is true today, but only because we only instantiate the wrapper from UnifiedSVMBase).

test_conformance/SVM/test_unified_svm_mem_cpy.cpp
```diff
@@ -54,9 +54,19 @@ struct UnifiedSVMOPs : UnifiedSVMBase
             void *dst_ptr = &dst->get_ptr()[offset];
 
             clEventWrapper event;
-            err = clEnqueueSVMMemcpy(queue, CL_BLOCKING, dst_ptr, src_ptr,
-                                     length, 0, nullptr, &event);
-            test_error(err, "clEnqueueSVMMemcpy failed");
+            if (!use_new_api)
```
bashbaug, line 57: Nitpicks:

* I'd prefer to test for the positive condition `if (use_new_api)` vs. the negative condition `if (!use_new_api)`.
* Could we use a more descriptive name than "new API"? Maybe something about "with properties"? Or "extension API", or "KHR API"? Or, if we want to invert the meaning, "core API"?

bashbaug (commented):
I don't have an easy way to test SVM types without the single address space capability, but I did run these tests with the unified SVM emulation layer and they're all still passing, except for one case that appears to be unrelated to these changes.

test_conformance/SVM/unified_svm_fixture.h
```diff
@@ -290,6 +366,27 @@ struct UnifiedSVMBase
...
+        const char* programString = R"(
+            kernel void test_WriteOp(global unsigned char* dst, const global unsigned char* src)
+            {
+                dst[get_global_id(0)] = src[get_global_id(0)];
+            }
+
+            kernel void test_ReadOp(const global unsigned char* src, global unsigned char* dst)
+            {
+                dst[get_global_id(0)] = src[get_global_id(0)];
+            }
```
bashbaug, line 378: Actually, thinking about this a bit more, do we really need separate "write" and "read" kernels?  Seems like we would be fine with just a single "copy" kernel that works in both directions.

test_conformance/SVM/unified_svm_fixture.h
joselopeqti: Would it be possible to keep it until that change gets into the main branch of the headers?

test_conformance/SVM/unified_svm_fixture.h
joselopeqti: Done

test_conformance/SVM/unified_svm_fixture.h
joselopeqti: Yes, done

test_conformance/SVM/unified_svm_fixture.h
joselopeqti: Done

test_conformance/SVM/unified_svm_fixture.h
joselopeqti: Done

test_conformance/SVM/unified_svm_fixture.h
joselopeqti: Done

test_conformance/SVM/test_unified_svm_setarg.cpp
joselopeqti: Yeah, this was not needed, it was committed by mistake.

test_conformance/SVM/unified_svm_fixture.h
joselopeqti: I would say I like the idea of keeping the interface more of less the same, even though right now it is only one kernel. Do we foresee a scenario where more kernels have to be used by the wrapper?

PR #2804, opened 2026-09-09
integer-ops: Optimize tests to improve runtime

rjodinchr:
This patch rationalizes the integer operations conformance tests by focusing on test data quality rather than raw random iteration quantity. By prioritizing high-value edge cases and removing redundant test suites, the overall workload is confidently reduced without sacrificing test integrity.

Specific changes include:
* Remove explicit `quick_*` test definitions and handle workload scaling dynamically using `gWimpyMode`.
* Replace hardcoded static `specialCaseList` arrays with mathematically dense edge cases generated dynamically via `GetIntSpecialValues`.
* Add `offset_limit` and `wimpy` parameters to `GetIntSpecialValues` to drastically prune the combinatorial tree of generated edge cases when fast/wimpy runs are requested.
* Rewrite `char` and `uchar` initialization to use a deterministic sliding 16-bit window instead of random data, systematically guaranteeing complete coverage for 8-bit interactions.


test_common/harness/conversions.h
```diff
@@ -251,9 +259,10 @@ template <typename T> std::vector<T> InitIntSpecialValues()
     return vec;
 }
 
-template <typename T> std::vector<T> &GetIntSpecialValues()
+template <typename T>
+std::vector<T> &GetIntSpecialValues(int offset_limit = 3, bool wimpy = false)
 {
-    static std::vector<T> vec = InitIntSpecialValues<T>();
+    static std::vector<T> vec = InitIntSpecialValues<T>(offset_limit, wimpy);
```
shajder: Effectively, only the first invocation on `GetIntSpecialValues` would honor `offset_limit`/`wimpy` arguments, is that right and intended?

test_common/harness/conversions.h
```diff
@@ -181,26 +182,33 @@ template <typename T> std::vector<T> InitIntSpecialValues()
...
         }
     };
 
     // Add all the values close to 0, MIN, and MAX.
-    for (unsigned i = 0; i < 256; i++)
-    {
-        push(i);
-    }
+    push((T)0);
+    push((T)1);
```
shajder, line 208: The old `for (i = 0; i < 256; i++) push(i)` sweep is replaced by just `push(0); push(1)`, unconditionally, not behind `wimpy` unlike the other changes around which are explicitly reduced only in wimpy mode. Is dropping this sweep in the full (non-wimpy) run intentional?

test_conformance/integer_ops/verification_and_generation_functions.cpp
```diff
@@ -280,28 +284,27 @@ verify_long(int test, size_t vector_size, cl_long *inptrA, cl_long *inptrB, cl_l
     if (count) return -1; else return 0;
 }
 
-void
-init_long_data(uint64_t indx, int num_elements, cl_long *input_ptr[], MTdata d)
+void init_long_data(uint64_t indx, uint32_t num_elements, cl_long *input_ptr[],
+                    MTdata d, int num_runs_shift)
```
shajder: `init_long_data` / `init_ulong_data` / `init_int_data` / `init_uint_data` / `init_short_data` / `init_ushort_data` are now nearly identical so it is worth to try to remove code duplication by collapsing into one templated `init_data<StorageType, clType>` instead of maintaining six copies of the same logic.

test_conformance/integer_ops/verification_and_generation_functions.cpp
```diff
@@ -13,18 +13,22 @@
...
 #include <sys/stat.h>
 
+#include <cassert>
 #include <cinttypes>
 
 #include "harness/conversions.h"
+#include "harness/parseParameters.h"
 
 extern     MTdata          d;
+#define MASK(n) ((1 << (n)) - 1)
```
shajder: Dead code?

test_conformance/integer_ops/verification_and_generation_functions.cpp
```diff
@@ -1367,31 +1368,38 @@ verify_ushort(int test, size_t vector_size, cl_ushort *inptrA, cl_ushort *inptrB
...
-            {
-                ((cl_ushort *)input_ptr[0])[index] = specialCaseList[x];
-                ((cl_ushort *)input_ptr[1])[index++] = specialCaseList[y];
-            }
+        cl_uint random = genrand_int32(d);
+        input_ptr[0][index] = bitcast<cl_ushort, cl_ushort>(random & 0xffff);
+        input_ptr[0][index + 1] = bitcast<cl_ushort, cl_ushort>(random >> 16);
+        random = genrand_int32(d);
+        input_ptr[1][index] = bitcast<cl_ushort, cl_ushort>(random & 0xffff);
+        input_ptr[1][index + 1] = bitcast<cl_ushort, cl_ushort>(random >> 16);
```
shajder: Overflow of `input_ptr[1]` when the remaining count is odd - the loop condition should be `index + 1 < num_elements`, not `index < num_elements`, is that right?

shajder (changes requested):
This PR changes testing scope in two independent ways, first the per-type shift sizes are now cut differently by domain size (long -1, int -2, short -3, char down to exhaustive). Second, the 0..255 base sweep in `InitIntSpecialValues` is dropped unconditionally, not just in wimpy mode, so even a full conformance run loses that coverage with no way to opt back in.

test_common/harness/conversions.h
rjodinchr: This is correct. It is indeed the intended behavior

test_common/harness/conversions.h
rjodinchr: Yes it is intentional. It drops some value, but I would argue that they are definitely not special enough to be kept, and it free the same for some important one.

Note that all those values get transformed into 3 other values (`v` => `~v`, `v^sign`, `~(v^sign)`).

| value | source |
|---|---|
| 0 | push(0) |
| 1 | push(1) |
| 2 | 2^1 |
| 3 | 3^1 |
| 4 | 2^2 |
| 5 | 5^1 |
| 6 | 5^1 + 1 |
| 7 | 7^1 |
| 8 | 2^3 |
| 9 | 3^2 |
| 10 | 3^2 + 1 |
| 11 | 3^2 + 2 |
| 12 | 3^2 + 3 |
| 13 | 2^4 - 3 |
| 14 | 2^4 - 2 |
| 15 | 2^4 - 1 |
| 16 | 2^4 |
| 17 | 2^4 + 1 |
| 18 | 2^4 + 2 |
| 19 | 2^4 + 3 |
| ... | *(20–21 missing)* |
| 22 | 5^2 - 3 |
| 23 | 5^2 - 2 |
| 24 | 5^2 - 1 |
| 25 | 5^2 |
| 26 | 5^2 + 1 |
| 27 | 3^3 |
| 28 | 3^3 + 1 |
| 29 | 2^5 - 3 |
| 30 | 2^5 - 2 |
| 31 | 2^5 - 1 |
| 32 | 2^5 |
| 33 | 2^5 + 1 |
| 34 | 2^5 + 2 |
| 35 | 2^5 + 3 |
| ... | *(36–45 missing)* |
| 46 | 7^2 - 3 |
| 47 | 7^2 - 2 |
| 48 | 7^2 - 1 |
| 49 | 7^2 |
| 50 | 7^2 + 1 |
| 51 | 7^2 + 2 |
| 52 | 7^2 + 3 |
| ... | *(53–60 missing)* |
| 61 | 2^6 - 3 |
| 62 | 2^6 - 2 |
| 63 | 2^6 - 1 |
| 64 | 2^6 |
| 65 | 2^6 + 1 |
| 66 | 2^6 + 2 |
| 67 | 2^6 + 3 |
| ... | *(68–77 missing)* |
| 78 | 3^4 - 3 |
| 79 | 3^4 - 2 |
| 80 | 3^4 - 1 |
| 81 | 3^4 |
| 82 | 3^4 + 1 |
| 83 | 3^4 + 2 |
| 84 | 3^4 + 3 |
| ... | *(85–96 missing)* |
| 97 | 10^2 - 3 |
| 98 | 10^2 - 2 |
| 99 | 10^2 - 1 |
| 100 | 10^2 |
| 101 | 10^2 + 1 |
| 102 | 10^2 + 2 |
| 103 | 10^2 + 3 |
| ... | *(104–121 missing)* |
| 122 | 5^3 - 3 |
| 123 | 5^3 - 2 |
| 124 | 5^3 - 1 |
| 125 | 5^3 |
| 126 | 5^3 + 1 |
| 127 | 2^7 - 1 |
| 128 | 2^7 |
| 129 | 2^7 + 1 |
| 130 | 2^7 + 2 |
| 131 | 2^7 + 3 |
| ... | *(132–239 missing)* |
| 240 | 3^5 - 3 |
| 241 | 3^5 - 2 |
| 242 | 3^5 - 1 |
| 243 | 3^5 |
| 244 | 3^5 + 1 |
| 245 | 3^5 + 2 |
| 246 | 3^5 + 3 |
| ... | *(247–252 missing)* |
| 253 | 2^8 - 3 |
| 254 | 2^8 - 2 |
| 255 | 2^8 - 1 |
| 256 | 2^8 |

test_conformance/integer_ops/verification_and_generation_functions.cpp
rjodinchr: They are not that nearly identical, but I've switched to a template based approach (`int_data`, `int_data_16bit` & `int_data_8bit`)

test_conformance/integer_ops/verification_and_generation_functions.cpp
rjodinchr: Indeed, thank you

test_conformance/integer_ops/verification_and_generation_functions.cpp
rjodinchr: good catch, thanks

test_conformance/integer_ops/verification_and_generation_functions.cpp
shajder: Version for `long/ulong/int/uint` passes at my machine:

```
template <typename T> T genrand(MTdata d);
template <> uint32_t genrand<uint32_t>(MTdata d) { return genrand_int32(d); }
template <> uint64_t genrand<uint64_t>(MTdata d) { return genrand_int64(d); }

template <typename T, typename clType>
void int_data(uint64_t indx, uint32_t num_elements, clType *input_ptr[],
              MTdata d, int num_runs_shift)
{
    const auto &specialValues = GetIntSpecialValues<T>(1, gWimpyMode);
    assert((1ULL << (num_runs_shift / 2)) >= specialValues.size());
    uint32_t index;
    const uint64_t num_specials = specialValues.size();
    const uint64_t total_special_pairs = num_specials * num_specials;
    for (index = 0;
         ((index + indx) < total_special_pairs) && index < num_elements;
         index++)
    {
        input_ptr[0][index] =
            bitcast<T, clType>(specialValues[(index + indx) % num_specials]);
        input_ptr[1][index] =
            bitcast<T, clType>(specialValues[(index + indx) / num_specials]);
    }
    for (; index < num_elements; index++)
    {
        input_ptr[0][index] = bitcast<T, clType>(genrand<T>(d));
        input_ptr[1][index] = bitcast<T, clType>(genrand<T>(d));
    }
}
```

test_common/harness/conversions.h
shajder: Understood, still, might be worth adding a simple log noting that the parameters are being ignored on subsequent calls.

rjodinchr: done

shajder: hmm, something is wrong at my end:

long_math...
Testing "+" test_integer_ops: /home/mahd/projects/khr/OpenCL-CTS/test_common/harness/conversions.h:277: std::__debug::vector<T>& GetIntSpecialValues(int, bool) [with T = long unsigned int]: Assertion `false' failed.

rjodinchr: Sorry about that, it should be fixed now.

rjodinchr: The bad news is that the CI did not catch it somehow. I was able to reproduce locally, but I would have expected the CI to reproduce that bug as well.

rjodinchr: https://github.com/KhronosGroup/OpenCL-CTS/actions/runs/34574406255/job/103185720488
The CTS is build in release, so the assertion did nothing.
But we have the error message:
https://github.com/KhronosGroup/OpenCL-CTS/actions/runs/34574406255/job/103185720488

rjodinchr: In fact, I'm not convinced about my implementation, I will change it later today.

rjodinchr: This is ready to be reviewed.

shajder (approved):
LGTM

bashbaug:
Merging as discussed in the September 22nd teleconference.

PR #2805, opened 2026-09-10
Fix test_vulkan_interop_image descriptor/command-buffer reuse hazard in useSingleImageKernel path

### LINKED ISSUE #2800: test_image_single_queue FAILED with --useSingleImageKernel when num2DImages > 1
Repro: ./test_vulkan --useSingleImageKernel test_image_single_queue

Number of images: 1 passes for all 21 formats. Number of images: 2
fails immediately at the first format/size (r8ui, 4x4) with "Source and destination buffers don't match".

Environment: rusticl -> Zink -> Lavapipe (software), Mesa 26.3.0-devel, LLVM 17.0.6, VK_LAYER_KHRONOS_validation 1.4.313, Vulkan instance 1.4.358. Not yet verified on hardware drivers.

shajder:
Fixes #2800

Environment: rusticl -> Zink -> Lavapipe (software), Mesa 26.3.0-devel, LLVM 17.0.6, VK_LAYER_KHRONOS_validation 1.4.313, Vulkan instance 1.4.358.

Fixes VUID-vkUpdateDescriptorSets-None-03047 and VUID-vkBeginCommandBuffer-commandBuffer-00049, only visible with options `--useSingleImageKernel --useValidationLayers` together.

`vkDescriptorSet` and `vkCopyCommandBuffer` were reused across the per-image loop in `run_test_with_one_queue` and `run_test_with_two_queue` without waiting for the previous iteration's dispatch to finish. Added a VulkanFence to synchronize reuse.

### DIFF
--- a/test_conformance/vulkan/test_vulkan_interop_image.cpp
+++ b/test_conformance/vulkan/test_vulkan_interop_image.cpp
@@ -283,8 +283,8 @@ int run_test_with_two_queue(
     VulkanCommandPool vkCommandPool(vkDevice);
     VulkanCommandBuffer vkCopyCommandBuffer(vkDevice, vkCommandPool);
     VulkanCommandBuffer vkShaderCommandBuffer(vkDevice, vkCommandPool);
-    std::shared_ptr<VulkanFence> vkImageDispatchFence(
-        new VulkanFence(vkDevice));
+    std::shared_ptr<VulkanFence> vkImageDispatchFence =
+        std::make_shared<VulkanFence>(vkDevice);
     VulkanQueue &vkQueue = vkDevice.getQueue(getVulkanQueueFamily());
 
     VulkanSemaphore vkVk2CLSemaphore(vkDevice, vkExternalSemaphoreHandleType);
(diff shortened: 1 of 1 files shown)


rjodinchr (commented):
Can you explain why we should use a `VulkanFence` and not just call `waitIdle`?
If we keep `VulkanFence`, I think we should use `std::make_shared` instead of `new`, and we could have something like that:
```
if (i2DIdx < vkImage2DList->size() - 1)
{
  vkQueue.submit(...);
  vkImageDispatchFence->wait();
  vkImageDispatchFence->reset();
}
```

shajder:
> Can you explain why we should use a `VulkanFence` and not just call `waitIdle`? 

`vkQueueWaitIdle` drains the whole queue; from Khronos's own guidance (Vulkan-Samples wait_idle [sample](https://github.com/KhronosGroup/Vulkan-Samples/tree/main/samples/performance/wait_idle)):

> Don't: Use vkQueueWaitIdle() or vkDeviceWaitIdle() unless absolutely necessary for less granular synchronisation.
> Do: Use Fences to asynchronously read back data to the CPU.

...

> If we keep `VulkanFence`, I think we should use `std::make_shared` instead of `new`...

Done.

rjodinchr (approved):

PR #2806, opened 2026-09-10
commonfns: fix signed-zero verification

svenvh:
Positive and negative zero compare equal in C++, which allowed results of 0 with the wrong sign to pass.  Fix with sign-aware comparisons.

Also fix smoothstep diagnostics to report arguments in the official function signature order.

This is preparatory work for improving the test's input ranges.

[run-test: test_commonfns]

### DIFF
--- a/test_conformance/commonfns/test_base.h
+++ b/test_conformance/commonfns/test_base.h
@@ -175,6 +175,17 @@ template <typename T> inline float conv_to_flt(const T &val)
         return (float)val;
 }
 
+// Unlike ordinary equality, distinguish positive and negative zero.
+template <typename Expected, typename Actual>
+inline bool fp_value_equals(const Expected &expected, const Actual &actual)
+{
+    const double expected_value = conv_to_dbl(expected);
+    const double actual_value = conv_to_dbl(actual);
+    if (expected_value != actual_value) return false;
+    return expected_value != 0.0
+        || std::signbit(expected_value) == std::signbit(actual_value);
+}
+
 template <typename T> inline half conv_to_half(const T &val)
 {
     if (std::is_floating_point<T>::value)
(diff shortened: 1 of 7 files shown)


test_conformance/commonfns/test_base.h
```diff
@@ -175,6 +175,25 @@ template <typename T> inline float conv_to_flt(const T &val)
...
+
+// Unlike ordinary equality, distinguish positive and negative zero.
+template <typename Expected, typename Actual>
+inline bool fp_value_equals(const Expected &expected, const Actual &actual)
+{
+    const double expected_value = conv_to_dbl(expected);
+    const double actual_value = conv_to_dbl(actual);
+    if (expected_value != actual_value) return false;
+    return expected_value != 0.0 || fp_signbit(expected) == fp_signbit(actual);
+}
```
rjodinchr: I don't think `fp_signbit` is really needed here.
`expected_value` and `actual_value` are already converted to `double` via `conv_to_dbl`.
For `cl_half`, `conv_to_dbl` calls `cl_half_to_float()`, which converts `0x8000` to `-0.0f` with bit 31 set. Casting to `double` preserves `-0.0`.
```suggestion
// Unlike ordinary equality, distinguish positive and negative zero.
template <typename Expected, typename Actual>
inline bool fp_value_equals(const Expected &expected, const Actual &actual)
{
    const double expected_value = conv_to_dbl(expected);
    const double actual_value = conv_to_dbl(actual);
    if (expected_value != actual_value) return false;
    return expected_value != 0.0 || [fp_signbit(expected) == fp_signbit(actual);](std::signbit(expected_value) == std::signbit(actual_value);)
}
```

svenvh: Nice catch, thanks!

rjodinchr (approved):

bashbaug:
Merging as discussed in the September 22nd teleconference.

PR #2809, opened 2026-09-11
ci: PoCL build changes to pass more tests

bashbaug:
This PR includes several changes that might enable more tests to pass with PoCL:

1. Update to LLVM 22. This didn't have as big of an effect as I thought it might have when I tested locally, but it couldn't hurt. We could also try moving to LLVM 23.
2. Add the `ENABLE_CONFORMANCE` flag according to the [PoCL documentation](https://portablecl.org/docs/html/conformance.html#pocl-conformance). This did seem to have a big effect, but it also has some tradeoffs.  In particular it disables cl_khr_fp16 support, so we should carefully consider whether this is the right tradeoff. If we decide not to do this, I'll just drop the second commit.

### DIFF
--- a/.github/actions/install-pocl-deps/action.yml
+++ b/.github/actions/install-pocl-deps/action.yml
@@ -10,6 +10,8 @@ runs:
     - name: Install dependencies
       shell: bash
       run: |
+        curl -L "https://apt.llvm.org/llvm-snapshot.gpg.key" | sudo apt-key add -
+        echo "deb https://apt.llvm.org/noble/ llvm-toolchain-noble-${{ inputs.llvm-version }} main" | sudo tee -a /etc/apt/sources.list
         sudo apt-get update
         sudo apt-get install -y \
           python3-dev libpython3-dev build-essential ocl-icd-libopencl1 \
--- a/.github/workflows/pocl.yml
+++ b/.github/workflows/pocl.yml
@@ -6,8 +6,8 @@ permissions:
   pull-requests: read
 
 env:
-  LLVM_VERSION: 20
-  TRANSLATOR_BRANCH: llvm_release_200
+  LLVM_VERSION: 22
+  TRANSLATOR_BRANCH: llvm_release_220
   POCL_VERSION: v7.2
   VULKAN_SDK_VERSION: 1.4.309.0
 
--- a/.github/workflows/pocl.yml
+++ b/.github/workflows/pocl.yml
@@ -123,6 +123,7 @@ jobs:
             -DLLVM_DIR=/usr/lib/llvm-${LLVM_VERSION}/lib/cmake/llvm \
             -DLLVM_SPIRV_LIB=${SPIRV_LLVM_TRANSLATOR_PATH}/lib/libLLVMSPIRVLib.a \
             -DLLVM_SPIRV_INCLUDEDIR=${SPIRV_LLVM_TRANSLATOR_PATH}/include/LLVMSPIRVLib \
+            -DENABLE_CONFORMANCE=ON \
             -DENABLE_ICD=ON
           cmake --build pocl/build
           cmake --install pocl/build


bashbaug:
Seems promising:

```
--- Test Differences ---
 FIX: [test_bruteforce] 'fract' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'ilogb' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'ldexp' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'log' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'logb' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'modf' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'nextafter' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'pown' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'powr' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'remainder' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'remquo' expected 'fail', but got 'pass'.
 FIX: [test_bruteforce] 'rootn' expected 'fail', but got 'pass'.
```

PoCL folks, do you have any advice how we should be setting `ENABLE_CONFORMANCE`?

rjodinchr:
I think we should not set `ENABLE_CONFORMANCE`, as it disables many features. I think we learn more with as many things enabled as possible.

I'm fine switching to a newer LLVM version.

bashbaug:
> I think we should not set `ENABLE_CONFORMANCE`, as it disables many features. I think we learn more with as many things enabled as possible.

Sure, sounds good, I've reverted the second commit.

It'd be nice to fix PoCL so `ENABLE_CONFORMANCE` isn't required and we get the best of both worlds - all features enabled and all tests passing!

rjodinchr:
It seems like the LLVM update is enough to fix some brute force test, that's nice!

bashbaug:
Hmm, I'm a little worried that we're hitting the PoCL cache, so the tests are still running with `ENABLE_CONFORMANCE`.  I didn't see these tests passing when I ran locally.

I'll see if I can force a rebuild.

bashbaug:
I forced a rebuild and we're back to the same set of passing and failing tests now, at least for the math tests.

rjodinchr (approved):

rjodinchr:
I'm fine landing this asap as it is just CI

bashbaug:
Merging as discussed in the September 15th teleconference.

pjaaskel:
Sorry for being late noticing this one: ENABLE_CONFORMANCE's semantics are supposed to be such that it exposes only features that are known to pass the conformance test and it might also make some shortcuts for speedup purposes. So if you have features that pass CTS with ENABLE_CONFORMANCE off, it should be fixed in PoCL to expose those features also when it's ON. Pinging Michal @franz .

bashbaug:
@pjaaskel @franz How about the other way around - is there any tracker to inventory and ideally address CTS tests that fail with ENABLE_CONFORMANCE off?  Not sure how you would want to track this, but I'd be happy to file a PoCL issue if that would be useful. Thanks!

pjaaskel:
I'm not aware of, @franz may have one. There's a list of unfinished extensions which can be browsed in CMakeLists.txt, but also some core features. Perhaps some of them are outdated and could be conformant, please file issues if you find such.

franz:
I don't think there is a tracker as such; as Pekka mentioned, CMakeLists.txt does most of the enabling/disabling of features based on ENABLE_CONFORMANCE, though some of it is also in the sources.

Probably simplest way is to get the full list is a `git grep ENABLE_CONFORMANCE` in the PoCL repository. 

Some of the unfinished / buggy features are probably easy to fix, but some are definitely not ( like full subgroup support).

The FP16 support has been fixed in the main branch (but this wasn't backported to 7.2) so that it passes all FP16 tests in CTS math_bruteforce.

PR #2815, opened 2026-09-17
Fix final test status in case of skipp and selection subtest

gwawiork:
In case of selection one sub test the Final status of test in case of SKIPP is not printed:
Example before fix
test_api get_device_info_31
Compute Device Name = Intel(R) Iris(R) Xe Graphics, Compute Device Vendor = Intel(R) Corporation, Compute Device Version = OpenCL 3.0 NEO , CL C Version = OpenCL C 1.2
Device latest conformance version passed: v2025-04-14-00
sizeof( void*) = 8  (host)
sizeof( void*) = 8  (device)
get_device_info_31...
get_device_info_31 skipped (requires at least OpenCL version 3.1, but the device reports OpenCL version 3.0)

after fix:
test_api get_device_info_31               
Compute Device Name = Intel(R) Iris(R) Xe Graphics, Compute Device Vendor = Intel(R) Corporation, Compute Device Version = OpenCL 3.0 NEO , CL C Version = OpenCL C 1.2 
Device latest conformance version passed: v2025-04-14-00
sizeof( void*) = 8  (host)
sizeof( void*) = 8  (device)
get_device_info_31...
get_device_info_31 skipped (requires at least OpenCL version 3.1, but the device reports OpenCL version 3.0)
SKIPPED 1 of 1 tests.


darmroz:
@bashbaug could you please take a look?

bashbaug (commented):
I'm not opposed to this, but I do wonder what problem this is intended to solve, and I wonder if there might be a more complete solution.

Is the issue that we used to print some sort of "PASSED" message when we skip all tests, but we don't any longer, after https://github.com/KhronosGroup/OpenCL-CTS/pull/2781 (`gTestsFailed + gTestsPassed` is zero when all tests are skipped, which suppresses all output)?  If so, should we count the number of skipped tests (in addition to the number of passed tests and failed tests) and include that in our test "count", so we will still print "PASSED" even if we skip all tests? This will effectively treat a test skip the same as a test pass, which we already do for the exit code.

Or, do we really want to know the number of skipped tests? If so, should we print the number of test skips even when we don't skip all tests? Should we pass the number of skipped tests to `print_results` and report it in some sort of standard format?

darmroz:
Yes, that is exactly the problem. Our CI system was not prepared for the behavior introduced after #2781.
I consider this change a hotfix that restores the previous behavior. A better solution around tracking and reporting skipped tests could certainly be discussed separately.

gwawiork:
> I'm not opposed to this, but I do wonder what problem this is intended to solve, and I wonder if there might be a more complete solution.
> 
> Is the issue that we used to print some sort of "PASSED" message when we skip all tests, but we don't any longer, after #2781 (`gTestsFailed + gTestsPassed` is zero when all tests are skipped, which suppresses all output)? If so, should we count the number of skipped tests (in addition to the number of passed tests and failed tests) and include that in our test "count", so we will still print "PASSED" even if we skip all tests? This will effectively treat a test skip the same as a test pass, which we already do for the exit code.
> 
> Or, do we really want to know the number of skipped tests? If so, should we print the number of test skips even when we don't skip all tests? Should we pass the number of skipped tests to `print_results` and report it in some sort of standard format?

consistency - each run should print any kind of summary status now if we choose case that is not supported we do not have summary status for such run. In all other cases (some test passed/skipped, all skipped ) we have.

output parsers compatibility - we parse test outputs and we are looking for output. Now we do not have chance to find any test status.

bashbaug:
OK. You may wish to look into the `CL_CONFORMANCE_RESULTS_FILENAME` environment variable, which outputs results for each sub-test as a JSON file, no log parsing required. Nevertheless, if we want to log _something_ even when we skip all tests, akin to what we did previously, I suggest one of the two solutions below. Which one we choose should depend whether we want to distinguish between SKIP and PASS tests:

1. If we don't care to distinguish between SKIP and PASS tests (the only important thing is that they did not FAIL), we can simply count the skipped tests (somehow), include this count in the "denominator", and we will print PASS like we did previously.

2. If we do care to distinguish between SKIP and PASS tests (similar to what's currently done in this PR), we need to decide when we want to print the number of SKIP tests. In this PR currently, we only print SKIP information if all tests are skipped, which seems strange to me. If we care about SKIP tests, shouldn't we print this even if some tests pass and some tests are skipped?

Engineering-wise, it would be really nice to eliminate the `gTestsPass` and `gTestsFailed` global variables, especially if we are already iterating over the test statuses, but we can do that in a follow-on PR if we want to limit the scope of this PR.

bashbaug (commented):
OK, I kind of get it now, this change is intended to mirror the output when an entire test suite is skipped, see for example:

https://github.com/KhronosGroup/OpenCL-CTS/blob/03ac70dbbdf7635ff0c51768e03747ab7f7cab3d/test_common/harness/testHarness.cpp#L193-L196

Which produces output like:

```
$ ./test_conformance/pipes/test_pipes 
Compute Device Name = Intel(R) Graphics [0x4626], Compute Device Vendor = Intel(R) Corporation, Compute Device Version = OpenCL 3.0 NEO , CL C Version = OpenCL C 1.2 
Device latest conformance version passed: v2026-04-28-00
sizeof( void*) = 8  (host)
sizeof( void*) = 8  (device)
Test skipped while initialization
SKIPPED 55 of 55 tests.
```

There are lots of additional improvements and unifications we could make in this area, but this PR does seem to be solving a problem in a reasonably consistent way without making things worse, so I'm fine with it.

Approved from my end, but needs a non-Intel reviewer before merging.

rjodinchr (approved):
LGTM

There are typos in the commit title: `skipp` -> `skip`, and `selection subtest` -> `selected tests` (or `selected subtests`).

